PR #22, opened 2016-03-24
More explanation on firefox version/error message

KensoDev:
Struggled with this today, I think it will help a lot of people running older versions of selenium-webdriver.

****Author is Not a native English speaker


felicianotech:
@KensoDev Thank you for your contribution. Your change will be live in the near future.

PR #23, opened 2016-03-24
Deploy staging version of docs to staging bucket

travis:
In preparation for a real staging env for the static site.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,8 +7,9 @@ dependencies:
     #- bundle exec jekyll build -s jekyll -d jekyll/_site/docs/
 
     # https://github.com/jekyll/jekyll/issues/4713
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -13,4 +14,5 @@ deployment:
   prod:
     branch: master
     commands:
+      - aws s3 sync jekyll/_staging_site s3://circle-staging-docs --delete
```
z00b: as i mentioned in the other pr, i'd use a bucket that is itselft a viable hostname, then you can just use static hosting.

travis: see other post - we need multiple buckets stitched together

ndintenfass:
Lgtm

z00b:
:+1:

PR #26, opened 2016-03-25
Temporarily grep for Jekyll build errors until they fix their exit codes.

felicianotech:
The `jekyll build` command seems to return exit code 1 for some errors but not others. These causes scenarios where some docs aren't being build due to YAML issues, Jekyll says there is an error, yet the build passes successfully anyway.

Opened an issue, turns out there's already an issue.  Until this is fixed, we grep the output of Jekyll build, which is saved into a file.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,11 @@ machine:
 dependencies:
   post:
     # https://github.com/jekyll/jekyll/issues/4122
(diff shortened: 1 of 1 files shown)


felicianotech:
Test failed because of the HTML. This means the change made here actually works. o.O

PR #29, opened 2016-03-28
Deprecated the NineFold guide.

felicianotech:
Added a `deprecated` front matter item with an explanation. This isn't technically supported in docs yet but will be.

Also removed the Ninefold link as it's now 404'ing.


ndintenfass:
lgtm

PR #30, opened 2016-03-29
Sync docs to prod

travis:
They won't show up on the prod site until we tweak nginx, but we can go directly to the bucket url to see them until we're ready for that.


ndintenfass:
lgtm

PR #33, opened 2016-03-29
Build tweaks

felicianotech:
- make `.jctl start` duplicate less work so it's faster
- added _config_production.yml so that we can change `site.url` to something real in production. Useful for sitemaps, rss feeds, anything that needs an absolute URL.
- set `JEKYLL_ENV` to production for CircleCI builds so that we can choose avoid Google Tag Manager in dev environments.


ndintenfass:
lgtm

PR #39, opened 2016-04-01
Prevent search form from posting to nothing.

felicianotech:
Prevents the user typing something in the search bar, clicking enter, and having the page reload for no reason.


travis:
lgtm

PR #41, opened 2016-04-04
Add Phabricator doc

appplemac:
Doc on how to make Phabricator work with Circle. This is orientated towards people who are already using Phabricator, so I have omitted general Phabricator setup and the Arcanist setup steps.


KunalJain:
:ship: :shipit:

PR #43, opened 2016-04-04
Updated Composer recommendation for GitHub API limit.

felicianotech:


ndintenfass:
I don't know a lot about Composer, but assuming the fundamental content is sound, the changes lgtm

felicianotech:
Rebasing.

I had this reviewed by a few PHP folks and they said it looks good. Added a specific note about when not to commit `composer.lock`.

PR #44, opened 2016-04-04
Gave sitemap.xml some smarts.

felicianotech:
Set the `lastmod` time in sitemap.xml to today, as a manual value. This is better than not having anything at all, which is where we've been at for awhile. In the future, will update `lastmod` dynamically by reading from Git.

The default `changefreq` value in sitemap.xml is 'monthly. An individual doc can now override that by using the `changefreq` Front Matter variable. For example, the Environment doc has been set to 'weekly' since it changes more often than other docs.

### DIFF
--- a/jekyll/_docs/environment.md
+++ b/jekyll/_docs/environment.md
@@ -1,9 +1,9 @@
 ---
 layout: classic-docs
-title: Test environment
+title: Test Environment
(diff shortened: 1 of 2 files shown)


ndintenfass:
lgtm

felicianotech:
Rebased.

PR #45, opened 2016-04-05
Move some images from frontend

kimh:
Copied images from https://github.com/circleci/frontend/tree/master/resources/public/img/outer/docs


felicianotech:
Good stuff. The "Getting Started" category icon URL didn't change. Is that intentional or just missed? Also I think the icons themselves could just be in the `img` directory, or maybe `img/icons`. The goal is to keep `img/docs` for images that appear in the doc content, such as a screenshot for a step or something.

kimh:
Good catch! I addressed your comment about the place to store the image as well.

felicianotech:
Awesome =) LGTM

PR #46, opened 2016-04-05
Mention the release of Phabricator where CircleCI integration is available

appplemac:
CircleCI integration was made available in the release *2016 Week 13* of Phabricator—older versions don’t have the *Build on CircleCI* item available in Harbormaster. We need to highlight that to eliminate all possible confusion.


KunalJain:
:shipit:

PR #49, opened 2016-04-05
Updated global nav, footer, and JS.

felicianotech:


travis:
should site.min.js be checked in?

felicianotech:
@travis I haven't figured out a good way to mirror some of the assets from the main site yet, and the asset URL to site.min.js now changes frequently. So I manually include the entire file in the repo to get it to work.

Before when the asset URL didn't changed, I just linked to it directly from the CDN.

travis:
ah, k - lgtm

PR #52, opened 2016-04-06
Add Redis to the list of services that can be started during the build

appplemac:
Redis is not being started by default in the new image releases, so it would be important to mention it in the list of services that can be started via `circle.yml`.


felicianotech:
LGTM

PR #53, opened 2016-04-06
Updated text of fastlane version to actual current version

rishimkumar:
said 1.64, should be 1.66 - thanks to user kitsunde!


appplemac:
Good point, 🚢

Issue #55, opened 2016-04-07
RE: jekyll/_docs/test-metadata.md


retorquere:
If I add `--format json --out $CIRCLE_TEST_REPORTS/cucumber/tests.cucumber` as per the suggestion in this document, I get no output from the build myself. Is there a way to collect these artifacts while still seeing the build progress?

bear:
@retorquere cucumber allows you to have multiple `--format` output selections, so this is possible:

```
--format pretty --format json --out $CIRCLE_TEST_REPORTS/cucumber/tests.cucumber
```

retorquere:
So it does! Would be useful to add to the docs I think.

bear:
we were discussing this very thing in our internal support/docs channel :)

I'm going to mark this as closed if that is ok?

retorquere:
absolutely, thank you for your help.

PR #59, opened 2016-04-08
add note for cucumber about multiple --format option

bear:
adding some text to cover the question raised in Issue #55


felicianotech:
LGTM

PR #61, opened 2016-04-11
When there is no output for the API indicate that explicitly instead of showing an empty box

ndintenfass:


kimh:
I don't know the story here but LGTM.

felicianotech:
LGTM too

PR #62, opened 2016-04-11
Update footer that encourages contributions

ndintenfass:
Change the footer that encourages contributions to have a clearer separation from the content of each document. Also, tweak the wording to make it clearer.


kimh:
Non-admins don't have the permission to directly edit files on Github, so maybe we can omit "edit this file" part?

screen shot 2016-04-11 at 3 09 46 pm

acinathan:
Merging this change, as it's still net better -- will fix that link later...

PR #67, opened 2016-04-12
Amend instructions for connecting to host services

ilkka:
The IP of `docker0` needs to be provided to the container for it to be able to connect to e.g. Postgres. One way is the `--add-host` flag for `docker run`. This change adds instructions for doing so, so that people don't need to go looking for this info themselves.

### DIFF
--- a/jekyll/_docs/docker.md
+++ b/jekyll/_docs/docker.md
@@ -353,4 +353,17 @@ test:
     - sudo /etc/init.d/postgresql restart
 ```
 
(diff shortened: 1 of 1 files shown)


kimh:
I suppose this archives the same thing as https://github.com/circleci/circleci-docs/pull/80 ? Sorry for not checking your PR first. I was reviewing in reverse order.

ilkka:
@kimh yeah I think so :) no worries, the main thing is the docs got 1% better.

PR #79, opened 2016-04-14
Clarify that the Phabricator Herald rule must be "Global"

epriestley:
A user ran into some confusion about this: the "Run build plans" action is not available for personal rules, so make it clear that the user must choose the "Global" rule type. Also changed the letter case of a couple of labels to match what appears in the UI.


felicianotech:
Hey, thank you for the contribution. Do you mind adding that Project Rules are supported as well? So that'll be global and project rules.

epriestley:
I think I'd have to implement object rules for revisions in Phabricator first?

nosupport

Object rules are available for commits, but this section is specific to revision rules.

felicianotech:
Hmm, fair enough. Thanks for pointing that out.

PR #80, opened 2016-04-14
Add example for add-host to connect to db

iGEL:
`ip addr show docker0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1` will figure out the IP address from the `docker0` interface.


kimh:
Thanks for the patch. LGTM

PR #82, opened 2016-04-14
Include steps to share .xcsheme file through xcode gui

andrewmcdonald2:


felicianotech:
Not sure why that piece was missing. Thank you so much for the contribution =)

PR #84, opened 2016-04-16
Fix search on Firefox by passing the event as an argument of the two …

ndintenfass:
…relevant functions in search.js


eric-hu:
Cloned, reproduced the error on the commit before this one, and confirmed this fixed it on Firefox.  Didn't have an issue in Chrome or Safari.

LGTM.

PR #86, opened 2016-04-18
Cleanup env page

kimh:
- Convert html to markdown
- Add CIRCLE_BUILD_IMAGE env var that will be available once a related PR is merged

### DIFF
--- a/jekyll/_docs/environment-variables.md
+++ b/jekyll/_docs/environment-variables.md
@@ -114,6 +114,10 @@ The total number of nodes across which the current test is running.
 
 The index (0-based) of the current node.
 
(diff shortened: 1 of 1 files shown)


kimh:
I found adding an empty line after each env var gives too much space, so I guess it's ok as is.

felicianotech:
I think this looks great. Using H2s instead of the H3s that were there is smart. Semantically, that's what it should be.

LGTM :)

kimh:
Now `CIRCLE_BUILD_IMAGE` env var should be available once my change is propagated to all machines. I think it's good to merge this now.

PR #87, opened 2016-04-20
Reflect the change in the iOS inference build command in the docs

appplemac:
We have recently changed the default iOS build command from `xctool` to `xcodebuild` (more details [here](https://discuss.circleci.com/t/xcodebuild-becomes-default-ios-build-command/3266)), so adding this change to the docs.

### DIFF
--- a/jekyll/_docs/ios-builds-on-os-x.md
+++ b/jekyll/_docs/ios-builds-on-os-x.md
@@ -66,12 +66,12 @@ customization options.
 
 By default, CircleCI will build your project with Xcode 7.0. You can select 7.1, 7.2 or 7.3.
 by specifying the version in a [circle.yml file]({{ site.baseurl }}/configuration/) in the root of your
-repo. For example, for 7.2, add the following:
+repo. For example, for 7.3, add the following:
 
 ```
 machine:
   xcode:
-    version: 7.2
+    version: 7.3
 ```
 
 ### CocoaPods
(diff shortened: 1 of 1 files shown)


jekyll/_docs/ios-builds-on-os-x.md
```diff
@@ -136,9 +136,9 @@ While CircleCI's inferred commands will handle many common testing patterns, you
 also have a lot of flexibility to customize what happens in your build.
 
 ### Build Commands
-CircleCI runs tests from the command line with the [`xctool`](https://github.com/facebook/xctool)
-command by default. We have found that `xctool` is more stable when testing with
-the iOS simulator than using `xcodebuild` directly.
+CircleCI runs tests from the command line with the [`xcodebuild`](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/xcodebuild.1.html)
+command by default. This is a tool developed by Apple, and we found it to be the most
```
rishimkumar: "... and we *find* it  to be the"  as opposed to the past-tense found.  but honestly ~~

jekyll/_docs/ios-builds-on-os-x.md
```diff
@@ -442,18 +443,18 @@ machine:
...
-      CODE_SIGN_IDENTITY=      # as it will be done in the deployment step
-      PROVISIONING_PROFILE=
-      -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest'
-      -sdk iphonesimulator
-      -workspace MyWorkspace.xcworkspace # your workspace name
-      -scheme "My Scheme" # your scheme name
-      build build-tests run-tests
+    - set -o pipefail &&
+      xcodebuild
+        CODE_SIGNING_REQUIRED=NO # no need to do code signing at this point
```
rishimkumar: i'm sad to say i dont know how the yml parser works - but just a check here as it appears to have a space in front of CODE_SIGNING_REQUIRED  - which the original text did not

rishimkumar:
igyb changes or not - just double check what i said.

jekyll/_docs/ios-builds-on-os-x.md
appplemac: @rishimkumar Extra space in this particular case is fine—just double-checked with a YAML parser ([here](http://yaml-online-parser.appspot.com)). Thanks for pointing this out.

Issue #85, opened 2016-04-16
gcloud update fails without sudo


kunchenguid:
In https://circleci.com/docs/google-cloud-platform/ and https://circleci.com/docs/google-auth/ there are a couple of places where the example suggested running
`gcloud components update`

this won't succeed without sudo. 

```
ERROR: (gcloud.components.update) You cannot perform this action because you do not have permission to modify the Google Cloud SDK installation directory [/opt/google-cloud-sdk].

Re-run the command with sudo: sudo /opt/google-cloud-sdk/bin/gcloud ...

gcloud --quiet components update kubectl returned exit code 1

Action failed: gcloud --quiet components update kubectl
```

the correct example should be:
`sudo /opt/google-cloud-sdk/bin/gcloud --quiet components update`

kunchenguid:
No wait.. even sudo won't work. Everything started to fail after running gcloud components update with sudo.

felicianotech:
Your original assertion was correct. Using sudo and the full path is the way to use it, ran into this issue myself on an upcoming guide. We'll need to update the docs you pointed out.

What is the new error you're getting with updating?

kunchenguid:
I ended up doing the following:

```
#!/bin/bash
sudo -E env "PATH=$PATH" gcloud --quiet components update kubectl
sudo chmod 757 /home/ubuntu/.config/gcloud/logs -R

gcloud auth activate-service-account --key-file ./my/key.json
gcloud config set project $GCLOUD_PROJECTID
gcloud container clusters get-credentials my-cluster \
  --project=$GCLOUD_PROJECTID --zone=$CLOUDSDK_COMPUTE_ZONE
```

It's really weird that "gcloud container clusters get-credentials" couldn't read the environment variables correctly and requires us to manually specify --project and --zone. but once specified, this is working. kubectl can work fine after these commands.

felicianotech:
@kunchenguid  Once again, thanks for pointing this out. I was aware of the need to use sudo and the entire path but didn't realize the gcloud binary also has trouble with environment variables.

Docs has been updated to reflect using the long, unwieldy command while we investigate getting the command to work the way it should from within the image.

PR #89, opened 2016-04-20
Updated gcloud command example to reflect real usage.

felicianotech:
Currently, you can't use `gcloud` on CircleCI without running as root and using the full path. Until this changes in the build images, the doc has been updated to reflect correct usage.


kimh:
LGTM.

I didn't know this issue. Do you know if we have a ticket for this?

felicianotech:
@kimh  I'm not sure. I do know that it was mentioned recently in #85. I've also dealt with it in two guides now and have seen a few mentions in https://discuss.circleci.com. I have it on my task link to investigate some more and then bother you about it =)

PR #93, opened 2016-04-25
Fix the syntax of the suggested iOS command

appplemac:
YAML does not accept comments in this particular position, so removing the comment.


levlaz:
LGTM 🚢

PR #94, opened 2016-04-25
Fix error when using Android build tools >= 23.0.2

epeterson320:
I followed the docs and was not able to install build tools 23.0.3. I fixed it following the advice from the forums here:

https://discuss.circleci.com/t/installing-android-build-tools-23-0-2/924

### DIFF
--- a/jekyll/_docs/android.md
+++ b/jekyll/_docs/android.md
@@ -24,6 +24,7 @@ installed, you can install it as part of your build with:
 ```
 dependencies:
   pre:
+    - echo y | android update sdk --no-ui --all --filter "tools"
     - echo y | android update sdk --no-ui --all --filter "package-name"
 ```
 


felicianotech:
Hi, that's for the PR. I'm not sure how this solves an issue. In the example here, "package-name" is a placeholder. That should be replaced with what you want to install. This could be `build-tools`, which is the most common scenario, or something else within the Android SDK.

epeterson320:
Thanks for the response. Sorry if I wasn't clear: I _am_ replacing `"package-name"` with the package I want to install, specifically `"build-tools-23.0.3"`. This failed until I first ran the command I added in the PR.

I'm not sure if you want the docs to always recommend installing `tools` first, or to add more explanation about when it's appropriate. I just wanted to open discussion around the issue that the current wording of "replace package-name with the package you want to install" does not work when someone installs build-tools >= 23.0.2.

felicianotech:
Yup, understood, thanks. The issue is that Android SDK Tools isn't update and the newer version has a different repository source, which includes build-tools-23.0.x.

We're planning on updating the version of Android SDK that is included but this PR will be good until that happens. Thank you.

PR #100, opened 2016-04-26
Fix docker exec workaround example

kevinroche:
Too many backticks were used for the docker exec workaround example, causing the curly braces and their contents to not be displayed.


felicianotech:
Great catch. Thank you.

PR #103, opened 2016-04-27
Update how cache works under parallel builds

lumberj:
https://discuss.circleci.com/t/cache-save-restore-algorithm/759


ndintenfass:
Thanks @lumberj

PR #120, opened 2016-05-04
Update toolset.

felicianotech:
Added some missing items to `.gitignore` to complement the fact that not everyone uses the same tools for docs.

Upgraded to Ruby 2.3.1 and Jekyll 3.1.3.


ndintenfass:
lgtm

PR #129, opened 2016-05-10
Use JUnit formatter for cucumber output

jamiecobbett:
We found that the JSON formatter meant that the "Test Summary" didn't include
any details about the failures (other than the filename), meaning you had to
dig through each of the container tabs to find the detail, whereas the JUnit
formatter means the details are included in the summary.


levlaz:
This is excellent @jamiecobbett thanks so much for the contribution!

felicianotech:
Thank you for your contribution @jamiecobbett. :)

PR #132, opened 2016-05-11
Update unrecognized-ruby-version.md

arcz:
Missing space


felicianotech:
Nice catch. Thank you.

PR #145, opened 2016-05-19
Getting Trusty versions fron versions.json

kimh:
Replacing trusty/versions.yml with trusty/versions.json which is
automatically generated by image-builder.

This is one further step toward CD of build image docs


ndintenfass:
lgtm -- Kinda wonder if given our audience we could just point people to the json itself instead of maintaining the page, but we can deal with that later.

felicianotech:
@ndintenfass As long as we have some form of datasource, I think there's a lot we can do with it in the future than simply listing versions on the build image pages.

@kimh I'm assuming we're switching to JSON here due to the image builder workflow correct? If so, LGTM.

kimh:
> Kinda wonder if given our audience we could just point people to the json itself instead of maintaining the page, but we can deal with that later

@ndintenfass I think having the version page is a bit nicer and more friendly for our users. Especially, when the cost of maintaining the page is very small.

> I'm assuming we're switching to JSON here due to the image builder workflow correct? If so, LGTM.

@felicianotech I'm switching to json from yaml because it's more programable IMO.

I'm merging now.

PR #149, opened 2016-05-22
Fix typo in "Fork PR builds"

laughinghan:
The above list of kinds of configuration data is now 5 not 4 items long.


felicianotech:
Nice catch @laughinghan. Thank you.

PR #151, opened 2016-05-22
Add some fb-adb shell commands

ismail-s:


felicianotech:
This looks great @ismail-s, thank you.

PR #153, opened 2016-05-23
Update PHP file paths. Small fixes.

### LINKED ISSUE #98: RE: jekyll/_docs/language-php.md
The documentation specifies that the php conf.d is located at ` ~/.phpenv/versions/$(phpenv global)/etc/conf.d` 

In circle.yaml, I had to use `/opt/circleci/php/$(phpenv global)/etc/conf.d` because the documented path threw `bash: line 1: ~/.phpenv/versions/$(phpenv global)/etc/conf.d/timezone.ini: No such file or directory`

### LINKED ISSUE #133: RE: jekyll/_docs/language-php.md
```
dependencies:
  pre:
    - echo "memory_limit = 64M" > ~/.phpenv/versions/$(phpenv global)/etc/conf.d/memory.ini
```

This way of adding custom configuration to PHP just doesn't work. No such directories in user's home and from everything it looks like it's not the correct way to add .ini configurations anymore.

Please advice/fix.

Tnx

felicianotech:
Update the paths referenced in the PHP doc. CircleCI images (starting with Ubuntu 14.04) now place languages and many other preinstalls in `/opt/circleci/<THING>`.

Fixed the formatting for headings.
-----
Fixes #98 
Fixes #133

### DIFF
--- a/jekyll/_docs/language-php.md
+++ b/jekyll/_docs/language-php.md
@@ -6,12 +6,15 @@ categories: [languages]
 description: Continuous Integration and Continuous Deployment with PHP
 ---
 
-Circle works seamlessly with all the tools and package managersthat
+CircleCI works seamlessly with all the tools and package managers that 
 help PHP developers test and deploy their code. We can generally infer
 most of your dependencies and test commands, but we also provide custom
-configuration via a `circle.yml` checked into your repo's root directory.
+configuration via a `circle.yml` file checked into your repo's root directory. 
+The examples here referring to PHP filepaths are for the Ubuntu 14.04 and newer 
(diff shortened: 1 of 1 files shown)


ndintenfass:
lgtm - only comment is that if someone is using Precise we are sorta leaving them out in the cold, no? I think we are hoping to move over to Trusty as the default for new projects soon, so perhaps not that big a problem

felicianotech:
@ndintenfass That was my idea. If we switch the default soon, then I think this is good. If we're using Precise as the default months from now, no.

felicianotech:
@ndintenfass I added more info for Precise, saying which path to use instead. That's my compromise :)

Merging when this passes.

PR #156, opened 2016-05-24
Exposing trusty versions.json as a public page

kimh:
I want to access Trusty `versions.json` directly via url `/docs/trusty_versions.json` for reusability. We copy the file during CI build  on the fly instead of having the separate file in order to avoid the duplication.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,7 +7,8 @@ dependencies:
     #- bundle exec jekyll build -s jekyll -d jekyll/_site/docs/
 
     # We want to expose the version file to public for the reusability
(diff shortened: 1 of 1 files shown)


kimh:
Ok, jekyll  doesn't allow symbolic link https://github.com/jekyll/jekyll/pull/315

ndintenfass:
Consider making a folder called "environments" and then putting this in there as trusty.json -- thinking ahead to having more environments and more automation.

ndintenfass:
Lgtm. I do wonder if the environment dir should be in source.

kimh:
Good catch. I checked in the directory. Merging now.

PR #161, opened 2016-05-27
Improve CircleCI Docs Search Experience

felicianotech:
Enables full text search for docs. Search priority is title, category, URL,
then doc contents.

Also, shows "No results found" when nothing comes up. Hopefully this reduces
the number of people thinking the search box is broken.


appplemac:
Code LGTM. Let’s iterate based on how it looks in production, if necessary.

PR #165, opened 2016-05-30
Add indication to not add a passphrase

kenden:
when creating an SSH key to be used to tag the github repo


felicianotech:
Thanks @kenden for this. Great point.

PR #169, opened 2016-06-01
Updating wrong PSQL version

kimh:
We were using a wrong command to get PSQL version which gets psql client version, but what we really need is server/daemon version. The command will be fixed at https://github.com/circleci/image-builder/pull/82


kimh:
Small change, so merging.

PR #180, opened 2016-06-06
Added mocha

drazisil:
Also added a section table of contents and re-alphabetized the customer runners are the list is getting a tad long.


drazisil:
pinging @felicianotech :)

felicianotech:
Hey @drazisil, you're the man! :)

One ask, could you switch from relative links (just #whatever) to absolute links (prefixing with `{{site.baseurl}}/test-metadata/#whatever`?

We do this because occasionally a doc will be picked up on a feed reader. When that happens, the relative links break because it becomes something like `https://feedly.com/#whatever` instead of pointing to the correct place.

Thank you =)

drazisil:
@felicianotech  Done.

felicianotech:
+1

PR #181, opened 2016-06-07
Unlock rspec_junit_formatter gem version

RobinDaugherty:
Version 0.2.3 has been released, and there doesn't seem to be any reason to lock the gem to a specific version. (The gem originally needed to be run from the CircleCI fork, but that's no longer the case.)

### DIFF
--- a/jekyll/_docs/test-metadata.md
+++ b/jekyll/_docs/test-metadata.md
@@ -21,7 +21,7 @@ For RSpec:
 Add this to your gemfile:
 
 ```
-gem 'rspec_junit_formatter', '0.2.2'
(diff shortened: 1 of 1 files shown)


felicianotech:
@eric-hu Does this make sense to you?

eric-hu:
@felicianotech Yes, this makes sense. Pinning dependency versions is a tradeoff between getting bug fixes automatically and preventing automatic inclusion of breaking changes.

In this case, the Junit XML spec is pretty established, so I doubt this gem will be updated for anything except bug fixes and Rspec compatibility.

@RobinDaugherty Thank you for spotting this and taking the time to update it.

PR #191, opened 2016-06-14
Escape curly braces

kimh:
Notice `Id` part doesn't show up in our doc.

image


kimh:
Confirmed it's fixed locally.

image

PR #196, opened 2016-06-15
Updating PSQL version now only 9.5 available

keybits:
Since the June 13 update of this image 9.4 is no longer preinstalled.


keybits:
Small change, so merging.

PR #197, opened 2016-06-15
Image Update 201606-01

kimh:
@keybits Thanks for updating psql version on Trusty in #196. Although manually editing the versions.json file works, it's better if you copy versions.json from image-builder artifact.

I just copied the file from https://circleci.com/gh/circleci/image-builder/596#artifacts The build created the image that we are now running in production, therefore the versions.json is the most accurate representation of what we actually have in the image.


felicianotech:
Too bad JSON doesn't allow comments. I would have put a comment at the top saying that it's generated, don't edit manually.

keybits:
Thanks for the info :-) I did wonder where you were grabbing the JSON from. Now 💡

PR #215, opened 2016-06-25
More ruby glob explanation

bellkev:
I always for get which flavors of globbing do/don't support multiple directory expansion. Make it clear that ours does.


felicianotech:
@bellkev You used the word "layes". Is that a typo?

bellkev:
Oops, fixed.

felicianotech:
LGTM if CI passes.

PR #218, opened 2016-06-27
Updated iOS Code Signing guide to describe distribution

KrauseFx:
Distribution was described in the table of contents, but wasn't actually written about - I added a new section 👍


felicianotech:

PR #240, opened 2016-07-15
Add Web Notifications documentation

AndresCuervo:


felicianotech:
Great stuff @AndresCuervo. For SEO purposes, we shouldn't have two H1s on the page, could you change the second one to an H2?

Otherwise, with passing CI, LGTM.

PR #243, opened 2016-07-18
Differences between Trusty and Precise

kimh:


felicianotech:
Great doc. After #244, this PR LGTM.

PR #245, opened 2016-07-18
Cocoapods -> CocoaPods

orta:
Woo!


felicianotech:
Thanks, I love PR like these.

orta:
I had looked a few times to see if I could make this change but never had the chance 👯

PR #252, opened 2016-07-25
Added --file flag to iOS app distribution with Hockeyapp

Tzeejay:
Had the issue with new iOS customer NBCUniversal. The `ipa` gem tries to upload the file (no idea what file) anyways and then just fails and directly prints the Hockeyapp server output into stdout.


kimh:
IGYB

PR #272, opened 2016-08-07
Fix whitespace in circle.yml documentation example

drewinglis:


felicianotech:

PR #296, opened 2016-08-18
Use recommended `--format json` output for cucumber in example

bcjordan:


felicianotech:
It looks like Cucumber's [CI page on GitHub](https://github.com/cucumber/cucumber/wiki/Continuous-Integration) recommends junit.

> If you’re using such a CI server we recommend you use Cucumber with the junit formatter.`

bcjordan:
Gotcha, in that case the preceding paragraph should be changed to reference junit and .xml (right now it mentions cucumber and json)

felicianotech:
Yes, great point. I opened up an [issue](https://github.com/circleci/circleci-docs/issues/354) to track that fix, thank you for pointing that out.

PR #297, opened 2016-08-19
Update gcloud upgrade instructions

eric-hu:
Related JIRA ticket: https://circleci.atlassian.net/browse/CIRCLE-695

### DIFF
--- a/jekyll/_docs/continuous-deployment-with-google-container-engine.md
+++ b/jekyll/_docs/continuous-deployment-with-google-container-engine.md
@@ -88,8 +88,8 @@ You should replace the values of these three variables for your specific project
 ```
 dependencies:
   pre:
-    - sudo /opt/google-cloud-sdk/bin/gcloud --quiet components update
-    - sudo /opt/google-cloud-sdk/bin/gcloud --quiet components update kubectl
+    - sudo /opt/google-cloud-sdk/bin/gcloud --quiet components update --version 120.0.0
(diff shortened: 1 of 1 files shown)


eric-hu:
Adding a little context:

Our Gcloud docs currently suggest upgrading gcloud and gcloud/kubernetes.  Without a version specified, this brings them to version 122 (last tested today).

A customer had an issue deploying with these tools at v122, but was able to successfully deploy using v120 . This version was installed at the suggestion of the Google team.

kimh:
LGTM. FWIW, gcloud is pretty fast to update version, so I haven't pinned which version to install in image-builder.

This PR also reminded me to do https://github.com/circleci/circleci-docs/pull/299

eric-hu:
@kimh Thanks for the PR review.

Note that this is not a "gcloud is outdated" issue. The customer in the ticket used `gcloud components update` and got version 122, which was broken for him.

Someone needs to investigate why that is.  It's possible we need to make other changes to the image.  It's also possible that it's a gcloud-only problem.

kimh:
Yup, Eric, I understood. In my impression, gcloud is a bit buggy. So, technically we can find a **golden** version that works perfect on Circle and pin the version. My comment earlier meant that we don't do it because they update and add features very fast and our customers want them sooner or later, so pinning version does not work very well.

Anyway, I think it will work again in 123 :)

PR #300, opened 2016-08-22
Update elasticsearch install instructions

mikesea:
This ES healthcheck is prone to failure if ES takes longer than 10 seconds to boot and responds with a connection refused error. Switch to `wget` for the healthcheck to take advantage of the `--retry-connrefused` option, which will treat connection refused errors as transient and retry-able.


felicianotech:

PR #302, opened 2016-08-26
Add note about authenticating w/ http basic

zzak:
Was a sad :panda_face: that I couldn't authenticate using `Authorization` header, such as:

```
curl -H "Authorization: Basic <my-token>" https://circleci.com/api/...
```


DanToml:
LGTM

PR #310, opened 2016-08-31
Update docker.md

gssbzn:
`docker build` now defaults to `--rm=true`, this has problems with intermediate images on circleci
This new option should fix many users having problems using docker cache


felicianotech:
This is a GREAT tip that definitely avoids all those error messages. Thanks a lot @gssbzn.

PR #327, opened 2016-09-16
Ensure that reports are written to a sub-directory under $CIRCLE_TEST_REPORTS

zzak:
This was a common mistake that I made the first few times trying out circle :(

Also made the sub-sections under "automatic test metadata collection" stand out more


eric-hu:
LGTM

PR #329, opened 2016-09-20
Fix issue with parenthesis appearing in Liquid if blocks.

felicianotech:
The combination of the additional parenthesis added during the CCIE Docs merge in addition to updated Jekyll causes Liquid to spit out errors. This removes those unnecessary parenthesis.


ndintenfass:
lgtm

PR #330, opened 2016-09-20
Update Gems.

felicianotech:
Testing in brand new VM (Vagrant). Initially, the updated version of Jekyll exposed some Liquid tag errors but those were fixed in #329.


keybits:
LGTM if tested locally with these gems.

PR #331, opened 2016-09-20
Add DB examples for MySQL, MongoDB, and PostgreSQL.

felicianotech:


jekyll/_docs/configuration.md
```diff
@@ -291,6 +291,9 @@ You can enable other databases and services from the `services` section:
 ```
 machine:
   services:
+    - mysql
+    - mongodb
+    - postgresql
```
z00b: this seems to be in conflict with the paragraph above that says they are running by default. doesn't putting them here try to start them?

felicianotech: Thought that might only apply to Ubuntu 12.04 (old default) but it applies to 14.04 also. :/

felicianotech:
Invalid.

PR #335, opened 2016-09-22
[iOS] Add getting started guide [CIRCLE-975]

DanToml:


felicianotech:
We should probably remove some of the overlap from this guide and https://circleci.com/docs/ios-builds-on-os-x/ then right? One section that jumped out immediately is "code signing".

DanToml:
@felicianotech definitely, got a partial rewrite of that in the works that makes it more of an advanced/FAQ type doc (carrying on with its theme of documenting versions etc)

jekyll/_docs/ios-getting-started.md
```diff
@@ -0,0 +1,184 @@
+---
+layout: classic-docs
+title: Getting started with iOS builds
+short-title: iOS Getting Started
+categories: [mobile-platforms]
+description: Testing iOS applications on CircleCI
+---
+
+To get your build running on CircleCI, you first need to [add your project to CircleCI](). When you've done this, GitHub/Bitbucket will start notifying us of changes to your repository so that we may perform builds.
```
felicianotech: Missing link

jekyll/_docs/ios-getting-started.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### Using your provisioning profile
+
+To use your provisioning profile with your CircleCI builds, you need to commit the `.mobileprovision` file to your repository. Your provisioning profiles will automatically be added to the `circle.keychain` at the start of the build.
+
+## Customising your build
+
+Although our inference will work for many cases, some teams may want to customise their build process to use custom tools or run their own scripts. This is done using the `circle.yml` file.
+
+If you wish to see a more detailed guide to the format, you can take a look at our [configuration sample](https://circleci.com/docs/config-sample/).
```
felicianotech: Should be a relative link.

DanToml:
@felicianotech should be good now

felicianotech:
Some concerns:

We refer to Apple's OS as macOS, which is the new name and fine. However the screenshots (and I assume the live website) still refers to it is OS X, which is inconsistent.

Actually, that's my only concern. Otherwise I like it. :)

felicianotech:
Btw, if we're actually running macOS Sierra now, maybe this should be changed in the other repo instead of this guide?

felicianotech:
The issue still needs to be addressed. Opened an Issue for it (ref).

PR #342, opened 2016-09-26
Save intermediate layer to allow caching

sketchytechky:
Since docker `0.9.0` all intermediate layers is removed by default.

https://github.com/docker/docker/issues/4291

As a result, the suggest patter with `docker save` will not work unless `rm=true` is specified.

This is a PR to address this.


sketchytechky:
Actually scratch this.  This caching method is out-dated probably much better to maintain this via Docker hub.

PR #343, opened 2016-09-26
Fix to `GOPATH` in machine environments

sjlu:
Go commands fail when GOPATH is prefixed with "~"

```
go: GOPATH entry cannot start with shell metacharacter '~': "~/.go_workspace"

go get -t -d -v ./... returned exit code 2

Action failed: go get -t -d -v ./...
```


felicianotech:
Thank you.

PR #345, opened 2016-09-27
add note about keeping sensitive info out of environment variables

afeld:


felicianotech:

PR #347, opened 2016-09-28
Clear up the suggestion to use `--rm=false` with `docker build`.

felicianotech:
This suggestion was added to the first example YAML but was missing from the rest.


keybits:
LGTM

PR #368, opened 2016-10-10
Better regex for every branch except master

jsonmaur:
Here is a better regex to match all branches except the master branch. The current regex in there will match any branch that even contains the word "master", which would not technically be everything but master ;)


eric-hu:
Confirmed.  Used [an online regex tester](https://regex101.com/).  The old regex will accept "develop", but not "master" or "notmaster".

The newer regex will accept "develop" and "notmaster", but not "master".

PR #375, opened 2016-10-17
Add Debugging section to CCIE docs

zzak:
Includes tips for Replicated provided by @kimh

### DIFF
--- a/jekyll/_docs/enterprise/debugging.md
+++ b/jekyll/_docs/enterprise/debugging.md
@@ -34,9 +34,9 @@ Please check if the file exists and if not, please create the file manually. The
 }
 ```
 
-### Restarting replicated and the CircleCI app
+### Restarting Replicated and the CircleCI app
 
-Please try restarting replicated services. You can do this by running the following commands on the service box.
+Please try restarting Replicated services. You can do this by running the following commands on the service box.
 
 ```
 sudo restart replicated-ui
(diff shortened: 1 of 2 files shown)


kimh:
I would say how to restart replicated apps is worth for being exposed in the doc but other things are a bit too ad-hoc to be in our public doc. I had to try these things in the previous debugging session because support bundle was missing many files but it normally should contain all these information (i.g. replicated version, the output of docker ps, etc).

felicianotech:
I made a small comment. So far, syntax wise the doc seems good. I wouldn't know too much on the topic itself.

zzak:
@kimh @felicianotech Updated, thanks for the :eyes:

kimh:
LGTM

jekyll/_docs/enterprise/debugging.md
```diff
@@ -0,0 +1,85 @@
...
+
+```
+sudo restart replicated-ui
+sudo restart replicated
+sudo restart replicated-agent
+```
+
+Then, go to your services box admin (i.e.: https://YOUR-CCIE-INSTALL:8800) and try restarting with "Stop Now" and "Start Now".
+
+### Trying to login replicated
```
ndintenfass: We should capitalize Replicated - here and in the line below

ndintenfass:
Given that this is all about Replicated I'd suggest we make the title Debugging Replicated or just Replicated -- as a general Debugging page this will disappoint, but as a way to get info about Replicated it seems reasonable (please capitalize Replicated, per my inline comments).

zzak:
@ndintenfass Thanks for your feedback!

bellkev:
Discussion of where logs are?

PR #378, opened 2016-10-18
Add note that Redis is not started on Ubuntu 14.04

ligaz:


zzak:
I think this is information is too version specific to go here, what about the other services listed and their possible gotchas? It's better to put this kind of information somewhere it can be easily collected and found.

ligaz:
Good point. Maybe we should delete the whole sentence about which services are started by default then?

felicianotech:
@ligaz For now I'd go with that.

felicianotech:
In the mean time, this PR does make an improvement so I'm going to just merge it.

PR #379, opened 2016-10-18
small markdown fixes, linky links, phrasing clarification

li3n3:
Some markdown readers seem fine with the spaces around `**`, but GH is not one of them! So be it.

Made CircleCI links into...links!


li3n3:
Oh, also — noticed that the `last_updated` is May 1, 2013. Clearly it's had some more recent modifications, though! This date doesn't appear in the rendered version of this doc, but I'm wondering if that needs updating, too (or where that information is used).

ndintenfass:
lgtm

ndintenfass:
For the last_updated, I'd go ahead and bump it - afaict it's not actually used anywhere, so it's just informational to the reader of the code, but better to keep it current than not.

PR #396, opened 2016-10-27
the line pwd: ../firefox-20 needs to be indented

emhopkins:
I had the same issue as this user (https://discuss.circleci.com/t/parallelism-and-bash-case-anybody-understand-this-error-message/1023) when running this (different example, but same problem).  This ran correctly for me when the line pwd: ../firefox-20 was indented two additional spaces.


felicianotech:

PR #400, opened 2016-11-01
Added segment page tracking

JustinC474:
We added all the meta data to the tracking properties, and make the doc's root subcategory "default".

For all other pages the category is `docs` and the subcategory is the url that follows with the leading and trailing slashes removed (ie: `i-am-some-article` or `enterprise/some-enterprise-article`)


annguy3n:
LGTM

trevor-circleci:
@JustinC474 LGTM

PR #402, opened 2016-11-01
[CIRCLE-1738] Initializing user before tracking

annguy3n:
Initializing user before tracking

**Note:** unable to test locally because `localhost:4000/api/v1/me` cannot be resolved. We can potentially set up nginx to route it to `cci.com/api/v1/me` but ¯\\_(ツ)_/¯. It seems running cci.com locally also cannot resolve `localhost:4000/api/v1/me`

### DIFF
--- a/jekyll/assets/js/main.js
+++ b/jekyll/assets/js/main.js
@@ -18,4 +18,10 @@ $( document ).ready(function() {
 	$("article h2, article h3, article h4, article h5, article h6").filter("[id]").hover(function () {
 		$(this).find("i").toggle();
 	});
+
+	$.getJSON("/api/v1/me").done(function (userData) {
+		$(document.body).addClass('loggedin');
+		analytics.identify(userData['analytics_id']);
+		window.userData = userData;
+	});
 });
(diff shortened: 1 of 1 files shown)


jekyll/assets/js/main.js
```diff
@@ -18,4 +18,10 @@ $( document ).ready(function() {
 	$("article h2, article h3, article h4, article h5, article h6").filter("[id]").hover(function () {
 		$(this).find("i").toggle();
 	});
+
+	$.getJSON("/api/v1/me").done(function (userData) {
+		$(document.body).addClass('loggedin');
+		analytics.identify(userData['analytics_id']);
+		window.userData = userData;
```
JustinC474: do we need this part? I know it was copied over from cci.com, but if we don't need it lets remove it

jekyll/assets/js/main.js
```diff
@@ -18,4 +18,10 @@ $( document ).ready(function() {
 	$("article h2, article h3, article h4, article h5, article h6").filter("[id]").hover(function () {
 		$(this).find("i").toggle();
 	});
+
+	$.getJSON("/api/v1/me").done(function (userData) {
+		$(document.body).addClass('loggedin');
```
JustinC474: same with this

PR #408, opened 2016-11-03
Update docs about using npm tokens with CircleCI.

dtinth:
I updated the docs to include the `circle.yml` configuration to authenticate using npm auth tokens.


felicianotech:
Looks good. Thank you for the contribution.

Issue #413, opened 2016-11-07
Provide more info about Android Support (_cci1/build-image-trusty.md)


mpp-doric:
It would good to provide more info around your `android extra` section. Currently there is `m2repository` and `support` listed. 

The `m2repository` is actually the support repo, and listed in the SDK tools as `Android Support Repository` with version `39.0.0`. Trusty has v `31.0.0` which should be listed here also really, as its out of date and generally needs updating. 

Not sure what `support` is referring to, but guess its the old static support files, in which case it can be removed.

doridori:
Corrected Google extras names also - unsure what versions your rocking tho!

ryanwohara:
This problem is fundamentally addressed in CircleCI 2.0.

PR #414, opened 2016-11-07
Update versions.json

doridori:
Correct Android Extras info #413


felicianotech:
Hi,

Thank you for the contribution. Small issue though. This file is actually generated from a different project (https://github.com/circleci/image-builder). We keep everything in-sync but copying that file over during update versus editing it directly.

doridori:
Ah ok - maybe you could patch the issues there then

PR #417, opened 2016-11-07
Fix reference to default Linux image (now 14.04).

felicianotech:
This doc referenced Ubuntu 12.04 as the default Linux image, which is no longer the case.

### DIFF
--- a/jekyll/_docs/build-image-trusty.md
+++ b/jekyll/_docs/build-image-trusty.md
@@ -6,9 +6,7 @@ description: Ubuntu 14.04 (Trusty)
 changefreq: "weekly"
 ---
 
-You can run your Linux builds on the default image, Ubuntu 14.04 (Trusty). You 
-can switch to Trusty from "Project Settings" -> "Build Environment" of your 
-project.
+The default image is Ubuntu 14.04 (Trusty), but you can always change the image directly by going to 'Project Settings' -> 'Build Environment'.
 
 Please note that you need to trigger a build by pushing commits to GitHub or 
 Bitbucket (instead of rebuilding) to apply the new setting.
--- a/jekyll/_docs/build-image-trusty.md
+++ b/jekyll/_docs/build-image-trusty.md
@@ -6,8 +6,9 @@ description: Ubuntu 14.04 (Trusty)
 changefreq: "weekly"
 ---
 
-You can run your Linux builds on Ubuntu 14.04 Trusty (the default). You can 
-switch to Trusty from "Project Settings" -> "Build Environment" of your project.
+You can run your Linux builds on the default image, Ubuntu 14.04 (Trusty). You 
+can switch to Trusty from "Project Settings" -> "Build Environment" of your 
+project.
 
 Please note that you need to trigger a build by pushing commits to GitHub or 
 Bitbucket (instead of rebuilding) to apply the new setting.
--- a/jekyll/_docs/build-image-trusty.md
+++ b/jekyll/_docs/build-image-trusty.md
@@ -3,11 +3,14 @@ layout: classic-docs
 title: Ubuntu 14.04 (Trusty)
 categories: [build-images]
 description: Ubuntu 14.04 (Trusty)
+changefreq: "weekly"
 ---
 
-You can run your Linux builds on Ubuntu 14.04 Trusty (default is Ubuntu 12.04). You can switch to Trusty from "Project Settings" -> "Build Environment" of your project.
+You can run your Linux builds on Ubuntu 14.04 Trusty (the default). You can 
+switch to Trusty from "Project Settings" -> "Build Environment" of your project.
 
-Please note that you need to trigger a build by pushing commits to GitHub or Bitbucket (instead of rebuilding) to apply the new setting.
+Please note that you need to trigger a build by pushing commits to GitHub or 
+Bitbucket (instead of rebuilding) to apply the new setting.
 
 ## Base
 


jekyll/_docs/build-image-trusty.md
```diff
@@ -3,11 +3,14 @@ layout: classic-docs
 title: Ubuntu 14.04 (Trusty)
 categories: [build-images]
 description: Ubuntu 14.04 (Trusty)
+changefreq: "weekly"
```
smart-alek: What is this for?

jekyll/_docs/build-image-trusty.md
```diff
@@ -3,11 +3,14 @@ layout: classic-docs
...
 ---
 
-You can run your Linux builds on Ubuntu 14.04 Trusty (default is Ubuntu 12.04). You can switch to Trusty from "Project Settings" -> "Build Environment" of your project.
+You can run your Linux builds on Ubuntu 14.04 Trusty (the default). You can 
+switch to Trusty from "Project Settings" -> "Build Environment" of your project.
 
-Please note that you need to trigger a build by pushing commits to GitHub or Bitbucket (instead of rebuilding) to apply the new setting.
+Please note that you need to trigger a build by pushing commits to GitHub or 
+Bitbucket (instead of rebuilding) to apply the new setting.
```
smart-alek: Why split across 2 lines?

jekyll/_docs/build-image-trusty.md
```diff
@@ -3,11 +3,14 @@ layout: classic-docs
 title: Ubuntu 14.04 (Trusty)
 categories: [build-images]
 description: Ubuntu 14.04 (Trusty)
+changefreq: "weekly"
 ---
 
-You can run your Linux builds on Ubuntu 14.04 Trusty (default is Ubuntu 12.04). You can switch to Trusty from "Project Settings" -> "Build Environment" of your project.
+You can run your Linux builds on Ubuntu 14.04 Trusty (the default). You can
```
smart-alek: Change to:

You can run your Linux build on the default, Ubuntu 14.04 Trusty. You can switch to Trusty from "Project Settings" -> "Build Environment" of your project.

jekyll/_docs/build-image-trusty.md
felicianotech: This is a signal to the generated sitemap (sitemap.xml for Google) on approximately how often the page might be updated. This in turn hints to search engines like Google how often their search spiders should come back and check the page.

jekyll/_docs/build-image-trusty.md
felicianotech: https://news.ycombinator.com/item?id=5436993

jekyll/_docs/build-image-trusty.md
```diff
@@ -6,8 +6,9 @@ description: Ubuntu 14.04 (Trusty)
 changefreq: "weekly"
 ---
 
-You can run your Linux builds on Ubuntu 14.04 Trusty (the default). You can 
-switch to Trusty from "Project Settings" -> "Build Environment" of your project.
+You can run your Linux builds on the default image, Ubuntu 14.04 (Trusty). You 
+can switch to Trusty from "Project Settings" -> "Build Environment" of your 
+project.
```
felicianotech: Some people may not longer be on Ubuntu 14.04 and want to switch back.

Some people who had an existing project before the switch was made, are still using Ubuntu 12.04 because that was the default. Now they may want to switch to the new default image.

The goal of this PR is to simply remove the reference of Ubuntu 12.04 being the default since that is now false.

smart-alek: Ok, if that's the case, then I think the best way to write this is:

"The default image is Ubuntu 14.04 (Trusty), but you can always change the image directly by going to 'Project Settings' -> 'Build Environment'."

This captures the (new) default and also shows people where to change that image if they need to.

jekyll/_docs/build-image-trusty.md
smart-alek: Right, but isn't that for code, not prose?

jekyll/_docs/build-image-trusty.md
felicianotech: Fixed in newest commit.

jekyll/_docs/build-image-trusty.md
```diff
@@ -3,11 +3,13 @@ layout: classic-docs
...
 categories: [build-images]
 description: Ubuntu 14.04 (Trusty)
+changefreq: "weekly"
 ---
 
-You can run your Linux builds on Ubuntu 14.04 Trusty (default is Ubuntu 12.04). You can switch to Trusty from "Project Settings" -> "Build Environment" of your project.
+The default image is Ubuntu 14.04 (Trusty), but you can always change the image directly by going to 'Project Settings' -> 'Build Environment'.
 
-Please note that you need to trigger a build by pushing commits to GitHub or Bitbucket (instead of rebuilding) to apply the new setting.
+Please note that you need to trigger a build by pushing commits to GitHub or
```
smart-alek: "To apply the new setting, you will need to trigger a build by pushing commits to GitHub or Bitbucket (instead of rebuilding)."

felicianotech: We should just do this in another PR instead of continuing to delay this one. That's outside the point of this PR.

This point of which image is default is incorrect right now.

PR #418, opened 2016-11-08
Update nightly-builds.md with link to alternative serverless.com approach.

godfreyhobbs:
Adding a link to an alternative serverless.com aws lambda based approach.  Let me know what you think.


felicianotech:
I would feel more comfortable linking to your project once it has had some time to mature.

PR #421, opened 2016-11-10
Protip for bundle check || install inferred step for Ruby applications

zzak:


felicianotech:
LGTM

jekyll/_docs/language-ruby-on-rails.md
```diff
@@ -45,6 +45,8 @@ dependencies:
     - bundle exec rake assets:precompile
 ```
 
+The default inferred step for Bundler is `bundle check || bundle install`. For reasons, we use `check` prior to `install` in order to skip unnecessary steps inherit from the `bundle install` command when all depedencies are cached or otherwise already installed and available on the system.
```
smart-alek: Typo: "inherit" should be "inherited"

li3n3: > For reasons

is this sass or just the most succinct way of explaining this?

PR #429, opened 2016-11-10
Fix typo in RoR doc.

felicianotech:


zzak:
Thanks @felicianotech and @smart-alek

PR #431, opened 2016-11-11
Enhanced search results with content snippets

keybits:
This adds an excerpt of the content of the suggested articles based on the full text search results of article content:

image


eric-hu:
@keybits for posterity, can you insert a screenshot?

felicianotech:
This is great. How difficult would it be to split on word vs character count?

keybits:
@felicianotech not sure, just wanted a quick improvement for now.

PR #433, opened 2016-11-11
Add note about activating required python version to deploy with gcloud

zzak:


felicianotech:
Considering WIP until @zzak checks in with @smart-alek & tech confirmed.

zzak:
@felicianotech @smart-alek Updated <3

PR #435, opened 2016-11-15
WIP: Correct default java version and add missings

drazisil:
Changed to java 7 as default, but everything else uses site.data.trusty.versions.summary, which is also wrong so checking to see where that comes from.

### DIFF
--- a/jekyll/_docs/build-image-trusty.md
+++ b/jekyll/_docs/build-image-trusty.md
@@ -74,15 +74,21 @@ actually installed correctly, we hardcode versions here.
 
 ### Java
 
-Default: `oraclejdk8`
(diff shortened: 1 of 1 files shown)


smart-alek:
@drazisil can you create a follow-up issue for site.data.trusty.versions.summary?

felicianotech:
@drazisil This is what happens.

`site.data.trusty.versions.summary` is Jekyll accessing the datafile `jekyll/_data/trusty/versions.json`. The file `versions.json` isn't suppose to be edited directly here but instead is created as an artifact when https://github.com/circleci/image-builder builds. That file should be copied over whenever a new image is deployed, which Kim usually handles. I'm not sure what his process is for that however.

PR #439, opened 2016-11-20
Go: document when symlinks in $GOPATH are created

dotdoom:


felicianotech:
Thank you for the contribution.

PR #444, opened 2016-11-22
Add Yarn doc.

felicianotech:


jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,92 @@
...
+layout: classic-docs
+title: "Install and Use Yarn (the NPM replacement) on CircleCI"
+short-title: "Install and Use Yarn on CircleCI"
+categories: [how-to]
+description: "How to modify `circle.yml` in order to install and use Yarn on CircleCI."
+---
+
+<img src="{{site.baseurl}}/assets/img/logos/yarn-logo.svg" style="display:block;margin:15px auto;width:40%;min-width:320px;" alt="Yarn Logo" />
+
+[Yarn][yarn-site] is an open-source package manager for JavaScript. Both Yarn itself as well as the packages it installs can be cached on CircleCI. This can potentially speed up builds but more importantly can remove network connectivity as a source of problems.
```
smart-alek: [Yarn][yarn-site] is an open-source package manager for JavaScript. CircleCI can cache both Yarn and its dependencies. This can potentially speed up builds but, more importantly, can reduce errors related to network connectivity."

smart-alek: I don't know enough about Yarn to know if this is true, but it's a pretty strong statement to say that Yarn removes network connectivity _entirely_ as a potential problem source. If that really is true, I would love more explanation on why that's the case.

jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,92 @@
...
+## Setup
+
+```
+machine:
+  environment:
+    YARN_VERSION: 0.17.8
+    PATH: "${PATH}:${HOME}/.yarn/bin"
+```
+
+The Yarn install script seen later can be used without a version. That will download the latest stable version of Yarn. Here a Yarn version is set in order to reduce the number of variables in a build.
```
smart-alek: I don't think it makes sense to mention a script that the reader hasn't seen yet. Better might be:

"We've set the Yarn version here to reduce the number of variables in the build, but you don't need to specify a version when actually installing Yarn; the default setting is the latest stable version."

jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,92 @@
...
+
+`~/.yarn/bin/` is added to the PATH as well. The Yarn install script also builds the path however in the scenario where we're running from cache, this is needed.
```
smart-alek: Although the Yarn install script builds the path, when we're running from the cache, we have to explicitly add `~/.yarn/bin/` to the PATH.

jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,92 @@
...
+    - |
+      if [[ ! -e ~/.yarn/bin/yarn || $(yarn --version) != "${YARN_VERSION}" ]]; then
+        curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version $YARN_VERSION
+      fi
+  cache_directories:
+    - ~/.yarn
+    - ~/.yarn-cache
+```
+
+Here in the `dependencies` section, Yarn is installed via an install script if and only if:
```
smart-alek: Here, the script installs Yarn if and only if:

jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,92 @@
...
+```
+dependencies:
+  override:
+    - yarn install
+test:
+  override:
+    - yarn test
+```
+
+When CircleCI detects a JavaScript based project, certain commands can be run such as `npm install` or `npm test`. To use Yarn instead, we override both the `dependencies` and `test` section.
```
smart-alek: When CircleCI detects a JavaScript project, certain commands (like `npm install` or `npm test`) might be run. To use Yarn instead of `npm`, we override both the `dependencies` and `test` sections.

jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Full Example
+
+Here is the full example `circle.yml` file:
```
smart-alek: Remove this line since the header already tells us what this is.

jekyll/_docs/install-and-use-yarn.md
felicianotech: Caching is what removes the network as an issue, not Yarn.

felicianotech:
I like your changes. Will update when they are applied.

felicianotech:
@smart-alek Comments addressed.

jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,88 @@
...
+## Setup
+
+```
+machine:
+  environment:
+    YARN_VERSION: 0.17.8
+    PATH: "${PATH}:${HOME}/.yarn/bin"
+```
+
+We've set the Yarn version here to reduce the number of moving parts in the build. Although the Yarn install script sets its `PATH`, when we're potentially running from cache, we have to explicitly add `~/.yarn/bin/`.
```
smart-alek: That last sentence still doesn't read exactly right to me. Right now, it's not super clear that we *are* running from cache. The phrase "when we're potentially running from cache" makes it sound like that's another case that *could* happen but isn't.

How about:

"Although the Yarn install script sets its `PATH`, running from cache requires adding `~/.yarn/bin/` explicitly."

jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,88 @@
...
+    - |
+      if [[ ! -e ~/.yarn/bin/yarn || $(yarn --version) != "${YARN_VERSION}" ]]; then
+        curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version $YARN_VERSION
+      fi
+  cache_directories:
+    - ~/.yarn
+    - ~/.yarn-cache
+```
+
+Here, the Yarn install script is run if and only if:
```
smart-alek: I can't remember if I suggested changing this line, but...I think the following reads better.

"Here, the Yarn install script runs if and only if:"

jekyll/_docs/install-and-use-yarn.md
```diff
@@ -0,0 +1,88 @@
...
+
+* Yarn isn't already installed.
+* The installed version of Yarn doesn't match the specified version in the machine section.
+
+`~/.yarn` is specified as a cached directory for Yarn itself. `~/.yarn-cache` is specific as a cached directory for Yarn packages.
```
smart-alek: Because these two sentences are so similar, I would bold the differences:

"**`~/.yarn`** is specified as a cache directory for **Yarn itself**. **`~/.yarn-cache`** is specified as a cache directory for **Yarn packages**."

jekyll/_docs/install-and-use-yarn.md
felicianotech: There is times where we wouldn't be running from cache. Your sentence still sounds better though.

Daniel15:
From the perspective of the Yarn team, we prefer the Debian packages rather than the installation script where possible. Is it possible to use that instead?

```yaml
dependencies:
  pre:
    # Install Yarn
    - sudo apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg
    - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
    - sudo apt-get update -qq
    - sudo apt-get install -y -qq yarn=0.17.10-1
```

Daniel15:
Ohh I see, you actually cache Yarn itself. That's a very interesting concept!

felicianotech:
@Daniel15 I'd have no problem uses packages if that's what the team prefers. My only two concerns for CircleCI customers would be speed and caching. The script we're using now (which I assumed was okay since it's within the official Yarn codebase) allows us to:

- cache Yarn itself since some teams might build 100 times a day. Downloading Yarn each and every single time is a waste of resources and time.
- avoid using apt-get. The script is slightly faster than having to run `apt-get update && apt-get install yarn`.

To be fair, I didn't actually measure that last point though. Any thoughts on how we can continue to do this another way? Is there a `.deb` we could simply use, cache, and then `dpkg -i` in a build? That should be faster than `apt-get`.

Daniel15:
Yeah, good points. Using the install script is totally fine,  no issues
with that. Your reasoning makes sense! :)

If you'd like to try using the Debian package directly, you can use a URL
like https://yarnpkg.com/downloads/0.17.10/yarn_0.17.10_all.deb. Also, if
you end up preinstalling Yarn on the build worker images (which would be
awesome), the Debian package would probably work best for that.

Does CircleCI use a caching proxy like apt-cacher-ng for Debian packages? I
imagine that'd help with perf of apt-get

felicianotech:
"Does CircleCI use a caching proxy like apt-cacher-ng for Debian packages?" Not that I'm aware of no. We're actually moving away from pre-installing anything with our new platform. We'll be moving to containers that people can specify themselves in the YAML file. Having Yarn pre-installed in the NodeJS Docker image for example would then be more useful in the future.

PR #445, opened 2016-11-22
Comment out extra builds we're not currently using.

felicianotech:


felicianotech:
My problem with that is we forget they exists. How often do we look through Git history for something without a note or is commented out?

smart-alek:
Maybe just a philosophical difference, then. I'm just more in favor of committing code to VCS if it's actually being run.

felicianotech:
So this doesn't get forgotten completely: https://github.com/circleci/circleci-docs/issues/451

PR #453, opened 2016-11-28
[WIP] First Draft. Adding advanced configuration page. 

GERey:


jekyll/_docs/enterprise/advanced-configuration.md
```diff
@@ -0,0 +1,224 @@
...
+layout: enterprise
+title: Advanced Configuration
+category: [resources]
+order: 0
+description: "Advance Configurations"
+hide: false
+---
+
+
+# Advanced Configuration
```
GERey: Made a test comment

jekyll/_docs/enterprise/advanced-configuration.md
```diff
@@ -0,0 +1,224 @@
...
+
+
+### Getting Started
+
+For the purposes of this demo and to follow the examples, we assume that we have an unauthenticated http proxy at `10.0.0.33:3128` and the services box is at `10.0.1.238`. We use `ghe.example.com` as our github enterprise host.
+
+In an ideal case, traffic to S3 will not be proxied, and instead be added to the NO_PROXY rules (to bypass it out add `s3.amazonaws.com,*.s3.amazonaws.com`  to `NO_PROXY`)
+
+Everything here must be done **before** installing CircleCI on fresh vm's/instances.
+
```
GERey: So far it applies to everything in Advanced config. I can move that closer to the top but I don't know that it will stay true in the long run.

nholthaus: It's not clear to me if this applies just to amazon proxy settings or also installations that are behind a corporate proxy. In the corporate case, adding s3 to `NO_PROXY` probably won't work since all internet traffic is typically routed through the proxy, or not at all.

If it's only amazon proxies, you may want to add corporate proxy install instructions as well. Working around the lack of proxy support in some of the scripts was the hardest part of the install process for me.

jekyll/_docs/enterprise/advanced-configuration.md
```diff
@@ -0,0 +1,224 @@
...
+```
+$ curl -o ./init-services.sh https://s3.amazonaws.com/circleci-enterprise/init-services.sh
+$ sudo -E bash init-services.sh
+```
+
+Note that this differs from the regular installation instructions by including the `-E` command with `sudo`. This allows the script to have access to the enviroment variables you exported above.
+
+Also note that when the instructions ask you if you use a proxy, they will also prompt you for the address. It is **very important** that you input the proxy in the following format `<protocol>://<ip>:<port>`. If you are missing any part of that, then apt-get won't work correctly and the packages won't download. 
+
+If for any reason you can't access the page the CircleCI replicated console, but the services machine seems to be running, try to ssh tunnel into the machine doing the following: `ssh -L 8800:<address you want to proxy through>:8800 ubuntu@<ip_of_services_box>`.
```
GERey: I'll switch them to placeholder ip's . I feel more comfortable having that, and I think it would be clearer

jekyll/_docs/enterprise/advanced-configuration.md
```diff
@@ -0,0 +1,224 @@
...
+    SERVICES_PRIVATE_IP=10.0.1.238 bash
+```
+
+### Installing CircleCI Builder
+This really depends on what type of distro you are installing CircleCI on, but the above instructions cover a basic kind of install.
+
+
+##GPU Install Instructions 
+
+This is only supported in our newest release ( 1.47.0 +). If you need help upgraging, please reachout to enterprise-support@circleci.com
```
GERey: Those changes will get pulled into Kevin's.

jekyll/_docs/enterprise/advanced-configuration.md
```diff
@@ -0,0 +1,224 @@
...
+
+First
```
GERey: this still needs to be addressed in this revision. I just wanted to commit what I had in the moment to do a less bulky change. As a note these instructions will also be changing to include GPU Docker support.

felicianotech:
@GERey Is this still a WIP?

PR #460, opened 2016-12-02
Simplify the overview, seperate the two different types of installs a…

GERey:
…nd initial readover to make instructions specific to the quirks of each install

### DIFF
--- a/jekyll/_docs/enterprise/docker-install.md
+++ b/jekyll/_docs/enterprise/docker-install.md
@@ -1,6 +1,6 @@
 ---
 layout: enterprise
-title: "Installing CircleCI"
+title: "Non-AWS"
 category: [installation]
 order: 3
 description: "How to install CircleCI Enterprise on any generic machine"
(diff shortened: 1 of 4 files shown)


bellkev:
"for environments not on AWS" ... "For other environments, contact us at <enterprise@circleci.com> for a guided installation. " Is a bit weird... Presumably we've covered everything with AWS/non-AWS?

jekyll/_docs/enterprise/docker-install.md
```diff
@@ -0,0 +1,121 @@
+---
+layout: enterprise
+title: "Installing CircleCI"
```
bellkev: Feels like this should be "non-AWS" or something, because as-is this will look like:

Installation
- AWS
- Installing CircleCI

Which seems weird.

GERey:
I agree with that. In the new PR we should talk about what those alternatives would be. I like the idea of AWS and `Generic Cloud Provider`.

PR #476, opened 2016-12-13
Adding a spelling fix and rewording a part to include further directions

GERey:


GERey:
Thanks @felicianotech !!

PR #479, opened 2016-12-14
Remove vague question/answers

smart-alek:
* bad question does not require answering


felicianotech:
Are you intending to delete this page? If so, do we want to redirect people somewhere? Last 30 days alone has 375 pageviews for this page:

docs-unusual-traffic

smart-alek:
@felicianotech Yep, intending to delete this. This question isn't specific enough for a developer to ask it. I believe people are clicking on it because it's listed in the Getting Started section, and this doc isn't linked to from anywhere else in the docs.

felicianotech:
@smart-alek With that being said, I took a look at how people are landing on the page and 92% of the traffic is coming from previous pages, which supports what you're saying.

PR #488, opened 2016-12-22
replace cache/yarn with yarn-cache

jasonLaster:
I believe this is more consistent with the description at the top for `~/.yarn-cache`.


felicianotech:
Thank you for this however the fix was actually the other way around. Where `~/.yarn-cache` is referenced at the top, that should have been `~/.cache/yarn`.

The referenced PR fixes this.

PR #495, opened 2016-12-28
Document implicit behavior caused by rvm on the platform

zzak:


zzak:
Also fixes a broken link caused by https://github.com/circleci/circleci-docs/commit/95da75915bee45717dcf850cf951b6d0b9186e05#diff-d0477d8327cc6bd2f42bdd0b33902230

felicianotech:
- The first time RVM is mentioned, can we link to the project?
- This isn't a blocker for me but more of a philosophy question: If a Discuss post is important enough that we'd want to link to it from Docs, is that a sign that that post should probably also be in Docs? I feel like link a Doc in Discuss makes sense but the other way round, I'm not too sure. @smart-alek, thoughts?

felicianotech:
@smart-alek Thoughts? Otherwise I'm just going to approve.

jekyll/_docs/language-ruby-on-rails.md
```diff
@@ -25,12 +25,24 @@ If you don't want to use the default, you can specify your version in `circle.ym
...
 machine:
   ruby:
-    version: rbx-2.2.6
+    version: 2.4.0
 ```
 
-Our [test environment doc]({{site.baseurl}}/language-ruby-on-rails/)
-covers more details about language versions and tools; it also explains how Circle
-works with testing tools that require a browser.
+In the current version of CircleCI, we use `rvm` to manage and install different versions of Ruby. This has a couple of side-effects that users should be aware of.
```
smart-alek: Suggested:

"In the current version of CircleCI, we use `rvm` to manage and install different versions of Ruby. This has a couple side-effects:"

jekyll/_docs/language-ruby-on-rails.md
```diff
@@ -25,12 +25,24 @@ If you don't want to use the default, you can specify your version in `circle.ym
...
+
+First, if the version specified in `circle.yml` isn't available then rvm will try to install it for you. If RVM cannot find a pre-compiled binary from the available mirrors, then it will download and compiled from source. Since, fetching the binary each time from a number of possible mirrors can often fail due to network instability; we highly recommend to cache this file. Please see the following guide on discuss in order to [cache a pre-compiled Ruby binary](https://discuss.circleci.com/t/caching-compiled-rubies-with-rvm/3636) and improve build performance and reliability.
```
smart-alek: Suggested:

"First, if the version specified in `circle.yml` isn't available, then rvm will try to install it. If RVM cannot find a pre-compiled binary from available mirrors, it will download and compile from source. Since fetching a binary from several possible mirrors is prone to failure caused by unstable networks, we highly recommend caching this file. To do this, please read our Discuss guide on how to [cache a pre-compiled Ruby binary](https://discuss.circleci.com/t/caching-compiled-rubies-with-rvm/3636) to improve build performance/reliability."

jekyll/_docs/language-ruby-on-rails.md
```diff
@@ -25,12 +25,24 @@ If you don't want to use the default, you can specify your version in `circle.ym
...
+
+Second, whenever the build changes into your project directory, RVM will try to read either `.rvmrc` or `.ruby-version` via the `cd` auto hook. In order to avoid this, and the aforementioned behavior, we recommend to remove these files by doing:
```
smart-alek: What is "the aforementioned behavior"? I already forgot. 😬 

"Second, whenever the build changes into your project directory, RVM will try to read either `.rvmrc` or `.ruby-version` via the `cd` auto hook. To avoid this and the aforementioned behavior, we recommend removing these files by running:"

jekyll/_docs/language-ruby-on-rails.md
```diff
@@ -25,12 +25,24 @@ If you don't want to use the default, you can specify your version in `circle.ym
...
+
+```
+rm $CIRCLE_PROJECT_REPONAME/.ruby-version
+```
+
+Last, you can configure RVM to not install versions of Ruby that don't exist on the system by doing:
```
smart-alek: Using "Last" here makes this sound like one of the side-effects, when really this is a precautionary thing you can do to prevent RVM from being too aggressive. Try:

"If RVM insists on installing certain versions, you can configure RVM to not install Ruby versions that don't exist on the system:"

smart-alek:
@felicianotech in regards to your question about Discuss, I think you're right about it needing to belong in the docs itself. I've been thinking a lot about what the criteria for putting something in Discuss should even be, and IMO a _guide_ belongs in documentation.

This might sound too simple, but I believe that if something is open for discussion, it belongs in discuss. So open questions, troubleshooting, feature requests. Everything else is being presented as a source of truth and belongs in our docs.

felicianotech:
@smart-alek I agree with what you said, with the addition of tiny "guides" which are somewhat temporary. Workarounds for example.

zzak:
@felicianotech @smart-alek I've made the requested changes, if you want I can include the discuss post as a new doc in this PR.. or submit it later. Which would you prefer?

smart-alek:
@zzak submit later works great for me.

zzak:
@smart-alek Thank you for the review!

PR #507, opened 2017-01-09
CIRCLE-2488 Doc for Build images 2.0

kimh:
Build image that customers can use in 2.0 is black box, so we'll document what's actually installed in each container. Whenever we update images, we will add a new page to .`/docs/build-images-2-0/` but won't delete old ones because customers can use any image in 2.0.

As a starter, I'll do with `ubuntu-14.04-XL` which is the most frequently uses image that we maintain. 

I haven't put `/docs/build-images-2-0/` in sidebar/navigation to hide the page for now because we are still in Closed Beta.


kimh:
@eric-hu that's correct. I'll make it public and easily noticeable once we are on Open Beta.

notnoopci:
Worth pointing out that this line of heavy container images are essentially deprecated in 2.0.  We will be moving to off the shelf images from dockerhub supplemented with many small Convenience Images model, e.g. images specifically for language/stack (e.g. an image go, another one for ruby 2.3) without needing RVM/rbenv/pyenv/phpenv/etc  and ability to upgrade a language stack without the other.  It's an explicit goal to move away from a single image to rule them all with all the pain we've had for years with managing them -- it'll be quite a shame to bring that image into CircleCI 2.0.

kimh:
I thought that decision was on hold although I'm very happy to get rid of the huge image. So can we just remove our mentioning of the image in our doc at all?

I wish I could know this before working on CIRCLE-2488 because I wasted time fighting with jekyll and a few changes in image-builder.

eric-hu:
@kimh your work on this was not wasted.  I've spent some time onboarding a couple of bigger customer projects.  This image will be necessary if we expect large customers to transition to 2.0.

One customer needed nodejs side by side with Ruby, along with custom postgres extensions, and a browser for integration testing.  I actually implemented it both ways--with the monolithic-image and with community DockerHub images + custom installs. (I hit a filesystem error, but didn't recognize it at the time).  In all, I spent a week doing this.

For a customer converting from 1.0 to 2.0, this image saves a lot of work that can be put off until later.  I agree it would be good for us in the long term to cut demand for these images.  Having tried a 2.0 conversion both ways, this monolithic image gives our larger customers a much, much easier transition into 2.0.

felicianotech:
I agree with Eric. While hopefully we won't need to use this work for long, currently, it's going to be a great resource for converting customers.

jekyll/_docs/build-image-ubuntu-14.04-XL-922-9410082.md
```diff
@@ -6,6 +6,11 @@ description: ubuntu-14.04-XL-922-9410082
 changefreq: "weekly"
 ---
 
+***Please note that we are planning to make the image depreciated on CircleCI 2.0 in the future.***
+
+The idea of monolithic build image doesn't fix the idea of CircleCI 2.0 quite well and we are currently
```
eric-hu: > depreciated

Do you mean deprecated?

> doesn't fix the idea of CircleCI 2.0 

@kimh Do you mean "doesn't fit the idea of CircleCI 2.0"?

kimh: You are right :( Thanks for pointing out. I'll fix it.

PR #508, opened 2017-01-09
Reverting to previous script

GERey:


ndintenfass:
This is sad-making, but seems OK for what it is.

PR #520, opened 2017-01-16
Because circleci-android23 image is gone, replace references with cir…

ngriffiths:
…cleci-android24. Also, handle the fact that `-no-audio` isn't a valid parameter (nor is `-noaudio`): http://stackoverflow.com/questions/40272884/running-android-emulator-with-noaudio-option-returns-qemu-system-i386-exe-au


felicianotech:
Thank you @ngriffiths.

PR #526, opened 2017-01-18
Add new Yarn doc.

### LINKED ISSUE #487: Update Yarn Doc to Reference Pre-installed Yarn
The next Ubuntu 14.04 image should include this merged PR: https://github.com/circleci/image-builder/pull/120

When that happens, using Yarn can be even simpler. The doc should then be updated.

felicianotech:
Fixes #487.


jekyll/_docs/yarn.md
```diff
@@ -0,0 +1,62 @@
+---
+layout: classic-docs
+title: "Using Yarn (the npm replacement) on CircleCI"
```
smart-alek: Why define Yarn here? I'm assuming people are already searching for Yarn and know what it is. Change to: "Using Yarn on CircleCI"

jekyll/_docs/yarn.md
```diff
@@ -0,0 +1,62 @@
...
+```yaml
+dependencies:
+  override:
+    - yarn
+test:
+  override:
+    - yarn test
+```
+
+When CircleCI detects a JavaScript project, certain commands (like `npm install` or `npm test`) might be run. To use Yarn instead of npm, we override both the `dependencies` and `test` sections.
```
smart-alek: I'd rather have the description before the code, but maybe that's just my preference. Otherwise, the code just sort of appears without any context.

jekyll/_docs/yarn.md
felicianotech: Was thinking of SEO.

jekyll/_docs/yarn.md
felicianotech: ??? I'm pretty sure we had this exact same convo before and you said the reverse, which is why I did it this way. :p

I can reverse them.

PR #553, opened 2017-01-24
Show how to call gym via fastlane in the iOS docs

appplemac:
Invoking `fastlane` directly is no longer recommended, we should suggest using `fastlane gym` everywhere instead.


marcomorain:
🚀

PR #554, opened 2017-01-25
Changed away from using init-services

GERey:


felicianotech:
@GERey bump :)

GERey:
ah yeah. I just need other enterprise folks to give me a heads up as well as it'll affect them too :)

PR #555, opened 2017-01-25
Add discuss guide for caching pre-compiled Ruby binaries with RVM

zzak:
/cc #495


zzak:
Thank you @felicianotech!

PR #557, opened 2017-01-25
Add in :vcs-type to examples

jxc:
I had some difficulty following the examples until I found a slight difference between the current API docs ([ref](https://github.com/circleci/circleci-docs/blob/master/jekyll/_docs/api.md)) and the example code. This PR is aimed at making the examples valid again.

Thanks for the consideration!


felicianotech:
I also updated the API version number for those two calls as well. Thank you for catching this.

Issue #543, opened 2017-01-23
RE: jekyll/_docs/yarn.md


callumlocke:
The [Circle CI page on using yarn](https://circleci.com/docs/yarn/) says:

> Yarn is pre-installed on CircleCI

So I configured Circle to use `yarn` to install dependencies, but I get this:

![image](https://cloud.githubusercontent.com/assets/250617/22203386/c1127a1a-e164-11e6-95da-afa03ca01730.png)

Here is my full `circle.yml` config:

```yml
machine:
  timezone: Europe/London
  node:
    version: stable
  environment:
    PATH: "${PATH}:${HOME}/${CIRCLE_PROJECT_REPONAME}/node_modules/.bin"

dependencies:
  cache_directories:
    - ~/.cache/yarn
    - bower_components
    - node_modules
  override:
    - yarn

test:
  override:
    - yarn test
  post:
    - yarn run preflight

deployment:
  s3: # this is just a custom name, could be anything
    branch: /.*/
    commands:
      - yarn run deploy:confirm

general:
  artifacts:
    - dist
```

felicianotech:
Which build image are you using? 14.04 or 12.04?

craigteegarden:
I had the same issue and fixed it by updating the project settings to use the 14.04 base image instead of the 12.04 base image.

I proposed an update to the docs here: https://github.com/circleci/circleci-docs/pull/567

PR #567, opened 2017-01-28
Update yarn.md to clarify yarn is only available in the 14.04 base image

craigteegarden:
The current document does not mention that yarn is only available in the 14.04 base image and is not present in the 12.04 base image. This causes an error (see issue referenced below) when following the instructions on this page with a project configured to use the 12.04 base image.

Addresses the issue in https://github.com/circleci/circleci-docs/issues/543


felicianotech:
There's a change coming to this doc (Yarn, and others) shortly to address this. I'll reference back to this PR when that happens (within 48 hours).

PR #571, opened 2017-01-31
Add OS version matrix widget.

### LINKED ISSUE #549: Create Software Version/OS Widget
First docs on specific software, such as Docker or Yarn, a small widget/table to clarify what versions are available on Ubuntu 14.04, Ubuntu 12.04, and OS X would clarify some many docs.

Would also help the following issues:

- #543 
- #542

felicianotech:
Fixes #549.

Added a "widget" to display which OS and version a particular software in on. This is useful for many docs but I noticed particular need with the Yarn doc. MANY people have tried to follow the instructions in Ubuntu 12.04 (Yarn isn't pre-installed there) or on OS X (not here either).

I made this a "widget" versus simply text so that it stands out more from the text and draws the user's attention. Also, in the future, it'll be easier to add software where we have multiple versions available, such as PHP, Python, or NodeJS.

The widget is responsive so works on desktop or mobile. The `WIDGETS.md` explains how to use this.

os-matrix-screenshot-desktop
os-matrix-screenshot-mobile


felicianotech:
Great, I caught an unnecessary period after your edits, but otherwise everything looks good.

PR #585, opened 2017-02-03
Update Docker Postgres example.

### LINKED ISSUE #512: RE: jekyll/_docs/docker.md
The section titled "Connecting to services outside of the container" has a code snippet that uses Postgres v9.4, which is no longer present. Both v9.5 and v9.6 seem to be on there, so it's not clear which one should be used. It would be nice if there was something a little handier to use here. I ended up solving this problem by copying to both the 9.5 and 9.6 versions since I wasn't sure which one would get used.

felicianotech:
Fixes #512.


felicianotech:
I don't see why not but I'll wait to get this test over the weekend first.

felicianotech:
These changes are actually included on our Posgres config already, in the image. People are following this example as if it's needed, which it's not. I'm going to remove it.

PR #594, opened 2017-02-06
clean up phrasing and other small edits

li3n3:
Lots of tiny little things.

### DIFF
--- a/jekyll/_docs/how-cache-works.md
+++ b/jekyll/_docs/how-cache-works.md
@@ -40,7 +40,17 @@ dependencies:
 Cache is saved **after** the `dependencies: override` phase and **before** the
 `test` phase. This means that the directory (or directories) you'd like to
 cache must exist after `dependencies: override`, so if the desired cache
-directory hasn't been created automatically, you'll need to create it manually.
+directory hasn't been created automatically, you'll need to create it manually,
+like this:
+
+```yaml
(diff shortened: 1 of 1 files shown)


li3n3:
@keybits rightly noted that it isn't currently clear that one must make the directory before it gets to that section (which runs immediately after dependencies > override). Have attempted to explain a little better, but would welcome another set of eyes!

jekyll/_docs/how-cache-works.md
```diff
@@ -37,7 +37,10 @@ dependencies:
     - "~/custom_2" # relative to the user's home directory
 ```
 
-Cache is saved **after** the `dependencies` phase and **before** the `test` phase.
+Cache is saved **after** the `dependencies: override` phase and **before** the
+`test` phase. This means that the directory (or directories) you'd like to
+cache must exist after `dependencies: override`, so if the desired cache
+directory hasn't been created automatically, you'll need to create it manually.
```
keybits: Suggest adding an example here - e.g:

```
dependencies:
  override:
    - mkdir ~/my_cache_dir
    # now add files and directories to the above directory
  cache_directories: 
    - "~/my_cache_dir
```

li3n3: so into this. Updated!

PR #595, opened 2017-02-06
Added missing close quote at line 52

ksbeattie:
The rendering on the main site was wrong (at least in my Firefox browser)

### DIFF
--- a/jekyll/_docs/how-cache-works.md
+++ b/jekyll/_docs/how-cache-works.md
@@ -49,7 +49,7 @@ dependencies:
     - mkdir ~/my_cache_dir
     # now add files and directories to the above directory
   cache_directories:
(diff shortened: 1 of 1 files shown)


felicianotech:
I'm not sure what you meant by:

> the rendering on the main site was wrong

but thank you for the fix.

ksbeattie:
Hmm, looks like that didn't fix it.

Here's a snippet of what [that page](https://circleci.com/docs/how-cache-works/) looks like to me:

how-cache-works

I figured it was that missing that quote that messed up the MD rendering, but that doesn't appear to be the case...

felicianotech:
Oooo okay I see what you mean now. No worries, I'll figure this out shortly and submit a fix. Thank you for the screenshot.

PR #625, opened 2017-02-10
Algolia docsearch

### LINKED ISSUE #617: Implement a New Search System

### LINKED ISSUE #618: CCIE Docs & Search
1) CCIE don't appear in internal search
2) The search bar on CCIE docs is basically useless ATM

This issue should be fixed with #617.

trevor-circleci:
@keybits This should do it!

Fixes #617 
Fixes #618


felicianotech:
I pushed a PR to remove some HTML and CSS for the old search results. Also, I turned debug mode on for local dev.

A couple of questions:

- are we not using the Jekyll plugin version then?
- the search results link to the live website in local dev? How can we fix this? This will make local testing difficult in the future.

keybits:
@felicianotech Thanks for the updates.

For now we are using Algolia Docsearch based on discussions with Algolia team. Their Jekyll plugin is being rewritten for a V1 release so we may revisit this in the future.

I don't think it's possible to have relative links for local testing since the result service is based on an index stored in Algolia. The index is regenerated at least every 24 hours. A benefit of having full results is the search can be used on other sites / pages. This allows us to do some useful things from within our app and during support requests. Is there a need to test search in a local only environment?

This is an initial pass and I have some incremental improvements planned. For now I believe it matches and exceeds the capabilities of the old search, so if we're happy I think this can be merged.

felicianotech:
That really sucks to not have search work right in local dev. That's my only complaint. :'(

PR #632, opened 2017-02-11
Add Beta CircleCI 2.0 Docs

felicianotech:
Adds the 2.0 docs section, a new Jekyll Collection, to the site. Doesn't list the section in the dropdown for now.

I've reviewed. Alek?

### DIFF
--- a/jekyll/_config_staging.yml
+++ b/jekyll/_config_staging.yml
@@ -1,1 +1,1 @@
-url: "https://static-staging.circleci.com"
+url: "https://cci.feliciano.tech"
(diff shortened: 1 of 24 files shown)


jekyll/_config_staging.yml
```diff
@@ -1 +1 @@
-url: "https://static-staging.circleci.com"
+url: "https://cci.feliciano.tech"
```
smart-alek: do we even need a staging URL now? what exactly is this for?

felicianotech: Staging is removed, at this point, this is just a string. Does nothing. Do you want me to keep the previous string instead? Neither does anything but I can do a final commit here if you want.

smart-alek: yeah I'd just leave this out if it doesn't do anything. :)

PR #634, opened 2017-02-11
Add Ava to the custom runner example

adieuadieu:
Adds example of how to collect test metadata with Ava

Prompted by https://twitter.com/circleci/status/830111551050829825


adieuadieu:
No problem! 👍

PR #649, opened 2017-02-16
Notify that we aren't accepting new docs right now.

felicianotech:


smart-alek:
@felicianotech Should this maybe go in our README? Or maybe update the footer?

felicianotech:
Not sure. I don't think it's necessary. We don't get whole new docs often as it is. If we did it at all, then I guess he README, I don't like the idea of it in the footer. However, I linked the contributing guide, which is a GitHub/FOSS standard, so would think that would be enough. Up to you.

PR #650, opened 2017-02-16
2.0 add faq

smart-alek:
This is the FAQ from Discuss, but heavily revised to keep it from being too "marketing-esque".

For example, I removed questions like "Why doesn't Build with SSH work?" because the answer is "It isn't implemented yet, sorry!"

### DIFF
--- a/jekyll/_cci2/faq.md
+++ b/jekyll/_cci2/faq.md
@@ -11,6 +11,12 @@ No, CircleCI 2.0 does not infer anything from your project.
 
 This means that no commands are run automatically. Don’t assume that a particular library/package is available unless it is bundled with the Docker image you are using!
 
+## My project is running on CircleCI 2.0, but the build is frozen!
+
+Builds often freeze due to syntax errors in `circle.yml`.
+
+Cancel the build, check your `circle.yml` for proper indents, and ensure that all stages/steps have the required keys.
+
 ## Can I use CircleCI 2.0 without creating base images?
 
 Yes, you can use* one of ours!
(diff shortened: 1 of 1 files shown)


felicianotech:
@smart-alek Looks like you're still pushing commits. Should I wait?

smart-alek:
nah you're all good, should be done.

jekyll/_cci2/faq.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Can I use CircleCI 2.0 without creating base images?
+
+Yes, you can use* one of ours!
+
+The `circleci/build-image:ubuntu-14.04-XL-922-9410082` image has the same content as the Ubuntu Trusty 14.04 image our web app uses. Just know that the image is fairly large (around 17.5 GB uncompressed), so it’s less ideal for local testing.
+
+The image defaults to running actions as the `ubuntu` user and is designed to work with network services provided by Docker Compose.
+
+Here’s a [list of languages and tools](https://circleci.com/docs/build-images-2-0/) included in the image.
```
felicianotech: This should be changed to `[list of languages and tools]({{site.baseurl}}/docs/build-images-2-0/)`.

jekyll/_cci2/faq.md
```diff
@@ -0,0 +1,68 @@
...
+
+For public images on Docker Hub, you can pull the image by prefixing the account/team username:
+
+```
+myUsername/couchdb:1.6.1
+```
+
+## Can I use the `latest` tag when specifying image versions?
+
+You _can_, but please don’t. The `latest` tag is mutable and doesn’t always refer to a consistent image SHA. CircleCI runs on a fleet of machines, so a mutable tag could be improperly cached. In that case, re-running a build wouldn’t be deterministic.
```
felicianotech: We don't need to do this now, but I think this can be better worded. In the larger Docker/DevOps community, it's common for `latest` to not be the same hash. It's the point and why people use it that way. Not using it on CircleCI is more about reducing variables to make builds more deterministic.

smart-alek: OK, yeah: is that not coming across clearly enough here?

felicianotech:
My bad with the URL piece. Everything looks good now.

PR #652, opened 2017-02-16
2.0 add env vars

smart-alek:
Adding available environment variables and descriptions for CircleCI 2.0.


felicianotech:
SEO/proper HTML wise those should not be headers. Especially 3rd level headers. I'm too tired to figure out what they should be though. :(

smart-alek:
@felicianotech maybe just bullets? they look a little small when I do that.

felicianotech:
Technically a definition list. But I don't remember how to do that in Markdown right now. I'd say this is fine for now and I can go back and take a look later if you want.

PR #663, opened 2017-02-23
Add collapsable sidebar menus

keybits:
This adds collapsable sidebar menus. Current page is highlighted and the menus remember state between page transitions.

Expand all / collapse all option is available.

Screencast to demonstrate:

Asset minification and optimisation will follow when other updates have been completed.


keybits:
A small todo is to reset the menu status when switching between sections. I think this can come later.

jekyll/_includes/global-footer.html
```diff
@@ -49,4 +49,38 @@
...
+            });
+
+          $("#expandAll").click(function(e) {
+            e.preventDefault();
+            $("#mysidebar").navgoco('toggle', true);
+            });
+
+        });
+
+    </script>
```
felicianotech: Can we move lines 54 - 85 into the main JS file for this site? This is cleaner, better organized, and we won't need jQuery's ready function in this snippet anymore.

keybits: Goo idea - done.

Issue #122, opened 2016-05-05
Add section the Artifacts Doc about Artifact Size Limitation 


levlaz:
@eric-hu and I confirmed that since curl does not support streaming uploads there is a system limitation of roughly 3GB for the max size of a single artifact.

PR #678, opened 2017-02-28
Use API 1.1 and explain retention and file size for artifacts

### LINKED ISSUE #148: Document limits on Build Artifacts size, retention, etc
RE: jekyll/_docs/build-artifacts.md

https://circleci.com/docs/build-artifacts/ doesn't mention any limits on the sizes of build artifacts (I see #122 mentions a 3GB size limit on individual artifacts; what about total size per build, or per project?), nor how long Circle promises to host the artifacts (e.g. only 3 months like (random example) [Ghost Inspector](https://ghostinspector.com/pricing/), or indefinitely but promise 6 months warning if you're gonna start taking them down, etc).

For my case I'm not actually worried about the size limit (though I would love to see it documented), just retention, especially since the filepaths are all `.../tmp/circle-artifacts.AvvFajf/...`, which isn't exactly resounding with permanence. Can I hotlink them, or should I be uploading them somewhere else? (Imgur probably)

There may be other limits worth documenting, too. (Rate limits on downloads? That's probably high enough not to be worth documenting.)

keybits:
Fixes #148 and #122

### DIFF
--- a/jekyll/_cci1/build-artifacts.md
+++ b/jekyll/_cci1/build-artifacts.md
@@ -25,20 +25,19 @@ You'll find links to the artifacts at the top of the build page. You can also co
 You can also access your artifacts in your browser with the following url:
 
 ```
-https://circleci.com/api/v1/project/:org/:repo/:build_num/artifacts/:container-index/path/to/artifact
+https://circleci.com/api/v1.1/project/:vcs-type/:org/:repo/:build_num/artifacts/:container-index/path/to/artifact
 ```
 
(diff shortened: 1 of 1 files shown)


felicianotech:
@tom The test passed, not sure why GitHub didn't get updated.

keybits:
Rebuilding as I think this got caught by S3 outage.

felicianotech:
@tom Did you touch Git at all to do that or just rebuilt for the CircleCI UI?

keybits:
Just a rebuild on CircleCI - we suggest this in support when people complain the the GitHub status didn't get updated.

PR #686, opened 2017-02-28
Update quickstarts with latest syntax.

technomancy:
Had this committed to the old repo earlier.

### DIFF
--- a/jekyll/_cci2/quickstart-javascript.md
+++ b/jekyll/_cci2/quickstart-javascript.md
@@ -41,7 +41,7 @@ jobs:
   build:
     working_directory: ~/cci-demo-react
     docker:
-      image: node:7.4.0
+      - image: node:7.4.0
 ```
 
 Now we’ll add several `steps` within the `build` stage.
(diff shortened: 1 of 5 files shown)


smart-alek:
LGTM

jekyll/_cci2/quickstart-javascript.md
```diff
@@ -54,42 +54,43 @@ jobs:
...
     docker:
-      image: node:7.4.0
+      - image: node:7.4.0
     steps:
-      - shell:
-        name: Pre-Dependencies
-        command: mkdir ~/artifacts
+      - run:
+          name: Pre-Dependencies
+          command: mkdir /home/ubuntu/cci-demo-react/artifacts
```
smart-alek: Should we change this back to `~/cci-demo-react/artifacts`? Or do we want to leave the `/home/ubuntu`?

jekyll/_cci2/quickstart-javascript.md
```diff
@@ -54,42 +54,43 @@ jobs:
...
           command: npm test
 ```
 
 Remember when we created a directory for our artifacts? CircleCI won't unless we tell it where to look.
 
 ```yaml
-      - artifacts_store:
-          path: ~/artifacts
+      - store_artifacts:
+          path: /home/ubuntu/cci-demo-react/artifacts
```
smart-alek: Same question about the root directory.

jekyll/_cci2/quickstart-javascript.md
technomancy: There are problems with `~` in local builds for the top-level `working_directory` setting in a job (not in steps) but these were coming from me trying to debug that failure and forgetting to change it back. I'll bring back the tilde here.

PR #692, opened 2017-03-01
Add why migrating to 2.0 page

kimh:
This page server as the first page in the migration section. The rationale is to give users benefits of moving from 1.0 to 2.0.

### DIFF
--- a/jekyll/_cci2/migration-why-2.md
+++ b/jekyll/_cci2/migration-why-2.md
@@ -6,37 +6,32 @@ categories: [migrating-from-1-2]
 order: 1
 ---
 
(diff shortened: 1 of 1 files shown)


keybits:
Thanks for writing this up @kimh I've edited based on some info I learned from Nathan this morning that we will use 'jobs' instead of 'builds' in 2.0.

I think we will not be using the word 'Pod' - but I'm not sure what that doc would be called. @ndintenfass  can you suggest (see last line of this doc).

smart-alek:
@kimh @keybits Revised my way through this; let me know if there are any glaring inaccuracies! I left the Pod line in for now, but it's definitely not in the configuration doc anymore, so this will probably need at least one more commit.

PR #695, opened 2017-03-01
Getting Started: Configuration

smart-alek:
This splits the current "Getting Started: Configuration" into 2 pieces: first is a high-level overview of the layout of `config.yml`. The second is a more detailed list of the built-in steps allowed for each job.

@technomancy @kimh


jekyll/_cci2/builtin-steps.md
```diff
@@ -0,0 +1,290 @@
...
+* `[environment]`: additional environmental variables, locally scoped to command
+    * must be a YAML map
+* `[background]`: boolean representing whether command should be run in the background
+    * default: `false`
+* `[pwd]`: absolute path to working directory
+    * defaults to the stage’s `workDir`
+    * if named path does not exist, errors thrown
+
+```yaml
+- type: shell
```
technomancy: This should be more like below instead:

```yaml
  - run:
      command: bundle check || bundle install
      shell: /bin/bash
      environment:
        FOO: bar
```

PR #700, opened 2017-03-02
Add configuration mapping

kimh:
I'm not sure if "Configuration Mapping" is a correct but I couldn't come up a better name. If anyone has a better idea, please just use it.


ndintenfass:
I'd consider "configuration equivalents" over "configuration mapping"

PR #710, opened 2017-03-03
Grouping dependencies, database, and test together

kimh:
@technomancy suggested to group dependencies, database, and test sections into the same category because they are just bunch of run steps.

Also, cache_save -> save_cache

### DIFF
--- a/jekyll/_cci2/migrating-from-1-2.md
+++ b/jekyll/_cci2/migrating-from-1-2.md
@@ -158,9 +158,7 @@ jobs:
       - run: git submodule sync && git submodule update --init # use submodules
 ```
 
-## Dependency
-
-### Override
+## Dependency/Database/Test
 
 **1.0**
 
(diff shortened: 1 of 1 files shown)


kimh:
Note to self: make sure to update "database" section for PG thing

technomancy:
This is great; thanks. Minor suggestion: "Database is not created automatically in 2.0." should probably be "The database is not ...".

Also I'm not 100% sure about this, but it seems that adding `save_cache` without ever restoring it doesn't benefit you, right?

Other than that it looks great.

kimh:
> Also I'm not 100% sure about this, but it seems that adding save_cache without ever restoring it doesn't benefit you, right?

This is correct but the purpose of this article is giving users a simple 1-to-1 mapping between 1.0 and 2.0 config. There is no config equivalent to `restore_cache` in 2.0. I think we should cover `restore_cache` in a different page dedicated to caching.

I'll add a note that users have to run `restore_cache` manually in 2.0 and link to more details.

kimh:
Addressed feedbacks. Merging.

PR #715, opened 2017-03-03
Use PG instead of sqlite

kimh:
To catch up https://github.com/circleci/cci-demo-rails/pull/3

### DIFF
--- a/jekyll/_cci2/language-ruby.md
+++ b/jekyll/_cci2/language-ruby.md
@@ -42,9 +42,12 @@ jobs:
     working_directory: ~/cci-demo-rails
     docker:
       - image: ruby:2.3
+      - image: postgres:9.4.1
+        environment:
+          POSTGRES_USER: root
 ```
 
-Normally, you’d also specify a database (DB) image here, but our app is using SQLite. SQLite is _so_ light that Rails will install it during setup, which means we don’t need to specify a DB image.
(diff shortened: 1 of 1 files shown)


keybits:
LGTM

jekyll/_cci2/language-ruby.md
```diff
@@ -42,9 +42,12 @@ jobs:
     working_directory: ~/cci-demo-rails
     docker:
       - image: ruby:2.3
+      - image: postgres:9.4.1
+        environment:
+          POSTGRES_USER: root
 ```
 
-Normally, you’d also specify a database (DB) image here, but our app is using SQLite. SQLite is _so_ light that Rails will install it during setup, which means we don’t need to specify a DB image.
+We use two Docker images here: ruby:2.3 as a primary build image and postgres:9.4.1 for database image because we use PostgreSQL in this demo project.
```
smart-alek: EDIT:

We use 2 Docker images here: `ruby:2.3` as the primary build image and `postgres:9.4.1` for the database image.

smart-alek: (I think it's implied that we're using PostgreSQL for the example project, but feel free to push back on that @kimh)

smart-alek:
Pushed a commit that addresses my review comments, but feel free to revert if you feel strongly @kimh.

PR #724, opened 2017-03-04
Remove configuration.md

smart-alek:
This article is outdated, has been split into 2 separate articles which
live in Getting Started (Basic Configuration & Built-in Steps).


smart-alek:
@keybits is going to remove this doc in a larger PR that renames several docs.

PR #727, opened 2017-03-06
Clarify DB section in the equiv

kimh:
The DB section is broken and inaccurate: `createdb` is missing in the ruby image. Also, you don't need to create a test DB by yourself in Rails.

### DIFF
--- a/jekyll/_cci2/migrating-from-1-2.md
+++ b/jekyll/_cci2/migrating-from-1-2.md
@@ -153,21 +153,21 @@ jobs:
 ```
 dependencies:
   override:
-    - <dependecy-install-command>:
+    - dependecy-install-command:
         timeout: 180
 
 database:
   override:
-    - <setup-db-command>
+    - setup-db-command
 
 test:
   override:
-    - <test-command>
+    - test-command
 ```
 
 **2.0**
 
-The `dependency`, `database`, and `test` sections are translated into a sequence of 'run' steps.
+The `dependency`, `database`, and `test` sections are translated into a sequence of `run` steps. To learn more about `run` steps, please checkout [this]({{ site.baseurl }}/2.0/configuration-reference/#run) page.
 
 ```
 version: 2
@@ -179,9 +179,9 @@ jobs:
       # In 2.0, you can use multiple different images and group them as a 'Pod'.
       # Because Docker containers normally don't include DBs, you need to bring
       # a build image that serves as DB container in your build.
-      - image: <primary-build-image>
-      - image: <db-image-1>
-      - image: <db-image-2>
+      - image: primary-build-image
+      - image: db-image-1
+      - image: db-image-2
 
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -197,10 +198,14 @@ jobs:
...
 
       # The DB is not automatically created in 2.0, so we manually set up a test DB
-      - run: sudo -u root createuser -h localhost --superuser ubuntu
-      - run: sudo createdb -h localhost test_db
+      # If your project is Rails, then 'db:create' should create a test DB for you.
       - run: bundle exec rake db:create db:schema:load
 
+      # You can also create a test DB manually if your project requires.
+      # We are installing postgresql package because we need 'createdb' command
+      - run: apt-get update -qq; apt-get install -y postgresql
```
zzak: Is `update` needed here? I know it's probably useful for demonstration, but I think it's unnecessary for documentation (perhaps the user updates earlier in their config, or they will install this directly on their own image).

kimh: Yeah, unfortunately you need. Otherwise, you are getting "no package" error :(

keybits: @kimh it would be better to specify a primary container image that comes with Postgres client. `apt-get update` isn't ideal in 2.0 docs for several reasons:

- it's slow
- it's ubuntu / debian specific
- it's mutable since apt repositories change. we are aiming for immutability to make things more reliable

we should move away from how we did things in 1.0.

keybits: @kimh it would be better to specify a primary container image that comes with Postgres client. `apt-get update` isn't ideal in 2.0 docs for several reasons:

- it's slow
- it's ubuntu / debian specific
- it's mutable since apt repositories change. we are aiming for immutability to make things more reliable

we should move away from how we did things in 1.0.

kimh:
Based on feedbacks, I think the mistake is trying to make an example actually work. However, this is not a goal of the equiv guide: it's there to give users a feeling about how to convert 1.0 to 2.0 config. 

The accuracy of commands (`bundle install` not working) or best practices (not to use `apt-get`) are not the scope of this guide: we should address them in tutorials and getting started guides.

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -181,29 +180,28 @@ jobs:
...
-      - run: sudo createdb -h localhost test_db
-      - run: bundle exec rake db:create db:schema:load
-
-      # Run any test commands
-      - run: bundle exec rspec
+      - run: <setup-db-command>
 
+      # Again, there is no inference in 2.0, so nothing to override here, either.
+      # Just use any test commands.
+      - run: <test-command>
```
keybits: I think this is confusing for people since this section will likely have multiple sub-sections, e.g:

```
      - run:
          name: Run tests
          command: lein test
          environment:
            DATABASE_URL: postgres://conductor:@localhost:5432/conductor_test
```

Can we make it clearer how the `run:` step is used?

kimh: Sure. I'll omit `environment:` part though, because it doesn't exist in 1.0 version. Instead, I added a link to a reference page of run step for more info.

kimh:
I'm merging this now but happy to address in subsequent PRs if there are more feedbacks.

ndintenfass:
I'm not sure I follow that @kimh -- if the practice is not recommended or there is a better path in 2.0 than 1.0 the docs should reflect that. No?

kimh:
@ndintenfass Yes, absolutely, but I think they have to be addressed in tutorials, not in equiv guide. IMO the equiv guide should provide simple mappings between 1.0 and 2.0 config.

eric-hu:
FYI: I think the merged changes is a good middle-ground.  It removes steps we advise against, like `apt-get update`, but gets the point across  👍

PR #728, opened 2017-03-06
Clarify DB section in the Rails guide

kimh:
We are using PG in our sample Rails project https://github.com/circleci/cci-demo-rails/pull/3 

Although we don't have to add a step in circle.yml (because Rails will take care of creating test db in PG), some clarification is useful.


zzak:
LGTM

PR #732, opened 2017-03-06
Replace dashes with underscores for new style config

keybits:


technomancy:
We still support this style for compatibility, but we want to move to the canonical version being to verb-first, so `store_artifacts`, `restore_cache`, etc. going forward for consistency.

PR #761, opened 2017-03-08
Fix unbalanced backtick

kimh:


kimh:
Small change. Merging.

PR #769, opened 2017-03-08
Fix/sitemap

### LINKED ISSUE #610: Sitemap Needs To Be Updated
The sitemap needs to be updated to support the new site architecture of having sections, and all the new docs.

felicianotech:
This fixes `sitemap.xml` for Docs. The sitemap file now properly supports Jekyll Collections. This will load all docs from within any collections available.

- specifically added the main index page, which we didn't have before
- entity escapes the URLs now since characters such as `&` are not allowed the the sitemap (and XML) syntax.
- ignores sections that are hidden (`selectorName == false`)
- ignores docs that are hidden (`hide == true`)
- validated output with W3C: https://validator.w3.org/
- uses Google's suggested scheme: https://www.sitemaps.org/protocol.html
---
Fixes #610 
---
@michaelcstearns Any notes you'd like to add?


michaelcstearns:
are we keeping the same /docs/sitemap.xml file path? if so, this should be fine. if not, we need to update robots and webmaster tools that are pointing to that.

keybits:
LGTM

@michaelcstearns yes, the file is in the same location.

felicianotech:
@michaelcstearns Yup, it's the same. Google already picked up the updated sitemap. I have a question that I'll ping you for though.

@keybits Thank you.

PR #773, opened 2017-03-08
Implement canonical URLs as absolute links - marketing asked for this

ndintenfass:


felicianotech:
This is usually done `{{ doc.url | prepend: site.baseurl | prepend: site.url }}` so that when/if we use staging, the URLs all still work. Same for when/if we return on external link checking (these would all fail locally).

ndintenfass:
in this case the only purpose of canonical header info is for search engines, so that shouldn't matter.

ndintenfass:
(and we wouldn't want a search engine indexing staging - and if that happened we'd actually want the hard-coded canonical URL)

felicianotech:
What I posted becomes the full absolute URL too. It's just flexible incase docs ever moves (ever the hostname or the directory).

With the feed file going, I guess it matters less though.

ndintenfass:
@felicianotech changed to your suggestion

PR #781, opened 2017-03-09
passing a map to machine

kimh:
Currently `machine` only takes a map. Giving a single value will lead to config error.

Bad
https://circleci.com/gh/kimh/picard-test/1033

Good
https://circleci.com/gh/kimh/picard-test/1034


keybits:
LGTM

PR #785, opened 2017-03-09
Add table of content on config-ref page

kenden:
Having a table of content allows to  quickly see which steos are available


keybits:
Thank you for the suggestion - much appreciated. We are implementing automatic table of contents for all doc pages in the next few days.

PR #794, opened 2017-03-09
Move beta banner up and proper layout.

felicianotech:
- moves the beta banner above the title
- moves the beta banner code to a Jekyll include
- use the normal layout `classic-docs` for the include instead of a nearly duplicate one
- adjust cci2 docs to use the original layout


keybits:
Tested locally and LGTM.

Thanks for the cleanup :-)

felicianotech:
Somehow there was a conflict. Merging once test pass.

PR #795, opened 2017-03-09
Create API section.

felicianotech:
This pulls put the API doc from 1.0 into it's own section. The API is for both 1.0 and 2.0. I have A LOT of plans for what the API docs could be, but this PR isn't it. It's just a first step to make sure the API is visible to both 1.0 and 2.0 doc readers.

The current idea is, the existing 1.0 API doc because the API reference page in the new section. Down the road, we'll improve the reference page (because it sucks), add examples of how to use the API in different languages with live examples.

ToDo:
- [x] create API category(ies)
- [x] remove API from our load-balancer redirects so that this page can actually be viewed
- [x] redirect 1.0 api doc to the new landing page


felicianotech:
I still need to fix the test but thoughts @ndintenfass, @tom?

felicianotech:
For this initial piece, nothing. Just additional PRs to keep making the section better. I just really would like Tom's :eyes: since he dealt wit the redirects on our front-end. Then we can merge this and our internal PR for the front-end.

keybits:
This is great. Just what we need for now.

PR #798, opened 2017-03-09
Switch example to use pre

mtogstad:
Suggesting overriding the inferred dependencies step without the context of what overriding does could (did :( ) cause someone to not notice they were reliant on inferred dependency steps later in their build.


smart-alek:
@mtogstad rather than change this step, do you think it would be more helpful to provide context around what overriding does instead?

PR #809, opened 2017-03-10
Clarify that ssh-agent isn't needed for normal SSH use.

technomancy:
The current wording is a bit unclear and could be read as implying that you need to set up an agent for any `ssh` key use.


felicianotech:
LGTM

PR #814, opened 2017-03-11
Removes "Requires Admin" doc

### LINKED ISSUE #406: RE: jekyll/_docs/requires-admin.md refers to page that no longer exists
"Manage Projects" doesn't seem to exist (anymore), though I got to take that opportunity to dive into the codebase with @drazisil (which was great!).

What page should this refer to now? `Add Projects`? `Projects` (which I don't have)? Something else?

I'm itching to rewrite this little doc in general. 😀

smart-alek:
This doc was last updated on Feb 3, 2013 and has no category. It also
refers to a page that no longer exists. This doc doesn't even need to
exist because this requirement should be clearly outlined in a Getting
Started guide.

Closes #406


kimh:
The update date is actually wrong. It's last modified at https://github.com/circleci/circleci-docs/pull/623

I would modify to point to a correct page instead of the non-existent one.

felicianotech:
Kim is right, use Git to determine when it was last updated. The `last_updated` Front Matter variable is a relic that we don't use anymore.

With that being said, I agree with the decision to kill it.

smart-alek:
Ah, thanks @kimh and @felicianotech! Makes a lot more sense to use Git to get the date of the last update.

smart-alek:
Going to just remove this one; a search for "admin permissions" still results in helpful pages.

PR #819, opened 2017-03-13
Try to make it more clear that rebuild won't apply

zzak:


smart-alek:
On a more serious note: this doc is only 20 lines long. Whenever I find a doc this short, I try to think how it could be combined with something else or if the doc needs to exist.

@zzak Can you make a case for keeping this doc around?

smart-alek:
Merging this for now and saving existential discussions for another time. :)

PR #823, opened 2017-03-13
Add notes on using non-pre-installed Node versions

jamesroutley:
The existing [documentation](https://circleci.com/docs/1.0/language-nodejs/#version) on picking a NodeJS version point the user to the list of versions pre-installed on the Ubuntu containers. These lists do not contain newer NodeJS (e.g. 6.10.0). I think this implies that you can't specify that version number in `circle.yml`. In fact you can, and CircleCI just downloads and uses the version specified.

This PR adds a sentence explaining that you can use non-pre-installed versions.

I've only tested this with `v6.10.0`, and it may not be true for all versions of NodeJS.

### DIFF
--- a/jekyll/_cci1/language-nodejs.md
+++ b/jekyll/_cci1/language-nodejs.md
@@ -16,7 +16,7 @@ commands that will run after a green build.
 
 ### Version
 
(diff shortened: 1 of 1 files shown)


jamesroutley:
Just found docs on this [here](https://circleci.com/docs/1.0/build-image-precise/#nodejs)

smart-alek:
Hey @jamesroutley, thanks for closing your own issue! You are a model contributor. :)

PR #824, opened 2017-03-13
some polishing for docker docs

smaant:


jekyll/_cci2/building-docker-images.md
```diff
@@ -52,22 +52,22 @@ jobs:
...
           docker push circleci/cci-demo-docker:$TAG
 ```
 
 Let’s break down what’s happening during this build’s execution:
 
-- All commands are executed in the [primary container][primary-container].
-- Once `setup_docker_engine` is called, a new remote environment is created, and your primary container is configured to use it.
-- All docker-related commands are also executed in your primary container, but building/pushing images and running containers happens in the remote Docker Engine.
-- We use project environment variables to store credentials for Docker Hub.
+1. All commands are executed in the [primary container][primary-container].
```
smart-alek: Should [primary container] be linking to the glossary?

ndintenfass: @smart-alek it's being defined at the bottom of the page using `[primary-container]: {{ site.baseurl }}/2.0/glossary/#primary-container`

smart-alek: ah shoot, missed that.

PR #827, opened 2017-03-13
Remove references to artifacts-store

smart-alek:
There was only one left in the Configuration Reference docs.


smart-alek:
Self-merging because this change is trivial.

PR #829, opened 2017-03-13
Fix 302s

smart-alek:
A few stragglers were lacking the trailing forward slashes.


smart-alek:
Self-merging since these links are technically broken.

PR #830, opened 2017-03-13
remove trailing slash from inner link

smart-alek:
Links to inner should not have a trailing slash.


smart-alek:
Minor change, self-merging.

PR #833, opened 2017-03-14
Analytics

trevor-circleci:
adding sections to analytics on docs. 

@ndintenfass, @keybits

### DIFF
--- a/jekyll/_includes/segment.html
+++ b/jekyll/_includes/segment.html
@@ -12,7 +12,7 @@
     "short-title": {{ page["short-title"] | jsonify }},
     "categories" : {{ page["categories"] | jsonify }},
     "description": {{ page["description"] | jsonify }},
-    "collection": {{ page.collection | jsonify }}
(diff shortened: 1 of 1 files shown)


jekyll/_includes/segment.html
```diff
@@ -10,7 +11,9 @@
     "title"      : {{ page["title"] | jsonify }},
     "short-title": {{ page["short-title"] | jsonify }},
     "categories" : {{ page["categories"] | jsonify }},
-    "description": {{ page["description"] | jsonify }}};
+    "description": {{ page["description"] | jsonify }},
+    "collection": {{ page.collection | jsonify }}
```
ndintenfass: Is there a reason you're changing style from the above?

trevor-circleci: @ndintenfass I thought the use of the old notation was strange there; less clean, IMO. But I'll re-do that for consistency.

PR #837, opened 2017-03-14
CIRCLE-3354 First writing about deployment

kimh:
I'm only addressing deployment part in this PR. Also, I renamed deployments-and-uploads.md to simply deployments.md because I thought the two topics have to deserve a dedicated page.

### DIFF
--- a/jekyll/_cci2/deployments-and-uploads.md
+++ b/jekyll/_cci2/deployments-and-uploads.md
@@ -1,8 +1,0 @@
----
-layout: classic-docs
-title: "Deployments and Uploads"
-short-title: "Deployments and Uploads"
-categories: [configuring-jobs]
-order: 98
-published: false
----
--- a/jekyll/_cci2/deployments.md
+++ b/jekyll/_cci2/deployments.md
@@ -6,9 +6,12 @@ categories: [configuring-jobs]
 order: 98
 ---
 
-There are two standard ways to do deployment in 2.0: using a [deploy]({{ site.baseurl }}/2.0/configuration-reference/#deploy) step and triggering a deploy job.
+There are 2 standard ways to deploy in 2.0:
 
-## Using a Deploy Step
+- use a [deploy step][deploy-step]
+- trigger a deploy job
+
+## Use a Deploy Step
 
 ```YAML
 version: 2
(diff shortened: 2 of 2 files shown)


jekyll/_cci2/deployments-and-uploads.md
```diff
@@ -1,8 +0,0 @@
----
```
ndintenfass: Rather than delete files, please add the `redirect` front-matter directive (eg value: `/docs/2.0/deployments/`) to avoid 404s.

jekyll/_cci2/deployments.md
```diff
@@ -0,0 +1,93 @@
...
+    working_directory: /tmp/my-project
+    steps:
+      - run: <do-some-stuff>
+      - deploy:
+          name: Maybe Triggering Deploy Job
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "master" ]; then
+                curl -u ${CIRCLE_API_TOKEN}: \
+                     -d build_parameters[CIRCLE_JOB]=deploy \
+                     https://circleci.com/api/v1.1/project/<vcs-type>/<org>/<repo>/tree/master
```
ndintenfass: This feels like a terrible pattern to codify as an official recommendation.

jekyll/_cci2/deployments-and-uploads.md
kimh: Sure, I can do that, but is it still necessary even if https://circleci.com/docs/2.0/deployments-and-uploads does not exist now?

jekyll/_cci2/deployments.md
kimh: You mean the fact that we ask customers to use `curl` to invoke a new job? I feel it's ok until we have a workflow but do you have any suggestions?

jekyll/_cci2/deployments-and-uploads.md
smart-alek: I'm curious about your answer, too @ndintenfass. Wouldn't this mean that our documentation would just infinitely grow? Do we just never delete articles that we rename? I'm mainly worried about directories filling up with dead articles that just redirect.

PR #838, opened 2017-03-14
CircleCI 2.0 PHP Improvement

hskrasek:
Instead of confirming the composer installer using a hardcoded hash, which will fail the next time composer updates, you can instead copy down a signature file and compare that way.

### DIFF
--- a/jekyll/_cci2/language-php.md
+++ b/jekyll/_cci2/language-php.md
@@ -48,9 +48,10 @@ jobs:
       - run:
           name: Install Composer
           command: |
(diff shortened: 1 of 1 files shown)


smart-alek:
Great addition @hskrasek; thanks for helping out! :1st_place_medal:

felicianotech:
Niceee. We need to do this in the actual demo project this came from as well.

smart-alek:
@felicianotech good point. Can you create an issue for that?

felicianotech:
@smart-alek I already did, it's "GitHub referenced" right above my comment.

PR #840, opened 2017-03-14
make audience/voice consistent across 2.0 index page

smart-alek:
It's weird to be talking about "we're" doing as CircleCI devs and then shift
to a question where "I'm" new to CircleCI. This change makes the voice
consistent across the doc.


kimh:
LGTM

smart-alek:
Thanks for all these reviews @kimh!

PR #852, opened 2017-03-15
[CIRCLE-3496] Stop requiring ESXI migrations in CCIE iOS setup guide

appplemac:
Starting on the next CCIE release, it will no longer be necessary to run ESXI migrations when setting up iOS infrastructure. The migrations will be run via `circleci run-migrations` instead.


keybits:
@appplemac LGTM

PR #855, opened 2017-03-15
Update building-docker-images.md punctuation mark

michelle-luna:
Add missing period at end of Overview section.


smart-alek:
Ah, those sneaky punctuation marks! Thanks @michelle-luna! 🙏

PR #862, opened 2017-03-16
Minor spelling fix (it's -> its)

relsqui:
Does what it says on the tin.


smart-alek:
Those pesky apostrophes invite themselves to every party! Thanks for catching @relsqui. :)

PR #865, opened 2017-03-16
Triggering job

kimh:
After some thinking, I came to think `deploy job` section in https://circleci.com/docs/2.0/deployments/ are not necessary to be there. I put there because that's the pattern we do in our internal repos but I don't think this is what most of customers should follow. In other words, most of deployments can be done with `deploy` step.

Documenting how to define/trigger non-build job is still useful, so I made a separate page for this.


kimh:
@ndintenfass maybe you want to take a look? I think change accords with some of thoughts that you raised earlier.

PR #870, opened 2017-03-16
update README to re-order sections, add two procedures and fix punctuation

michelle-luna:
Updates to re-order sections of README to put easy way first, add procedures to fill gaps, and fix some pesky punctuation marks/initial caps.


felicianotech:
@ndintenfass @smart-alek Thoughts? I like the change. My only hesitation is the whole GitHub Desktop bit.

michelle-luna:
Hi @felicianotech thanks for this feedback. I agree on the GitHub Desktop steps, those should be git command(s). Push and then request-pull? I wasn't sure enough to guess the start and end on the first try. I would welcome edits that form those commands correctly.

smart-alek:
Looks good for now @michelle-luna! Thanks for this cleanup; I'll be adjusting some of the language and probably swap the GitHub desktop piece out as well. :)

PR #871, opened 2017-03-17
Add a reference to artifact API

kimh:
Only adding a reference to artifact build because accessing them via file path is broken in 2.0.


kimh:
Small change: merging.

PR #872, opened 2017-03-17
docker layer caching

smaant:


jekyll/_cci2/docker-layer-caching.md
```diff
@@ -0,0 +1,24 @@
...
+
+If you run many parallel jobs for the same project (the same or different branches) and each of them are requesting reusable environment none of them will be blocked and all will be provided with a Remove Docker Environment (but probably not all of them will have cached layers). Please note that this feature is in Beta right now and we might change this behaviour later.
+
+The second option `exclusive` indicates whether you want to allow running parallel jobs in the same Remote Docker Environment simultaneously or not. Disabling it (`exclusive: false`) will make parallel jobs of the same project share one Remote Docker Environment. This way you can reduce chances of getting a clean Remote Docker. However you need to be cautious and make sure that your project is tolerant to shared Docker environment. For instance, if you build images with a mutable tags (like `latest`) then shared Docker environment might be a bad idea. In case of `exclusive: false` we will allow only two parallel jobs to use the same Docker environment. Please note that this feature is in Beta right now and we might change this behaviour later.
```
eric-hu: > For instance, if you build images with a mutable tags (like `latest`) then shared Docker environment might be a bad idea.

Why is this _more_ of a bad idea with layer caching and a Remote Docker Environment?  Also, I'm still not sure I understand what the `exclusive` flag is supposed to do.  It sounds like it shifts the odds of two non-deterministic outcomes?

kimh:
(Only reviewing PR from technical perspective)

Is the term "Docker Layer Caching" common enough? We internally use the phrase but "Docker Image Caching" sounds more familiar to me.

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -0,0 +1,24 @@
+---
+layout: classic-docs
+title: "Docker Layer Caching"
+short-title: "Docker Layer Caching"
+order: 5
+---
+
+{% include beta-premium-feature.html%}
```
ndintenfass: In this call we should add `feature='Docker Layer Caching'`, so that label appears in the banner.

smaant: Thanks, I intentionally dropped it to avoid repetition, because the whole doc is for this feature and just right above there will doc title. But if you think it still needs to be there, let me know, I'll add it.

smaant:
@kimh I think these two terms are close enough to each other, "docker layer caching" is more technically correct, probably.

jekyll/_cci2/docker-layer-caching.md
smaant: @eric-hu yeah, I though it might be a bit confusing :( can you give me your current understanding you got from this doc? So I would know how to change it.

eric-hu: @smaant sure.  Let me see...

> The second option `exclusive` indicates whether you want to allow running parallel jobs in the same Remote Docker Environment simultaneously or not. 

So my understanding with `exclusive: true` is if 2 jobs are running at the same time, we will try to avoid putting them on the same machine.  _"Try to"_ seems key here: you could still get that behavior because we're not providing any guarantees with this flag.  Is that right?

The example with mutable tags seems like it could still be a problem with non-simultaneous jobs.  Suppose three jobs are running on the same Remote Docker Environment and they use mutable tags, with this ordering:

A) Job 1, latest tag.  Runs and finishes
B) Job 2, versioned tag forces an update, which is the new latest.  Runs and finishes.
C) Job 3, latest tag.  Runs and finishes.

This would be problematic if A, B and C ran at the same time or ran serially on the same machine.  Can you construct an example like that where it would be problematic if concurrent, but not serial?

kimh:
> you are probably know that Docker image consists of layers

@smaant Yeah, I admit that it's technically more accurate but I wonder the statement above is true for everybody: I believe not everyone understands that images are composed of layers.

The following google queries show how the terms are popular:

**Docker Layer Caching**
https://www.google.com/search?safe=off&gl=us&hl=en&pws=0&site=webhp&source=hp&q=docker+layer+caching&oq=docker+lay&gs_l=hp.3.0.35i39k1j0l9.3253.4234.0.5370.12.12.0.0.0.0.152.982.4j5.9.0....0...1.1.64.hp..3.7.845.0..0i67k1j0i131k1j0i20k1.tC8hZILksi8

**Docker Image Caching**
https://www.google.com/search?safe=off&gl=us&hl=en&pws=0&site=webhp&q=docker+image+caching&oq=docker+image+caching&gs_l=serp.3..35i39k1.56465.56934.0.57134.5.5.0.0.0.0.79.352.5.5.0....0...1.1.64.serp..0.5.352...0i7i30k1.2cgfpBbnDx4

Whenever the top hit is "Scholarly articles for..." in Google, I wonder if I'm googling with wrong keywords.

Maybe we can bring this up in #docs and how other folks think about it?

jekyll/_cci2/docker-layer-caching.md
smaant: @eric-hu I'll try to update doc and ask you again, if you don't mind

smaant:
@kimh I think it's easier to explain this feature using "Layers" term. Also, it's kind of advanced feature, so we can expect some level of understanding from users using it. I don't mind asking #docs  opinion about this though.

jekyll/_cci2/docker-layer-caching.md
smaant: @eric-hu I've updated doc a bit, can you please tell me what is your current understanding of `exclusive` option? Also, what is the exact problem you see in your example above?

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -7,18 +7,32 @@ order: 5
...
     reusable: true    # default - false
     exclusive: true   # default - true
 ```
 
-Please note, that this feature requires your project to be whitelisted. Please reach out to your support representative in order to enable it.
+## **`reusable`**
+
+With `reusable: true`, CircleCI will try to reuse your Remote Docker Environment; every layer you built in a previous job will be accessible there. We do our best to reuse the previous environment when we can, but in some cases your job may find itself in a clean environment, even if you've specified `reusable: true`.
+
+If you run many parallel jobs for the same project, and each job requests a reusble environment, all of them will be provided with a Remote Docker Environment. However, not all of them will have cached layers, although this behavior is subject to change over the course of the Beta.
```
smaant: > not all of them will have cached layers

This's not exactly true, now it sounds like for sure some parallel jobs won't have a cached layer. However they all **might** have it, it just depends on some factors.

jekyll/_cci2/docker-layer-caching.md
eric-hu: Quoting since this is merged already:

> If you specify exclusive: true, your job is guaranteed to have an exclusive Remote Docker Environment that other jobs can’t access.
>
> If you specify exclusive: false, a project’s parallel jobs will be able to share the same Remote Docker Environment. This method lets you reduce the chances of receiving a Remote Docker Environment without a cache. If you choose this option, ensure that your project can handle concurrent operations in Docker Engine.
> 
> For example, if you build images with mutable tags like latest, then a shared Docker environment could be a bad idea. With exclusive: false, we only allow 2 parallel jobs to use the same Docker environment.

Okay, this is clearer to me now.  To restate this and confirm I understand: if you use immutable tags, and you're certain your jobs can otherwise run in the same Docker environment as each other, you can use `exclusive: false` safely.  You'll benefit from a higher chance of cache-hits.

If you're unsure, you should use `exclusive: true`, as you risk running your jobs with inconsistent environments and creating hard-to-debug problems.  Is that about right?

Nice-to-have additional questions:
With `exclusive: false`, is every 2nd parallel job guaranteed to land on the machine of the 1st parallel job?  Does this apply to every 2N and 2N-1 job as well?

smaant: Yes, your understanding is correct. Additional question - yes if all jobs have exactly the same execution time and none of them fails (or all of them fails after exactly the same time), in the real life users should not have any expectations about what machine will be used for each particular job.

jekyll/_cci2/docker-layer-caching.md
smart-alek: Ah, got it. Want me to swap this copy back @smaant?

PR #877, opened 2017-03-20
[CIRCLE-3453] document no_output_timeout

notnoopci:
`run` step can take a `no_output_timeout` modifier

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -190,6 +190,7 @@ shell | N | String | Shell to use for execution command (default: `/bin/bash -e`
 environment | N | Map | Additional environmental variables, locally scoped to command
 background | N | Boolean | Whether or not this step should run in the background (default: false)
 working_directory | N | String | What directory to run this step in (default:  [`working_directory`](#jobs) of the job)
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -190,6 +190,7 @@ shell | N | String | Shell to use for execution command (default: `/bin/bash -e`
 environment | N | Map | Additional environmental variables, locally scoped to command
 background | N | Boolean | Whether or not this step should run in the background (default: false)
 working_directory | N | String | What directory to run this step in (default:  [`working_directory`](#jobs) of the job)
+no_output_timeout | N | String | Time duration where the command can run without output. The is decimal number with unit suffix, such as "20m", "1h", "5s" (default: 10 minutes)
```
felicianotech: "The is decimal number" <- small grammar error there

felicianotech:
@notnoopci I pointed out a small grammar error. I could fix it myself but, more importantly, you mention decimal but all the examples are exclusively integers. Are decimal values such as `2.5m` allowed? If so, that should be an example. Otherwise I think we should say integers to be more specific.

notnoopci:
That should have been `The string is ...` - and yes decimal is ok so `2.5m` would work.  Updated example.

felicianotech:
LGTM. Merging after passed test.

PR #881, opened 2017-03-20
[CIRCLE-3110] Add example of pre-starting simulator

appplemac:
The customers who want to pre-start the simulator can run into the case where there are two simulators with the same name, a phone-only one and a phone-plus-watch one.

Adding an example and explaining how to uniquely identify a phone-only simulator (which is what the customers need most commonly).


felicianotech:
LGTM

keybits:
LGTM

PR #882, opened 2017-03-20
Pipefail machine

smaant:


jekyll/_cci2/configuration-reference.md
```diff
@@ -244,7 +249,12 @@ Each `run` declaration represents a new shell. It's possible to specify a multi-
       make test
 ```
 
-Note that our default `shell` has the `-e` option, which causes commands to:
+##### **Shell options**
+**Note** that our default `shell` has a few options enabled by default.
```
smart-alek: I think it's a bit strange to put a note immediately after a section header. Were you concerned that people weren't seeing it?

smaant: This's pretty important section, but I guess we would be fine without this "Note that"

jekyll/_cci2/configuration-reference.md
```diff
@@ -244,7 +249,12 @@ Each `run` declaration represents a new shell. It's possible to specify a multi-
...
+
+**`-e` option**
```
smart-alek: Does this also need to be `-e pipefail` (like `-o pipefail`)?

smaant: No

jekyll/_cci2/configuration-reference.md
```diff
@@ -260,13 +270,28 @@ So if in the previous example `mkdir` failed to create a directory and returned
...
+For example:
+``` YAML
+- run: make test | tee test-output.log
+```
+
+If `make test` fails `-o pipefail` option will cause whole step to fail. Without `-o pipefail` step will always stay successful because the result of whole pipeline will be determined by the last command (`tee test-output.log`) which will always return zero status regardless of `make test`.
+
+Note that even if `make test` fails the rest of pipeline will be execute.
+
+If you want to avoid this behaviour you can specify `set +o pipefail` in command or override whole `shell` (see example above).
```
smaant: The one at the end of "-e" section, it's logically about the same thing - how to change behavior back

smart-alek:
@smaant revised and left a few comments/questions for you (they might be trapped in that outdated area, though).

smaant:
Thanks @smart-alek, good corrections.

PR #885, opened 2017-03-20
add premium banner

smart-alek:
As part of the preparation for 2.0 pricing, we want to warn users that a feature that they're using might wind up getting charged down the road.

Addresses [this ticket](https://circleci.atlassian.net/browse/CIRCLE-3493).

### DIFF
--- a/jekyll/_cci2/local-jobs.md
+++ b/jekyll/_cci2/local-jobs.md
@@ -6,6 +6,8 @@ categories: [configuring-jobs]
 order: 30
 ---
 
+{% include beta-premium-feature.html %}
+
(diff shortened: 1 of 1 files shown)


keybits:
Is the `circleci` CLI going to be a premium feature? I thought it running local jobs via the CLI was to be freely available to everyone.

smart-alek:
@ndintenfass also mentioned that this was going to be free, but I was basing this off of a list of potentially premium features that Rishi gave me.

But since this is the only spot I found that also matched one of the listed features, I'm going to close this PR since we're all in agreement about keeping this as a free feature.

PR #886, opened 2017-03-20
clearly call out parallelization gotchas, cleanin'

li3n3:
Add new information from @eric-hu about parallelization, which routinely trips up even very experienced developers. Very open to reorganization if folks have opinions on that; going for "done is better than perfect" here, but "better" is even better than "good."

I am bummed about mentioning a backlogged bug in the docs, but the alternative is a logical assumption ("commands that appear to run in all nodes do indeed run in all nodes") being secretly wrong, which I'm even less enthusiastic about.

There were mixed header tags and Markdown in this file, which leads to an unclear hierarchy on the rendered page (e.g. `Splitting your test suite` and `Using the files configuration modifier` were the same level, even though one is logically under the other).

Since this is a Markdown file, I would rather go for Markdown, but I don't know how to do that and maintain the anchored links. If someone knows how to do that, let's do it; in the meantime, I've swapped existing Markdown headers (e.g. `###`) for HTML.

As an aside, I am super curious why we appear to sometimes call containers "containers," and sometimes call them "nodes." I've stuck with what's internally consistent in this doc.

### DIFF
--- a/jekyll/_cci1/parallel-manual-setup.md
+++ b/jekyll/_cci1/parallel-manual-setup.md
@@ -12,7 +12,7 @@ you've overridden our test commands, you'll still be able to set up parallelism 
 To begin with, you'll need to turn on parallelism from your project's settings page.
 Go to **Project Settings > Parallelism** to adjust the settings.
 
-<h2 id="files-splitting">Splitting your test suite</h2>
+## Splitting your test suite
 
 When you use CircleCI's parallelization, we run your code on multiple separate VMs.
 To use parallelism, you make your test runner run only a subset of tests on each VM.
@@ -22,17 +22,16 @@ configuration modifier (a very simple and straightforward way for most use cases
 using parallelism environment variables (aimed at more complex scenarios).
 
 Either way, there are a few things to keep in mind:
-* `test` and `deploy` sections in your `circle.yml` are single-node only by
+
+- `test` and `deploy` sections in your `circle.yml` are single-node only by
   default and will only run on node 0. However, you can override `test` sections
   per command (add `:` to the end of each command you'd like to parallelize, then
-  add a double-indented `parallel: true` below that). Alas, no parallel deploys.
-
-* Inferred `test` commands will automatically parallelize.
-
-* Currently, commands that only run in node 0 will erroneously show up in the
+  add a double-indented `parallel: true` below that).
+- Inferred `test` commands will automatically parallelize.
+- Currently, commands that only run in node 0 will erroneously show up in the
   UI as running on all nodes.
 
-<h3>Using the `files` configuration modifier</h3>
+### Using the `files` configuration modifier
 
 Parallelizing test runners that accept file names is straightforward!  The `files` modifier
 can list paths to the test files, and CircleCI will run the test runners with different test files in each node.
@@ -48,7 +47,7 @@ test:
           - spec/**/*.rb          # or a glob (ruby globs)
 ```
 
-**Note:**
+**Note:**  
 The `**` pattern matches multiple directory layers with Ruby globs. You can test glob patterns locally with e.g. `ruby -e "puts Dir.glob('**/*.py')"`.
 
 In this example, we will run `bundle exec rspec` in all nodes appended with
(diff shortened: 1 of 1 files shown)


li3n3:
Would also love feedback about other docs that the parallelization info might fit in (in addition to this one). It's sneaky enough that most people on our team get tripped up about it; I'd love it to be more discoverable.

li3n3:
> Since this is a Markdown file, I would rather go for Markdown, but I don't know how to do that and maintain the anchored links. If someone knows how to do that, let's do it; in the meantime, I've swapped existing Markdown headers (e.g. ###) for HTML.

On second thought, this would be totally fine in Markdown. The issue would be breaking existing links to those anchors. So, HTML it is unless someone wants to change that!

jekyll/_cci1/parallel-manual-setup.md
```diff
@@ -16,11 +16,23 @@ Go to **Project Settings > Parallelism** to adjust the settings.
...
+  per command (add `:` to the end of each command you'd like to parallelize, then
+  add a double-indented `parallel: true` below that). Alas, no parallel deploys.
+
+* Inferred `test` commands will automatically parallelize.
+
+* Currently, commands that only run in node 0 will erroneously show up in the
+  UI as running on all nodes.
 
-## Using the `files` configuration modifier
+<h3>Using the `files` configuration modifier</h3>
```
eric-hu: You can change this to `###` for an H3.  Other places use an HTML tag if they need an HTML attribute.

jekyll/_cci1/parallel-manual-setup.md
```diff
@@ -16,11 +16,23 @@ Go to **Project Settings > Parallelism** to adjust the settings.
...
 There are two mechanisms for splitting tests among nodes: using the `files`
 configuration modifier (a very simple and straightforward way for most use cases), or
-using parallelism environment variables (aimed for  more complex scenarios).
+using parallelism environment variables (aimed at more complex scenarios).
+
+Either way, there are a few things to keep in mind:
+* `test` and `deploy` sections in your `circle.yml` are single-node only by
+  default and will only run on node 0. However, you can override `test` sections
+  per command (add `:` to the end of each command you'd like to parallelize, then
+  add a double-indented `parallel: true` below that). Alas, no parallel deploys.
```
eric-hu: > Alas, no parallel deploys.
I would word this differently.

Single node deploys are intentional and a good thing.  You only want one machine pushing code to production at a time.

jekyll/_cci1/parallel-manual-setup.md
```diff
@@ -144,7 +156,7 @@ fi
 
 This script partitions the test files into N equally sized buckets, and calls "test-runner" on the bucket for this machine. Note that you will still need to include `parallel: true` in `circle.yml` with this script.
 
-### Test suite split with knapsack gem
+<h3>Test suite split with knapsack gem</h3>
```
eric-hu: Actually, coming back around to this--I've seen advice inside and outside CCI discouraging style changes unless it's to make it consistent with the rest of the code.  In this case, I get that you're going for consistent `h2`, `h3` or `##`, `###` usage.

It looks like [other](https://raw.githubusercontent.com/circleci/circleci-docs/master/jekyll/_cci1/build-image-precise.md) [files](https://raw.githubusercontent.com/circleci/circleci-docs/eaad6390837e64fd962bcb92db3c8c34b32eecce/jekyll/_api/v1-reference.md) use the style "prefer hashes for headers unless an HTML id or class are needed".  I would say these `###` to `h3` changes are unnecessary, but your `h2` to `h3` change makes sense.

keybits:
Thanks for clarifying this Liene :-)

My commit confirms that we use Markdown for all formatting in docs. The title anchor links will be auto-generated by the markdown renderer at build time.

Also removed 'alas no parallel deploys' as described by Eric.

keybits:
@li3n3 this LGTM if you want to merge :-)

li3n3:
The magic of submitting a PR and going on vacation!

PR #889, opened 2017-03-21
Improve 'make this document better' links

### LINKED ISSUE #706: Update text for "Open an issue about this page"
Most issues that get created by our users have titles in the format of:

"RE: jekyll/_docs/<name_of_doc>"

While this does tell us _where_ the issue is, it doesn't do a good job of summarizing the issue for development and maintenance. Instead, let's have the issue pull just the name, then require the issue creator to specify what the problem is. Like so:

"<name_of_doc>: <action_item>"

keybits:
closes #706

This simplifies some text, adds a link to Discuss on all docs pages and makes the link to a new issue fill out the 'body' rather than the title:

image

This means the person submitting the issue will have to create a title (thus preventing lots of unhelpfully titled issues with file name links) while automatically providing a link to the referring page.

### DIFF
--- a/jekyll/_includes/edit-on-github.html
+++ b/jekyll/_includes/edit-on-github.html
@@ -1,10 +1,12 @@
 <br />
 <hr />
-<h2>Help make this document better...</h2>
-<p>This guide, as well as the rest of our docs, are open-source and available on <a href="{{ site.gh_url }}">GitHub</a>. We welcome <a href="{{ site.gh_url }}/blob/master/CONTRIBUTING.md&#35;contributing-to-circleci-docs">your contributions</a> to make these docs better.
+<h2>Help make this document better</h2>
+<p>This guide, as well as the rest of our docs, are open-source and available on <a href="{{ site.gh_url }}">GitHub</a>. We welcome your contributions.
 </p>
 <ul>
-	<li><a href="{{ site.gh_url }}/issues/new?title=RE%3A%20jekyll/{{ page.path }}" target=_blank>Open an issue about this page</a> to leave feedback.
-	<li><a href="{{ site.gh_url }}/blob/master/jekyll/{{ page.path }}">Suggest an edit of this page</a> if you have a way to make it better.
+	<li>Please read the <a href="{{ site.gh_url }}/blob/master/CONTRIBUTING.md&#35;contributing-to-circleci-docs" target="_blank">contributing guide</a>.
+	<li><a href="{{ site.gh_url }}/blob/master/jekyll/{{ page.path }}" target="_blank">Suggest an edit to this page</a>.
+	<li><a href="{{ site.gh_url }}/issues/new?body=This issue is about <https://circleci.com/docs{{ page.url }}>" target="_blank">Open an issue about this page</a> to report an error or problem.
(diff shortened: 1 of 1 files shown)


felicianotech:
Couple of notes:

- since you are putting in text for the body, which includes spaces, in the URL, we should URL encode the address. Chrome, and most browsers will understand and automatically do it for us when it loads the page but technically it's an improper URL.
- what do you think about using the filepath instead of the URL? My thought it, the filepath tells us exactly which file the issue is about, making it easier to find it and edit/fix it. The URL still needs to be "interpreted" before finding the file in the filesystem/git.
- should we include the contributing link here? GitHub will automatically show the user and tel them they should read it, as seen in your screenshot. Removing it could make this blurb slimmer.
- DISCUSS! I don't know what took us so long to include Discuss here. This can help ward of so many questions that belong in Discuss rather than GitHub. Thank you. =)

felicianotech:
Oh also, yes having that prefilled text in the body is so much better than the title. Nice.

jekyll/_includes/edit-on-github.html
```diff
@@ -1,10 +1,12 @@
...
+<h2>Help make this document better</h2>
+<p>This guide, as well as the rest of our docs, are open-source and available on <a href="{{ site.gh_url }}">GitHub</a>. We welcome your contributions.
 </p>
 <ul>
-	<li><a href="{{ site.gh_url }}/issues/new?title=RE%3A%20jekyll/{{ page.path }}" target=_blank>Open an issue about this page</a> to leave feedback.
-	<li><a href="{{ site.gh_url }}/blob/master/jekyll/{{ page.path }}">Suggest an edit of this page</a> if you have a way to make it better.
+	<li>Please read the <a href="{{ site.gh_url }}/blob/master/CONTRIBUTING.md&#35;contributing-to-circleci-docs" target="_blank">contributing guide</a>.
+	<li><a href="{{ site.gh_url }}/blob/master/jekyll/{{ page.path }}" target="_blank">Suggest an edit to this page</a>.
+	<li><a href="{{ site.gh_url }}/issues/new?body=This issue is about <https://circleci.com/docs{{ page.url }}>" target="_blank">Open an issue about this page</a> to report an error or problem.
```
smart-alek: "error" is a type of "problem", so maybe just shorten to "problem"?

jekyll/_includes/edit-on-github.html
```diff
@@ -1,10 +1,12 @@
...
+	<li><a href="https://discuss.circleci.com/" target="_blank">Visit our forum 'Discuss'</a> to ask question and search for other solutions.
```
smart-alek: "Question[s]" plural.

keybits:
Thanks for the suggestions. I think I addressed them all. New issue now looks like this:

image

Kept a link to 'contributing' next to 'Suggest an edit' since the edit screen doesn't display the 'contributing' banner.

jekyll/_includes/edit-on-github.html
keybits: Thanks - fixed

jekyll/_includes/edit-on-github.html
keybits: Thanks - fixed.

felicianotech:
"Kept a link to 'contributing' next to 'Suggest an edit' since the edit screen doesn't display the 'contributing' banner."

nice

PR #890, opened 2017-03-21
Allow search engines to start indexing 2.0 docs.

felicianotech:
This allows search engines like Google to index docs from the 2.0 section.

Also, the change in `jekyll/_cci1/index.md` is actually a bugfix I caught while I was in that file.


keybits:
LGTM

PR #888, opened 2017-03-21
renamed `setup_docker_engine` step

smaant:
`setup_docker_engine` step renamed into `setup_remote_docker`
Do not merge until https://github.com/circleci/build-agent/pull/318 released

PR #894, opened 2017-03-21
Revert "renamed `setup_docker_engine` step"

smart-alek:
Whoops, got trigger happy again. Reverts circleci/circleci-docs#888


smart-alek:
@smaant Sorry about this! When you're all ready with circleci/build-agent#318, then we'll revert this revert. 😢

PR #900, opened 2017-03-22
Change references to working_directory that made it sound optional. 

ndintenfass:


keybits:
LGTM

jekyll/_cci2/configuration-reference.md
```diff
@@ -31,7 +31,7 @@ Key | Required | Type | Description
 docker | Y <sup>(1)</sup> | List | Options for [docker executor](#docker-executor)
 machine | Y <sup>(1)</sup> | Map | Options for [machine executor](#machine-executor)
 steps | Y | List | A list of [steps](#steps) to be performed
-working_directory | Y | String | What directory to run the steps in (default: depends on executor)
+working_directory | Y | String | What directory to run the steps in. Can be short-handed as `workDir`.
```
keybits: Reading again - I think we should not mention `workDir` even as an option. It's old-style and will be removed at some point in the future. from Mahmood: "we will continue to support the deprecated `workDir` until we decide to sunset "old" syntax" < but I don't think we should include both versions in the docs.

felicianotech: I agree in not mentioning it. Less trouble in the future.

ndintenfass: The issue is that people are confused by snippets in the wild.  The thinking here is it's more important to prevent confusion then worry about some nebulous future where we may deprecate this token.

felicianotech: In that case, I think instead of "Can be short-handed as `workDir`", something like "previously `workDir`" or something similar is better. Still leaves a trail for people coming from older syntax, but doesn't appear to promote or support its use.

ndintenfass:
@felicianotech good idea - made change, now merging

PR #910, opened 2017-03-22
Add limitation of go-git

kimh:
Currently, we don't support source caching with go-git. This is because go-git doesn't support `git reset` command which is needed in order to handle checkout with source caching. Although I'm working on finding a solution,  currently we error out when source caching is found.


felicianotech:
Fixed the link. LGTM.

keybits:
LGTM - thanks Kim :-)

PR #918, opened 2017-03-23
The reference docs should link to the executor description page.

technomancy:
Also clarify that the primary container is the only place that steps can run currently.


technomancy:
Cool; thanks.

PR #919, opened 2017-03-23
Add heroku.com to ssh known host

browniebroke:
Without it, my build would be stuck on:
```
The authenticity of host 'heroku.com (50.xx.x.xx.xx)' can't be established.
RSA key fingerprint is 8b:48:5e:67:0e:c9:16:47:32:f2:87:0c:1f:c8:60:ad.
Are you sure you want to continue connecting (yes/no)?
```


keybits:
Thank you for adding this - much appreciated.

PR #922, opened 2017-03-24
split local dev setup out of README

smart-alek:
Addresses [this ticket](https://circleci.atlassian.net/browse/CIRCLE-3649).


README.md
```diff
@@ -1,91 +1,8 @@
 # CircleCI Documentation [![CircleCI Build Status](https://circleci.com/gh/circleci/circleci-docs.svg?style=shield)](https://circleci.com/gh/circleci/circleci-docs) [![GitHub license](https://img.shields.io/badge/license-MIT-blue.svg)](https://raw.githubusercontent.com/circleci/circleci-docs/master/LICENSE) [![CircleCi Community](https://img.shields.io/badge/community-CircleCI%20Discuss-343434.svg)](https://discuss.circleci.com)
 
-This is the public repository for <https://circleci.com/docs/>, a static website generated by [Jekyll](https://jekyllrb.com/). If you find any errors or have requests, you can set up a [local environment](#local-development) and make changes by following our [Contributing Guide](CONTRIBUTING.md).
+This is the public repository for <https://circleci.com/docs/>, a static website generated by [Jekyll](https://jekyllrb.com/). If you find any errors in our docs or have suggestions, please follow our [Contributing Guide](CONTRIBUTING.md) to submit an issue or pull request.
```
ndintenfass: We should probably add basic info on how to work on the repo to the README by convention -- feel free to just repeat your paragraph in CONTRIBUTING -- "If you already have a stable Ruby environment and feel comfortable installing dependencies, follow our [Jekyll guide](README-Jekyll). Otherwise, you can follow our [Vagrant guide](README-Vagrant) to set up a clean environment with all the project's neccessary dependencies."

README-Vagrant.md
```diff
@@ -0,0 +1,52 @@
+# Using Vagrant for Local Development
+
+If you don't want to install Jekyll, you can use Vagrant to set up a clean environment with all of the project's necessary dependencies.
+
+## Prerequisites
+- Vagrant: [download directly](https://www.vagrantup.com/downloads.html), `brew cask install vagrant`, or `sudo apt-get install vagrant`
+- VirtualBox: [download directly](https://www.virtualbox.org/wiki/Downloads), `brew cask install virtualbox`, or `sudo apt-get install virtualbox`
+
+## macOS Setup
```
felicianotech: This seems kind of out of no where with respect to heading placement.

felicianotech: If we are to have this, it should be under a heading such as `OS Specific Instructions`, and we should have Linux instructions as well.

Also, maybe @ndintenfass can comment here, but if they want to use Brew to install these tools, they likely already know how to install/already have brew installed.

If they are the type of user to need instructions on installing it (working with xcode, etc) then 1) I think we should link to their instructions and 2) They probably don't want/need brew and should just install these tools from the websites, like most users who don't already use Brew would.

smart-alek:
@felicianotech it's definitely hard to know how long a README should be. We want people to feel like it's easy to contribute, and @ndintenfass and I both felt that the length of README made it seem harder to get set up than it actually is.

re: your comments:

- I agree that the "macOS" header is a bit strange. Maybe we'll yank it out on the next iteration.
- I agree on the note about Brew. I really think we should be linking out to external docs when we can, which is what I did for the Jekyll way of things.
- I don't think we need a separate README for just altering things on GitHub. That's outlined pretty clearly in CONTRIBUTING.

Feel free to add some more PRs after this gets merged if you feel strongly about the other issues. I gotta move on. :)

felicianotech:
@smart-alek I wasn't suggesting a separate readme file (remember, I don't even want the multiple we have now). I was suggesting mentioning that in the main one, as a choice among the local Jekyll and Vagrant options.

smart-alek:
@felicianotech oh got it, makes sense; I'll add that.

ndintenfass:
@smart-alek is this ready to go?

smart-alek:
@ndintenfass yep, was just awaiting approval!

PR #927, opened 2017-03-24
bring back parallelism doc

smart-alek:
Bringing back doc that was linked in many places.


smart-alek:
Merging since this is producing several 404s.

PR #931, opened 2017-03-25
Adjust database command

dwightwatson:
First, indicate that the command is only relevant for Laravel (`artisan` is the Laravel command line utility).

Second, the commands can be merged into one using `migrate --seed`.

Third, `--force` will prevent Laravel asking "are you sure" before migrating, and have it just go ahead.

### DIFF
--- a/jekyll/_cci2/language-php.md
+++ b/jekyll/_cci2/language-php.md
@@ -187,15 +187,13 @@ Now that Composer has been installed, let's use it to install the project's depe
           command: php composer.phar install
 ```
 
-Next, let's initialize required database (DB) tables and seed it with initial data.
+Next, let's initialize required database tables and seed it with initial data (assuming we're using Laravel).
 
 ```YAML
 ...
       - run:
           name: Initialize Database
-          command: |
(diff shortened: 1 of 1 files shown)


ndintenfass:
@smart-alek this is perhaps our catalyst for making the configs in docs embedded versions of what's in Github, as accepting this PR (which we should) would make the code in docs different than in the sample project, which would be less than good.

smart-alek:
@ndintenfass yep. Changes to the sample projects should be submitted as PRs in their dedicated repos, and then those changes should propagate to docs.

Either way, thanks for the contrib @dwightwatson! :)

felicianotech:
For the record, the example project was specifically written for Lumen at the time, not Laravel (though they overlap heavily) and PHP in general.

Nathan, I agree.

Alek, we were taking about having the changes go through automatically due to embeds vs having to make these changes in multiple locations.

michelle-luna:
So, is the next step to update this doc with the embed instead?

michelle-luna:
this guide has been reworked for the new demo, so closing this PR now, thank you!

PR #932, opened 2017-03-26
Remove errant typos about Go projects in PHP and Python sample pages

ndintenfass:


ndintenfass:
self-merging, as this is trivial and embarrassing

PR #945, opened 2017-03-30
Destination is not mandatory anymore in store_artifacts step

kimh:


jekyll/_cci2/configuration-reference.md
```diff
@@ -503,7 +503,7 @@ Step to store artifacts (for example logs, binaries, etc) to be available in the
 Key | Required | Type | Description
 ----|-----------|------|------------
 path | Y | String | Directory in the primary container to save as build artifacts
-destination | Y | Prefix added to the artifact paths in the artifacts API
+destination | N | Prefix added to the artifact paths in the artifacts API
```
smaant: Could you please also describe default value in case of not specified?

ndintenfass: Near as I can tell, the behavior is to put the artifacts in the same path as the working directory?

smaant: Not working directory, but the `path`

kimh:
I'm working on something else now and will come back to take care of feedbacks here

kimh:
Feedbacks were addressed. Merging.

PR #946, opened 2017-03-30
Add AWS CLI page

kimh:
Now 2.0 uses AWS credentials configured from the project setting page.

I'm not sure if having a dedicated page for this is good since the page is very sparse. If there are better places, please let me know.

### DIFF
--- a/jekyll/_cci2/configure_aws_cli.md
+++ b/jekyll/_cci2/configure_aws_cli.md
@@ -1,0 +1,11 @@
+---
+layout: classic-docs
+title: "Configuring AWS CLI"
+short-title: "Configuring AWS CLI"
+categories: [configuring-jobs]
+order: 90
+---
+
+You can add your AWS credentials from **Project Settings > AWS Permissions** page.
+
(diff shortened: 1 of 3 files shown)


keybits:
Thanks for this Kim. I think it should go in a doc called 'Deployment Integrations'.

I was already thinking we should update the 2.0 sidebar menu sections. I would like a 'Deploying' section that would likely have the following pages:

## Deploying

- Deployment Integrations (AWS, Heroku etc)
- Deploying Manually (currently called 'Deployments')
- Building Docker Images (move from the 'Using Docker' section)

Then we can discuss the various deployment targets (e.g. GAE, GKE etc) that we would like examples for in the 'Deploying manually' section.

@ndintenfass thoughts?

ndintenfass:
I do think for now a section on Deployment would make a lot of sense, but if it's quicker we could do a single page under Configuration for Deployment, where this is a section, and then blow it out to pages in a section as that grows.

kimh:
@keybits @ndintenfass I created a "Deploying" section and add AWS CLI stuff in a dedicated "Deployment Integrations" page. Let me know what you think.

keybits:
Thanks Kim. LGTM.

Once merged I'll add in the other pages for this section and fill out the page added by this PR with a bit more info.

PR #954, opened 2017-04-04
fixed the location of the libphp5.so

kanhaic:
The location provided of libphp5.so provided is incorrect. just copying this example fails. I have sshed into the vm and found the correct location that worked for me.


keybits:
Thank you for this. You're correct with the new path - the old value was for the now outdated 12.04 image.

PR #959, opened 2017-04-05
Add how to set timezone in Docker to 2.0 FAQ

keybits:


kimh:
It would be great if you can add this to https://circleci.com/docs/2.0/migrating-from-1-2/ as a context of converting `machine:timezone` config in 1.0.

keybits:
Thanks Kim - will do the above

PR #961, opened 2017-04-06
ios-build-on-osx article spelling updates

JamesRagnar:
A couple spelling errors


zzak:
Thank you @JamesRagnar! Nice catch :)

PR #962, opened 2017-04-06
Fix chown command for GKE tutorial

gkop:
Beginning 2017-04-05, Circle no longer seems to have a `/home/ubuntu` directory, so `.kube/` is no longer located there. I found `.kube/` now in `~/.kube`. By chowning the new `.kube/` location, the kubectl command functions correctly again, rather than complaining about missing credentials.


gkop:
Actually this wasn't what fixed it for me. Running kubectl as sudo once does something that makes kubectl work subsequently without sudo. As a hacky fix I'm just going to run kubectl once as sudo :/

PR #968, opened 2017-04-10
wget instead of curl

chrisclark:
curl dies immediately if the connection is refused (e.g. as the app is starting). wget can retry on connection refused, which is much more robust during an app boot.


keybits:
Thank you for suggesting this. I agree this is a better approach. (Some extra context on `wget` retry behaviour here: https://gist.github.com/flomotlik/6212725)

PR #970, opened 2017-04-10
First pass at auto-documenting CircleCI Images

ndintenfass:
Add plugin to pull external json into site.data, then pull the convenience image data, and add a page that renders it


ndintenfass:
Feedback needed on:

1) Layout and information -- should we be including the detailed description?

2) This happens at render time, not in the browser -- the nice thing is the site is still wholly static, but also means if the data changes we need to re-render to freshen it. Doing it in the browser will require a very different approach.

keybits:
LGTM. I think for now it's fine as it is.

re: 1. less information is fine - people will need to click through to Docker Hub anyway

re: 2. I like that it keeps the site static. We ca look into automatically triggering rebuilds later.

PR #977, opened 2017-04-12
Add a third party info include banner

keybits:
Adds an include that will display like this:

image

Call it with:

```
{% include third-party-info.html app='GitHub'%}
```


keybits:
@ndintenfass feedback on wording appreciated.

PR #978, opened 2017-04-12
Add a note about language v database images

keybits:
It would be nice to eventually split the 'available' images listing so it differentiates between 'primary' images and 'service' images. This note clarifies things for now.

### DIFF
--- a/jekyll/_cci2/circleci-images.md
+++ b/jekyll/_cci2/circleci-images.md
@@ -26,6 +26,9 @@ NOTE: We strongly recommend that you lock your image to a particular tag to avoi
 
 
 ## Available Images
+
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/circleci-images.md
```diff
@@ -26,6 +26,9 @@ NOTE: We strongly recommend that you lock your image to a particular tag to avoi
 
 
 ## Available Images
+
+**Note:** The 'language' images would usually be used as your 'primary' container. The 'database' images are best used as a secondary 'service' container.
```
kimh: ~~ I prefer to saying 'container 0' instead of 'primary' these days.

PR #950, opened 2017-04-02
Add examples how to use balancing libraries for 2.0

ArturT:
I've updated info for https://circleci.com/docs/2.0/parallelism-faster-jobs/

PR #985, opened 2017-04-13
Add examples how to use balancing libraries for 2.0

keybits:
We added a Jekyll include after the original PR, so this is a rebase of: https://github.com/circleci/circleci-docs/pull/950

Thanks for your input @ArturT


ArturT:
Awesome! :)

PR #1017, opened 2017-04-28
Fix incorrect indent in config.yml example

andrewtamura:
Commit 2245bc8d409e0c3180bec6179ca49be065e589b2 introduced an extra white space.


smart-alek:
Thanks for the fix @andrewtamura!

PR #1023, opened 2017-05-06
Add Premium Support links to match inner

zzak:
This patch adds three things:

* Missing link to Premium Support in the header nav (to match inner)
* Link "Support" in "/docs" index page to Premium Support (Only non-linked header on page)
* Adds "Premium Support" link to footer (This is new, open to removing it)

screenshot from 2017-05-09 02-23-45

screenshot from 2017-05-09 02-26-09


keybits:
LGTM

PR #1025, opened 2017-05-08
Update sidebar.html

thiemo:
Changed the drop down menu so that it respects collections:output


felicianotech:
@thiemo  This change works,, but why do you think it's necessarily? None of the collections that CircleCI Docs currently use have output set to false. Thoughts?

Thanks in advance.

felicianotech:
@thiemo bump

thiemo:
@felicianotech you're right, but on the other hand showing or hiding the collections is indeed implemented on the overview page [here](https://circleci.com/docs/) although no collection is set to false at this time :-) To me it was a matter of consistency

felicianotech:
Not sure where you're pointing to. Th 2.0 index page is manually made (for some reason). I don't see us using collections that aren't outputted in the near future so I'm going to go ahead and close this one.

PR #1029, opened 2017-05-10
Clarify that we look for skip in title and description

keybits:


keybits:
Thanks for the suggestions @michelle-luna 

I made a more extensive rewrite of the document. I didn't add it to the troubleshooting guide since the topic is self-contained and the issue the customer was having didn't result in any error messages. The customer referred to this page since it didn't explicitly say that we ignore `ci skip` in the commit description.

The note about Git titles and descriptions is, I think, useful here since the naming around this is not consistent and I've seen customers unsure about it.

PR #1031, opened 2017-05-10
Update beta-premium-feature.html for clarity

michelle-luna:
Clarify premium banner


michelle-luna:
And this should be WIP because the other doc that includes this banner is for non-paid customers (use-custom-images), so I need to make another banner for that first.

michelle-luna:
It is actually Machine Executor and I think the better approach is to simply not use a banner in that section of the executor-types.md file. A banner for one section of one article doesn't make good sense to me. I will just make updates in the text and see what folks think.

PR #1035, opened 2017-05-10
Move custom-images under docker category

zzak:
Current:
screenshot from 2017-05-10 14-46-44

This moves the custom images doc out of the "configuing jobs" category which is already growing a bit long in the tooth. Also, the only relevant part to actual configuration here is where you can specify the custom image, but I honestly forgot where this page was -- it makes more sense under "docker" IMHO <3


zzak:
Thank you @felicianotech @michelle-luna :bow:

PR #1042, opened 2017-05-14
Note that gcloud may deploy extra artifacts

vochicong:
During the testing phase, extra artifacts (not git commited) may be generated on CircleCI,
and will be deployed by `gcloud app deploy` to App Engine.
This behavior is normally not expected, so it should be noted.


vochicong:
You're welcome 👍

PR #1048, opened 2017-05-18
Local dev with docker-compose and build with CircleCI 2.0

keybits:
This adds support for developing locally with Docker.

README files have been updated and references to Vagrant removed.

Support for building and testing using CircleCI 2.0 is included and 1.0 config has been removed.

For local development with `docker-compose up` we're using the official Jekyll Docker image since it exposes volumes and is small (based on Alpine).

For testing on CircleCI we're using the CircleCI Ruby 2.3.3 convenience image.


keybits:
Note: once this is merged we can delete the `cci2-workflow` branch as this supersedes that.

keybits:
Rationale for using the Jekyll official Docker image for local dev includes:

- it has all dependencies baked in. CircleCI local builds don't use cache so dependencies would have to be installed on each run if we used the CircleCI image
- it is 275MB smaller than the CircleCI image since it's Alpine based so for people on slow connections it's quicker to get up and running
- it has a volume exposed to aid local development - exposing volumes on CircleCI 2.0 does not work since there is no 'host' volume. If we used the CircleCI image we'd need to adapt it to expose a volume locally and run differently in a local v CI environment (this could potentially be done in the future) 
- updating for new versions of Ruby or Jekyll should be straightforward - the official Jekyll image will be updated and the CircleCI image will be updated as new versions of Ruby are released. All we need to do is change the image version numbers in `.circleci/config.yml` and `docker-compose.yml`.

README-local-development.md
```diff
@@ -0,0 +1,43 @@
...
+
+The recommended way to work on the CircleCI docs locally is to use Docker.
+
+1. Install Docker for you platform: <https://docs.docker.com/engine/installation/>
+2. Clone the CircleCI docs repo: `git clone https://github.com/circleci/circleci-docs.git`
+3. `cd` into the directory where you cloned the docs
+4. run `docker-compose up`
+5. The docs site will now be running on <http://localhost:4000/docs/>
+
+**Note:** If you want to submit a pull request to update the docs, you'll need to [make a fork]((https://github.com/circleci/circleci-docs#fork-destination-box) of this repo and clone your version in step 2 above. Then when you push your changes to your fork you can submit a pull request to us.
```
michelle-luna: extra open parentheses here?

keybits: Thanks and good catch :-) Fixed.

Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '2.3.1'
+ruby '2.3.3'
 
-gem 'jekyll'
-gem 'html-proofer'
```
felicianotech: Why are we setting these versions here when we're using a lock file?

felicianotech: This is suppose to be what it was changed too, not the old lines. :/

keybits: I believe it's useful to specify versions in the Gemfile - see: https://everydayrails.com/2012/09/11/bundler-rails-specify-versions.html

Especially in a project where we're expecting people will install locally and might run `bundle update`.

felicianotech: I know what it's used for. If they run Bundle update, their local lockfile gets updated and we'll see that in the PR. This doesn't stop that but holds us back from keeping uptodate with these Gems when we do run it.

PR #1061, opened 2017-05-19
Update 2.0 language docs

keybits:
This fixes links to all the main language demos.

The language docs themselves have been simplified since they contained inaccurate and outdated information. We will expand the details in the docs over the coming days.

I'm hoping to get this merged on the principle that it's better to have limited but accurate information instead of extensive outdated information.

### DIFF
--- a/jekyll/_cci2/index.md
+++ b/jekyll/_cci2/index.md
@@ -11,12 +11,12 @@ Welcome to CircleCI 2.0 Beta! The Beta release of CircleCI 2.0 includes many imp
 
 Code that builds on Linux will generally build on CircleCI 2.0. For some language versions, CircleCI provides demo applications and Docker images for primary container use, as follows: 
 
-- **Clojure 1.8 and later**, see <https://github.com/circleci/cci-demo-clojure> and <https://hub.docker.com/r/circleci/clojure>
-- **Go 1.7 and later**, see <https://github.com/circleci/cci-demo-go> and <https://hub.docker.com/r/circleci/golang>
-- **Node.js 4 and later**, see <https://github.com/circleci/cci-demo-react> and <https://hub.docker.com/r/circleci/node>
-- **PHP 5 and later**, see <https://github.com/circleci/cci-demo-lumen> and <https://hub.docker.com/r/circleci/php>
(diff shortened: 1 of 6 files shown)


jekyll/_cci2/index.md
```diff
@@ -11,12 +11,12 @@ Welcome to CircleCI 2.0 Beta! The Beta release of CircleCI 2.0 includes many imp
 
 Code that builds on Linux will generally build on CircleCI 2.0. For some language versions, CircleCI provides demo applications and Docker images for primary container use, as follows: 
 
-- **Clojure 1.8 and later**, see <https://github.com/circleci/cci-demo-clojure> and <https://hub.docker.com/r/circleci/clojure>
-- **Go 1.7 and later**, see <https://github.com/circleci/cci-demo-go> and <https://hub.docker.com/r/circleci/golang>
-- **Node.js 4 and later**, see <https://github.com/circleci/cci-demo-react> and <https://hub.docker.com/r/circleci/node>
-- **PHP 5 and later**, see <https://github.com/circleci/cci-demo-lumen> and <https://hub.docker.com/r/circleci/php>
-- **Python 2 and later**, see <https://github.com/circleci/cci-demo-flask> and <https://hub.docker.com/r/circleci/python>
-- **Ruby 2 and later**, see <https://github.com/circleci/cci-demo-rails> and <https://hub.docker.com/r/circleci/ruby> 
+- **Clojure 1.8 and later**, see <https://github.com/CircleCI-Public/circleci-demo-clojure-luminus> and <https://hub.docker.com/r/circleci/clojure>
```
technomancy: Minor nitpick: there's no need to specify a version of Clojure we support; any version will work.

keybits: Agree - will fix in future update

michelle-luna: People do need the version numbers and they leave the docs site trying to find them in the usability study. They are looking for it at the start of their evaluation of the product. So, I need to see a few more studies conducted with this content intact.

technomancy: I see; that makes sense. In that case instead of removing it we should specifically state "versions 1.0 and later" in order to not imply that there is a minimum version required.

PR #1065, opened 2017-05-19
Add gotcha for modifying `$PATH` on 2.0

zzak:


jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -101,6 +101,39 @@ jobs:
...
+    docker:
+      - image: buildpack-deps:jessie
+    environment:
+       BASH_ENV: ~/.bashrc
+  steps:
+    run: echo "export PATH=/path/to/foo/bin:$PATH" > ~/.bashrc
+    run: some_program_inside_of /path/to/foo/bin
+```
+
+In this example, we've added `/path/to/foo/bin` to the `$PATH` which will get evaluated when a command is executed. You can image that `some_program_inside_of` is an executable within that path, and we don't need to refer to it by the full path name.
```
michelle-luna: image -> imagine?

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -101,6 +101,39 @@ jobs:
...
+version: 2
+jobs:
+  build:
+    working_directory: /tmp
+    docker:
+      - image: buildpack-deps:jessie
+    environment:
+       BASH_ENV: ~/.bashrc
+  steps:
+    run: echo "export PATH=/path/to/foo/bin:$PATH" > ~/.bashrc
```
notnoopci: actually, I would recommend changing the line to

```
echo 'export PATH=/path/to/foo/bin:$PATH' >> $BASH_ENV
```

Also, note `>>` (to avoid overwriting the file) and use of single-quotes `'` to allow for delayed expansion if that's desired.

zzak: Aww yeah, good point. Thanks for the review! :bow:

jekyll/_cci2/migrating-from-1-2.md
zzak: :+1:

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -101,6 +101,39 @@ jobs:
...
+
+```YAML
+version: 2
+jobs:
+  build:
+    working_directory: /tmp
+    docker:
+      - image: buildpack-deps:jessie
+    environment:
+       BASH_ENV: ~/.bashrc
```
notnoopci: `BASH_ENV` is already defined by default, you don't need to set it

felicianotech: It's set to some file in `/tmp` though. Check with `machine` as well as a Docker image.

notnoopci: What's downside of it being in `/tmp`?

felicianotech: Not sure. It's tmp? In the above example, the one in `/tmp` has some random ID generated in the file name. The user would have to find that and add to it versus something they could include in their repo?

notnoopci: I don't think users need to find it though or check it into the repository (`~/.bashrc` isn't in the repo either).  Users who need to modify it can run commands like `echo 'export FOO=BAR' >> $BASH_ENV` without needing to know where it lives.

felicianotech: Good point. If that works, I like that better.

PR #1068, opened 2017-05-19
Fix doc boog where rspec formatter missing from config

zzak:
Originally pointed out here by @conwayje:
https://discuss.circleci.com/t/metadata-collection-parallelization-leads-to-time-out-for-rspec/12720

Unless you specify a formatter which actually outputs to STDOUT, which junit formatter doesnt,
your rspec command will eventually be killed if it takes longer than 10 minutes (without output).

Also, this patch addresses unnecessary require and instructions for rspec_junit_formatter gem


keybits:
Good catch :-) LGTM

PR #1073, opened 2017-05-23
2.0 Documentation for Google Cloud/GKE

sha1sum:
This PR introduces Google Cloud configuration documentation for CCI 2.0 for Google Cloud Platform (in the way of general authentication instructions for service accounts) and Google Container Engine.

This document draws on various sources, most notably the [1.0 documentation for GKE](https://circleci.com/docs/1.0/continuous-deployment-with-google-container-engine/).

### DIFF
--- a/jekyll/_cci2/google-container-engine.md
+++ b/jekyll/_cci2/google-container-engine.md
@@ -1,0 +1,78 @@
+---
+layout: classic-docs
+title: "Google Container Engine"
+short-title: "Google Container Engine"
+categories: [deploying]
+description: "Setting up continuous deployment with Google Cloud Platform's Container Engine with CircleCI 2.0."
+order: 30
+---
+
+In order to use Google Cloud, you will need to ensure that the [Google Cloud SDK](https://cloud.google.com/sdk/) is installed on your primary container.
+
+## Prerequisites
+
+This document makes the following assumptions:
+
+1. You have a working knowledge of Docker and building Docker images.
+1. You already have a GCP project registered. Keep the project name handy.
+1. A Container Engine cluster has already been created in your GCP project. Keep the cluster name handy.
+1. Your repository has already been configured as a CircleCI 2.0 project.
+
+## Selecting a Base Image
+If Debian is acceptable as a base for your custom primary container, Google's [`google/cloud-sdk`](https://hub.docker.com/r/google/cloud-sdk/) image can be a good base image to use.
+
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/google-container-engine.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Configuring `gcloud`
+
+As part of your deployment commands, configure `gcloud` to use the newly-configured service account and set up the appropriate defaults:
+
+```sh
+gcloud auth activate-service-account --key-file ${HOME}/gcp-key.json
+gcloud --quiet config set project ${GOOGLE_PROJECT_ID}
+gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
+gcloud --quiet config set container/cluster ${GOOGLE_CLUSTER_NAME}
```
ahmetb: this line should not be necessary

jekyll/_cci2/google-container-engine.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Configuring `gcloud`
+
+As part of your deployment commands, configure `gcloud` to use the newly-configured service account and set up the appropriate defaults:
+
+```sh
+gcloud auth activate-service-account --key-file ${HOME}/gcp-key.json
+gcloud --quiet config set project ${GOOGLE_PROJECT_ID}
+gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
```
ahmetb: you can add `--zone` to get-credentials below and avoid this line, too.

jekyll/_cci2/google-container-engine.md
sha1sum: If multiple clusters are present and users are performing cluster operations from CircleCI, then this is necessary to set the proper cluster for the operations. This was taken from [the GKE documentation on cluster operations](https://cloud.google.com/container-engine/docs/clusters/operations#switching_between_clusters) and I would argue that it's better left in because it does not harm anything and also ensures that the cluster operations performed are being executed against the proper cluster.

For instance, we're using this to create resources on a GKE cluster, and we have multiple clusters.

jekyll/_cci2/google-container-engine.md
sha1sum: @ahmetb Is this something I should be taking action on or information only?

jekyll/_cci2/google-container-engine.md
ahmetb: @sha1sum you're parsing cluster name to the get-credentials command below, so this line I commented on is completely unnecessary.

PR #1076, opened 2017-05-23
Update provision-builder script references

bellkev:
Looks like a reference to an older provision-builder script got missed.


GERey:
lgtm

PR #1078, opened 2017-05-24
Migration section

michelle-luna:
This is a huge wad of edits to create a new Migration section, using most of the content we had already, but moved around a bit with more formal, declarative language and instructions. 

Migration
-Migration FAQ (removed troubleshooting and added 'Why Migrate?' content from the walkthrough)
-Sample 2.0 config.yml File (new overview of jobs and steps with javascript demo app config.yml example)
-Project Tutorial (renamed walkthrough and edited to streamline it and have sections build on each other a bit more, while removing info covered in the getting started docs and moving manual deploy steps to the end)
-Migrating from 1.0 to 2.0 (re-ordered and changed tack to first add version and re-nest all sections, then search and replace the changed/equivalent keys)

thanks,
Michelle

### DIFF
--- a/jekyll/_cci2/faq.md
+++ b/jekyll/_cci2/faq.md
@@ -12,7 +12,7 @@ order: 1
 
 ## Why Migrate from CircleCI 1.0 to 2.0?
 
-- No more queued jobs! CircleCI 2.0 includes a significant rewrite to resource utilization that eliminates waiting for the last Job to use available containers by breaking Jobs into Steps. 
+- CircleCI 2.0 includes a significant rewrite of container utilization to run more jobs faster and to prevent available containers from sitting idle. 
 - In 2.0, Jobs are broken into Steps. You compose these Steps within a Job at your discretion, giving you greater flexibility to run your build the way you want. 
 - 2.0 Jobs support almost all public Docker images. You can also create a custom image and run jobs on it or compose multiple images together (like MySQL 5.7 + Redis 3.2) and run jobs on them as if they were a single image.
 - If your job uses Docker, CircleCI runs your job on a dedicated VM so you can access all of Docker’s features. 
@@ -107,5 +107,6 @@ jobs:
 In this example, we're setting the timezone for both the primary image and an additional mySQL image.
 
 A full list of available timezone options is [available on Wikipedia](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).
+
 
 
--- a/jekyll/_cci2/faq.md
+++ b/jekyll/_cci2/faq.md
@@ -1,41 +1,62 @@
 ---
 layout: classic-docs
-title: "Frequently Asked Questions (FAQ)"
-short-title: "FAQ"
-description: "Frequently asked questions about CircleCI 2.0"
-categories: [getting-started]
-order: 50
+title: "Migration FAQ"
+short-title: "Migration FAQ"
+description: "Frequently asked questions about CircleCI 2.0 Migration"
+categories: [migration]
+order: 1
 ---
+
+* Contents
+{:toc}
 
 ## Can I try CircleCI 2.0 while still using 1.0?
 
 Yep! If you're not ready to fully commit to 2.0, you can easily try it while still building on 1.0:
 
-Create a new branch and add 2.0 configuration as described [here](https://circleci.com/docs/2.0/project-walkthrough/). When you push the branch with 2.0 configuration, your project will build on CircleCI 2.0. All other branches will continue building on CircleCI 1.0.
+Create a new branch and add 2.0 configuration as described in the [2.0 Project Tutorial](https://circleci.com/docs/2.0/project-walkthrough/). When you push the branch with 2.0 configuration, your project will build on CircleCI 2.0. All other branches will continue building on CircleCI 1.0.
 
 If you'd like to completely revert to 1.0 configuration, simply replace `.circleci/config.yml` with a 1.0 `circle.yml` file.
 
-## What operating systems does CircleCI support?
+## What operating systems does CircleCI 2.0 support?
 
-**Linux:** CircleCI is flexible enough that you should be able to build most applications that run on Linux. These don't have to be web applications!
+- **Linux:** CircleCI is flexible enough that you should be able to build most applications that run on Linux. These don't have to be web applications!
 
-**iOS:** Building iOS apps is not yet supported on CircleCI 2.0. Please refer to our documentation for [iOS on 1.0]({{ site.baseurl }}/1.0/mobile/) until 2.0 support is available.
+- **iOS:** Building iOS apps is not yet supported on CircleCI 2.0. Please refer to our documentation for [iOS on 1.0]({{ site.baseurl }}/1.0/mobile/) until 2.0 support is available.
 
-**Android:** There is currently no official support for Android on CircleCI 2.0, but users have still been able to build Android apps on 2.0. [This post](https://discuss.circleci.com/t/thank-you-and-android-build-example/11298) is an excellent example of one our users successfully building an Android app on 2.0.
+- **Android:** There is currently no official support for Android on CircleCI 2.0, but users have still been able to build Android apps on 2.0. [This post](https://discuss.circleci.com/t/thank-you-and-android-build-example/11298) is an excellent example of one our users successfully building an Android app on 2.0.**Note that this example does not use the emulator, so it's not affected by incompatibility with Docker.**
 
-**Note that this example does not use the emulator, so it's not affected by incompatibility with Docker.**
+- **Windows:** We do not yet support building and testing Windows applications.
 
-**Windows:** We do not yet support building and testing Windows applications.
+## How do I migrate from Jenkins to CircleCI 2.0?
+
+Start with the [Hello World doc]({{ site.baseurl }}/2.0/hello-world/), then add `steps:` to duplicate your project exactly as it is in Jenkins, for example:
+
+    steps:
+      - run:
+        command: "Add any bash commands you want here"
+    test:
+      - run:
+        command: |
+          "More arbitrary multi-line bash"
+        command: |
+          "Probably copy-pasted from 'Execute Shell' on Jenkins"
 
 ## Does CircleCI 2.0 run inference commands?
 
 Currently, CircleCI 2.0 doesn't infer anything from your project, but we have plans to replace our inference system before moving out of Beta. Until then, you'll need to manually configure all jobs.
 
-## My project is running on CircleCI 2.0, but the build is frozen!
+## Can I use CircleCI 2.0 without creating base images?
 
-Builds often freeze due to syntax errors in `config.yml`.
+Yes, you can use* one of ours!
 
-Cancel the build, check your `config.yml` for proper indentation, and ensure that all jobs and steps have the required keys.
+The `circleci/build-image:ubuntu-14.04-XL-922-9410082` image has the same content as the Ubuntu Trusty 14.04 image our web app uses. Just know that the image is fairly large (around 17.5 GB uncompressed), so it’s less ideal for local testing.
+
+The image defaults to running actions as the `ubuntu` user and is designed to work with network services provided by Docker Compose.
+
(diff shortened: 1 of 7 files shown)


jekyll/_cci2/faq.md
```diff
@@ -1,41 +1,69 @@
...
+categories: [migration]
+order: 1
 ---
 
+* Contents
+{:toc}
+
+## Why Migrate from CircleCI 1.0 to 2.0?
+
+- No more queued jobs! CircleCI 2.0 includes a significant rewrite to resource utilization that eliminates waiting for the last Job to use available containers by breaking Jobs into Steps.
```
keybits: I'm not sure we can promise 'No more queued jobs' - that will still happen if we have infrastructure issues or if the customer is submitting more builds then their plan allows. This sounds like it's about removing 'lock-step'. I don't have good suggestions on how to explain this here, but suggest removing 'No more queued jobs!' for now.

jekyll/_cci2/faq.md
```diff
@@ -1,41 +1,69 @@
...
+order: 1
 ---
 
+* Contents
+{:toc}
+
+## Why Migrate from CircleCI 1.0 to 2.0?
+
+- No more queued jobs! CircleCI 2.0 includes a significant rewrite to resource utilization that eliminates waiting for the last Job to use available containers by breaking Jobs into Steps. 
+- In 2.0, Jobs are broken into Steps. You compose these Steps within a Job at your discretion, giving you greater flexibility to run your build the way you want.
```
keybits: It's my understanding that steps still run in sequence - which is the same as 1.0.

Workflows will allow the flexibility suggested here, but that's not possible yet. Steps will still be in sequence but 'jobs' will be composable in arbitrary orders.

jekyll/_cci2/faq.md
michelle-luna: thanks, how about 'CircleCI 2.0 includes a significant rewrite of container utilization to run more jobs faster and to prevent available containers from sitting idle.'

jekyll/_cci2/faq.md
michelle-luna: yes, this is a bit of spin on configuring jobs and steps as a benefit, but I think it is still a reason to migrate to 2.0. Thanks for this feedback, keep it coming!

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -3,332 +3,133 @@ layout: classic-docs
...
-machine:
-  timezone: Europe/Dublin
-```
-
-**2.0**
-
-You can set the timezone in Docker images with the `TZ` environment variable. The value should be something like: `TZ="/usr/share/zoneinfo/America/Los_Angeles"`. A full list of available timezone options is [available on Wikipedia](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).
-
-In your `.circleci/config.yml` it would look like:
+4. Replace `machine:` with the following two lines, nesting all of the following sections under `build`:
```
michelle-luna: per conversation with Nathan, this might be possible for some users, but machine isn't required, so I will change this to make sure users know they must have the jobs: build: line for 2.0.

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -3,332 +3,133 @@ layout: classic-docs
...
-    working_directory: ~/your-dir
-    environment:
-      TZ: "/usr/share/zoneinfo/America/Los_Angeles"
 ```
 
-Note that in the above example we are setting the timezone for both the primary image and an additional mySQL image.
+## Search and Replace Keys
 
-### Global Environment Variables
+- Replace the language and version associated with your application, for example:
```
michelle-luna: per conversation with Nathan, language and version is also not required in 1.0, so I will rewrite this as 'you must add' for 2.0

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -3,332 +3,133 @@ layout: classic-docs
...
+```
+     - deploy:
 ```
 
-**2.0**
-
-Currently, 2.0 doesn't support automatic deployment via integrations (like the above Heroku example).
-
-You can write your own manual `deploy` steps as shown in the [Configuration Reference]({{ site.baseurl }}/2.0/configuration-reference/#deploy).
+**Note:** Currently, 2.0 Beta does not include integration support. To write your own manual `deploy` steps, refer to the [Configuration Reference]({{ site.baseurl }}/2.0/configuration-reference/#deploy).
```
michelle-luna: per Nathan, this should be referred to as 'continuous deployment' instead of integration support.

jekyll/_cci2/faq.md
```diff
@@ -1,41 +1,69 @@
...
 
 Currently, CircleCI 2.0 doesn't infer anything from your project, but we have plans to replace our inference system before moving out of Beta. Until then, you'll need to manually configure all jobs.
 
-## My project is running on CircleCI 2.0, but the build is frozen!
+## Can I use CircleCI 2.0 without creating base images?
+
+Yes, you can use* one of ours!
 
-Builds often freeze due to syntax errors in `config.yml`.
+The `circleci/build-image:ubuntu-14.04-XL-922-9410082` image has the same content as the Ubuntu Trusty 14.04 image our web app uses. Just know that the image is fairly large (around 17.5 GB uncompressed), so it’s less ideal for local testing.
```
keybits: minor suggestion. 17.GB is VERY large IMO.

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -3,332 +3,131 @@ layout: classic-docs
...
+     ```
+5. Add the language and version to your configuration using either the `docker:` and `- image:` keys in the example or by setting `machine: true`. If your configuration includes language and version as shown for `ruby:` below, replace it as shown.
+     ```
+       ruby:
+         version: 2.3
+     ```
+     Replace with the following two lines:
+     ```
+         docker:
+           - image: ruby:2.3
```
keybits: Suggest using the CircleCI image here: `circleci/ruby:2.3`

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -3,332 +3,131 @@ layout: classic-docs
...
+     ```
+     Or
+
+     ```
+    machine: true
```
keybits: not sure we should be suggesting to use `machine: true` without context. People should default to `docker` unless they really need the machine executor.

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -3,332 +3,131 @@ layout: classic-docs
...
+     checkout:
+       post:
+     ```
+     With the following:
+     ```
+         steps:
+           - checkout
+           - run:
+     ```
+7. Validate your YAML at <http://codebeautify.org/yaml-validator> to check the changes.
```
keybits: We should perhaps recommend using the local build tool to do the validation? `circleci config validate` : https://circleci.com/docs/2.0/local-jobs/#usage

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -3,332 +3,131 @@ layout: classic-docs
...
-
-You can write your own manual `deploy` steps as shown in the [Configuration Reference]({{ site.baseurl }}/2.0/configuration-reference/#deploy).
+**Note:** Currently, 2.0 Beta does not include continuous deployment support. To write your own manual `deploy` steps, refer to the [Configuration Reference]({{ site.baseurl }}/2.0/configuration-reference/#deploy).
 
-If you **add SSH keys** via the CircleCI UI 'Project Settings > SSH Permissions' screen for deployment with 2.0, you must **explicitly add them to your job with: `- add_ssh_keys`** in `.circleci/config.yml`. See [Configuration Reference add SSH keys]({{ site.baseurl }}/2.0/configuration-reference/#add_ssh_keys). The documentation also explains how to activate `ssh-agent` so that you can `ssh-add` your keys for tools that require them to be available via an agent.
+- Validate your YAML again at <http://codebeautify.org/yaml-validator> to check the changes. Fix up any issues and commit the updated `.circle/config.yml` file. CircleCI automatically starts a test run on CircleCI 2.0 with a 2.0 icon on the builds page for that run.
```
keybits: As above - suggest using local build tool for this: https://circleci.com/docs/2.0/local-jobs/#usage (possibly as well as an online yaml validator)

jekyll/_cci2/faq.md
```diff
@@ -1,41 +1,69 @@
...
 
+* Contents
+{:toc}
+
+## Why Migrate from CircleCI 1.0 to 2.0?
+
+- CircleCI 2.0 includes a significant rewrite of container utilization to run more jobs faster and to prevent available containers from sitting idle. 
+- In 2.0, Jobs are broken into Steps. You compose these Steps within a Job at your discretion, giving you greater flexibility to run your build the way you want. 
+- 2.0 Jobs support almost all public Docker images. You can also create a custom image and run jobs on it or compose multiple images together (like MySQL 5.7 + Redis 3.2) and run jobs on them as if they were a single image.
+- If your job uses Docker, CircleCI runs your job on a dedicated VM so you can access all of Docker’s features.
```
keybits: This is incorrect. We only use a dedicate VM is executory type is machine. Suggest removing this line since the  machine executor is for special use cases.

jekyll/_cci2/faq.md
```diff
@@ -1,41 +1,69 @@
...
 ---
 
+* Contents
+{:toc}
+
+## Why Migrate from CircleCI 1.0 to 2.0?
+
+- CircleCI 2.0 includes a significant rewrite of container utilization to run more jobs faster and to prevent available containers from sitting idle. 
+- In 2.0, Jobs are broken into Steps. You compose these Steps within a Job at your discretion, giving you greater flexibility to run your build the way you want. 
+- 2.0 Jobs support almost all public Docker images. You can also create a custom image and run jobs on it or compose multiple images together (like MySQL 5.7 + Redis 3.2) and run jobs on them as if they were a single image.
```
keybits: Combining multiple images in Docker wouldn't be recommended for an example like this. The example suggests combining two database images but you wouldn't do this. You would keep them separate and call them both as 'service' containers in the build.

e.g. `.circleci/config.yml` might look like this:

```
jobs:
  build:
    working_directory: ~/main
    docker:
      - image: python:3.6.0
      - image: mysql:X.Y.Z
      - image: redis:X.Y.Z
```

Suggest change to:

> 2.0 Jobs support almost all public Docker images. You can also create a custom image with your own dependencies specified (see: https://circleci.com/docs/2.0/custom-images/)

With 'multi stage builds' (https://docs.docker.com/engine/userguide/eng-image/multistage-build/#use-multi-stage-builds) this recommendation can be changed a bit. But that's not release yet.

jekyll/_cci2/faq.md
keybits: 👍

keybits:
Few more comments, otherwise LGTM.

jekyll/_cci2/migrating-from-1-2.md
michelle-luna: fixed, thanks.

jekyll/_cci2/migrating-from-1-2.md
michelle-luna: removed, thanks

jekyll/_cci2/migrating-from-1-2.md
michelle-luna: It looks like the CLI requires you to install docker, so I hesitate to add it here based on my conversations with Nathan this week around Docker installation for folks who are just getting started with 2.0, but I will try to make the CLI content and the config option more obvious in a future PR for sure.

jekyll/_cci2/faq.md
michelle-luna: thanks, removed.

jekyll/_cci2/faq.md
michelle-luna: thanks, replaced with your rewrite.

PR #1084, opened 2017-05-25
Corrected command to get builder secret

drazisil:
The prior command no longer worked due to a location change of where we save the secrets


keybits:
LGTM

PR #1087, opened 2017-05-25
Top level v1 doc update

michelle-luna:
Just a little love for the 1.0 collection, using the landing page to provide the language versions and limited set of links, overview information, and adding TOCs to the long getting started docs

### DIFF
--- a/jekyll/_cci1/index.md
+++ b/jekyll/_cci1/index.md
@@ -26,7 +26,7 @@ For some language versions, CircleCI provides demo applications, as follows:
 - **PHP 5 and later**, see the [PHP Guide]({{ site.baseurl }}/1.0/language-php/)
 - **Python 2.6.6 and later**, see the [Python Guide]({{ site.baseurl }}/1.0/language-python/)
 - **Ruby 1.8 and later**, see the [Ruby/Rails Guide]({{ site.baseurl }}/1.0/language-ruby-on-rails/) 
-- **Scala 0.13.11 and later**, see the [Scala Guide]({{ site.baseurl }}/1.0/scala/)
+- **Scala 0.13.11 and later**, see the [Scala Guide]({{ site.baseurl }}/1.0/language-scala/)
 
 Build projects in C, C#, C++, Clojure, Elixir, Erlang, Groovy, Haxe, Javascript, Perl, Rust, Scala and many more.
 
--- a/jekyll/_cci1/index.md
+++ b/jekyll/_cci1/index.md
@@ -17,18 +17,18 @@ For example, after a software repository on GitHub or Bitbucket is authorized an
 
 ## Programming Language Support
 
-For some language versions, CircleCI provides demo applications, as follows:  
+CircleCI 1.0 supports the use of any language version that is installable on Ubuntu 14.04. Additional versions may be installed at runtime, see the [Ubuntu 14.04 Trusty doc]({{ site.baseurl }}/1.0/build-image-trusty/#programming-languages) for the defaults. For some language versions, CircleCI provides demo applications, as follows:  
 
 - **Go 1.5.3**, see the [Golang Guide]({{ site.baseurl }}/1.0/language-go/)
 - **Haskell 7.4 and later**, see the [Haskell Guide]({{ site.baseurl }}/1.0/language-haskell/)
 - **Java JDK6 and later**, see the [Java Guide]({{ site.baseurl }}/1.0/language-java/)
 - **Node.js all versions**, see the [Node Guide]({{ site.baseurl }}/1.0/language-nodejs/)
 - **PHP 5 and later**, see the [PHP Guide]({{ site.baseurl }}/1.0/language-php/)
-- **Python 2.6.6 and later**, see the [Python Guide]({{ site.baseurl }}/1.0/language-python/)
+- **Python 2.x or 3.x**, see the [Python Guide]({{ site.baseurl }}/1.0/language-python/)
 - **Ruby 1.8 and later**, see the [Ruby/Rails Guide]({{ site.baseurl }}/1.0/language-ruby-on-rails/) 
 - **Scala 0.13.11 and later**, see the [Scala Guide]({{ site.baseurl }}/1.0/language-scala/)
 
-Build projects in C, C#, C++, Clojure, Elixir, Erlang, Groovy, Haxe, Javascript, Perl, Rust, Scala and many more.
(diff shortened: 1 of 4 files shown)


jekyll/_cci1/index.md
```diff
@@ -5,39 +5,39 @@ title: "CircleCI 1.0 Documentation"
...
-
-	{% if docs_found > 3 %}
-		{% assign more = docs_found | minus: 3 %}
-		<li><em><a href="{{ site.baseurl }}/1.0/{{ catFile }}/">{{ more }} more</a></em></li>
-	{% endif %}
-	</ul>
-</div>
-{% endif %}
-{% endfor %}
+CircleCI is a modern continuous integration and continuous delivery (CI/CD) platform. The hosted solution is available at <https://circleci.com/>. The [CircleCI Enterprise](https://circleci.com/enterprise/) solution is installable inside your private cloud or data center. Both are free to try for two weeks. If you are new to CircleCI, check out the [Getting Started with CircleCI]({{ site.baseurl }}/1.0/getting-started/).
```
keybits: .com is 'free forever' for Linux with one container: https://circleci.com/pricing/

jekyll/_cci1/index.md
```diff
@@ -5,39 +5,39 @@ title: "CircleCI 1.0 Documentation"
...
+
+- **Go 1.5.3**, see the [Golang Guide]({{ site.baseurl }}/1.0/language-go/)
+- **Haskell 7.4 and later**, see the [Haskell Guide]({{ site.baseurl }}/1.0/language-haskell/)
+- **Java JDK6 and later**, see the [Java Guide]({{ site.baseurl }}/1.0/language-java/)
+- **Node.js all versions**, see the [Node Guide]({{ site.baseurl }}/1.0/language-nodejs/)
+- **PHP 5 and later**, see the [PHP Guide]({{ site.baseurl }}/1.0/language-php/)
+- **Python 2.6.6 and later**, see the [Python Guide]({{ site.baseurl }}/1.0/language-python/)
+- **Ruby 1.8 and later**, see the [Ruby/Rails Guide]({{ site.baseurl }}/1.0/language-ruby-on-rails/) 
+- **Scala 0.13.11 and later**, see the [Scala Guide]({{ site.baseurl }}/1.0/language-scala/)
+
```
keybits: The language versions supported on 1.0 are different to 2.0. On 1.0 we have preinstalled versions as described here:

https://circleci.com/docs/1.0/build-image-trusty/#programming-languages

They are the currently supported versions, but it does change.

That said, customers can (and do) install their own versions of languages at runtime. So any language version they can install on Ubuntu 14.04 (which is basically any version you want) is possible.

jekyll/_cci1/index.md
michelle-luna: thanks, I think it is ok to just answer the high-level question about free trial on this page without going into more details (people come to this page for the version numbers, so I'm assuming if they get that bit of info, plus the free trial bit they may go to sign up from here rather than to google to do more searching).

jekyll/_cci1/index.md
```diff
@@ -5,39 +5,39 @@ title: "CircleCI 1.0 Documentation"
...
+CircleCI automates build, test, and deployment of software for mobile, enterprise, and web applications. 
+
+![CircleCI Example Flow with GitHub]({{ site.baseurl }}/assets/img/docs/how_it_works.png)
+
+For example, after a software repository on GitHub or Bitbucket is authorized and added as a project to the circleci.com SaaS application, every new commit triggers a build and notification of success or failure through webhooks. CircleCI also supports Slack, HipChat, Campfire, Flowdock, and IRC notifications. Code coverage results are available from the details page for any project for which a reporting library is added.
+
+## Programming Language Support
+
+For some language versions, CircleCI provides demo applications, as follows:
```
keybits: We don't have up-to-date demo applications for 1.0 :-(

jekyll/_cci1/index.md
michelle-luna: I did change these versions to reflect what is currently listed as supported in the 1.0 docs, so I think based on your comment, this is correct? Or should I add a line of text referring to the build-image-trusty page? Thanks!

jekyll/_cci1/index.md
michelle-luna: Thanks, we do have something for them to start with, which is better than nothing I think. If these should be archived because they are so out-of-date as to be harmful, I'm ok with that as well.

jekyll/_cci1/index.md
keybits: I would suggest something like this:

```
Preinstalled language versions for CircleCI 1.0 are listed here: <https://circleci.com/docs/1.0/build-image-trusty/#programming-languages>

You can install other versions at runtime. **You can use any version of a language that can be installed on Ubuntu 14.04.** If you would like more control over the version of a language that is preinstalled, we recommend CircleCI 2.0: <https://circleci.com/docs/1.0/build-image-trusty/#programming-languages>

Language documentation is available for:

- **Go**, see the [Golang Guide]({{ site.baseurl }}/1.0/language-go/) 
- **Haskell**, see the [Haskell Guide]({{ site.baseurl }}/1.0/language-haskell/)
- **Java**, see the [Java Guide]({{ site.baseurl }}/1.0/language-java/)
- **Node.js*, see the [Node Guide]({{ site.baseurl }}/1.0/language-nodejs/)
- **PHP 5**, see the [PHP Guide]({{ site.baseurl }}/1.0/language-php/)
- **Python**, see the [Python Guide]({{ site.baseurl }}/1.0/language-python/)
- **Ruby**, see the [Ruby/Rails Guide]({{ site.baseurl }}/1.0/language-ruby-on-rails/) 
- **Scala**, see the [Scala Guide]({{ site.baseurl }}/1.0/language-scala/)
```

keybits: (I'm uncomfortable with how the version numbers are stated for 2.0 on https://circleci.com/docs/2.0/ - it's not accurate since other versions are possible if you build a Docker image with the version you want. Also developers wouldn't phrase it like 'Python 2 and above' since noone writes Python 1.0 code. The statement seems redundant. We should state that we support building for all language version on 2.0 - that's the flexibility you get with Docker)

keybits:
Comments above, otherwise LGTM

jekyll/_cci1/index.md
michelle-luna: Thanks, I changed it to Python 2.x or 3.x which is how they have it on python.org docs and added the link with the statement about ubuntu 14.04.

PR #1088, opened 2017-05-25
adding new nav to docs

trevor-circleci:
adding optimzely JS tag to docs, as well as adding the new nav, to sync it up with outer. 

The nav is hidden, to be triggered by an optimizely experiment for A/B testing.


felicianotech:
The "hotspot" for clicking on the nav logo/link doesn't cover the whole logo. This is specifically for the new nav. The original (but updated) nav seems to be working correctly.

Also, I don't think the new nav should be fixed. It takes up a lot of valuable room when trying to browse docs for information.

Everything else seems to be good. The new nav looks amazing. Happy to see docs promoted to the top-level.

keybits:
@trevor-circleci dropdowns are slightly misaligned when I run this locally:

image

On outer they are aligned correctly:

image

keybits:
LGTM

PR #1089, opened 2017-05-25
Add a video walkthrough of the single-box install

bellkev:
Here's what it looks like in place:

screen shot 2017-05-25 at 4 47 46 pm


keybits:
Awesome - thanks Kevin :-) LGTM

PR #1090, opened 2017-05-26
Change glossary from static page to a collection

ndintenfass:


ndintenfass:
This ended up a bit more complicated because we use collections for pages -- the order the collections are processed was an issue in calling the collection "glossary". The work-around was to call it "_glossary" instead, so it was definitely processed first (seems Jekyll uses alpha order). The "_" prefix for collections that come before the pages (so that they can be used in pages) works reasonably for collections needed within the "page" collections we use for the main content.

PR #1095, opened 2017-05-29
Fix references to .circle that should be .circleci

ndintenfass:


ndintenfass:
self-merging because it's wrong

PR #1100, opened 2017-05-31
Update API list variables example.

felicianotech:


jekyll/_data/api.yml
```diff
@@ -498,8 +498,8 @@ list_environment_variables:
   try_it: false
   url: "/api/v1.1/project/:vcs-type/:username/:project/envvar"
   method: "GET"
-  description: "Lists the environment variables for :project"
-  response: '[{"name":"foo","value":"xxxx"}]'
+  description: "Lists the environment variables for :project. For security reasons, only four 'x's and the last 4 characters of the value will be returned. This is the same as viewing existing environment variables via the CircleCI webapp."
```
michelle-luna: Thanks, for clarity, suggest rewriting this to: Returns four `x` characters plus the last four alphanumeric characters of the value, consistent with the display of environment variable values in the CircleCI application.

felicianotech: I'm fine with this except "alphanumeric" isn't correct. It'll return any supported characters. I'm not 100% sure that those are but I strongly think it's any ASCII characters.

You okay with "alphanumeric" -> "ASCII" ?

michelle-luna: Yep, works for me, thanks!

felicianotech: Done

michelle-luna:
LGTM

PR #1107, opened 2017-06-01
Update local-jobs.md

michelle-luna:
Qualify working_directory for clarity and add limitation and workaround for env vars.

### DIFF
--- a/jekyll/_cci2/local-jobs.md
+++ b/jekyll/_cci2/local-jobs.md
@@ -79,7 +79,7 @@ Although running jobs locally with `circleci` is very helpful, there are some li
 
 ### Setting Environment Variables
 
-We don't yet have an ideal solution for setting environment variables, but there is a workaround. Define which variables you want to export in a file within your repository (don't commit it - just have it in the directory). Then set BASH_ENV to that filename in your `.cirecleci.config.yml` file. The file will be automatically sourced on each command as long as BASH_ENV is correct.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/local-jobs.md
```diff
@@ -87,12 +91,10 @@ Caching is not currently supported in local jobs. When you have either a `save_c
 
 ### Relative Path for working_directory
 
-Using a relative path for `working_directory` is not currently supported in local jobs. If a relative path is used in `working_directory`, then `circleci` returns an error and immediately exists.
+The `working_directory:` key in your `.cirecleci.config.yml` file must not use a relative path for local jobs. If a relative path is used in `working_directory`, then `circleci` returns an error and immediately exists. To workaround this problem, change the value of the  `working_directory:` key in your `.cirecleci.config.yml` file to use an absolute path.
```
keybits: Typo: `exists` should be `exits`

jekyll/_cci2/local-jobs.md
```diff
@@ -77,6 +77,10 @@ The CLI tool automatically checks for updates and will prompt you if one is avai
 
 Although running jobs locally with `circleci` is very helpful, there are some limitations.
 
+### Setting Environment Variables
+
+We don't yet have an ideal solution for setting environment variables, but there is a workaround. Define which variables you want to export in a file within your repository (don't commit it - just have it in the directory). Then set BASH_ENV to that filename in your `.cirecleci.config.yml` file. The file will be automatically sourced on each command as long as BASH_ENV is correct.
```
keybits: Suggest to be super clear this is for local builds:

"We don't yet have an ideal solution for setting environment variables on local builds, ..."

I've not had a chance to check, but if `BASH_ENV` is set in `.circleci/config.yml` what happens when that file doesn't exist on a non-local build?

PR #1109, opened 2017-06-02
Update caching.md

eric-hu:
Add more detail to source caching section.

Consider this is a first draft, very much open to suggestions.  I'd like to give our users this additional information to reason about configuring their builds.


keybits:
LGTM

PR #1112, opened 2017-06-02
Initial update of 2.0 caching doc

keybits:
WIP - needs copy editing and links updated to working versions.

### DIFF
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -1,53 +1,148 @@
 ---
 layout: classic-docs
-title: "Caching in CircleCI"
+title: "Caching For Faster Jobs"
 short-title: "Caching"
 description: "Caching strategies to increase build speeds"
 categories: [configuring-jobs]
 order: 20
 ---
 
-Caching is one of the most effective ways to reduce job times, but it can also be subtle. This article will outline some strategies you can use to increase caching effectiveness in CircleCI.
+Caching is one of the most effective ways to make jobs faster on CircleCI. To get the best performance can take a bit of planning and fine tuning. There is an art to effective caching. This document gives you strategies to optimise caching to speed up your jobs.
 
-## Partial Cache Restores
+## Types of cache
 
-In most cases, partial cache hits are preferable to zero cache hits. If you get a cache miss, you’ll have to reinstall all your dependencies, which can result in a significant performance hit.
+Thing you might choose to include:
 
-The alternative, then, is to get a good chunk of your dependencies from an older cache instead of starting from scratch. Here are a few specific ways you can do that:
+- Dependencies
+- Source code
 
-### Use Epoch Wisely
+Your dependencies are the libraries and tools your project needs to build, test and deploy. The source is the version controlled source code for your project.
 
-When defining a unique identifier for the cache, be careful about overusing {% raw %}`{{ epoch }}`{% endraw %}. If you limit yourself to {% raw %}`{{ .Branch }}`{% endraw %} or {% raw %}`{{ checksum "filename" }}`{% endraw %}, you’ll increase the odds of a job hitting the cache.
+We will cover [dependency caching]() in detail since it where the biggest performance gains are and has more variables and options. [Source caching]() is covered at the end of the document.
 
-### Define Multiple Keys for Cache Restores
+**Note:** On CircleCI 2.0 the Docker images you use for you jobs are automatically cached on our infrastructure where possible. This document is about manually caching dependencies and source code.
 
-Having multiple key candidates for restoring a cache increases the odds of a partial cache hit. Instead of:
+---
+# Caching Dependencies
+ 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/caching.md
```diff
@@ -1,74 +1,172 @@
...
+
+The Docker image(s) you use for your jobs will usually have tools preinstalled that are generic for building projects using the language the image is focused on. For example the `circleci/ruby:2.4.1` image has useful tools like git, openssh-client and gzip preinstalled. So these types of tools that are not explicitly required for your project are good candidates for being stored on the Docker image.
+
+## How do I know what the path is for my dependencies?
+
+The various language dependency managers (e.g. `npm` or `pip`) each have their own paths where dependencies are installed. See our Language guides and demo projects for the specifics for your stack: <https://circleci.com/docs/2.0/demo-apps/>
+
+## Using cache keys and template values to control what is cached and restored
+
+This section is where you will learn about the art of caching. We say this because caching is a balance between getting maximum performance (using a full cache for every build) and reliability (not using an out-of-date or inappropriate cache).
```
notnoopci: reliability trade-off is more damaging, e.g. getting unclean/corrupted build; build using stale dependencies

jekyll/_cci2/caching.md
```diff
@@ -1,74 +1,172 @@
...
+
+When storing a new cache, the `key` value may contain special template values:
+
+Template | Description
+----|----------
+{% raw %}`{{ .Branch }}`{% endraw %} | the VCS branch currently being built
+{% raw %}`{{ .BuildNum }}`{% endraw %} | the CircleCI build number for this build
+{% raw %}`{{ .Revision }}`{% endraw %} | the VCS revision currently being built
+{% raw %}`{{ .Environment.variableName }}`{% endraw %} | the environment variable `variableName`
+{% raw %}`{{ checksum "filename" }}`{% endraw %} | a base64 encoded SHA256 hash of the given filename's contents. This should be a file committed in your repo. Good candidates are dependency manifests, such as `package.json`, `pom.xml` or `project.clj`. It's important that this file does not change between `restore_cache` and `save_cache`, otherwise the cache will be saved under a cache key different than the one used at `restore_cache` time.
```
notnoopci: we don't need to specify the full details of the hash - it should be sufficient to say "a checksum of the file - so you get a new cache key if file change"

jekyll/_cci2/caching.md
```diff
@@ -1,74 +1,172 @@
...
 {% raw %}
 ```YAML
 - restore_cache:
     keys:
-      - projectname-{{ .Branch }}-{{ checksum "package.json" }}
-      - projectname-{{ .Branch }}
-      - projectname-master
+      # Find a cache corresponding to this specific package.json checksum
+      # when this file is changed, this key will fail
+      - v1-projectname-npm-deps-{{ .Branch }}-{{ checksum "package.json" }}
```
notnoopci: if using checksum - you most likely don't want to depend on branch.  If `project.json` of two builds are the same, they should use the same cache regardless of the branch most of the time.

notnoopci: Also, there is no point in adding the `projectname` IMO.  The cache key is local to the project already.

jekyll/_cci2/caching.md
```diff
@@ -1,74 +1,172 @@
...
+ * {% raw %}`myapp-{{ .Branch }}-{{ checksum "package.json" }}`{% endraw %} - same as the previous one, but each branch will generate separate cache
+ * {% raw %}`myapp-{{ epoch }}`{% endraw %} - every build will generate separate cache
+
+While choosing suitable templates for your cache `key`, keep in mind that cache saving is not a free operation, because it will take some time to upload the cache to our storage. So it make sense to have a `key` that generates a new cache only if something actually changed and avoid generating a new one every build.
+
+## Tips and potential pitfalls with dependency caching
+
+### Partial cache restores are better than no cache restore
+
+In most cases, partial cache hits are preferable to zero cache hits. If you get a cache miss, you’ll have to reinstall all your dependencies, which can result in a significant performance hit.
```
notnoopci: I think this depends in the circumstances.   Partial caches in python/php can be pretty bad, e.g. your code may end up depending on a dependency that's not inlined with your declared dependencies and the problem won't surface until you run a build without a cache.  Other languages build tools (e.g. java build tools) are smarter about it. 

If the dependencies change infrequently, I would suggest going for a zero cache hit first; and only supporting partial caching if the cache miss costs end up being significant over time.

keybits:
@notnoopci thanks for the comment :-) will update when I get back from vacation.

jekyll/_cci2/caching.md
michelle-luna: I'm making these changes and will put up a PR later today

jekyll/_cci2/caching.md
michelle-luna: ok, fixed

jekyll/_cci2/caching.md
michelle-luna: ok, I removed the second item in the list,  removed branch from the first item, and removed project name, let me know if that is not correct

jekyll/_cci2/caching.md
michelle-luna: thanks, I added this

PR #1119, opened 2017-06-06
Add missing top level 'jobs' and 'version' keys.

frewsxcv:


felicianotech:
@frewsxcv If you're going to do that, we might as well add the version key as well and make it a complete config. Do you mind adding that last bit?

felicianotech:
You are awesome. Thank you!

PR #1131, opened 2017-06-08
Restructure config sections

michelle-luna:
This wad adds a new category for Virtualization and changes Configuring Jobs to Configuration. The goal of the structure is to help a new person understand virtualization in 2.0 and their options first then to start at the beginning with configuration. I also retitled many articles to make them more meaningful and action-oriented.

Virtualization:
-Executor Types -> Configuring Test Machine Images
-CircleCI Images -> Using Pre-Built CircleCI Docker Images
-Building Custom Docker Images
-Using docker-compose -> Installing and Using docker-compose
-Docker Layer Caching-> Enabling Docker Layer Caching
-Using Private Images

Configuration:
-Running Jobs Locally -> Using the CircleCI Command Line Interface (CLI)
-Configuration Reference-> Writing Jobs with Steps 
-Environment Variables -> Using CircleCI Environment Variables
-Caching -> Caching Dependencies and Source Code
-Parallelism for Faster Jobs -> Merging and Splitting Tests
-Uploading Build Artifacts
-SSH Access to Jobs -> Debugging Jobs Over SSH
-Defining Multiple Jobs

### DIFF
--- a/jekyll/_cci2/artifacts.md
+++ b/jekyll/_cci2/artifacts.md
@@ -4,7 +4,7 @@ title: "Uploading Build Artifacts"
 short-title: "Uploading Build Artifacts"
 description: "Example of uploading artifacts created during a build"
 categories: [configuring-jobs]
-order: 60
+order: 70
 ---
 
 Sometimes, you'll want to upload artifacts created during builds so you can view them later. The following is an example of how to do that:
--- a/jekyll/_cci2/building-docker-images.md
+++ b/jekyll/_cci2/building-docker-images.md
@@ -1,21 +1,16 @@
 ---
 layout: classic-docs
-title: "Building Docker Images on CircleCI 2.0"
-short-title: "Building Docker Images"
+title: "Building Custom Docker Images"
+short-title: "Building Custom Docker Images"
 description: "How to build Docker images and access remote services"
-categories: [deploying]
-order: 20
+categories: [virtualization]
+order: 30
 ---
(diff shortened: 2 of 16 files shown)


felicianotech:
I'm not a fan of having that `Virtualization` category. I don't think it fits. Problem is, I don't know what I would like in its place.

michelle-luna:
Right, It isn't perfect, but I think it is better than "Using Docker", and is a very well-understood term for this bucket (Docker refers to itself as a virtualization technology when writing generally about this category as well), so I would like to try it.

ndintenfass:
+1 that "Virtualization" is not a good label for things like "Building Docker Images on CircleCI 2.0" -- A huge part of why 2.0 was built in the first place was to dramatically improve our Docker support across the board, so burying this under that title I think will confuse more than elucidate. That said, we have enough Docker-related things that aren't really the same set of issues (eg: building inside a given image is a very different thing to want to do than build images during a job) that a single "Docker" category might also not make sense. But, I do think those who want to do Dockery things will have a much easier time parsing "Docker" vs "Virtualization"

michelle-luna:
Again, it isn't perfect, but it is well-understood term for people who don't know Docker. People who want to do Docker things will see the titles of the Docker articles.

michelle-luna:
Hi,
Researching more over this weekend, what about using Containerization for this section instead?
Thanks,
Michelle

PR #1134, opened 2017-06-10
Identify the default elasticsearch version

AnvilStriker:
The variable referenced by the source file (`site.data.trusty.versions.summary.elasticsearch`) apparently isn't bound, resulting in the absurd statement in the rendered page: 

"The default version of Elasticsearch is ."


felicianotech:
That's the correct one. Thank you.

PR #1135, opened 2017-06-12
Adding job.shell to reference page

kimh:
Now we support top level shell option with `job.shell`.


kimh:
ping

ndintenfass:
@kimh are you still waiting for a lgtm on this? If so: I did not actually test code using the new syntax personally, bu lgtm

PR #1137, opened 2017-06-12
More structure updates

michelle-luna:
Hi,
After thinking more about Nathan's verbal suggestions last week and because more small sections will help the structure while we have only two levels, I have added two more categories:

Optimization (for the caching and perf content)
Troubleshooting (for the CLI and SSH content, but I'm sure this will grow over time)

And I fixed some categories that I messed up on the first pass.

Thanks,
Michelle

### DIFF
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -3,7 +3,7 @@ layout: classic-docs
 title: "Caching Dependencies and Source Code"
 short-title: "Caching Dependencies and Source Code"
 description: "Caching Dependencies and Source Code"
-categories: [configuring-jobs]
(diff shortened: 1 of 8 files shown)


keybits:
Looks good. Two suggestions sent by DM still stand.

Also:

image

^ I think 'Docker Layer Caching' should be the last item in that section. Typically you'd address the other two areas first.

image

^ The 'Using Docker' section currently has no articles. Should it be removed? Or should some of these be in that section:

image

michelle-luna:
Thanks Tom, I made the re-ordering changes and commented out the Using Docker category. I'm going to leave the retitling and migration section I have so far to see what feedback from the week brings.

PR #1139, opened 2017-06-12
 Restrict a step to run only on one container

yonigoldberg:
Not sure if this is the best place to add this, as the header changed from `Parallelism for Faster Jobs` to `Merging and Splitting Tests`

- Will be also nice to add how to make this step to wait until the tests are done in the other container. Any suggestions?

### DIFF
--- a/jekyll/_cci2/parallelism-faster-jobs.md
+++ b/jekyll/_cci2/parallelism-faster-jobs.md
@@ -96,3 +96,16 @@ The following libraries have built-in support for the CircleCI environment varia
 # create a bottleneck on the CI node
 - run: bundle exec rake knapsack_pro:queue:rspec
 ```
+
+### Restricting a step to run on only one container
+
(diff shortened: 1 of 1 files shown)


ndintenfass:
@jonathang have you seen our new Workflows feature (announced after you put up this PR)? It's likely the best path to chaining various jobs together. It is also likely a better way to segment out things that should run on one container rather than across a set of parallel containers. The change proposed is a reasonable approach, and I'd be happy to approve and merge, but would want to know if you'd change the recommendation to solving this problem using workflows. See: https://circleci.com/docs/2.0/workflows/

jonathang:
Thanks @ndintenfass - since I wrote this PR, I have change the build to use workflows which seems much less hacky. And I also removed all the places in the code that used this type of logic. 

I feel good  about closing this pr.

Thanks!
Yoni

PR #1140, opened 2017-06-12
Michelle luna patch 1

michelle-luna:


jekyll/_cci2/faq.md
```diff
@@ -107,6 +107,37 @@ In this example, we're setting the timezone for both the primary image and an ad
 
 A full list of available timezone options is [available on Wikipedia](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).
 
+## Workflows
+
+### Can I use the API with Workflows?
+Not yet, but we are working on that functionality.
+ 
+### Can I run the Auto-cancel feature with Workflows?
```
KunalJain: Probably be 'use' rather than 'run'.

jekyll/_cci2/workflows.md
```diff
@@ -0,0 +1,205 @@
...
+ 
+## Running a Workflow from a Failed Job
+
+It is possible to rerun a job that failed in the middle of a workflow and continue the rest of the workflow using the Rerun button on the Workflows page of the CircleCI application.
+
+![Rerunning a Workflow from a Failed Job]({{ site.baseurl }}/assets/img/docs/restart_from_failed.png) 
+
+To rerun a workflow from the failed job:
+
+1. From the Builds page, select the project name.
```
KunalJain: Don't think we need this. User will have first class way to navigate to workflows.

jekyll/_cci2/workflows.md
```diff
@@ -0,0 +1,205 @@
...

+2. Click the Workflows tab to display workflows for your project.
```
KunalJain: Click on the 'Workflows' icon on the main nav on the left.
Click on the project name to display workflows for your project.

jekyll/_cci2/workflows.md
```diff
@@ -0,0 +1,205 @@
...
+The following example shows a workflow configured with jobs on three branches: Dev, Stage, and Pre-Prod. 
+
+![Branch-Level Job Execution]({{ site.baseurl }}/assets/img/docs/branch_level.png) 
+
+The following `config.yml` snippet is an example of a workflow configured for branch-level job execution:
+
+```
+workflows:
+  version:2
+  dev_stage_pre-prod_master_deploy:
```
KunalJain: Maybe the name can be 'dev_stage_pro-prod'

jekyll/_cci2/workflows.md
```diff
@@ -0,0 +1,205 @@
...
+            - build
+      - test2:
+          requires:
+            - test1
+      - deploy:
+          requires:
+            - test2	
+```
+
+The dependencies are defined by setting the `requires:` key as shown. The `deploy:` job will not run until the `build` and `test` jobs complete successfully. A job must wait until all upstream jobs in the dependency graph have run. So, the `deploy` job waits for the `test2` job, the `test2` job waits for the `test1` job and the `test1` job waits for the `build` job.
```
gordonsyme: The second sentence mentions a `test` job which isn't present in the config. Might be clearer to drop the `test2` job and rename `test1` to `test`.

jekyll/_cci2/workflows.md
```diff
@@ -0,0 +1,205 @@
...
+              only: stage
+      - test_pre-prod:
+          filters:
+            branches:
+              only: pre-prod
+```
+
+## Workspaces
+
+Workflows that include jobs running on multiple branches may require artifacts to be shared using workspaces. To persist an artifact from a job and make it available to other jobs, configure the job to use the `persist_to_workspace` key where the value is a directory inside the project’s working directory. Artifacts of the job will be saved to this directory until the job is rerun and new artifacts are created.
```
gordonsyme: Workspaces are generally useful for workflows not just when branch filters are involved. E.g. Scala projects typically require lots of CPU for compilation but tests can parallelise across containers well. You could use a larger container for a `build` job and then save the compiled artifacts into the workflow's workspace so that the test containers can use the compiled Scala from the `build` job.

jekyll/_cci2/configuration-reference.md
```diff
@@ -555,6 +555,31 @@ fingerprints | N | List | List of fingerprints corresponding to the keys to be a
...
+Used for orchestrating all jobs. Each workflow consists of the workflow name as a key and a map as a value. A name should be unique within the current `config.yml`. Jobs are run in parallel by default, so you must explicitly require any dependencies by their job name.
+
+Key | Required | Type | Description
+----|-----------|------|------------
+version | Y | String | Should currently be 2
+name | N | String | Title of the workflow to be shown in the CircleCI UI
+jobs | Y | List | A list of jobs to run with their dependencies
+requires | N | List | A list of jobs that must succeed for the job to start
+branches | N | Map | A map defining rules for execution on specific branches
+only | N | String | A branch name
```
gordonsyme: We'll need to split this into multiple tables or use some other method since the keys are only applicable at different parts of the YAML hierarchy.

The top level keys for workflows config are `version` and `jobs`

A job can have the keys `requires` and `filters`

`filters` can have the key `branches`

`branches` can have the keys `only` and `ignore` which either map to a single string naming a branch (or a regexp to match against branches, has to be enclosed with `/`s) or map to a list of such strings.
* Any branches that match `only` will run the job.
* Any branches that match `ignore` will not run the job.
* If neither `only` nor `ignore` are specified then all branches will run the job.
* If both `only` and `ignore` are specified the `only` overrides `ignore`."

keybits:
This looks great - love the diagrams. In addition to Gordon and Kunal's comments, might we consider a warning / info box on the status of workflows? My impression is that it's more 'beta' than 2.0 in general and we should manage expectations, especially around the current UI.

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -45,6 +45,39 @@ CircleCI 2.0 introduces the requirement that you create a configuration file (`.
      ```
 7. Validate your YAML at <http://codebeautify.org/yaml-validator> to check the changes. 
 
+## Steps to Configure Workflows
+
+Optionally configure workflows, using the following instructions:
+
+1. To use the Workflows feature for job orchestration, first split your build job into multiple jobs, each with a unique name.
+
+2. To persist a job artifact, add the `persist_to_workspace:` key at the end of the job, under `steps:`. Then, to get the saved artifact for use by another job, add the `attach_workspace:` key with a nested `at:` key defined as the directory where the artifact was saved.
```
gordonsyme: ~ The `at` property names a directory to which the workspace will be "mounted" (it's not really mounted, contents are copied in). So it doesn't have to be the same directory that was saved via `persist_to_workspace` but it is handy if it is.

E.g. if all jobs use `/tmp/workspace` for the workspace location then everything is nice and consistent.

jekyll/_cci2/workflows.md
```diff
@@ -0,0 +1,205 @@
...
+            - echo-output
+
+  downstream:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+
+      - run: |
+          if [[ `cat workspace/echo-output` == "Hello, world!" ]]; then
```
gordonsyme: ~ the command needs to be `cat /tmp/workspace/echo-output` because the workspace was attached to `/tmp/workspace`

jekyll/_cci2/workflows.md
```diff
@@ -26,6 +26,8 @@ Select the Failed workflow to see the status of each job as shown in the next sc
 
 ![CircleCI Workflows Page]({{ site.baseurl }}/assets/img/docs/workflow_detail.png)
 
+**Note**: The workflows user interface is subject to change during the Beta and does not reflect the final version of the user experience for this feature. It is a work in progress and we welcome your feedback!
```
KunalJain: Thank you for adding this!

jekyll/_cci2/workflows.md
KunalJain: ^ Can delete point 1.

KunalJain:
Lgtm - Added 2 more comments.

PR #1142, opened 2017-06-14
docker-compose doesn't fit under deployment

zzak:
I put it under "containerization" instead


keybits:
Agree. LGTM, but you may wish to get confirmation from @michelle-luna

michelle-luna:
I think Nathan made the argument with me that this was only used in deployment so I moved it (maybe I misunderstood him)? I'm fine with this change if he agrees as well.

ndintenfass:
The one I was most concerned with was https://circleci.com/docs/2.0/building-docker-images/ - which is, I think, in the right place now. No strong opinion on docker-compose.

ndintenfass:
@zzak which sounds like you are go for :shipit:

PR #1145, opened 2017-06-14
2.0 parallelism is only to be set via config.

gordonsyme:
Respecting parallelism set via the API (or project settings) leads to
unexpectedly parallel jobs.

This is pending acceptance of PRs in other repos.


KunalJain:
👍

PR #1160, opened 2017-06-15
Doc restructure 1 0

michelle-luna:
Hi,
This set of changes is to improve the structure of the 1.0 docs collection to the following:

Getting Started
Migration
Build Environments
Configuration 
Optimization
Integrations
Language Guides
Reference
Troubleshooting

### DIFF
--- a/jekyll/_cci1/android.md
+++ b/jekyll/_cci1/android.md
@@ -2,7 +2,7 @@
 layout: classic-docs
 title: Test Android Applications
 short-title: Android
-categories: [mobile-platforms]
+categories: [build-images]
 description: Testing Android applications on CircleCI
 ---
 
--- a/jekyll/_cci1/background-process.md
+++ b/jekyll/_cci1/background-process.md
@@ -1,8 +1,9 @@
 ---
 layout: classic-docs
-title: Start background processes from circle.yml
-categories: [how-to]
+title: Starting Background Processes
+categories: [configuration]
 description: "How to start a background process from circle.yml"
+order: 50
 ---
 
 Starting a background process from [circle.yml]( {{ site.baseurl }}/1.0/configuration/)
--- a/jekyll/_cci1/background-process.md
+++ b/jekyll/_cci1/background-process.md
@@ -1,9 +1,9 @@
 ---
 layout: classic-docs
 title: Starting Background Processes
-categories: [configuration]
+categories: [configuration-tasks]
 description: "How to start a background process from circle.yml"
-order: 50
+order: 56
 ---
 
 Starting a background process from [circle.yml]( {{ site.baseurl }}/1.0/configuration/)
--- a/jekyll/_cci1/code-coverage.md
+++ b/jekyll/_cci1/code-coverage.md
@@ -1,9 +1,10 @@
 ---
 layout: classic-docs
-title: Generating code coverage metrics
-short-title: Code Coverage
-categories: [how-to]
+title: Generating Code Coverage Metrics
+short-title: Generating Code Coverage Metrics
+categories: [configuration]
 description: Generating code coverage metrics
+order: 30
 ---
 
 Code Coverage tells you how much of your application is tested.
--- a/jekyll/_cci1/code-coverage.md
+++ b/jekyll/_cci1/code-coverage.md
@@ -2,9 +2,9 @@
 layout: classic-docs
 title: Generating Code Coverage Metrics
 short-title: Generating Code Coverage Metrics
-categories: [configuration]
+categories: [configuration-tasks]
 description: Generating code coverage metrics
-order: 30
+order: 50
 ---
 
 Code Coverage tells you how much of your application is tested.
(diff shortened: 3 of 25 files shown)


keybits:
I feel we should be more cautious restructuring 1.0 docs than 2.0 docs because:

1. People are used to the structure and know where to look for things
2. There is long standing SEO for pages that people find via search engines

That said, I agree that the structure can be improved :-) Two questions:

- Why a 'migration' section? What are people migrating from?
- Why remove the 'mobile platforms' section? Our mobile offering is something that not all CI companies have people testing mobile apps have specific requirements and settings. iOS / macOS I think at least deserves its own section since it's a different build image, different hardware, different pricing and different environment. Android builds on the same 14.04 image as everything else, so it's not technically a 'Build Environment'. Perhaps we could rename 'Mobile Platforms' to 'Mobile'. Where it is in the menu structure (under languages) feels right.

A section that is sorely missing from 1.0 docs and is still worth adding is 'Databases' - all other CI / CD companies have this section. People need better information on how to configure MySQL, PostgreSQL, Redis etc on our platform.

michelle-luna:
Hi Tom,

Thanks for this feedback, much appreciated. I agree on being more careful here, and it is good to let 1.0 users know that docs are changing and to get them familiar with the structure they will see in 2.0. I didn't really retitle too many articles in this set of changes, so SEO should not be broken. The main thing I did was to reduce the number of articles in Getting Started to just three, to keep from overwhelming new people. 

Migration is for people coming from Jenkins or any other CI solution, so it has that article and the Sample Config and the GitHub info. Again, keeping the top categories as simple as possible. I changed Build Images to 'Build Environments' so I could put Android there with the rest and it gets the mobile platforms 'above the fold' with the goal of better visibility than before for the reason you stated. 

The major jumble of the current organization is How-To which is a catch all for 30+ docs. So, I put the docs most closely related to initial configuration in Configuration, leaving only Test and Deployment docs in the Integrations section. I moved one doc into Languages and renamed it Language Guides and largely left Reference and Troubleshooting alone for now.

Thanks,
Michelle

felicianotech:
I agree with literally every point that Tom made.

Without going crazy with it, I think changing some of the titles can be okay, I just think it's crucial we change as little URLs as possible.

The number 1 section missing from 1.0 definitely is databases. My questions is, is at too late? I know 2.0 is still in beta but I'm using it for all of my projects and it wells well enough that I don't know how much effort we should be putting into it. If anything, API docs, which is for 1.0 and 2.0 needs the most work. @keybits Is that new guy still doing the changes he wanted to do? If not, then maybe I/we should refocus on that like I was a few months ago.

@michelle-luna I will say, the giant how-to section definitely could be broken up. Especially since just changing the category won't change any URLs. :)

michelle-luna:
Ah, do we have content for a databases section? Even if it is just drafts? We can add that section to 1.0 if we have docs already, otherwise, we can plan to create new database content for 2.0. 

I agree 2.0 is the priority, the idea with this PR is to make 1.0 less overwhelming while most new customers are still going there and to also make some consistency between 1.0 and 2.0 categories & content. 

So, for this review, do the changes I made break anything or render anything incorrect? Even if it is not perfect, does it make things better than they are today?

Thanks,
Michelle

felicianotech:
I ran this branch locally. My biggest gripe is the reference section. Many things which I see belonging to the reference section such as the entirety of `circle.yml` was removed whereas "Polling project status using CCMenu, CCTray, etc." is in reference which is something I don't think belongs there.

Also, maybe it's just me but a lot of these PRs are huge so it's harder to discuss all the changes at once. Especially when most are good but there's one particular for discussion.

michelle-luna:
Hi, thanks for the feedback on these long PRs! Nothing was removed, just moved, the Sample circle.yml file is in the Migration section now.

The structure organizes docs in order of how you would use them from top to bottom, not by putting similar articles together. So, things that a new user would need at the beginning of using CircleCI for any platform go at the top and progress from there (in learning more advanced topics and in moving through the BTD process).

cheers,
Michelle

ndintenfass:
What you said is worth pushing beyond the scope of this PR -- "The structure organizes docs in order of how you would use them from top to bottom, not by putting similar articles together." -- I think part of the lengthy discussion on this PR is that a narrative-style structure is a high-level change in how to think about what the sections are for. Going towards a linear structure instead of a topic-driven structure could definitely end up with different sections and ordering. I think this is also part of why there has been some discussion on "Migration" -- I think in our internal myopia we see reference for getting started as what you'd do for a new project, whereas the premise of "Migration" is that most people reading these docs for the first time are migrating from some other system to CircleCI. Migration is the first step on your journey from where you were to our promised land. In any case, putting the question out there: should this merge and then more changes from that state can be proposed as subsequent PRs, or has feedback here changed your mind on current state of this @michelle-luna? Or is there value in this continuing as an open discussion?

michelle-luna:
Hi,

Thanks for this discussion, it is always helpful and this PR wasn't ready to go yet anyways. I think it is ready now and I feel it adds value and makes things better, though not perfect.  

-Michelle

ndintenfass:
Seems like this is your call to make @michelle-luna -- at this point it seems this is all a matter of judgement calls.

PR #1161, opened 2017-06-16
Working directory is mandatory, not optional

binkley:
Saying "In each job, we have the option of specifying a working_directory" is misleading: my build immediately broke without providing a working directory in `config.yml`.


keybits:
Good catch and thank you for the PR :-)

PR #1164, opened 2017-06-16
Add a hidden doc for the cohort of customers upgrading from 1.46.5 an…

ndintenfass:
…d earlier


ndintenfass:
Self-merging as this is a non-destructive, hidden change.

PR #1170, opened 2017-06-20
Workspaces and path fix

michelle-luna:
add workspaces info and fix migration doc for $PATH


jekyll/_cci2/configuration-reference.md
```diff
@@ -539,6 +539,48 @@ The directory layout should match the [classic CircleCI test metadata directory
...
+``` YAML
+- attach_workspace:
+    at: /tmp/file-name
+```
+
+Note the following distinctions between Artifacts, Workspaces, and Caches:
+
+| Type      | lifetime        | Use                      | Example |
+|-----------|-----------------|------------------------------------|---------
+| Artifacts | Months to years | Preserve long-term artifacts. |  Linked list in the Artifacts tab of the Build details under the `tmp/circle-artifacts.<hash>/container`   or similar directory.   |
```
KunalJain: Artifacts can be update lifetime to be 'months'.

jekyll/_cci2/configuration-reference.md
```diff
@@ -539,6 +539,48 @@ The directory layout should match the [classic CircleCI test metadata directory
...
+##### _Example_
+
+``` YAML
+- attach_workspace:
+    at: /tmp/file-name
+```
+
+Note the following distinctions between Artifacts, Workspaces, and Caches:
+
+| Type      | lifetime        | Use                      | Example |
```
lewang: I love the idea of this comparison table.  Just last night I had a conversation with a user about why caches and workspaces both exist.

jekyll/_cci2/configuration-reference.md
```diff
@@ -539,6 +539,48 @@ The directory layout should match the [classic CircleCI test metadata directory
...
+|-----------|-----------------|------------------------------------|---------
+| Artifacts | Months to years | Preserve long-term artifacts. |  Linked list in the Artifacts tab of the Build details under the `tmp/circle-artifacts.<hash>/container`   or similar directory.   |
+| Workspaces | Duration of workflow        | Move temporary files around for use by jobs run later in the workflow. | The `persist_to_workspace: <directory>` key under jobs, refer to the file with `attach_workspace: at: <directory>`.    |
```
lewang: At a product level, workspaces need a longer storage guarantee.

The reason for workspaces to exist is build repeatability.  I.e. when you rerun a workflow, the job is going to have the same workspace as the last run.  So they have to last as long as we expect people to want to rerun builds.

jekyll/_cci2/configuration-reference.md
lewang: "Linked list in the " -> "Available under"

linked-list has special meaning in CS context.

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -56,7 +56,7 @@ Optionally configure workflows, using the following instructions:
 
 1. To use the Workflows feature for job orchestration, first split your build job into multiple jobs, each with a unique name.
 
-2. To persist a job artifact, add the `persist_to_workspace:<directory>` key at the end of the job, under `steps:`. Then, to get the saved artifact for use by another job, add the `attach_workspace:` key with a nested `at:<directory>` key, ideally  defined as the directory where the artifact was saved.
+2. To persist a temporary file to a location between jobs, add the `persist_to_workspace:<directory>` key at the end of the job, under `steps:`. Then, to get the saved temporary file for use by another job, add the `attach_workspace:` key with a nested `at:<directory>` key, ideally defined as the directory where the artifact was saved.
```
lewang: "at the end of the job" not necessary.  `persist_to_workspace` can be called whenever.

They can also be called multiple times (I think, don't quote me).

jekyll/_cci2/configuration-reference.md
michelle-luna: thanks, I made these changes

jekyll/_cci2/configuration-reference.md
michelle-luna: Should this content change? I can't tell from your comment if you want the words to reflect this.

jekyll/_cci2/migrating-from-1-2.md
michelle-luna: thanks, I fixed the first problem, will leave the rest to be sure.

lewang:
I started making a PR for this, but realized I was being too prescriptive.
Instead I'm going to list out some points about caches, workspaces, and
artifacts here.  @michelle-luna We could talk synchronously if this is not
clear or need more details.

## cache

- Store a hierarchy of files under a **key**.
- Cache key is namespaced to the project
- Retrieval is prefix-match based.  (this point needs more attention in the
  docs, but not necessarily in this PR).
- Last writer to the cache wins.
    - Consider a workflow of J1 -> J2 -> J3
    - If J1 and J3 writes to the same cache key, a rerun of J2 will see J3's
      changes since it writes last.
    - In fact, any job that runs inside this project will always see the
      latest write.

### use case

Caches are appropriate for acretion only data, e.g. Ruby Gems (~/.gem), Maven (~/.m2).

For example when you bump versions of a Gem, the ~/.gem contains both the old
and new versions and the cache is made more useful by the addition of data.

## workspace

- Workflows aware storage mechanism.
- Stores data unique to the job, which may be needed in downstream jobs.
- No key required, a job has one implicit workspace when it runs in a workflow
  (although using it is optional).
- In the fan-in case, workspaces of upstream jobs are merged.  Note: name
  conflicts will cause error currently.
- When a job finishes, its workspace is "frozen" and won't be changed.  The
  same workspace will be used if a downstream job is rerun.
    - Consider a workflow of J1 -> J2 -> J3
    - Any rerun of J2 will see the same workspace as the original run.

### use case

Use workspaces to pass along data unique to this run and are needed for downstream jobs.

For example, you have a "build" job that builds a jar and saves it to a
workspace.  "build" can fan out into "integration-test", "unit-test", and
"code-coverage" to run those tests in parallel.

## artifacts

- Expose information about the job through files.

michelle-luna:
ok, I've added this content and will add the cache info in a different PR. Thank you! Michelle

jekyll/__glossary/workspace.md
```diff
@@ -0,0 +1,4 @@
+---
+term: Workspace
+---
+A workspace is a workflows-aware storage mechanism. A worspace stores data unique to the job, which may be needed in downstream jobs.
```
lewang: worspace

jekyll/_cci2/configuration-reference.md
```diff
@@ -539,6 +539,51 @@ The directory layout should match the [classic CircleCI test metadata directory
...
+Any rerun of J2 will see the same workspace as the original run. Use workspaces to pass along data unique to this run and which are needed for downstream jobs.
+
+Note the following distinctions between Artifacts, Workspaces, and Caches:
+
+| Type      | lifetime        | Use                      | Example |
+|-----------|-----------------|------------------------------------|---------
+| Artifacts | Months          | Preserve long-term artifacts. |  Available in the Artifacts tab of the Build details under the `tmp/circle-artifacts.<hash>/container`   or similar directory.     |
+| Workspaces | Duration of workflow        | Move temporary files around for use by jobs run later in the workflow. | The `persist_to_workspace: <directory>` key under jobs, refer to the file with `attach_workspace: at: <directory>`.    |
+| Caches    | Months          | Store dependency or source code data in fast memory for quick retrieval.          |  The `save_cache` job step with a `path` to a list of directories to add and a `key` to uniquely identify the cache (for example, the branch, build number, or revision).   Restore the cache with `restore_cache` and the appropriate `key`. |
```
lewang: "fast memory" ... "quick retrieval" is not correct.  Caches are not faster than workspaces.

lewang:
> Caches are appropriate for acretion only data, e.g. Ruby Gems (/.gem), Maven (/.m2).

A more readable way to say this:

-----

Use the cache to store data that makes your job faster, but in the case of a cache miss the job will still run successfully.  Examples are: Npm, Gem, Maven packages directories.

jekyll/_cci2/configuration-reference.md
lewang: I think say something like:

Non-vital data that may help the job run faster.  E.g. Npm, Gem packages.

jekyll/__glossary/workspace.md
michelle-luna: thanks, fixed

jekyll/_cci2/configuration-reference.md
michelle-luna: thanks, fixed.

PR #1179, opened 2017-06-22
Fix mistype in workflow example

asarturas:
Seems like job with no config should be without colon. I was getting parsing error until removed trailing colon:
`* error parsing config: yaml: line 100: mapping values are not allowed in this context`
part of failing config example (lines 95-105):
```
workflows:
  version: 2
  build_and_deploy:
    jobs:
      - build:
      - publish_api_client:
          requires:
            - build
          filters:
            branches:
              only: master
```


zzak:
Thank you!

PR #1199, opened 2017-06-27
Individual pages for each API call

TylerConlee:
Very work-in-progress, but the initial reorganization of the API docs into something more navigable.


michelle-luna:
closing this PR since it is from June, feel free to re-open (but best to just open a new one at this point), trying to do some cleanup on all these stuck PRs, thanks!

PR #1222, opened 2017-07-01
Approval job

michelle-luna:


KunalJain:
lgtm

jekyll/_cci2/workflows.md
```diff
@@ -90,6 +91,36 @@ workflows:
...
+  build-test-and-approval-deploy:
+    jobs:
+      - build
+      - test1:
+          requires:
+            - build
+      - test2:
+          requires:
+            - test1
+      - test3:
```
gordonsyme: There's another bit of "specialness" about approval jobs. Regular jobs _must_ name a job that exists in the job definitions outside the workflows config.
Approval jobs _must not_ name a real job.

KunalJain: Does this not makes it clear? The `type: approval` key is a special job type that is **only** added under in your `workflow` key.

jekyll/_cci2/workflows.md
```diff
@@ -17,8 +17,9 @@ A workflow is a set of rules for defining a collection of jobs and their run ord
 - Fan-out to run multiple jobs in parallel for testing various versions  
 - Fan-in for deployment to separate platforms with automated triggers 
 - Branch-level job execution with artifact sharing
+- Manual job approval to control deployments to separate environments
```
Peeja: It looks like this list started as a list of verbs and turned into a list of nouns somewhere during the ambiguous "fan-out" and "fan-in". Can we make them line up again? If we make them all verbs we can change "if you need to meet any of the following requirements:" to simply "if you need to:".

jekyll/_cci2/workflows.md
gordonsyme: Not really, the consequence of `test3` being an approval job is there _must not_ be a regular job named `test3`.

jekyll/_cci2/workflows.md
```diff
@@ -90,6 +91,36 @@ workflows:
 ```
 
 The dependencies are defined by setting the `requires:` key as shown. The `deploy:` job will not run until the `build` and `test1` and `test2` jobs complete successfully. A job must wait until all upstream jobs in the dependency graph have run. So, the `deploy` job waits for the `test2` job, the `test2` job waits for the `test1` job and the `test1` job waits for the `build` job.
+
+## Holding a Workflow for a Manual Approval
+
+Workflows may be configured to wait for manual approval of a job before continuing by using the `type: approval` key. The `type: approval` key is a special job type that is **only** added under in your `workflow` key. This enables you to configure a job with `type:approval` in the workflow before a set of parallel jobs that must all wait for manual approval. Jobs run in the order defined until the workflow processes a job with the `type: approval` key followed by a job on which it depends as in the following `config.yml` example:
```
Peeja: > Workflows may be configured to wait for manual approval of a job before continuing by using the `type: approval` key.

This is quite passive. How about:

> You can add a `type: approval` job to your workflow to make other jobs wait for manual approval before continuing.

---

> This enables you to configure a job with `type: approval` in the workflow before a set of parallel jobs that must all wait for manual approval.

This isn't really right: this says that the configuration enables you to add configuration. I'm not sure a sentence is actually necessary here, though; I'd just cut it. The example below is going to get the point across best.

---

> Jobs run in the order defined until the workflow processes a job with the `type: approval` key followed by a job on which it depends as in the following `config.yml` example:

This is actually incorrect: jobs don't run in the order defined at all, they run in dependency order. How about:

> Any job which depends on an Approval job will wait to run until that Approval job has been manually approved, as in the following `config.yml` example:

jekyll/_cci2/workflows.md
```diff
@@ -90,6 +91,36 @@ workflows:
...
+          requires:
+            - test1
+      - test3:
+         type: approval
+         requires:
+           - test1
+           - test2
+      - deploy:
+          requires:
+            - test3
```
Peeja: This is actually not a great example, as it's three Approval jobs chained together. Also, the UI is currently broken because there's a bug in workflows which begin with an Approval job. (Also, beginning with an Approval job is a bit of an odd thing to do in the first place.) Let's rite a more canonical example, preferably in a controlled test repo and not in our frontend repo.

jakajancar: I have a use-case for beginning with an approval job:

Unless branch is master/staging-*/deploy-*/..., require approval before running tests (substitute for branch filtering in CircleCI 1).

jakajancar: Hmm, was the UI fixed? seems to work for me even if the first step is approval!

idugalic:
I have implemented simple pipeline/workflow demonstrating this feature: [https://github.com/idugalic/spring-boot-webpack-es6-react-poc/blob/master/.circleci/config.yml](https://github.com/idugalic/spring-boot-webpack-es6-react-poc/blob/master/.circleci/config.yml)
workflow

jekyll/_cci2/workflows.md
michelle-luna: great, thanks I've added this requirement and changed the name of this job to 'hold'

jekyll/_cci2/workflows.md
michelle-luna: thanks, fixed

jekyll/_cci2/workflows.md
michelle-luna: Hi, thanks very much for these suggestions.

jekyll/_cci2/workflows.md
michelle-luna: thanks, I will use the image from idugalic

azr:
We have an issue on this.

We want our feature branches to be manually deployed, we also prevent PRs from being accepted if circle ci is not *green*, but with the on hold solution, the build always stays yellow.

Would it be possible to have "green hold steps" ?

ps: this seemed an appropriate place to comment, sorry if not.

Peeja:
@azr That's great feedback, thanks! Would you post that over on our [discussion board](https://discuss.circleci.com/c/feature-requests)? It'll probably get more traction over there.

azr:
I commented here: https://discuss.circleci.com/t/manually-triggered-deployment-stages/5515/25, may be a new post will get more traction

Peeja:
Thanks! I'll make sure the team sees it.

azr:
Also made a new feedback on the board : https://discuss.circleci.com/t/green-successful-hold-step-manually-triggered-deployments-that-succeed/17261 sorry it feels spammy :)

Peeja:
No worries, it's good to have the feedback!

PR #1227, opened 2017-07-03
First draft of Contexts documentation

ndintenfass:


jekyll/_cci2/contexts.md
```diff
@@ -0,0 +1,44 @@
...
+  version: 2
+  my-workflow:
+    jobs:
+      - run-tests:
+            context: org-global
+```
+
+## Environment Variables in Contexts
+Contexts have a set of environment variables defined as name/value pairs. When a job is run with a Context the environment variables of the Context are injected at runtime, overwriting any project-level environment variables that had been set in the Project Settings. However, environment variables defined in the job configuration itself will overwrite those of the Context.
+
```
rishimkumar: The last line makes it confusing. I wonder if it's easier to say something like this.  Here's a suggestion:

Contexts include environment variables defined as name/value pairs. When a job invokes a Context, the environment variables are injected at runtime.

By adding in Contexts, there are now three ways to set environment variables.
1) Project Settings: These can be set in the UI. 
2) Contexts are injected at runtime and can overwrite variables set in Project Settings.
3) Variables defined directly in the job configuration have always been possible and these variables take precedent over all others.

ndintenfass: Yeah, that is probably better -- one issue is we have a doc about Env Vars that doesn't actually explain 2.0 config vars.

jekyll/_cci2/contexts.md
```diff
@@ -0,0 +1,48 @@
...
+            context: org-global
+```
+
+## Environment Variables in Contexts
+Contexts have a set of environment variables defined as name/value pairs. When a job is run with a Context the environment variables of the Context are injected at runtime. The order of precedence is important to remember when working with environment variables that might overlap in name. Following is the order, with later injection always taking precedence when the steps of your job run:
+1. Project-level environment variables set on the Project Settings page.
+2. Context environment variables (assuming the user has access to the Context).
+3. Environment variables set in your config YAML with the `environment` key of a job definition.
+4. Environment variables declared inside a shell command in a `run` step of a job.
+
```
AlexanderMann: I think an example here might help. Not blocking, just a nit.

jekyll/_cci2/contexts.md
```diff
@@ -0,0 +1,48 @@
...
+categories: [configuring-jobs]
+order: 41
+---
+
+**NOTE:** As of July 2017 Contexts are launched in their MVP state. Some future features of Contexts are called out below as "Coming Soon". We encourage your feedback either on [our Discuss forum](https://discuss.circleci.com/t/contexts-feedback/13908) or in the Pull Request for this document.
+
+<hr>
+
+## Contexts Overview
+Contexts provide centrally managed, secured resources for use in jobs. They are primarily used for sharing environment variables across projects.
```
rishimkumar: Kinda buries the lede. whenever you talk about it internally you speak way more about the future of Context.  "Sharing env vars across projects" isn't really the primary use - it's the current v1 use.

here's some proposed text.


Contexts provide a new way to think about managing your build environments.

Contexts involve Resources (e.g., secrets/passwords/environment variables, fleets/machines) and Security Rules (which objects (e.g., projects/groups) can use which resources).  It is centrally managed at the organization level in the UI and gets invoked through config on a job-by-job basis

Contexts are designed to make it more convenient to add, update, and share environment variables within an organization.  As opposed to setting environment variables in the UI on a project-level, Contexts are set in UI at an organization level and are invoked by jobs through configuration files. 

This leads to new opportunities related  to security and permissions such that it will be easy to manage individual/group access to resources (secrets, fleets, machines, etc) wholly outside of your VCS-provider and such that organizations can prevent code (on a per-job basis) from running on certain resources.

PR #1231, opened 2017-07-04
Swap 3000 and 8080

mikeschinkel:
All your examples show Apache listening on `8080`, so of course as I read your example I was trying to go to `localhost:3000` and it kept failing. It was not until I wasted a lot of time and realized I needed to read your text very carefully that I finally realized that you had swapped the ports for this example, for some reason.  Basically I am suggesting you follow the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) here.


michelle-luna:
fixes in a new PR

PR #1250, opened 2017-07-10
Banner landing page edits

michelle-luna:
remove banner, restyle sections and focus only on docs sections in new order on page


KunalJain:
lgtm

PR #1270, opened 2017-07-11
Update postgres example to current syntax

### LINKED ISSUE #1269: Make configuration steps more consistent
This issue is about <https://circleci.com/docs/2.0/postgres-config/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/postgres-config.md>)

The page contains several configurations examples that have `steps` written in different ways, but still achieve the same thing (I think so). At least for me that is confusing, as I start to wonder if there is some difference in the different ways of writing the commands.

## Things to do

 - Remove unneeded `type` from steps, example:
```
    steps:
      - type: checkout
```
should rather be:
```
    steps:
      - checkout
```

 - Convert `type: shell` to `run` steps, example:
```
      - type: shell
        name: Install Ruby Dependencies
        command: bundle install
```
should rather be:
```
      - run:
          name: Install Ruby Dependencies
          command: bundle install
```

Most likely there are other pages with similar inconsistency issues, and those should be changed as well.

marcomorain:
Fixes #1269

### DIFF
--- a/jekyll/_cci2/postgres-config.md
+++ b/jekyll/_cci2/postgres-config.md
@@ -43,20 +43,17 @@ jobs:
         - POSTGRES_USER=ubuntu
         - POSTGRES_DB=db_name
     steps:
-      - type: checkout
-      - type: shell
-        name: Install System Dependencies
+      - checkout
+      - name: Install System Dependencies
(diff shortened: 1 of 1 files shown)


marcomorain:
@keybits could you review this please?

keybits:
LGTM 👍

michelle-luna:
thank you!

PR #1282, opened 2017-07-12
Update workflows.md

michelle-luna:
make the defaults key more clear


kenden:
Thanks!
Minor critic: this block of text is a bit compact, hard to read.
Put the *Note* on a different line/paragraph?

Also, a question: is it possible to define several `defaults`?
Would it look like this?
```
default1: &deploy
default2: &git_tag
default3: &build_code
```

Or maybe like this?
```
defaults: 
  - &deploy
  - &git_tag
  - &build_code
```
I think the second option wouldn't work because of spaces but... maybe?

michelle-luna:
Thanks, I moved the Note after the example to make it easier to read. Either way you define the defaults will work, probably simpler in your first example because you would have some keys defined after each that would be harder to nest under a list.

kenden:
Thanks! LGTM

PR #1283, opened 2017-07-12

michelle-luna:
add info about branches


michelle-luna:
Merging to prevent folks from doing the wrong thing, but please add comments I can incorporate tomorrow.

PR #1285, opened 2017-07-13
Make working_directory optional

kimh:
Now we have default `working_directory` `~/repo`. Best viewed with `?w=1`


ndintenfass:
lgtm

felicianotech:
Not a big fan of using `repo` as the default but at least it's short.

kimh:
@michelle-luna 

FYI: I changed the default to `~/project` and updated the PR.

> Does that mean for every job that does not define working_directory, steps are run in the /repo directory?

Yes (now it's `~/project`, though).

>Why would someone define a working directory other than /repo? Thanks!

The value of defining the top level working directory is a bit subtle. Some cases I can think of are

- `~/project` is already created in customer's container and conflict with our default (remember that `~/project` must be an empty directory)
- `~` `($HOME)` is not writable in customer's container. In this case, `checkout` step fails, so you want to change to something else e.g. `/tmp/project`

Why was `working_directory` required if little use?? It's a bit of history: we wanted to make every configuration as explicit as possible in the beginning but found out the usability hurts from many customers feedbacks.

However, the real value of defining own `working_directory` comes from each `run` step like the following.

```
steps:
  - checkout
  - run:
      command: echo unit test
      working_directory: unit_tests

  - run:
      command: echo integ test
      working_directory: integ_tests
```

With this, we checkout to `~/project` and the tests commands are run on `~/project/unit_tests` and `~/project/integ_tests`.

Let me know if you have more questions.

keybits:
LGTM - this is a good update

keybits:
@michelle-luna can you ping me on Slack when this is merged and I'll make a Discuss post about it too. Thanks!

PR #1291, opened 2017-07-13
Change language about "build" job to be more accurate

ndintenfass:


keybits:
How about a compromise?

"The `build` job is the default job and is required if you are not using Workflows. If you are using Workflows, a job named `build` is optional. See the [Workflows]({{ site.baseurl }}/2.0/workflows) documentation for detailed information."

Issue #1289, opened 2017-07-13
Find command help on page is for specific UID


eric-hu:
This issue is about <https://circleci.com/docs/2.0/high-uuid-error/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/high-uuid-error.md>)

If I try to reproduce the output on this docs page, I get the same output:

```
root@624866325c2e:~# docker run -it circleci/doc-highid sh

# # # find / \( -uid 1000000 \)  -ls 2>/dev/null
       50      0 -rw-r--r--   1 veryhigh veryhigh        0 Jul  9 03:05 /file-with-high-id

```

However, a user typically doesn't know the high UID/GID users in a Docker image.  They expect to copy and paste this command and see all users with a high UID.  If I use a lower number from the above example, I get no matches:

```
# find / \( -uid 999999 \) -ls 2>/dev/null
#
```

PR #1296, opened 2017-07-14
Improve UID troubleshooting doc

### LINKED ISSUE #1288: Find command is platform-specific
This issue is about <https://circleci.com/docs/2.0/high-uuid-error/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/high-uuid-error.md>)

The find command `find / \( -uid 1000000 \)  -ls 2>/dev/null` is for BSD and/or GNU find.  This shows no results for `find` on an alpine image with a high GID/UID.

kimh:
Fixes #1288 #1289

Also, fixing the wrong file name. I understand changing file name is not good in general in SEO perspective, but the page is new and not critical, so I think should be fine.


kimh:
Note to self: update the places that have links to the page once this is deployed.

PR #1308, opened 2017-07-16
Fix workflow version and jobs nest level

inuscript:
It is mislead for workflow document for nested version and jobs.


michelle-luna:
Thanks for this PR. I think the nesting is correct as-documented, asking for a couple of other reviewers to check this out.

PR #1311, opened 2017-07-17
More edits to workspaces

michelle-luna:


jekyll/_cci2/configuration-reference.md
```diff
@@ -650,7 +650,7 @@ Note the following distinctions between Artifacts, Workspaces, and Caches:
 | Type      | lifetime        | Use                      | Example |
 |-----------|-----------------|------------------------------------|---------
 | Artifacts | Months          | Preserve long-term artifacts. |  Available in the Artifacts tab of the Build details under the `tmp/circle-artifacts.<hash>/container`   or similar directory.     |
-| Workspaces | Duration of workflow        | Move temporary files around for use by jobs run later in the workflow. | The `persist_to_workspace: <directory>` key under jobs, refer to the file with `attach_workspace: at: <directory>`.    |
+| Workspaces | Duration of workflow        | Move temporary files around for use by jobs run later in the workflow. | The `persist_to_workspace:` key under jobs, refer to the file with the `attach_workspace:` key.    |
```
gordonsyme: It would be more accurate to say "attach the workspace in a downstream container with the `attach_workspace:` step."

`attach_workspace` copies and re-creates the entire workspace content when it's run.

jekyll/_cci2/workflows.md
```diff
@@ -224,9 +224,9 @@ For example, Scala projects typically require lots of CPU for compilation in the
 
 A second example is a project with a `build` job that builds a jar and saves it to a workspace. The `build` job fans-out into the `integration-test`, `unit-test`, and `code-coverage` to run those tests in parallel using the jar.
 
-To persist data from a job and make it available to other jobs, configure the job to use the `persist_to_workspace` key where the value is a directory inside the project’s working directory. Data from the job will be saved to this directory until the job is rerun and new data is created.
+To persist data from a job and make it available to other jobs, configure the job to use the `persist_to_workspace` key. Data from the job will be saved until the job is rerun and new data is created.
```
gordonsyme: Maybe something like:

> Files and directories named in the `paths:` property of `persist_to_workspace` will be uploaded to the workflow's temporary workspace and made available for subsequent jobs to use.

jekyll/_cci2/workflows.md
```diff
@@ -224,9 +224,9 @@ For example, Scala projects typically require lots of CPU for compilation in the
```
gordonsyme: I think we need an explanation of how workspaces work so people can grok the mental model, maybe

> Each workflow has an associated workspace which can be used to transfer files to downstream jobs as the workflow progresses.
> The workspace is an additive-only store of data. Jobs can persist data to the workspace, this archives the data and creates a new "layer" in an off-container store.
> Downstream jobs can attach the workspace to their container filesystem, this downloads and unpacks each "layer" based on the ordering of the upstream jobs in the workflow graph.

jekyll/_cci2/configuration-reference.md
michelle-luna: ok, thank you!

jekyll/_cci2/workflows.md
michelle-luna: great, I added this at the top of the section

PR #1315, opened 2017-07-18

michelle-luna:


michelle-luna:
Just some editorial changes, but want to make sure this is still correct and see if there are other concepts we should describe a bit more? I wonder why someone would choose the various options on this page, maybe we could add some examples or scenarios for the second option for true and false?

ryanwohara:
This is no longer correct because the bottom of the Dockerfile is what changes. It's confusing because when you stack the layers, the bottom of the Dockerfile becomes the top layer; it's basically the inverse after things are compiled. 

Anything you change in a Dockerfile invalidates the cache of that line and everything after it.

PR #1316, opened 2017-07-18

michelle-luna:
clarify env var template


ryanwohara:
The information is accurate.
`the` should not be capitalized for consistency with the other lines.

PR #1318, opened 2017-07-19
Using a private image on the `machine`

ryanwohara:
It came to my attention we don't currently document the ability to use the `machine` executor for private images. We will also need to elaborate on the `executor-types.md` page.


ita47726:
Okey friend! ;) I understand it

keybits:
Awesome - this is much needed. LGTM

PR #1335, opened 2017-07-23
Add quotes to checksum parameter

lnikkila:
Omitting the quotes seems to fail with something like:

    function "Dockerfile" not defined


michelle-luna:
this is fixed in the current file, so closing this PR, thank you!

PR #1354, opened 2017-07-27
Update local-jobs.md

michelle-luna:


myabc:
@michelle-luna thanks!

PR #1365, opened 2017-07-31
Document private docker image support

notnoopci:


felicianotech:
@notnoopci Is this config change all this is needed or is `setup_remote_docker` still needed?

notnoopci:
@felicianotech this makes private image support on bar with public image in docker executor.  `setup_docker_remote` isn't needed for simply using private images but will be needed to build images, run special docker commands, etc.

PR #1368, opened 2017-07-31
Fix indent space characters

inuscript:
I found broken yml indent.
orchestrating_workflows_-_circleci

And it's occure space character


michelle-luna:
thank you, this is fixed in the current file, so closing this PR.

PR #1371, opened 2017-07-31
Contexts update

michelle-luna:
please review for accuracy and omissions


jekyll/_cci2/contexts.md
```diff
@@ -5,23 +5,24 @@ short-title: "Using Contexts"
...
+Contexts are created on the Settings page of the CircleCI application, in the Organization section. After a context is set in the application it may be configured in the workflows section of the `config.yml` file for a project.
 
-<hr>
+To use environment variables set on the Contexts page, the person running the workflow must be a member of the organization for which the context is set and the rule must allow access to all projects in the org. The default context name is `org-default` and the default rule allows access to all projects in the org.
 
-## Contexts Overview
-Contexts provide centrally managed, secured resources for use in jobs. They are currently primarily used for sharing environment variables across projects.
+## Using a Context
 
-_**Coming Soon:** We will be adding new capabilities to Contexts such the ability to route jobs to certain workers, share encrypted files like certificates across jobs, and new kinds of security rules for who can execute jobs in a Context. What capabilities would you like to see? Let us know in [our Discuss forum](https://discuss.circleci.com/t/contexts-feedback/13908)._
```
ndintenfass: Why are we removing this?

jekyll/_cci2/contexts.md
```diff
@@ -32,21 +33,9 @@ workflows:
...
 4. Environment variables declared inside a shell command in a `run` step of a job.
 
-## Creating and Editing Contexts
-You can manage your organization's Contexts from the Organization Settings page. 
-
-To create a new context hit the "Add Context" button from the Contexts page.
-
-You can edit existing Contexts by clicking them in the list on the Context page.
-
-## Context Security
```
ndintenfass: Can we keep these preview sections?

jekyll/_cci2/contexts.md
```diff
@@ -5,23 +5,24 @@ short-title: "Using Contexts"
...
 
-<hr>
+To use environment variables set on the Contexts page, the person running the workflow must be a member of the organization for which the context is set and the rule must allow access to all projects in the org. The default context name is `org-default` and the default rule allows access to all projects in the org.
 
-## Contexts Overview
-Contexts provide centrally managed, secured resources for use in jobs. They are currently primarily used for sharing environment variables across projects.
+## Using a Context
 
-_**Coming Soon:** We will be adding new capabilities to Contexts such the ability to route jobs to certain workers, share encrypted files like certificates across jobs, and new kinds of security rules for who can execute jobs in a Context. What capabilities would you like to see? Let us know in [our Discuss forum](https://discuss.circleci.com/t/contexts-feedback/13908)._
+1. Navigate to the Settings > Contexts page in the CircleCI application. The Create a Context button appears.
```
ndintenfass: This is only true if someone has the org-centric UI on - will that launch too?

jekyll/_cci2/env-vars.md
```diff
@@ -16,6 +16,10 @@ This document describes using environment variables in CircleCI in the following
 
 To add keys or secret environment variables to your private project, use the Environment Variables page of the Build > Project > Settings in the CircleCI application. The value of the variables are neither readable nor editable in the app after they are set. To change the value of an environment variable, delete the current variable and add it again with the new value. It is possible to add individual variables or to import variables from another project. Do **not** add keys or secrets to a public CircleCI project.
 
+### Adding Global Environment Variables
+
+To add global environment variables that may be shared across projects, use the Settings > Contexts page of the CircleCI application. See the [Contexts]( {{ site.baseurl }}/2.0/contexts/) documentation for instructions.
```
ndintenfass: This only applies when the org-centric UI is on. Will that be launched in time?

jekyll/_cci2/workflows.md
```diff
@@ -107,6 +107,38 @@ The dependencies are defined by setting the `requires:` key as shown. The `deplo
 
 See the [Sample Sequential Workflow config](https://github.com/CircleCI-Public/circleci-demo-workflows/blob/sequential-branch-filter/.circleci/config.yml) for a full example.
 
+## Job Contexts Example
+
+The following example shows a workflow with four sequential jobs that use shared environment variables. 
+
+The following `config.yml` snippet is an example of a sequntial job workflow configured to use the resources defined in the `org-global` context:
+
+```
```
ndintenfass: I'd suggest showing an example where not every job uses the context. Perhaps in this case just the test1 and test2 would be appropriate.

jekyll/_cci2/contexts.md
```diff
@@ -32,21 +33,9 @@ workflows:
...
 ```
 
-## Environment Variables in Contexts
-Contexts have a set of environment variables defined as name/value pairs. When a job is run with a Context the environment variables of the Context are injected at runtime. The order of precedence is important to remember when working with environment variables that might overlap in name. Following is the order, with later injection always taking precedence when the steps of your job run:
+**Note:** Environment variables are considered in a specific order, as follows:
 1. Project-level environment variables set on the Project Settings page.
 2. Context environment variables (assuming the user has access to the Context).
-3. Environment variables set in your config YAML with the `environment` key of a job definition.
+3. Environment variables set in the `config.yml` file with the `environment` key of a job definition.
 4. Environment variables declared inside a shell command in a `run` step of a job.
```
marcomorain: The (existing) wording here is ambiguous. There are two was a user could specify an env var in a `run` step:
- inside the shell command like `FOO=bar make install`
- in the `environment` Key of a `run` step.

marcomorain: Sorry - can't edit on my phone - I meant "two ways"

jekyll/_cci2/configuration-reference.md
```diff
@@ -717,6 +717,14 @@ Key | Required | Type | Description
 requires | N | List | A list of jobs that must succeed for the job to start
 {: class="table table-striped"}
 
+##### **`contexts`**
+Jobs may be configured to use global environment variables set for an organization, see the [Contexts]({{ site.baseurl }}/2.0/workflows) document for adding a context in the application settings.
+
+Key | Required | Type | Description
+----|-----------|------|------------
+context | N | String | The name of the context. The default name is `org-global`.
```
marcomorain: The name of the key changes here - under "Jobs" it's referred to as `contexts` and here it's referred to as `context`.

jekyll/_cci2/contexts.md
rishimkumar: Yeah I'd leave this please.

jekyll/_cci2/contexts.md
rishimkumar: That is coming soon but i dont fully understand your concern here.  for clarity lets just write "Organization Settings" instead of Settings and it's agnostic.

jekyll/_cci2/env-vars.md
rishimkumar: same as above. am i missing something? you can get to org settings regardless of what UI you are using.

rishimkumar: @ndintenfass

jekyll/_cci2/contexts.md
ndintenfass: Your suggested change would solve my concern

jekyll/_cci2/env-vars.md
ndintenfass: As it's written, if someone did NOT have org ui the instruction to go to "Settings" is ambiguous.

jekyll/_cci2/contexts.md
michelle-luna: This content doesn't belong in the docs. When is "soon"? Why are the docs asking me a question when I'm here looking for answers? Can we put this in a blog post instead?

jekyll/_cci2/configuration-reference.md
michelle-luna: thanks, fixed

jekyll/_cci2/contexts.md
michelle-luna: I'd rather this go into a blog post or on discuss itself, when we have a new feature or piece of functionality, we add it to the docs.

jekyll/_cci2/contexts.md
michelle-luna: When I do this myself, I just click Settings and then Contexts, is there a case when the user  would click something called Organization Settings?

jekyll/_cci2/workflows.md
michelle-luna: ok, changed

jekyll/_cci2/contexts.md
michelle-luna: thanks, I made this change

jekyll/_cci2/configuration-reference.md
michelle-luna: fixed, thanks

jekyll/_cci2/contexts.md
ndintenfass: The point is to try engaging people earlier in the process and give them a feel for how this feature will be evolving.

jekyll/_cci2/contexts.md
ndintenfass: When the top-bar UI is not on there is "Settings" in the nav.

jekyll/_cci2/contexts.md
michelle-luna: I understand the point, but this engagement should happen in a blog or on discuss, we add info to the docs when we have the features available (docs are for features that are available). The feeling that a feature will be changing is not appropriate for documentation, when you read docs you want the feeling that features are stable and will work as-documented (not that they will change soon).

jekyll/_cci2/env-vars.md
michelle-luna: I think I need a screenshot to understand what change you want me to make here.

jekyll/_cci2/contexts.md
ndintenfass: Makes me sad, but it's your call to make. I'd like to try to convince you otherwise, but we can take that offline.

rishimkumar: i agree with Michelle's premise but I felt Nathan's remarks mostly fit within that - i.e., I don't think we're too far off. 
I suggest we leave a pointer to the Discuss post and the text on the docs say something akin to "We're interested in your feedback on how Contexts can evolve. Read more about our proposed roadmap and offer suggestions here"

michelle-luna: thanks, I've added a send feedback section with this sentence and a link to the discussion post.

Issue #1362, opened 2017-07-29
Examples for storing test results are out of date


radarsh:
This issue is about <https://circleci.com/docs/2.0/collect-test-data/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/collect-test-data.md>)

The examples listed no longer work without modifications. For instance, in case of Gradle builds, the example shows:

```yml
steps:
  - run: |
       mkdir -p /junit/
       find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} /junit/ \;
       when: always
  - store_test_results:
       path: /junit
  - store_artifacts:
       path: /junit      
```

I had to modify it to look like this:

```yml
steps:
  - run:
       name: Save test results 
       command: |
         mkdir -p ~/junit/
         find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
       when: always
  - store_test_results:
       path: ~/junit
  - store_artifacts:
       path: ~/junit
```

To summarise:

1. The shorthand run command syntax does not support `when: always`. So, the step was never being run. This is well [documented](https://circleci.com/docs/2.0/configuration-reference/#shorthand-syntax).
    > In this case, command and name become the string value of run, and the rest of the config map for that run have their default values.

2. The build is not executed as root. It's as a `circleci` user. So we should either use `sudo mkdir /junit` or simply create the directory inside the home directory (`~/junit`).

michelle-luna:
thanks, much appreciated I think I fixed the rest of the examples in this PR https://github.com/circleci/circleci-docs/pull/1383

PR #1379, opened 2017-08-02
Update gradle test results reporting for 2.0

erichelgeson:
see https://github.com/circleci/circleci-docs/issues/1362


marcomorain:
@erichelgeson does leaving out the `find` command work for you? Something like this:

```
steps:
  - store_test_results:
       path: build/test-results
  - store_artifacts:
       path: build/test-results
```

It's less general but neater.

erichelgeson:
The `find` is required for me since I have multi-project builds with multiple test phases, eg, 3 projects, with unit and integration.

PR #1388, opened 2017-08-05
Adding workflows

KunalJain:
Adding workflows to run deployment

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,7 @@ jobs:
           path: run-results/
           destination: run-results
   deploy:
+      - run:
           name: Deploy to S3 if tests pass and branch is Master
           command: aws s3 sync jekyll/_site/docs s3://circle-production-static-site/docs/ --delete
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -45,11 +45,32 @@ jobs:
...
           path: run-results/
           destination: run-results
-      - deploy:
+      - persist_to_workspace:
+          root: ~/circleci-docs/jekyll/_site 
+          paths: 
+            - docs
+  deploy:
+    docker:
+      - image: python:2.7
```
ProjectFrank: we can change this back to `circleci/python:2.7`

.circleci/config.yml
```diff
@@ -45,11 +45,32 @@ jobs:
...
+    working_directory: ~/circleci-docs  
+    steps:
+      - attach_workspace:
+          at: ~/circleci-docs/jekyll/_site 
+      - run:
+          name: Install awscli
+          command: pip install awscli
```
ProjectFrank: we should make this `sudo pip install awscli`

PR #1392, opened 2017-08-07
Improve wf overview

michelle-luna:
This update changes the emphasis from flexibility to speed in the overview, simplifies the first example and does some restructuring to simplify the TOC and put some of the features that enable efficiency at the top.


KunalJain:
lgtm

PR #1399, opened 2017-08-08
Improve job and build descriptions

marcomorain:
- The `jobs` stanza is a map, not a list.
- Be more clear about what jobs are run when.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -34,11 +34,19 @@ The `version` field is intended to be used in order to issue warnings for deprec
 
 ## **`jobs`**
 
-Each job is an item in the `jobs` list.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -34,11 +34,19 @@ The `version` field is intended to be used in order to issue warnings for deprec
 
 ## **`jobs`**
 
-Each job is an item in the `jobs` list.
+A build is comprised of 1 or more named jobs. These jobs are specified in the
```
KunalJain: A workflow is comprised of 1 or more names jobs.

marcomorain: There is not always a workflow though...

KunalJain: In 2.0 world everything is a job. I would really like to move away for build.

PR #1406, opened 2017-08-09
CIRCLE-5713 - Some clarifications on how to migrate

gordonsyme:
* Say where to set parallelism to 1 when migrating to 2.0
* Add an example for how to replace `checkout: post:` commands , it's not immediately clear that each command from the `checkout: post:` list should be a separate step otherwise.

### DIFF
--- a/jekyll/_cci2/migrating-from-1-2.md
+++ b/jekyll/_cci2/migrating-from-1-2.md
@@ -16,7 +16,7 @@ If you do not have a `circle.yml` file, refer to the [Sample 2.0 `config.yml` Fi
 
 ## Overview
 
-CircleCI 2.0 introduces the requirement that you create a configuration file (`.circleci/config.yml`), and it adds new required keys for which values must be defined. This release also allows you to use multiple jobs in your configuration. **Note:** If you configure multiple jobs, it is important to have parallelism set to `1` to prevent duplication of job runs.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -16,7 +16,7 @@ If you do not have a `circle.yml` file, refer to the [Sample 2.0 `config.yml` Fi
 
 ## Overview
 
-CircleCI 2.0 introduces the requirement that you create a configuration file (`.circleci/config.yml`), and it adds new required keys for which values must be defined. This release also allows you to use multiple jobs in your configuration. **Note:** If you configure multiple jobs, it is important to have parallelism set to `1` to prevent duplication of job runs.
+CircleCI 2.0 introduces the requirement that you create a configuration file (`.circleci/config.yml`), and it adds new required keys for which values must be defined. This release also allows you to use multiple jobs in your configuration. **Note:** If you configure multiple jobs, it is important to set the project parallelism to `1` on the project settings page to prevent duplication of job runs.
```
KunalJain: Users don't have to set parallelism in UI anymore for 2.0. `parallelism` is set at a job level.

michelle-luna: This is migration, so if they had it set to something other than 1 before, the idea was to have them set it back to 1 when we added this I believe. If that is not the case, maybe we should say the UI setting for 1.0 will be ignored in 2.0 (or whatever we actually do?)

gordonsyme: I just checked by running a build, we ignore project parallelism. I'll update the note.

PR #1422, opened 2017-08-11
Be more explicit about the default shell.

marcomorain:


jekyll/_cci2/configuration-reference.md
```diff
@@ -316,7 +316,11 @@ Each `run` declaration represents a new shell. It's possible to specify a multi-
 
 ##### _Default shell options_
 
-Our default `shell` has a few options enabled by default:
+The default value for the `shell` option is either `/bin/bash -eo pipefail` or
+`/bin/sh -eo pipefail`, depending on whether `/bin/bash` is present or not in
+the build container. The default shell is not a login shell (`--login` or `-l`
```
notnoopci: I think the following is a tad clearer: "The default value of `shell` option is `/bin/bash -eo pipefail` if `/bin/bash` is present in build container.  Otherwise it is `/bin/sh -eo pipefail`."

marcomorain: agreed

marcomorain:
@ryanwohara says:

> rohara Only other thing I'd add to the shells section is how you can define `shell: sudo bash -eo pipefail` to avoid prefixing everything in a section with sudo

jekyll/_cci2/configuration-reference.md
ataggart: What is the significance to the user of the default shell not being a login shell?

eric-hu: A login shell will look for `~/.bash_profile`, `~/.bash_login`, `~/.profile` and load the first one it finds.  A non-login shell won't.

eric-hu:
It looks like this is hanging on an API call from CircleCI.  Is there anything we can do to refresh this?

jekyll/_cci2/configuration-reference.md
ataggart: Cool, I get that. There are multiple ramifications of being or not being in a login shell. It wasn't clear to me _why_ you are letting them know it's not a login shell. E.g., is it to remind them that their `.profile` won't be sourced? Some of our users (even if they do know that fact) might not make those connection on their own.

eric-hu: That's a really good point.  I might fall into the category of "not knowing what connections to make" as well.  Besides sourcing these login files, are there other ramifications of running a script in a login shell vs not?

michelle-luna:
sorry this took so long to get back to, fixed in a new PR https://github.com/circleci/circleci-docs/pull/1645, so closing this one

PR #1438, opened 2017-08-16
update private images doc to remove setup_remote_docker step

michelle-luna:


jekyll/_cci2/private-images.md
```diff
@@ -20,7 +30,6 @@ jobs:
     working_directory: ~/my_app
     steps:
       - checkout
-      - setup_remote_docker
```
ryanwohara: Removing setup_remote_docker means we can't use the subsequent `docker` commands.

michelle-luna: hmm, so should I leave that example as-is?

iynere:
Hmmmm I might change the entire example used on this page... The whole point is no longer having to run Docker commands manually in order to use private images, so I think having an example with `docker run` and `docker build` kind of dilutes that.

jekyll/_cci2/private-images.md
iynere: I would delete this from the second example with the `machine` executor:
```
 # build and test application
      - run: |
          docker build -t my-app .
          # assuming that our app expects to have DB on localhost
          docker run --network container:db my-app test
```
and for the first example, just copy the bit from the bottom of this section: https://circleci.com/docs/2.0/configuration-reference/#docker--machine-executor

iynere:
looks good to me !

PR #1461, opened 2017-08-23
Removing tabs

kimh:
Sadly 2.0 build falls back to 1.0 when there are tabs in config yml. Therefore, copy/paste these examples won't run on 2.

### DIFF
--- a/jekyll/_cci2/workflows.md
+++ b/jekyll/_cci2/workflows.md
@@ -76,7 +76,7 @@ workflows:
             - test1
       - deploy:
           requires:
-            - test2	
+            - test2
 ```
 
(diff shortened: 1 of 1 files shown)


kimh:
@michelle-luna I may forget to come back to this PR, so please merge away if looks good for you. thank you!

jekyll/_cci2/workflows.md
```diff
@@ -168,14 +168,14 @@ workflows:
       - test1:
           requires:
             - build
-	  context: org-global  
+    context: org-global
```
nj: Was somehow got replaced with the wrong amount of spaces, making it appear in an odd way.

Fixed in #1536

PR #1462, opened 2017-08-23
Update location of test timing data.

abernix:
The previous suggestion to check in `/.circleci-task-data/circle-test-results/` doesn't seem to be correct as it never exists (and appears to be incorrect anyway as I wouldn't expect the data to be mounted in the root of the drive).

Judging by the environment variable `CIRCLE_INTERNAL_TASK_DATA` and the actual discovery of a `results.json` in this corrected path with the timing information, I believe this is correct.


michelle-luna:
this is fixed in the current file, thank you!

PR #1468, opened 2017-08-24
Injected variables don't override project settings

brandondean:
When triggering a build that contain injected environment variables with the same key as a variable defined in the project settings, the project settings will take precedence. This expectation should be noted in the documentation.

This has been my experience thus far, and would be delighted to learn this isn't the case, but it should be noted.

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -116,7 +116,7 @@ export qux="{\"quux\": 1}"
 export list="[\"a\", \"list\", \"of\", \"strings\"]"
 ```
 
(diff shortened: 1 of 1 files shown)


marcomorain:
@brandondean thanks for pointing this out. This is not intentional (a bug). I've logged an issue for our platform team to address this.

marcomorain:
Logged in our internal JIRA as CIRCLE-6434

brandondean:
@marcomorain Solid. I'm going to leave this PR open until it gets fixed, but you are welcome to close it if you feel otherwise.

michelle-luna:
Closing because I had to fix this in https://github.com/circleci/circleci-docs/pull/1644 (adding workflows config won't allow checks to pass for PR before that change)

PR #1469, opened 2017-08-24
Update available 'machine' images

keybits:


keybits:
This page will need updating again after 2017-09-03 when the new `latest` image goes live - see: https://discuss.circleci.com/t/new-docker-versions-for-machine-images-on-circleci-2-0/15686

PR #1493, opened 2017-08-31
Update to official Elixir docker image

paxer:
update Elixir docker image to the official address,  add `run: mix local.hex` and `run: mix local.rebar` since both are required to present in a common Elixir application for `mix deps.get`


paxer:
@michelle-luna 👍   https://github.com/CircleCI-Public/circleci-demo-elixir-phoenix/pull/1

PR #1499, opened 2017-09-01
Update oom with instructions for OCaml/Reason

dwwoelfel:
Tested this on a private OCaml project running on Circle. Thanks to @stedolan for the explanation and @sgrove for testing.


marcomorain:
Hi @dwwoelfel and @stedolan! Two former colleagues in a single PR :D

PR #1522, opened 2017-09-06
Fix the side nav to make the markup and jekyll code cleaner, and chan…

ndintenfass:
…ge JS to make menus expand by default


keybits:
LGTM

PR #1540, opened 2017-09-08
add indication for variable names & workflow name

michelle-luna:


jekyll/_cci2/configuration-reference.md
```diff
@@ -718,9 +722,9 @@ Key | Required | Type | Description
 jobs | Y | List | A list of jobs to run with their dependencies
 {: class="table table-striped"}
 
-#### **`build`**
+#### **<`build`>**
 
-A unique name for your job.
+A job name that exists in your `config.yml`.
```
ataggart: This looks great! It may be redundant with the higher-up change, but leaving "unique" here might not be a bad idea.

michelle-luna: Thanks, I removed it because it actually *must* be a name that is already in your config.

PR #1542, opened 2017-09-08
Update google-container-engine.md

michelle-luna:


djmc:
What is the point of `base64`ing the GCR credentials then, when it just leads up to storing the raw JSON file in an environment variable? Adding this moots all the steps prior.

renaudguerin:
Seconding what @djmc said, the need for base64 encoding JSON inside environment variables should be clarified.

PR #1551, opened 2017-09-09
WIP: first draft of vpn support

GERey:


GERey:
This would probably be better in a different format so I'll put it on the back burner

PR #1552, opened 2017-09-09
Document support for AWS ECR Authentication

notnoopci:


jekyll/_cci2/configuration-reference.md
```diff
@@ -150,6 +151,19 @@ jobs:
...
 ```
 
+If you are using a private image hosted on [AWS ECR](https://aws.amazon.com/ecr/), we must authenticate using the AWS credentials.  By default, CircleCI uses AWS credentials provided through the project AWS Permissions settings or by setting the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` project environment variables.  Users can also set the credentials by using `aws_auth` field like the following:
+
+```
+jobs:
+  build:
+    docker:
+      - image: account-id.dkr.ecr.us-east-1.amazonaws.com/org/repo:0.1
+        auth:
```
keybits: looks like this should be `aws_auth`?

felicianotech: Looks like it. I'll update in a few.

notnoopci: Thanks for catching this - correct it's `aws_auth`

PR #1554, opened 2017-09-11
Update the "exclusive" section

eric-hu:
Add suggestions for how users should configure jobs with the `exclusive` option.  More context here: https://circleci.slack.com/archives/C0B6W2U0L/p1505132677000540


eric-hu:
Thanks for the consistency check @michelle-luna .  Does this look good to you for merging?

PR #1382, opened 2017-08-03
Migrate mobile 2

michelle-luna:
Please review for accuracy and omissions, I migrated all of this over from the existing 1.0 docs without knowing anything about how it will actually work on 2.0. So, it is just a first start and doesn't include the code signing document, but wanted to get it rolling.

PR #1557, opened 2017-09-12
mobile v2 docs

michelle-luna:
Here is the update to my original mobile migration PR with all the changes from the https://github.com/circleci/circleci-docs/pull/1382 feedback. (I had to start again because of the circleci-docs workflows config addition in between). Thanks for the reviews, please double check this is all still correct and the I interpreted comments appropriately. Also, I added the ios-game config.yml (minus the 8.2.1 job) to the sample-config.yml doc in case we want to document it, and both examples for ios need some intro content and config comments. Thanks!


DanToml:
@michelle-luna I just pushed a few changes for things that have changed in 2.0 or in the years since the docs were originally written. Would love a review or fixup on grammar/wording

michelle-luna:
Thanks Dani! These changes look good to me.

jekyll/_cci2/sample-config.md
```diff
@@ -230,3 +230,166 @@ workflows:
...
+      - deploy:
+          requires:
+            - build-and-test
+          filters:
+            branches:
+              only: master
+
+```
+
+## Sample Linux and macOS Configuration File
```
marcomorain: @DanToml let's use your Swift Lint example here?

PR #1559, opened 2017-09-13
Fix up some example macOS config.yml 

marcomorain:
Hi @michelle-luna, I've made some quick updates targeting your mobile docs branch.

The changes are mostly about the new `config.yml` syntax, and limiting the examples to xcode `9.0` and `8.3.3`.

We also do not have support for running builds by specifying only the `x.y` build number at present, so I have removed that advice.


jekyll/_cci2/ios-tutorial.md
```diff
@@ -52,10 +52,6 @@ jobs:
 
 Refer to <https://codesigning.guide/> for details.
 
-## Using A Provisioning Profile
```
marcomorain: We don't have provisioning profile support in 2.0

PR #1565, opened 2017-09-14
add note about java and erlang to resource class section

michelle-luna:
CIRCLE-4998

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -259,6 +259,8 @@ large       | 4.0 | 8GB
 xlarge      | 8.0 | 16GB
 {: class="table table-striped"}
 
+**Note:** Java and Erlang languages that introspect the `/proc` directory for information about CPU count may require additional configuration to attain benefits from the CircleCI 2.0 resource class feature. Users of Java and Erlang languages should pin their CPU count to their guaranteed CPU resources. 
+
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -259,6 +259,8 @@ large       | 4.0 | 8GB
 xlarge      | 8.0 | 16GB
 {: class="table table-striped"}
 
+**Note:** Java and Erlang languages that introspect the `/proc` directory for information about CPU count may require additional configuration to attain benefits from the CircleCI 2.0 resource class feature. Users of Java and Erlang languages should pin their CPU count to their guaranteed CPU resources.
```
eric-hu: > to attain benefits from the CircleCI 2.0 resource class feature

This makes it sound like an optional performance optimization.  From what I recall, this is performance related, but programs with this issue will request 32 cpu cores and run far slower than they would requesting 1 core.  I think this should include something about programs that run unusually slow.

Also, I've now seen similar behavior with a Javascript tool as well.  Perhaps this should be more general: note that we've seen it with Java and Erlang, but it could happen with other language runtimes too.

michelle-luna: thanks! I've made some edits to address your changes.

eric-hu:
Looks good to me

PR #1581, opened 2017-09-18
[CIRCLE-6790] Update IAM policy document

levlaz:
The policy does not work in the current state. Including these two additional permissions will allow users to use our recommended IAM policy to install CircleCI Server on AWS with terraform.


bellkev:
LGTM

PR #1582, opened 2017-09-18
adding scss from outer; updating nav

trevor-circleci:
nav needed to be updated, but more styles needed to be pulled for the new nav to render appropriately on the docs site. 

I've added those styles - many, in a first step towards eventual sync with outer.
ticket: https://circleci.atlassian.net/browse/MKTG-600


eric-hu:
This changes a pretty big set of files.  This only affects rendering on mobile for the hamburger menu?

trevor-circleci:
@eric-hu no it's not just fixing mobile nav; it's importing most of the styles from outer, so we can start standardizing it; it also fixes the font rendering issues we were having, and adds some necessary JS for the mobile nav to work properly.

There is some bloat with the changes to the gemfile - it turns out that wasn't necessary; same with some config file changes. 

I'll revert some of those now.

eric-hu:
@trevor-circleci It looks like from the Gemfile.lock that gems are still being updated in "Files changed".  Were there ~300 files pulled from outer, or are these leftover from gem installs?

PR #1590, opened 2017-09-21
Clarify jobs/workflows caching example

eric-hu:
Addressing 2 issues:

1.  Old example implies caches are re-writeable: 

> Cache is written in chronological order. Consider a workflow of Job1 -> Job2 -> Job3. If Job1 and Job3 write to the same cache key, a rerun of Job2 may use the changes written by Job3 because Job2 ran last.

I reworked this example to emphasize that workflow jobs can share caches, and a race condition to be aware of.

2. Added an example of another kind of job cache race condition: a cache not yet existing when it's requested.  This came up in this ticket:

https://circleci.zendesk.com/agent/tickets/23696/events

Let me know if 1) changed the intent of the original doc too much and I can cut that part out.


keybits:
LGTM

PR #1594, opened 2017-09-22
WIP: Add a section for automatic config translation

kimh:
We are about to present the automatic config translation feature to customers. Please don't merge until all preparation on my side is done.

Best viewed with `?w=1`


cayennes:
Hi

This is *not* ready to become not-WIP; the feature isn't ready.  Also the interface is changing so the documentation will need to be rewritten (and I don't think there was even an intermediate point when it ever worked exactly as expressed here).  Sorry for the confusion.

cayennes:
Oops, I see that you were changing the title to rather than from WIP.  Thanks.

PR #1596, opened 2017-09-22
Update github-security-ssh-keys.md

eric-hu:
Remove link that's no longer valid with org-centric UI.  The link works, but takes me to a page that I can't add a project from.

Add suggestion about using incognito window since many users log out of only one site and get confused.


eric-hu:
Also worth noting, this README page applies to 1.0 and 2.0.  (The README URL says 1.0).  Just mentioning it because we've made 1.0 work a lower priority.

PR #1603, opened 2017-09-26
Update Docker Layer Caching

heug:
Add/edit instructions for machine executor
Remove references to `reusable` and `exclusive` which are now deprecated

PR #1604, opened 2017-09-26
Update to use setup_docker_engine

heug:
Remove references to setup_remote_docker
Add text on ability to run on machine executor

### DIFF
--- a/jekyll/_cci2/api-job-trigger.md
+++ b/jekyll/_cci2/api-job-trigger.md
@@ -60,7 +60,7 @@ jobs:
       - image: ruby:2.4.0
     working_directory: /
     steps:
-      - setup_remote_docker
+      - setup_docker_engine
       - run: echo "deploy section running"
 ```
 
--- a/jekyll/_cci2/building-docker-images.md
+++ b/jekyll/_cci2/building-docker-images.md
@@ -14,7 +14,7 @@ This document explains how to build Docker images for deploying elsewhere or for
 
 ## Overview
 
-To build Docker images for deployment, you must use a special `setup_remote_docker` key which creates a separate environment for each build for security. This environment is remote, fully-isolated and has been configured to execute Docker commands. If your build requires `docker` or `docker-compose` commands, add the `setup_remote_docker` step into your `.circleci/config.yml`:
+To build Docker images for deployment, you must use a special `setup_docker_engine` key which creates a separate environment for each build for security. This environment is remote, fully-isolated and has been configured to execute Docker commands. If your build requires `docker` or `docker-compose` commands, add the `setup_docker_engine` step into your `.circleci/config.yml`:
 
 ```YAML
 jobs:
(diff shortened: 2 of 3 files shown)


felicianotech:
We would need to change this everywhere then right? The old name is here: https://circleci.com/docs/2.0/docker-layer-caching/

Also, from that doc, it says:

> Because Docker Layer Caching uses the setup_remote_docker step, it is not compatible with the machine executor.

Since `setup_docker_engine` can now be used in `machine`, does that mean layer caching can be to? or does that exception still apply?

felicianotech:
Nevermind. You did this across two PRs: https://github.com/circleci/circleci-docs/pull/1603

jekyll/_cci2/building-docker-images.md
```diff
@@ -14,20 +14,22 @@ This document explains how to build Docker images for deploying elsewhere or for
...
 
-      - setup_remote_docker
+      - setup_docker_engine
 ```
 
-When `setup_remote_docker` executes, a remote environment will be created, and your current [primary container][primary-container] will be configured to use it. Then, any docker-related commands you use will be safely executed in this new environment.
+When `setup_docker_engine` executes, a remote environment will be created, and your current [primary container][primary-container] will be configured to use it. Then, any docker-related commands you use will be safely executed in this new environment.
 
-*Note: `setup_remote_docker` is not curently compatible with the `machine` executor.*
+`setup_docker_engine` is now enabled automatically when initializing a machine executor.
```
smaant: I think I understand the idea behind this sentence but think it might be a bit confusing. Machine executor and `setup_remote_docker` are two separate options that enable you running docker command during the build. But they work differently and machine executor doesn't have limitations of remote docker. If we want to say what users can use machine executor when they need to run docker, I think, it might be better to just say it.

jekyll/_cci2/building-docker-images.md
```diff
@@ -59,24 +62,25 @@ jobs:
...
 ```
 
 Let’s break down what’s happening during this build’s execution:
 
 1. All commands are executed in the [primary container][primary-container].
-2. Once `setup_remote_docker` is called, a new remote environment is created, and your primary container is configured to use it.
-3. All docker-related commands are also executed in your primary container, but building/pushing images and running containers happens in the remote Docker Engine.
-4. We use project environment variables to store credentials for Docker Hub.
+2. Once `setup_docker_engine` is called, a new remote environment is created, and your primary container is configured to use it.
+3. https://circleci.com/docs/2.0/docker-layer-caching/
```
smaant: Would it be rendered just as is or replaced with page title? I mean we can add more user-friendly text, i.e. "Read more about Docker Layer Caching".

jekyll/_cci2/building-docker-images.md
michelle-luna: Is the `setup_remote_docker` key supported with `machine`? Or should we keep this note about it not being compatible?

smaant: It makes no sense to call setup_remote_docker from machine executor, so I would keep this comment but we can update it a bit to say that machine executor allows running docker commands by itself.

michelle-luna: ok, thanks, will do

jekyll/_cci2/building-docker-images.md
michelle-luna: fixed

jekyll/_cci2/api-job-trigger.md
```diff
@@ -60,7 +60,7 @@ jobs:
       - image: ruby:2.4.0
     working_directory: /
     steps:
-      - setup_remote_docker
+      - setup_docker_engine
```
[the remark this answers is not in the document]
michelle-luna: got it, thanks

michelle-luna:
closing and re-doing this in a new PR

PR #1620, opened 2017-09-27
Added notes regarding submodules

drazisil:
We do not automatically checkout git submodules. This causes confusion to customers who use them. While this is documented in the 1.0 docs https://circleci.com/docs/1.0/manually/#checkout it is completely missing from 2.0. Correcting this.


felicianotech:
I'm not a reviewer so I'll just leave this here, LGTM.

PR #1622, opened 2017-09-27
Change out base/generic Docker image tag recommendation.

felicianotech:
We currently recommend the Docker image `docker:17.05.0-ce-git` as the base, generic, use this if you just need a basic Docker image with git support.

That version of Docker is no longer supported. Docker has a release cycle that includes longer, "stable" versions and intermediate release images. Very similar to the model Ubuntu uses.

The Docker Hub repo has a tag `docker:stable-git` that will move in between stable, longer supported releases only and not intermediate releases (which Docker 17.05 was).

### DIFF
--- a/jekyll/_cci2/executor-types.md
+++ b/jekyll/_cci2/executor-types.md
@@ -21,7 +21,7 @@ jobs:
     docker:
       - image: buildpack-deps:trusty
 ```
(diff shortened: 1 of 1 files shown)


eric-hu:
LGTM

PR #1629, opened 2017-09-28
Dlc and reusable updates

michelle-luna:


jekyll/_cci2/docker-layer-caching.md
```diff
@@ -17,15 +17,11 @@ Consider reusing the unchanged layers to significantly reduce image build times.
 
 ``` YAML
 - setup_remote_docker:
-    reusable: true    # default - false
+    docker_layer_caching: true
```
smaant: default is also `false`

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -17,15 +17,11 @@ Consider reusing the unchanged layers to significantly reduce image build times.
...
     exclusive: true   # default - true
```
smaant: we need to remove `exclusive`, not there'll only `docker_layer_caching` option

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -17,15 +17,11 @@ Consider reusing the unchanged layers to significantly reduce image build times.
...
 ```
 
-## **`reusable`**
-
-When `reusable` is set to `true`, CircleCI will try to reuse your Remote Docker Environment. That is, every layer you built in a previous job will be accessible in the remote environment and CircleCI will attempt to reuse the previous environment when it is possible. However, in some cases your job may run in a clean environment, even if the configuration specifies `reusable: true`.
-
-If you run many parallel jobs for the same project, and each job requests a reusable environment, all of them will be provided with a Remote Docker Environment. However, not all of the jobs will have cached layers, although this behavior is subject to change in a future update.
+**Note:** Previously the `docker_layer_caching` was called `reusable`. The `reusable` key is deprecated in favor of the `docker_layer_caching` key. 
 
 ## **`exclusive`**
```
smaant: we need to remove this whole section (and probably add note that it was here before, but not anymore, now all VMs will be treated as `exclusive: true`.

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -17,15 +17,11 @@ Consider reusing the unchanged layers to significantly reduce image build times.
...
 ``` YAML
 - setup_remote_docker:
-    reusable: true    # default - false
+    docker_layer_caching: true
     exclusive: true   # default - true
 ```
 
-## **`reusable`**
-
-When `reusable` is set to `true`, CircleCI will try to reuse your Remote Docker Environment. That is, every layer you built in a previous job will be accessible in the remote environment and CircleCI will attempt to reuse the previous environment when it is possible. However, in some cases your job may run in a clean environment, even if the configuration specifies `reusable: true`.
```
smaant: we still need to document what `docker_layer_caching` does and what to expect from it.

smaant:
This PR https://github.com/circleci/circleci-docs/pull/1603 has a correct approach (excluding `setup_docker_engine` thing), requires some editing though.

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -17,20 +17,19 @@ Consider reusing the unchanged layers to significantly reduce image build times.
 
 ``` YAML
 - setup_remote_docker:
-    reusable: true    # default - false
-    exclusive: true   # default - true
+    docker_layer_caching: true # default - true  
 ```
 
-## **`reusable`**
+When `docker_layer_caching` is set to `true`, CircleCI will try to reuse your Remote Docker Environment. That is, every layer you built in a previous job will be accessible in the remote environment and CircleCI will attempt to reuse the previous environment when it is possible. However, in some cases your job may run in a clean environment, even if the configuration specifies `docker_layer_caching: true`.
```
smaant: I would rather say something like "CircleCI will try to reuse Docker Images (layers) from your previous builds"  instead of "CircleCI will try to reuse your Remote Docker Environment"

smaant: "and CircleCI will attempt to reuse the previous environment when it is possible" seems like a repetition and I would probably avoid using phrases like "reuse environment" because it might give wrong expectations.

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -17,20 +17,19 @@ Consider reusing the unchanged layers to significantly reduce image build times.
...
 
-When `reusable` is set to `true`, CircleCI will try to reuse your Remote Docker Environment. That is, every layer you built in a previous job will be accessible in the remote environment and CircleCI will attempt to reuse the previous environment when it is possible. However, in some cases your job may run in a clean environment, even if the configuration specifies `reusable: true`.
+If you run many parallel jobs for the same project that depend on the same environment, all of them will be provided with a Remote Docker Environment. Docker Layer Caching guarantees jobs to have exclusive Remote Docker Environments that other jobs cannot access. However, not all of the jobs will have cached layers, although this behavior is subject to change in a future update.
```
smaant: "However, not all of the jobs will have cached layers, although this behavior is subject to change in a future update." -> "However, not all of the jobs might have cached layers and not all of the jobs will have identical cache."

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -17,20 +17,19 @@ Consider reusing the unchanged layers to significantly reduce image build times.
...
 
-If you run many parallel jobs for the same project, and each job requests a reusable environment, all of them will be provided with a Remote Docker Environment. However, not all of the jobs will have cached layers, although this behavior is subject to change in a future update.
+**Note:** Previously the `docker_layer_caching` was called `reusable`. The `reusable` key is deprecated in favor of the `docker_layer_caching` key. 
 
-## **`exclusive`**
+In addition, the `exclusive` option is deprecated in favor of all VMs being treated as exclusive. This indicates that jobs are guaranteed to have an exclusive Remote Docker Environment that other jobs cannot access when using `docker_layer_caching` with `machine`.
```
smaant: " with `machine`" should not be here, just "other jobs cannot access when using `docker_layer_caching`."

smaant: Also, I would probably join this whole paragraph with a previous one inside one Note

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -17,20 +17,19 @@ Consider reusing the unchanged layers to significantly reduce image build times.
...
 
-The second option (`exclusive`) indicates whether you want to allow parallel jobs to run simultaneously in the same Remote Docker Environment. Jobs for which `exclusive:` is set to `true` are guaranteed to have an exclusive Remote Docker Environment that other jobs cannot access.
-
-If you set `exclusive:` to `false`, the project's parallel jobs will be able to share the same Remote Docker Environment. This method lets you reduce the chances of receiving a Remote Docker Environment without a cache. If you choose this option, ensure that your project can handle concurrent operations in Docker Engine.
+``` YAML
```
smaant: Here we need to add a section "Docker Layer Caching in `machine` executor". And say that DLC is available not only for Remote Docker, but also for `machine` executor, that it works exactly the same way and enables as in the example below.

smaant: Also at the beginning of this article, there's a sentence "Because Docker Layer Caching uses the `setup_remote_docker` step, it is not compatible with the `machine` executor." - we need to remote it.

smaant: Also, before a paragraph "Consider reusing the unchanged layers" we need to add a title "Docker Layer Caching in Remote Docker"

jekyll/_cci2/docker-layer-caching.md
michelle-luna: great, I made the first change and removed the second.

jekyll/_cci2/docker-layer-caching.md
michelle-luna: thanks, changed to "However, all of the jobs may not have cached layers and not all of the jobs will have identical cache."

jekyll/_cci2/docker-layer-caching.md
michelle-luna: fixed, thank you!

smaant:
LGTM overall, some concerns if "However, all of the jobs may not have cached layers and not all of the jobs will have identical cache" give a right idea

PR #1641, opened 2017-10-04
Update executor-types.md

iynere:
```
machine: true
  image: etc
```

isn't valid yaml! it's gotta be:

```
machine:
  enabled: true
  image: etc
```


jekyll/_cci2/executor-types.md
```diff
@@ -59,7 +59,8 @@ The following example specifies the default `machine` image.
 ```YAML
 jobs:
   build:
-    machine: true
+    machine:
+      enabled: true
```
ndintenfass: Pretty sure you don't need the "enabled" key here.

iynere: oh tru !! that makes sense, so it's more like the docker syntax

iynere:
incorporated nathan's feedback & cleaned up wording

eric-hu:
LGTM

PR #1659, opened 2017-10-11
Update sample-config.md

jenniesyip:
- Refactor using `{{ .Environment.CIRCLE_SHA1 }}` var vs writing to a file
- Clean up `restore_cache` key maps since the example is only checking for one `key`


marcomorain:
Awesome, thanks @jenniesyip

PR #1661, opened 2017-10-12
Clarify workflow-job isolation

eric-hu:
Source: https://discuss.circleci.com/t/does-circleci-2-0-work-with-monorepos/10378/8?u=eric

CC @notnoopci and @smaant -- I didn't see `circleci step halt` in our docs, so I just want to make sure we're okay with our users running this command.  If so, we should also document that command.

### DIFF
--- a/jekyll/_cci2/faq.md
+++ b/jekyll/_cci2/faq.md
@@ -160,8 +160,23 @@ Yes, we are currently working on that functionality.
 Splitting `config.yml` into multiple files is not yet supported.
  
 ### Can I build only the jobs that changed?
-No.
- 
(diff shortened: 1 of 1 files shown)


eric-hu:
@zzak responding to your Slack comment here:

> eric I think this patch is interesting, probably don't want to put such a hack in the FAQ but maybe link to it elsewhere? imo I think the answer is still "no" with an asterisk 

What part of this seems like a hack to you?

I'm now debating between closing this PR or modifying it with a link.  There was some discussion in Slack about how `circleci step halt` is inconsistent with the rest of the config, and was meant to be a stopgap:

https://circleci.slack.com/archives/C106KF6V6/p1508426982000562

smaant:
I was just about to refer to this exact conversation in slack, I think it's @notnoopci call to decide here

PR #1668, opened 2017-10-13
Adding in the new workflow env vars

edkellena:
Hey, I noticed some new env vars that weren't documented so thought I'd throw them in here.

Cheers.

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -268,3 +268,27 @@ The repository name of the project being tested, i.e. “bar” in circleci.com/
 **CIRCLE_INTERNAL_TASK_DATA**
 
 The directory where test timing data can be found.
+
+**CIRCLE_STAGE**
+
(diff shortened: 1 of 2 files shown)


marcomorain:
Hi @edkellena, thanks for the PR.

Thanks for the ECR docs, that's a great addition to our docs 👍 .

Some of these environment variables are internal IDs that we don't really want to document. The workspace and workflow IDs are exposed as env vars to the build since as a temporary measure. We plan to pass these values around behind the scenes in upcoming releases.

I think that `CIRCLE_STAGE` and `CIRCLE_WORKFLOW_ID` are the interesting var here.

Do you have a use for any of the others?

marcomorain:
Hi @edkellena 

I'm going to close this issue for now without merging it. We appreciate the time you've spent making the PR (and others). In this case I want to keep these env vars somewhat hidden while we are iterating on the mechanics of workflows.

Marc

PR #1672, opened 2017-10-15
Heroku deployment fixes

RobinDaugherty:
- Use DNS host key verification for Heroku deploy. This is the real answer, rather than using whatever host key is presented, Heroku has DNS keys containing the host key.
- Remove `sudo` from the commands that place the heroku CLI. It isn't needed.
- Instead of using a git remote, just use a `git push` command with the URL in it.
   - Use SSH to connect to the git repo. There was no reason to set up the SSH key since it wasn't being used with the `https` protocol.
   - Since SSH is used to connect, the credentials for `git.heroku.com` are not needed in `.netrc`.
   - Force-push to Heroku to ensure that the deployment will occur even if the commit is not a fast-forward of the branch currently deployed to Heroku.


marcomorain:
Thanks @RobinDaugherty

PR #1673, opened 2017-10-15
/circleci/config.yml → .circleci/config.yml

84d010m08:
I fixed that `.circleci/config.yml` seems to be correct.


felicianotech:
Thank you for this.

PR #1674, opened 2017-10-16
[CIRCLE-3287] Document downgrade to MySQL 5.6 on 1.0

zzak:


felicianotech:
Should we be continuing to add to the 1.0 docs like this rather than urge 2.0?

michelle-luna:
Thanks for doing this, 2.0 is priority, unless tickets are being filed on something.

zzak:
@felicianotech I hear you, and I'm not trying to encourage or discourage against 2.0, but this doc was pretty much already written just needed to :shipit: and my bad for taking so long :bow:

PR #1682, opened 2017-10-18
~/circleci/config.yml ➡️ ~/.circleci/config.yml

84d010m08:
I thought that the dot was missing, so I changed it to `~/.circleci/config.yml`.


felicianotech:
Thank you.

PR #1689, opened 2017-10-22
Fix yaml syntax error at sync git submodule

sawadashota:
Just remove `:` because of yaml syntax error.


marcomorain:
Thanks @sawadashota

PR #1695, opened 2017-10-24
Add info about remote image limits for server

michelle-luna:


ndintenfass:
lgtm -- the phrase "in your remote Docker environment" is slightly awkward, but I couldn't come up with something clearly better.

PR #1697, opened 2017-10-25
Escape {{ checksum }} blocks in markdown

eric-hu:
{{ ... }} blocks are swallowed by Jekyll's liquid parser.  Liquid
attempts to interpret what's inside these blocks and throws errors on
`jekyll serve` startup.  The generated html is missing what we intend as
`{{ checksum (something) }}`.

This should be addressed in future edits by wrapping such text with

{% raw %} ... {{ checksum ... }} {% endraw %}


marcomorain:
LGTM

PR #1705, opened 2017-10-26
documents Vault HA configuration

stongo:
create a proper header for replicated configuration
document using custom CAs


stongo:
Logically, I think all the preceding content fit in to a "Operating" mongodb and postgres category. The new section is more of an `application integration` category

michelle-luna:
thank you!

PR #1709, opened 2017-10-27
enquote xcode version number

li3n3:
Xcode version number needs to be quoted as a string, or this example will fail.


zzak:
@li3n3 that is pretty great catch, I wonder if it's getting cast as a float and then fails down the line? :thinking:

li3n3:
@zzak there is definitely some reason for it that @DanToml caught; we were attempting to use what's in docs and failed a build. Quickly resolved because of Dani's sharp attention!

PR #1719, opened 2017-10-31
Update php version to use non-deprecated version

dgallegos:
Updated php to 7.0.24 because 7.0.4 is now deprecated.


felicianotech:
This is an example and does not actually list supported PHP versions. The links to the Ubuntu 12.04 and 14.04 versions list does that.

PR #1725, opened 2017-11-01
Update install options

bellkev:


ndintenfass:
lgtm - a bit terse, perhaps, but less likely to confuse.

PR #1726, opened 2017-11-02
Use the proper vocabulary wrt docker images

dougalcorn:
In the first example using the `machine: true` the comments refer to a docker container when I think you mean docker image. You don't build containers or push them to docker hub. You build images and run containers.


michelle-luna:
thank you!

PR #1736, opened 2017-11-07
Install docker using the convenience script

axelssonHakan:


felicianotech:
Is there a way to manually set the version to be installed with this method?

smart-alek:
@felicianotech Doesn't [look like it](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#upgrade-docker-ce-1).

"The script does not provide options to specify which version of Docker to install, and will install the latest version that is released in the “edge” channel."

felicianotech:
Yup, it doesn't. There's probably a cleaner solution to installing than what we have, but I don't think this PR is it due to:

1. The inability to specify a specific version to install (we want deterministic builds).
1. The Docker Docs stating that the "convenience script" method is designed for edge/testing cases and not recommended for production.

PR #1737, opened 2017-11-07
Adding a link to the testing iOS doc

emmabot:
Seemed like we were missing a link there.


felicianotech:
Thanks @emmabot. This is great but we should use relative links instead of absolute links when the doc is internal. I'll switch it over for you. :smile:

PR #1743, opened 2017-11-09
Another pass on iOS docs

appplemac:
* Linking to our own code signing doc now that we have it.
* Removing a few things that are no longer relevant.
* Linking to our example project.

### DIFF
--- a/jekyll/_cci2/testing-ios.md
+++ b/jekyll/_cci2/testing-ios.md
@@ -501,17 +501,12 @@ projects. Here are the most frequent of those:
 cases CircleCI may not correctly detect the Xcode workspace, project, or
 scheme. Instead, you can specify these through environment variables.
 
-### Constraints on macOS-Based Builds
-There are a few features normally available on CircleCI's standard
-Linux containers that are not available for macOS builds at the moment:
+### Constraints on macOS-based Builds
 
-* Parallelism is not supported
-* While the general `config.yml` file structure will be honored in macOS-based builds
-[configuration options]( {{ site.baseurl }}/2.0/configuration-reference/), the following sections of
-`circle.yml` will not work correctly:
-  * `machine: services`
-  * `machine: <language>`, where `<language>` is any language mentioned
-    in the [Configuration doc]( {{ site.baseurl }}/2.0/configuration-reference/)
+These are the features normally available on CircleCI 2.0
+Linux environment that are not available for macOS builds at the moment:
+
+* Parallelism is not supported.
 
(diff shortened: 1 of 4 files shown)


jekyll/_cci2/testing-ios.md
```diff
@@ -520,17 +501,12 @@ iPhoneSimulator: Timed out waiting 120 seconds for simulator to boot, current st
...
+Linux environment that are not available for macOS builds at the moment:
 
-* Parallelism is not supported
-* While the general `config.yml` file structure will be honored in macOS-based builds
-[configuration options]( {{ site.baseurl }}/2.0/configuration-reference/), the following sections of
-`circle.yml` will not work correctly:
-  * `machine: services`
-  * `machine: <language>`, where `<language>` is any language mentioned
-    in the [Configuration doc]( {{ site.baseurl }}/2.0/configuration-reference/)
+* Parallelism is not supported.
```
DanToml: Parallelism technically works, but there's no support for it inside Xcode... Might be better phrasing this as a section on "spitting up your builds" and including strategies for breaking up your tests into separate targets & jobs?

jekyll/_cci2/testing-ios.md
```diff
@@ -520,17 +501,13 @@ iPhoneSimulator: Timed out waiting 120 seconds for simulator to boot, current st
...
+We suggest the following strategies:
 
-* Parallelism is not supported
-* While the general `config.yml` file structure will be honored in macOS-based builds
-[configuration options]( {{ site.baseurl }}/2.0/configuration-reference/), the following sections of
-`circle.yml` will not work correctly:
-  * `machine: services`
-  * `machine: <language>`, where `<language>` is any language mentioned
-    in the [Configuration doc]( {{ site.baseurl }}/2.0/configuration-reference/)
+* Using concurrent jobs for building with different Xcode versions
```
michelle-luna: The existing 2.0 documentation doesn't talk about concurrent jobs (maybe it should?). Is the suggestion here to configure a workflow with parallel jobs? If this is what you mean, it would be better to state it as parallel jobs and point to https://circleci.com/docs/2.0/workflows/#workflows-configuration-examples

KunalJain:
@appplemac Users have been asking for react native doc. Should we add an example or add a link to this repo? https://github.com/facebook/react-native/blob/master/.circleci/config.yml

appplemac:
@KunalJain We should _not_ link to the yaml of the react-native project, mostly because that’s not a react native app. I’ll make a react native example next week and will add it separately.

appplemac:
@michelle-luna Addressed your comment, would appreciate another look.

PR #1772, opened 2017-11-16
Update note re: sudo to Heroku deployment info

eric-hu:
A user running our convenience image said that running the script with `sudo` failed later on in the script with SSH-ing into Heroku.  Re-arranging this to support our convenience image users first


iynere:
lgtm !

PR #1784, opened 2017-11-22
WIP: Gpu vm service docs

GERey:


zzak:
LGTM but on `vm-service-custom-image` we probably want to turn the URLs into actual links and not code blocks

PR #1785, opened 2017-11-22
New ha config

stongo:
Server has been refactored to be future proof and easier to customize


MikeyYeahYeah:
On Ubuntu 16.04, there is a `/usr/local/share/ca-certificates` with an s.  The last line of the doc has `/usr/local/share/ca-certificate` which is missing the S.  Not sure if this is a typo or just inconsistency across different distributions.  If it is the latter, might be worth making a quick mention.

ndintenfass:
@MikeyYeahYeah perhaps you can combine this PR with https://circleci.atlassian.net/browse/CIRCLE-7748

PR #1814, opened 2017-12-07
Change docker_layer_caching default comment

fernfernfern:
This is misleading. Docker layer caching is not set to true by default. You need to explicitly add the configuration option and set it to true


michelle-luna:
thanks this was fixed in a previous commit, so closing.

Issue #1323, opened 2017-07-20
Permission denied error when attempting docker-compose installation 


MisterGlass:
This issue is about <https://circleci.com/docs/2.0/docker-compose/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/docker-compose.md>)

Following the example in this doc, I received an error `/bin/bash: line 1: /usr/local/bin/docker-compose: Permission denied`. After some googling, I found the following advice, which seems to install docker-compose properly:
https://github.com/docker/machine/issues/652#issuecomment-243215931

I would open a PR to update the doc but I am not sure this method is desirable.

levlaz:
I install latest docker-compose with pip and I have never had this issue. How are you installing it?

i.e. `pip install docker-compose`

MisterGlass:
@levlaz with CURL as described in the beginning of the documentation

levlaz:
Thanks. I'll figure out why we're recommending that method. :)

sysradium:
@levlaz I guess because of this:
```
Successfully built PyYAML docopt texttable dockerpty
Installing collected packages: PyYAML, colorama, docopt, texttable, requests, six, websocket-client, docker-pycreds, docker, jsonschema, dockerpty, cached-property, docker-compose
Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/local/lib/python3.6/site-packages/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/usr/local/lib/python3.6/site-packages/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/usr/local/lib/python3.6/site-packages/pip/req/req_install.py", line 851, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/usr/local/lib/python3.6/site-packages/pip/req/req_install.py", line 1064, in move_wheel_files
    isolated=self.isolated,
  File "/usr/local/lib/python3.6/site-packages/pip/wheel.py", line 345, in move_wheel_files
    clobber(source, lib_dir, True)
  File "/usr/local/lib/python3.6/site-packages/pip/wheel.py", line 323, in clobber
    shutil.copyfile(srcfile, destfile)
  File "/usr/local/lib/python3.6/shutil.py", line 121, in copyfile
    with open(dst, 'wb') as fdst:
[...]

PR #1827, opened 2017-12-12
Fix permission denied error of install docker-compose #1323

qq52184962:
As issue #1323 mentioned, using curl to download and directly pull files into `/usr/local/bin/docker-compose` will result in `permission denied error`.


OskarStark:
is this already released on the docs website?

PR #1835, opened 2017-12-13
Added instructions on test metadata with pytest

drazisil:


drazisil:
@michelle-luna I'm not sure why this is failing the proofer.

PR #1838, opened 2017-12-14
Update caching.md

iynere:
added example of source caching


eric-hu:
This example does do source caching, but the example should include a caveat that if you want to try source caching, it's best to implement it and actually compare checkout times with and without source cache.  Due to the way Github is optimized versus our general purpose cache, a clone might be faster.

iynere:
@eric-hu added a commit w/ yr suggestions, thanks !! ^_^

PR #1840, opened 2017-12-14

iynere:
add info re: `setup_remote_docker`


ryanwohara:
LGTM assuming the Jekyll markdown is good. It _looks_ right...

iynere:
added a new commit with your suggestions !

PR #1598, opened 2017-09-22
$CIRCLE_INTERNAL_TASK_DATA

ryanwohara:
The introduction of $CIRCLE_INTERNAL_TASK_DATA.

Test timing data is not stored in the same location between executor types.

PR #1841, opened 2017-12-14
Update parallelism-faster-jobs.md

iynere:
clarification about the necessity of using `store_test_results` to store timing data


bf4:
Are there any examples of anyone using timing data via CIRCLE_INTERNAL_TASK_DATA?  I've searched the discuss, the github, the web, and cannot find nary a one. Just a few mentions.  (I also commented in #1598)

PR #1849, opened 2017-12-15
Iynere test results workflows patch

iynere:
we were documenting the timing-based test-splitting + workflows issue slightly inaccurately—it's not that `store_test_results` doesn't work w/ workflows (it does—otherwise Test Summary wouldn't work at all); it's just that using `circleci tests split --split-by=timings` doesn't work w/ workflows. so really, the only place we need that note is on the Running Tests in Parallel page, imo at least...

### DIFF
--- a/jekyll/_cci2/collect-test-data.md
+++ b/jekyll/_cci2/collect-test-data.md
@@ -9,7 +9,7 @@ order: 34
 
 *[Test]({{ site.baseurl }}/2.0/test/) > Collecting Test Metadata*
 
-CircleCI collects test metadata from XML files and uses it to provide insights into your build. This document describes how to configure CircleCI to output test metadata as XML for some common test runners and store reports with the `store_test_results` step. **Note:** This step is not supported with Workflows. To see test result as artifacts, upload them using the `store_artifacts` step. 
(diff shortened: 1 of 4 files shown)


michelle-luna:
Thank you! I believe Kunal and I worked with Anton on these notes because of customer problems with it with multiple jobs in a workflow, so we must get their reviews before removing these notes. If it works, we should add information that specifically describes exactly how to use it and what the results will be from your testing of it (that would help Kunal and Anton think through this again if they can see what it was that you were able to do).

michelle-luna:
Here is another related PR where we added the FAQ in Sept. for background, if we removed or change these notes, we should change the FAQ as well. https://github.com/circleci/circleci-docs/pull/1564/files

iynere:
thanks michelle ! i will make the change in that other file as well & add it to this PR

iynere:
ok done ! ok to merge now ?

michelle-luna:
I want engineering and PM review of this before we remove the notes, thank you for your patience!

iynere:
no worries, i appreciate your thoroughness ! 👍

KunalJain:
LGTM. Thank you!

iynere:
at this point, this PR is going to conflict with at least one other open PR : https://github.com/circleci/circleci-docs/pull/1888

i'll update this one so it doesn't conflict with #1888—they both should be merged, i think

iynere:
ok i ended up having to modify both PRs lol, they both should be good to go now tho 👍

michelle-luna:
thank you!

PR #1872, opened 2017-12-22
Update CIRCLE_TOKEN env var reference

heug:
Changed for consistency, users are instructed elsewhere in docs to set the env var `CIRCLE_TOKEN` to be their token, we shouldn't have them change it to a query string with their `CIRCLE_TOKEN` as they would need to change it again to do normal API things.


levlaz:
LGTM

PR #1888, opened 2018-01-08
Update store_test_results in config reference

iynere:
it's 1) confusing to link to a part of our 1.0 documentation that mentions `$CIRCLE_TEST_REPORTS` when this env var doesn't even exist in 2.0, and 2) probably just not ideal in general to be linking to our 1.0 documentation to explain things about 2.0, right ?

i took the info [here](https://circleci.com/docs/1.0/test-metadata/#metadata-collection-in-custom-test-steps) & just restated it, more or less, in a way that makes sense for 2.0


michelle-luna:
woot!

Issue #1891, opened 2018-01-08
Explain Why dockerize Command is Required in Circle CI 2.0 Ruby Docs


sshaw:
The [Ruby docs](https://github.com/circleci/circleci-docs/blob/76b628e82b24614d0ee225b283665477b0e15ea7/jekyll/_cci2/language-ruby.md#sample-configuration) use several Docker containers but doesn't explain how they're connected. The `dockerize` command is used, which *could be* to be doing this, but there's no  explanation whatsoever.

smart-alek:
Hey @sshaw, thanks for your comment/question!

The use of `dockerize` is actually separate from the 2 Docker images used as dependencies (Ruby and Postgres). `dockerize` is a [utility tool](https://github.com/jwilder/dockerize) for simplifying the containerization of applications.

By using the `docker` executor type, as we are in this example, we're telling CircleCI to use the first image you specify as the **primary container**. So when we use `dockerize` later, we're just telling CircleCI to hold on starting that container for 1 minute so the database has time to fully set up.

I've added a sentence about `dockerize` to the doc for clarification, which you can check at #1894

sshaw:
I'd link to [the glossary definition of primary container](https://circleci.com/docs/2.0/glossary/#primary-container). 

Another thing that may be worth stating is  that CircleCI is implicitly connecting the containers. I.e., There's a no port mapping or any Docker Compose type directives in the config, so it's not clear what needs to be done if I wanted to connect a 3rd Docker container that listens on port N. This, coupled with `dockerize`, can be a bit confusing.

smart-alek:
Updated with a link to primary container's definition!

So does your issue have more to do with understanding the implicit connections among multiple Docker containers? Or is it about understanding why `dockerize` is used?

sshaw:
Both. I thought `dockerize` may be doing some implicit connecting but I see now it's now. 

Arguably this info can be covered elsewhere.

smart-alek:
Nice! And I do think that the information on implicitly connecting containers _is_ covered [elsewhere](https://circleci.com/docs/2.0/executor-types/#using-multiple-docker-images); it's just not called out in the Ruby language guide.

If you're happy with the explanation of `dockerize` I've added, I'll go ahead and consider this case closed! :)

sshaw:
🚢 it❗️

PR #1894, opened 2018-01-09
Add context around `dockerize` command

smart-alek:
Potential resolution for #1891.

### DIFF
--- a/jekyll/_cci2/language-ruby.md
+++ b/jekyll/_cci2/language-ruby.md
@@ -247,7 +247,7 @@ steps:
 ```
 {% endraw %}
 
-Now we can setup our test database we'll use during the build. We're using the `dockerize` [utility](https://github.com/jwilder/dockerize) to delay the [primary container]({{ site.baseurl }}/2.0/glossary/#primary-container)'s main process until the database service is available.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/language-ruby.md
```diff
@@ -247,7 +247,7 @@ steps:
 ```
 {% endraw %}
 
-Now we can setup our test database we'll use during the build.
+Now we can setup our test database we'll use during the build. We're using the `dockerize` [utility](https://github.com/jwilder/dockerize) to delay the [primary container]({{ site.baseurl }}/2.0/glossary/#primary-container)'s main process until the database service is available.
```
michelle-luna: to avoid using 'we' and for clarity change to -> The next section sets up the test database. It uses the `dockerize` [utility](https://github.com/jwilder/dockerize) to delay starting the main process of the  [primary container]({{ site.baseurl }}/2.0/glossary/#primary-container) until after the database service is available.

otherwise it looks good, thank you!

smart-alek: Ha, the slow conversion from first-person plural is agonizing but delightful!

zzak:
LGTM

PR #1902, opened 2018-01-10
remove superflous comma

mmell:


smart-alek:
I've never seen such a brazenly superfluous comma! BEGONE WITH THEE! 🔨

PR #1904, opened 2018-01-10
Adding the beginnings of the 2.0 troubleshooting page

GERey:


jekyll/_cci2/troubleshooting.md
```diff
@@ -0,0 +1,76 @@
+---
+layout: classic-docs
+title: "Troubleshooting CircleCI 2.0"
+category: [administration]
+order: 12
+description: "Troubleshooting steps to take in CircleCI 2.0"
+---
+
+### Debugging Queuing builds.
```
ndintenfass: Should this be an H2 instead of H3?

GERey:
It's worth noting there is some overlap in this document and https://circleci.com/docs/2.0/nomad/ . The main reason I created this document was to function more a pure troubleshooting  ( do this if you see this ) kind of page, and these are the steps in the order that I most often go through.

PR #1909, opened 2018-01-11
Draft hidden ldap page

michelle-luna:


jekyll/_cci2/authentication.md
```diff
@@ -0,0 +1,40 @@
...
+ categories: [administration]
+ hide: true
+ order: 9
+---
+
+This document describes how to enable, configure, and test CircleCI to authenticate users with OpenLDAP or Active Directory credentials.
+
+## Prerequisites
+
+- Install and configure an LDAP instance and populate the database with your users and credentials.
```
smaant: I would probably not recommend calling LDAP a database, it might be a bit confusing. Since setting up LDAP by itself is out of scope we can just say what prerequisite is having up and running LDAP server/AD

jekyll/_cci2/authentication.md
```diff
@@ -0,0 +1,40 @@
...
+- Install a new instance of CircleCI 2.0 with no existing users using the [Installing CircleCI 2.0 on Amazon Web Services with Terraform]({{ site.baseurl }}/2.0/aws/) document.
```
smaant: I would recommend highlighting the fact that LDAP should not be used for existing installations. Also, we need a disclaimer that this feature is experimental and in alpha stage.

jekyll/_cci2/authentication.md
```diff
@@ -0,0 +1,40 @@
...
+4. Select the encryption type (plain text is not recommended).
+5. Fill in the Search user field with the LDAP admin username using the format `cn=<admin>,dc=<example>,dc=<org>` replacing `admin`, `example`, and `org` with appropriate values for your datacenter.
+6. Fill in the Search password field with the LDAP admin password.
+7. Fill in the User search DN field with an approrpiate value using the format `ou=<users>` replacing `users` with the value used in your LDAP instance.
+8. Fill in the Username field with an approriate unique identifier used for your users, for example, `mail`.
+9. (Optional) Fill in the Test username and Test password fields with a test email and password for an LDAP user you want to test.
+10. Save the settings.
+
+A user who logs in will be redirected to the Accounts page of the CircleCI application with a Connect button that they must use to connect their GitHub account. After they click Connect, an LDAP section with their user information (for example, their email) on the page will appear and they will be directed to authenticate their GitHub account. After authenticating their GitHub account users are directed to the Builds page to use CircleCI.
```
smaant: We might want to mention somewhere (maybe in prerequisites) that GitHub still has to be configured and GitHub still will be used as a source of organizations and projects that user has access to.

smaant: Also, we need to mention that if LDAP is enabled it will be the only option to authenticate on CircleCI.

jekyll/_cci2/authentication.md
```diff
@@ -0,0 +1,40 @@
...
+
+**Caution:** After a user has authenticated with LDAP and connected their GitHub account, removing their LDAP will not prevent them from logging in by using cookies.
```
smaant: A couple of comments here:
1. `removing their LDAP` is not a very clear wording, I would recommend something like "disabling/removing user from LDAP/AD"
2. `will not prevent them from logging in` - this is not exactly correct, a user still will be able to access CircleCI as long as that user stays logged in (because of cookies). As soon as user logs out or cookie got expired he/she won't be able to login back. Also, make sense to mention that a better handling for this case will be soon.

jekyll/_cci2/authentication.md
```diff
@@ -0,0 +1,40 @@
+---
+ layout: classic-docs
+ title: "Authentication"
+ short-title: "Authentication"
+ description: "Configuring LDAP Authentication"
+ categories: [administration]
+ hide: true
```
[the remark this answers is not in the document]
michelle-luna: Ok, thanks.  For the demo tomorrow, I will keep it hidden, then next week I will add it to the navigation.

jekyll/_cci2/authentication.md
michelle-luna: thanks, I changed to "Install and configure your LDAP server and Active Directory."

jekyll/_cci2/authentication.md
michelle-luna: Thanks, I added: "**Note:** LDAP is not supported with existing installations, only clean installations may use LDAP. The LDAP authentication feature is currently only a preview."

jekyll/_cci2/authentication.md
michelle-luna: Ok, I changed the other note to this: "**Note:** After completing this configuration, all users must log in to CircleCI with their LDAP credentials. After logging in to CircleCI, each user will then click the Connect button on the Accounts page to connect and authenticate their GitHub account."

jekyll/_cci2/authentication.md
michelle-luna: thanks, here is the updates text: "**Caution:** After a user has authenticated with LDAP and connected their GitHub account, disabling or removing a user from LDAP/AD will not log them out. A user who is removed from LDAP/AD will be able to access CircleCI as long as that user stays logged in (because of cookies). As soon as the user logs out or the cookie expires, they will not be able to log back in. A better handling for this case will be implemented as an update to this feature preview."

smaant: Cool, we've made some more research on that and it would be good to do some clarification otherwise it may sound scarier than it really is. "A user who is removed from LDAP/AD will be able to access CircleCI as long as that user stays logged in (because of cookies). However, users ability to see projects or run builds will be defined by his/her permissions in GitHub, if they're synced with LDAP/AD then automatically user will lose authorization to do anything in CircleCI as well."

Issue #1862, opened 2017-12-20
Be more specific about how cache writes and fallbacks interact


jfirebaugh:
This issue is about <https://circleci.com/docs/2.0/caching/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/caching.md>)

I'd appreciate some clarification about how cache writes and prefix matching interact. Specifically, given the following:

```
      - restore_cache:
          keys:
            - gem-cache-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
            - gem-cache-{{ arch }}-{{ .Branch }}
            - gem-cache
            
      - run: bundle install --path vendor/bundle
      
      - save_cache:
          key: gem-cache-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
          paths:
            - vendor/bundle
```

If the restore step finds no matches in the first two keys, but several potential matches with the prefix `gem-cache`, which one is chosen? (I'm hoping it's most recently written.)

jfirebaugh:
I just noticed "most recent" is implied by a comment in one of the examples:

> ```# Find the most recent cache used from any branch```

If this is in fact guaranteed, could the prose spell it out? Thanks!

smart-alek:
Hi @jfirebaugh, thanks for the feedback! I've cleaned up some of the language around caching. Let me know if #1910 makes things clearer!

PR #1910, opened 2018-01-12
Clarify cache behavior

smart-alek:
This PR focuses on clarity around the behavior of cache behavior. Potential response to #1862

### DIFF
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -65,7 +65,7 @@ As in CircleCI 1.0, it is possible and oftentimes beneficial to cache your git r
 
 In this example, `restore_cache` looks for a cache hit from the current git revision, then for a hit from the current branch, and finally for any cache hit, regardless of branch or revision. When CircleCI encounters a list of `keys`, the cache will be restored from the first match. If there are multiple matches, the **most recent match** will be used.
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/caching.md
```diff
@@ -59,13 +60,16 @@ As in CircleCI 1.0, it is possible and oftentimes beneficial to cache your git r
           paths:
             - ".git"
 ```
+
 {% endraw %}
 
-In this example, `restore_cache` looks for a cache-hit from the current git revision, then for a hit from the current branch, then, most broadly, for any cache-hit at all, regardless of branch or revision. If your source code changes frequently, or changes significantly across branches, you may want to remove the third `restore_cache` key, or even both the second *and* third keys, resulting in a more granular source cache that only looks for matches within the current branch and git revision.
+In this example, `restore_cache` looks for a cache hit from the current git revision, then for a hit from the current branch, and finally for any cache hit, regardless of branch or revision. When CircleCI encounters a list of `keys`, the cache will be restored from the first match. If there are multiple matches, the **most recent match** will be used.
+
+If your source code changes frequently, we recommend using fewer, more specific keys. This produces a more granular source cache that will update more often as the current branch and git revision change.
```
michelle-luna: Just a style nit, which is to change 'we recommend' to 'consider'. Otherwise this looks great, thanks for improving this!

PR #1919, opened 2018-01-16
Update aws.md

iynere:
fixing some awkward wording


iynere:
the 'access' bit doesn't quite seem necessary, as the IAM permissions requirements are explicitly given just a few bullet points down

PR #1925, opened 2018-01-17
adding script tags to proper location

trevor-circleci:
see ticket: https://circleci.atlassian.net/browse/CIRCLE-8444

This PR addresses the problem by re-adding the js tags that where removed when the footer was changed. (I accidentally removed the tags - doing so because they were improperly placed at the bottom of the footer tag instead of in the site layout).


smart-alek:
@AndresCuervo can you take a look here too?

smart-alek:
Cursory check on my local machine was all good. Anchor links are now displaying on mouseovers.

AndresCuervo:
Yep, double checked on mine, LGTM @trevor-circleci !

PR #1929, opened 2018-01-18
Test splitting is now available in Workflows

ryanwohara:


michelle-luna:
Also, do you all think we should changelog this for existing users?

ryanwohara:
That sounds positive :+1:

PR #1935, opened 2018-01-20
Fix quotes in postgres config examples

JonAbrams:
The wrong type of quote characters exist in examples. Copy-pasting leads to much confusion and frustration!


levlaz:
LGTM. Thanks Jon.

PR #1953, opened 2018-01-24
adding jobs to nav

trevor-circleci:
see ticket: https://circleci.atlassian.net/browse/MKTG-729

Adding jobs to the company dropdown in the nav.


trevor-circleci:
@smart-alek : the jobs dropdown was in the main site's nav, but not on docs. If you look at the code, on this pr, it's an addition.

smart-alek:
Ah ok, so why do you have to update in two places for the main nav in docs if it's already in the footer? And why do docs not inherit the same main nav as the rest of the site?

trevor-circleci:
Because docs is hosted on a separate site from outer, and it turned out to be a non trivial thing to share a component across servers on a flatfile system. 

and the footer is a whole different element - so it needs to be updated separately.

smart-alek:
Ha, gross. Thanks for the details @trevor-circleci  :shipit:

PR #1959, opened 2018-01-28
Update how to save minitest-ci reports

techgurupezza:
Currently Minitest setup is invalid and does not work with 2.0. With the new updates this allows a user to auto upload the reports to circleci test summary without error as, by default the reports are generated and saved to a folder reports under test folder in the repo. These updates have been tested and are valid for 2.0 builds.


smart-alek:
Awesome @techgurupezza. Thanks for cleaning up all that whitespace, too! 🥇

PR #1961, opened 2018-01-29
Make searching for 'set path' easier.

felicianotech:


felicianotech:
True. If you think of a better place, perhaps doc, we can link back to here as well. Let me know. I don't mind doing the PRs if I know where (structurally) the changes need to go.

PR #1967, opened 2018-01-30
Android: Clarify "emulator not supported"

jeremy-w:
It was unclear at first reading what "not supported" meant:

- Explicitly blacklisted, so don't even try?
- Not supported out of the box, but feel free to yak-shave?
- Not technically possible?

Language from the 1.0 docs clarifies that the correct choice is
"not technically possible," due to virtualization issues.

The language added is drawn from:
https://github.com/circleci/circleci-docs/blob/9b8e6b6fd055ca78ec20b558589f7f0bd1dfdb1b/jekyll/_cci1/android.md#starting-the-android-emulator


michelle-luna:
Thank you! I made a minor editorial change to this clarification based on a conversation among our support team today on this topic to be consistent with what they tell customers.

PR #1971, opened 2018-01-30
add link to dockerfile wizard

michelle-luna:


jekyll/_cci2/custom-images.md
```diff
@@ -20,6 +20,10 @@ Watch the following video for a detailed tutorial of customizing Docker images.
 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/JYVLeguIbe0" frameborder="0" allowfullscreen></iframe>
 
+## How to Use the CircleCI Dockerfile Wizard
```
smart-alek: This seems a little misleading. Rather than showing *how* to use the CircleCI Dockerfile Wizard, we're just linking and telling users to follow the instructions there.

It feels like the wizard is another (recommended?) method of creating a custom image. So maybe we write something like: "There are two ways to create a custom image: with our Dockferfile Wizard or manually."

jekyll/_cci2/custom-images.md
```diff
@@ -20,6 +20,10 @@ Watch the following video for a detailed tutorial of customizing Docker images.
...
+
+Refer to the [`dockerfile-wizard` GitHub repository of CircleCI Public](https://github.com/circleci-public/dockerfile-wizard) for instructions to clone and use the wizard to create a Dockerfile to generate your custom image.
```
smart-alek: Even if we keep the above heading, I'd move all of this beneath the Custom Image Overview since the docker image coming out of the wizard is _still_ a custom image — just with more handholding and less manual input.

jekyll/_cci2/custom-images.md
michelle-luna: Right, I will qualify the heading like we do for the video heading

jekyll/_cci2/custom-images.md
michelle-luna: Hmm, yeah, I thought about that too

jekyll/_cci2/custom-images.md
```diff
@@ -14,19 +14,26 @@ This document describes how to create and use custom Docker images with CircleCI
...
-<iframe width="560" height="315" src="https://www.youtube.com/embed/JYVLeguIbe0" frameborder="0" allowfullscreen></iframe>
-
-## Custom Image Overview
 CircleCI 2.0 gives you access to the power and flexibility of Docker. One of the ways you can take advantage of this is to create custom Docker images for your jobs. Following are the benefits of creating a custom image:
 
 1. Faster job execution because you can preinstall all the tools you require, eliminating the need to install them on each job run
 2. A more concise and easier to maintain CircleCI `config.yml` file
 
-In this document we will give a walkthrough of how to create a custom image. In most cases you'll want to have a custom image for your [primary container]({{ site.baseurl }}/2.0/glossary/#primary-container) so that is the focus of this document. But, you can easily apply this knowledge to create images for supporting containers as well.
+### CircleCI Dockerfile Wizard
```
smart-alek: Should this heading be the same size as "Creating a Custom Image Manually"?

PR #1990, opened 2018-02-05
Update collect-test-data.md

iynere:
add info on how to have Jest properly allocate CPU resources to avoid degraded performance in CI environments

### DIFF
--- a/jekyll/_cci2/collect-test-data.md
+++ b/jekyll/_cci2/collect-test-data.md
@@ -347,6 +347,8 @@ Then form your command in your config to output using the reporter.
 
 A really good read on this can be found [here](https://www.viget.com/articles/using-junit-on-circleci-2-0-with-jest-and-eslint).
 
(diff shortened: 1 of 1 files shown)


felicianotech:
Wasn't the fix to switch from `--runInBand` to `--maxWorkers=2`? Or am I reading that Issue thread wrong?

iynere:
i believe you may be misreading it....

> there's a bug in Circle builds where running jest ***without*** `--runInBand` attempts to use the number of CPUs of the machine rather than the VM (which would make things worse). The `--runInBand` flag ***"fixes"*** this by forcing it to one worker

that said—

> you should be able to use `-w` for up to the number of CPUs given to the Circle VM

so, if a customer knows they have `N` CPUs, then they can use `--maxWorkers=N`

felicianotech:
Yeah that's what they said but then look at the PR they did to fix the issue: https://github.com/facebook/react/pull/11983/files

I think the problem was that `--runInBand` isn't seeing the right number of vCPUs which is why setting it explicitly was done.

iynere:
`--runInBand` doesn't look for number of vCPUs, it forces Jest to use a single worker

iynere:
the PR used `--maxWorkers=2` because in that particular case, they knew that their CircleCI build had 2 vCPUs (which is not always the case)

felicianotech:
I was under the impression that two vCPUs were the minimum offering by apparently (https://circleci.com/docs/2.0/configuration-reference/#resource_class) it's not.

Your method looks best then. :+1:

PR #1991, opened 2018-02-05
Alter CPU wording for resource classes.

felicianotech:
`CPU` -> `vCPUs` is more for accuracy, less important.

The `1.0` -> `1` change I think is useful because there's no such thing as a fraction of a CPU or vCPU.


felicianotech:
Not a problem. Want me to do that in this PR? Another one? It seems you requested changes but then approved so I'm not sure which you'd like me to do.

michelle-luna:
Merging to get these fixes across the line, we can add more consistency changes in the next round of edits. Thank you!

PR #1994, opened 2018-02-06
Add circleci prefix to images

tvuotila:
I tried to install PostgreSQL with `- image: postgres:10.1-postgis`. It failed. What worked for me was `- image: circleci/postgres:10.1-postgis`. I assume, the other images work the same way.


smart-alek:
@tvuotila Excellent, thanks for the catch!

PR #1997, opened 2018-02-06
Minor edits for Jest section

smart-alek:
I should have added these comments while #1990 was in review, but I merged before I made the edits. These comments just clean up this section and remove some links so people don't get scared off by the `--runInBand` issue.


iynere:
LGTM !

PR #1998, opened 2018-02-06
Update building-docker-images.md

iynere:
add specs for remote docker env


iynere:
gleaned from running `cat /proc/cpuinfo`, `cat /proc/meminfo`, `fdisk -l /dev/sda` while SSH'd into the remote docker env

iynere:
🎉

PR #2001, opened 2018-02-06
Update Jekyll and Ruby versions.

felicianotech:
Update Jekyll and Ruby to newer versions.


felicianotech:
Nothing specific that I know of. I'm sure there's several bug fixes and maybe a few Jekyll features in tow.

This PR is to:

1. Just stay up to date and not fall too far behind.
2. Useful in case we use a lighter weight Jekyll Docker image (which uses Ruby 2.4.3) which I'm testing locally right now.

smart-alek:
Cool, feel free to merge whenever you're ready.

Issue #1985, opened 2018-02-01
Missing artifacts explanation


hamczu:
This issue is about <https://circleci.com/docs/2.0/deployment-integrations/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/deployment-integrations.md>)

In the setup described in the docs "deploy" job doesn't have access to any files created in "build" job. There section about artifacts and passing them between the jobs is missing.

hamczu:
Oh, I see - I should be using "workspaces" but still the setup mentioned in the docs simply doesn't work.

smart-alek:
Hey @hamczu, glad you stumbled on the workspaces bit, but I'm going to reopen this as an issue because that example should **also** be using them.

Thanks for bringing this up!

PR #2002, opened 2018-02-07
Rename deploy file and remove video

smart-alek:
This is a preliminary cleanup PR for the resolution of #1985. Highlights include:

- convert underscores to dashes to fit docs naming conventions
- remove link to distracting video
- clean up **Overview** and remove from TOC
- small changes for consistency and reworking

Note that the general example in the **Overview** is still (a) too general and (b) not using workspaces. I'll be adding a follow-up PR to address both of those things.


jekyll/_cci2/deployment-integrations.md
```diff
@@ -6,20 +6,17 @@ categories: [deploying]
 order: 10
 ---
 
-<iframe width="560" height="315" src="https://www.youtube.com/embed/nQo1dngizvA" frameborder="0" allowfullscreen></iframe>
-
-This document describes using the `deploy` step with example instructions in the following sections:
+CircleCI can be configured to deploy to virtually any service. This document provides instructions and examples for the following platforms:
```
michelle-luna: Let's replace this with something.

jekyll/_cci2/deployment-integrations.md
```diff
@@ -6,20 +6,17 @@ categories: [deploying]
...
-
-It is possible to deploy to any service by adding commands to `.circleci/config.yml` and setting secrets on the Project Settings > Environment Variables page of the CircleCI application. Available deployment targets include Azure, Google (App Engine, Container Engine, and Cloud) and many others. 
+## Overview
+{:.no_toc}
 
-Add a job to your `config.yml` to set up conditional deployment for your application. The following example uses a [workflow job filter]({{ site.baseurl }}/2.0/workflows/#using-contexts-and-filtering-in-your-workflows) to check that the current branch is the `master` branch before running any deploy commands. Without this workflow configuration, `<my-deploy-commands>` would be executed every time this job is triggered.
+To deploy your application, add a [job]({{ site.baseurl }}/2.0/jobs-steps/#jobs-overview) to `.circleci/config.yml`. If your deploy job uses any output from previous jobs, you can share that data by [using workspaces]({{ site.baseurl }}/2.0/workflows/#using-workspaces-to-share-data-among-jobs). You will also have to set any necessary secrets on the **Project Settings > Environment Variables** page of the CircleCI application.
```
michelle-luna: I think it would help to have a similar example that uses workspaces after the first simple example in this doc, so I would move this sentence about workspaces after and then use a modified config example with comments to describe it from the workspaces section.

smart-alek: Agree! But this first simple example doesn't actually work in its current form. My next PR is going to add workspaces to this first example, so we have a general sample at the top before getting into any platform-specific instructions.

michelle-luna: Sounds good

PR #2011, opened 2018-02-08
Add (fixed) homepage welcome image.

felicianotech:
Fixed the homepage image dimensions issue. I don't have Safari but I checked on Google Chrome, Chromium, Firefox

@michelle-luna I changed the image slightly. The original image was 2500x720 at 679.6KB, which is slightly large (file size wise) for a web page image. The new image is 2000x576 at 164.8KB.

The file size change is separate from the dimensions fix so I can undo it if you deem necessary.


michelle-luna:
Closing this for now since brand has a concern about using this image, in any size, on the docs site.

PR #2015, opened 2018-02-09
Update elixir language guide

smart-alek:
Updated the Phoenix language guide to be simply an annotated configuration.


smart-alek:
Yeah @zzak, we decided that the explanation was easier to read when we didn't break up pieces of the YAML. There will still probably be a generic walkthrough elsewhere but, for now, all the language "guides" are going to head in this direction.

PR #2016, opened 2018-02-10
Propose fixing two typos

jeis2497052:
OK to remove trailing white space?


smart-alek:
@jeis2497052 Yes indeed! Usually I keep the newlines at the end of files, but all the trailing whitespaces you found were aching for removal. Thank you very much. :)

PR #2035, opened 2018-02-19
Docs: Fix Github query search link

dijonkitchen:


dijonkitchen:
No problem @michelle-luna . The original link didn't pre-populate the search field for me, but the fix does.

PR #2036, opened 2018-02-20
docs: minor typo

flyinprogrammer:
We meant to reference the build job, not the test job in this workflow.

we might consider changing:

```
only: /^config-test.*/
```

to

```
only: /^v.*/
```

To represent an actual workflow of building and testing on every push, but only releasing on tags that start with `v` - as in: `v0.0.1`.

### DIFF
--- a/jekyll/_cci2/workflows.md
+++ b/jekyll/_cci2/workflows.md
@@ -295,7 +295,7 @@ workflows:
               only: /.*/
       - deploy:
           requires:
-            - test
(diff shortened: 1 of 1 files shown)


smart-alek:
Hey @flyinprogrammer, thanks for the catch! And that's a great idea re: the more authentic workflow experience. I'm going to file a separate ticket for that, though. And approve this in the meantime. ;)

PR #2057, opened 2018-02-23
Fix blank white line below footer.

### LINKED ISSUE #1978: There's Empty Whitespace At The Bottom of Every Page
Under every page, after the CircleCI footer, there's a strip of whitespace. This looks to be unintentional.

felicianotech:
Fixes #1978.

### DIFF
--- a/jekyll/assets/css/main.scss
+++ b/jekyll/assets/css/main.scss
@@ -345,3 +345,8 @@ code {
 	display: grid;
 	grid-template-columns: 50% 50%;
 }
+
+/* A fix for https://github.com/circleci/circleci-docs/issues/1978 */
+body > img:last-of-type{
+	display: none;
+}


jekyll/assets/css/main.scss
```diff
@@ -345,3 +345,8 @@ code {
 	display: grid;
 	grid-template-columns: 50% 50%;
 }
+
+/* A fix for https://github.com/circleci/circleci-docs/issues/1978 */
+body > img:last-of-type{
+	display: none;
```
trevor-circleci: Ah, yes. This will work for now, and we can implement it. But it's considered a sloppy coding practice to target elements like this: with a generic identifier that could change over time. 

For instance, say we removed this tracking pixel in the future, then the next image down would be affected. Or, say we added a new tracking pixel, or the position of this one changed in the dom; then this image would no longer be hidden.

That being said, it works for now. Might as well merge it and watch it. But it's not an ideal solution.

@smart-alek

PR #2062, opened 2018-02-23
Move add ssh keys step up

### LINKED ISSUE #1848: Move add_ssh_keys ahead of call to setup_heroku.sh
This issue is about <https://circleci.com/docs/2.0/deployment_integrations/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/deployment_integrations.md>)

Following the instruction I kept running into an issue around `heroku.com` not being a known host.

```
The authenticity of host 'heroku.com ([IP address])' can't be established.
RSA key fingerprint is [key fingerprint].
Are you sure you want to continue connecting (yes/no)?
```

1. The documentation states to add the SSH key for the hostname `git.heroku.com`, however the host given in the message states `heroku.com`. Should this be updated?
2. Even after adding an SSH key for `heroku.com` it wouldn't work. Only once I had moved the `add_ssh_keys` to be run _before_ `bash .circleci/setup-heroku.sh` it would work.

smart-alek:
Fixes #1848

### DIFF
--- a/jekyll/_cci2/deployment-integrations.md
+++ b/jekyll/_cci2/deployment-integrations.md
@@ -161,6 +161,9 @@ jobs:
           fingerprints:
             - "48:a0:87:54:ca:75:32:12:c6:9e:a2:77:a4:7a:08:a4"
       - run:
+          name: Run Setup Script
+          command: bash .circleci/setup-heroku.sh
+      - run:
           name: Deploy Master to Heroku
           command: |
             git push --force git@heroku.com:$HEROKU_APP_NAME.git HEAD:refs/heads/master
--- a/jekyll/_cci2/deployment-integrations.md
+++ b/jekyll/_cci2/deployment-integrations.md
@@ -150,7 +150,7 @@ This file runs on CircleCI and configures everything Heroku needs to deploy the 
 ```yaml
 version: 2
 jobs:
-  # build and test jobs go here
+  # jobs for building/testing go here
   deploy-job:
     docker:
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/deployment-integrations.md
```diff
@@ -109,24 +109,26 @@ To deploy to Azure, use a similar job to the above example that uses an appropri
...
-       login $HEROKU_LOGIN
-       password $HEROKU_API_KEY
-     EOF
-
-     cat >> ~/.ssh/config << EOF
-     VerifyHostKeyDNS yes
-     StrictHostKeyChecking no
-     EOF
-     ```
+
```
michelle-luna: I think it is correct not to have the tabs in this yaml, but it does break the step numbering, so we should not use steps here.

jekyll/_cci2/deployment-integrations.md
```diff
@@ -143,49 +145,45 @@ This file runs on CircleCI and configures everything Heroku needs to deploy the
...
+          requires:
+            - build-job
+          filters:
+            branches:
+              only: master
+```
 
-Refer to the full example in the [2.0 Project Tutorial]( {{ site.baseurl }}/2.0/project-walkthrough/) for additional details.
+For additional details,
+refer to the full example in the [2.0 Project Tutorial]({{ site.baseurl }}/2.0/project-walkthrough/)
```
michelle-luna: needs a period

PR #2075, opened 2018-02-28
Remove known high container UUID error doc

### LINKED ISSUE #1639: Docs missing valid UID range
This issue is about <https://circleci.com/docs/2.0/high-uid-error/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/high-uid-error.md>)

This page does a good job of explaining why high user IDs are problematic. However, it simply says that you must "change the ownership of the file" - without indicating what a _valid uid / gid range actually is_. It'd be good to include that so image maintainers don't have to guess.

(as an aside, just running `cat /etc/passwd` may be enough in many cases to find the high ID).

### LINKED ISSUE #1786: Bad assumption
This issue is about <https://circleci.com/docs/2.0/high-uid-error/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/high-uid-error.md>)

I hit the error yesterday and whilst the page was very helpful in explaining the problem, it didn't get me to where I could fix it.

Neither the find command suggested, nor any other approach, could find me any file with a too-high UID.

In the end, I discovered that it was one of the interim steps in building the container that was creating the bad file, but that file was gone by the time the final image was generated.

To fix it, I had to modify the "RUN" step in the Dockerfile, adding "&& chown -R root:root /root" to the problem step. This meant that the problem file was created and fixed all in one step and thus didn't create a bad interim.

The assumption in the docs is that the bad file will still exist in the image when it's run, but this is clearly not necessarily the case. Updating to explain that a bad file can be created, removed, and still cause a problem; and how to debug & fix this, would be very helpful.

smart-alek:
Fixes #1639 and fixes #1786 

@keybits has offered to move known issues to the Knowledge Base, and the high UUID error is a great candidate for the switch. In the meantime, I'm removing this article to prevent further confusion.


smart-alek:
@michelle-luna Agree that we should improve the steps per the issues filed. I believe @keybits is working on that this week, although he might just be transferring the doc to the KB.

But I feel like there's a bigger question around _where_ to put writing about known issues. In the same way that hacks and workarounds live in Discuss, solutions to known issues are ways of dealing with limitations. The article on the UUID error is not a guide on how to use features; it's a recommended process for handling a known issue.

In my mind, the KB is a great place for collecting these in one place with recommended actions. Ideally, issues in the KB would have a limited lifespan — they would eventually be fixed.

In terms of removing it, I think that having an article that sets an expectation of "this will help me" and then doesn't is _worse_ than not having the article at all. Removing the doc doesn't fix the issues linked above — those absolutely still need to be addressed. But this is a case where Support needs to establish a recommended workaround and document that on Discuss or the KB.

michelle-luna:
No, this doc should not move to the KB, the KB should point people to the docs as I said in my first comment. Known issues documentation (conceptual information and steps) go in circleci-docs, the title of the problem with a summary and link to the doc goes in the KB.

smart-alek:
@michelle-luna So all of these should have only titles and links to circleci-docs? https://support.circleci.com/hc/en-us/sections/115003756627-Troubleshooting-and-Known-Issues

michelle-luna:
Titles with a summary and a link to the docs, yes! Thank you!

smart-alek:
All right, going to fix this up instead! 🙃

PR #2083, opened 2018-02-28
Updated deploy config for heroku

AnthonyNahas:
The deploy workflow to heroku is only described by supporting ssh.
What about https? The doc for deploying to heroku do not support https which is even easier 
and do not required so much effort to get the deploy mechanism work like adding a bash script... adding a ssh private key and more..
Instead of that we use one single step that include the heroku api key as well the heroku app name and voilà we get our work deployed to heroku !!

the main step is the following:
```
git push --force https://heroku:$HEROKU_API_KEY@git.heroku.com/$HEROKU_APP_NAME.git HEAD:refs/heads/master
```

the above configuration is tested and it works great!

### DIFF
--- a/jekyll/_cci2/deployment-integrations.md
+++ b/jekyll/_cci2/deployment-integrations.md
@@ -189,6 +189,63 @@ workflows:
               only: master
 ```
 
+If you prefer to deploy to Heroku using https, please consider this circleci (2.0) configuration which is even easier:
+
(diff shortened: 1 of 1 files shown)


AnthonyNahas:
one additional remark: the bash script to setup heroku didn't work properly...

michelle-luna:
Can you say more about the problem with the bash script? We should get it fixed since it is part of our main Linux tutorial in the circleci-public repo. Thanks for your time to help out with some more details!

PR #2096, opened 2018-03-02
Fix link in elixir docs

lasseebert:
The link to the example repo is broken. I think this is the right one.


smart-alek:
Thanks for spotting that @lasseebert!

PR #2098, opened 2018-03-02
Update status badges source

### LINKED ISSUE #2097: Status Badges links incorrect
This issue is about <https://circleci.com/docs/2.0/status-badges/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/status-badges.md>)

Hello! It looks like the URLs to the status badges are outdated -- the image URLs are 404.

smart-alek:
Fixes #2097 

Rereading this makes me wonder why we don't just default to the shield style for status badges. Those seem to be the industry standard, so it's odd to force users to do extra work to get that.

### DIFF
--- a/jekyll/_cci2/status-badges.md
+++ b/jekyll/_cci2/status-badges.md
@@ -5,7 +5,7 @@ title: Embedding Build Status Badges
 
 [Basics]({{ site.baseurl }}/2.0/basics/) > Embedding Build Status Badges
 
-This document describes how to create a badge that displays your project's build status (passed or failed) in a README or other document.
+This document describes how to create a badge that displays your project's build status.
 You can generate code for the following formats: Image URL, Markdown, Textile, Rdoc, AsciiDoc, reStructuredText, and pod.
 
 ## Overview
--- a/jekyll/_cci2/status-badges.md
+++ b/jekyll/_cci2/status-badges.md
@@ -36,7 +36,7 @@ and create  a label for the token.
 If you created a token in the previous step,
 select the token you want to use in the _API Token_ dropdown menu.
 5. Select the appropriate language from the _Embed Code_ dropdown menu.
-6. Copy and paste the generated link to the document where you want to display the status badge.
+6. Copy and paste the generated link in the document where you want to display the status badge.
 
 ## Customization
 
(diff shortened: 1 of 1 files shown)


smart-alek:
Added an idea to change the default status badge style: https://circleci-internal.ideas.aha.io/ideas/CCI-I-265

jekyll/_cci2/status-badges.md
```diff
@@ -5,28 +5,43 @@ title: Embedding Build Status Badges
...
+click on the _API Permissions_ in the _Permissions_ section.
+Click _Create Token_,
+choose the _Status_ scope,
+and create  a label for the token.
+4. (Optional)
+If you created a token in the previous step,
+select the token you want to use in the _API Token_ dropdown menu.
 5. Select the appropriate language from the _Embed Code_ dropdown menu.
-6. Copy and paste the generated link to the web page or document where you want to display the status badge.
+6. Copy and paste the generated link to the document where you want to display the status badge.
```
michelle-luna: to the document->in the document

jekyll/_cci2/status-badges.md
```diff
@@ -5,28 +5,43 @@ title: Embedding Build Status Badges
 
 [Basics]({{ site.baseurl }}/2.0/basics/) > Embedding Build Status Badges
 
-This document describes how to embed a badge that shows the build status (passed or failed) for your project in your README or in another document. You can generate code for the following formats: Image URL, Markdown, Textile, Rdoc, AsciiDoc, reStructuredText, and pod. 
+This document describes how to create a badge that displays your project's build status.
```
michelle-luna: Again, I don't understand the rewrite of this intro to include less information. Leave as-is please.

smart-alek: The information is included in the _Overview_ section below. In my mind, the introduction should be as brief as possible, letting the _Overview_ do the bulk of the explaining.

Maybe I'm confused about the purposes of the introduction and the overview. Could you clarify what they should be for? And should every document have both?

michelle-luna: I see, I think the original intro sentence is better than the one in the Overview with the use of the word 'virtually' which could be confusing and because it doesn't say what status values are possible (it just says 'build status' which might not be clear to some users). Yes, each article needs both and they should ideally have different information in them, the first sentence tells you why you would read the document (this document describes xyz), the second is a conceptual overview of the topic.

PR #2122, opened 2018-03-07
Create new troubleshooting section

smart-alek:
This is a preliminary PR before the article outlining the known issue with Container UUIDs is updated. The slug for the server troubleshooting has been updated, and the new troubleshooting page is an index page for known issues.


smart-alek:
@michelle-luna vetoed this Troubleshooting section, so the Container UUID Error will exist in Reference for the time being.

PR #2128, opened 2018-03-08
Revise documentation of bad UID remapping error

### LINKED ISSUE #1639: Docs missing valid UID range

### LINKED ISSUE #1786: Bad assumption

smart-alek:
Full revision of the infamous high UUID error. Fixes #1786 and fixes #1639

### DIFF
--- a/jekyll/_cci2/high-uid-error.md
+++ b/jekyll/_cci2/high-uid-error.md
@@ -8,28 +8,23 @@ order: 21
 ---
 *[Reference]({{ site.baseurl }}/2.0/reference/) > Debugging Container ID Cannot Be Mapped to Host ID Error*
 
-When user namespace remapping fails,
-CircleCI will fail to start the affected container.
-This document explains the problem
-and shows you how to fix it.
+When starting a container,
+you may see this error message:
+
+```
+failed to register layer: Error processing tar file (exit status 1): container id 1000000 cannot be mapped to a host id
+```
+
+This document explains the problem and shows you how to fix it.
+
+## Overview
+
+You will see this message when CircleCI is unable to start your container
+
+The user namespace (userns), is a feature of the Linux kernel that adds
 
 The error `container id 1000000 cannot be mapped to a host id` indicates that CircleCI could not start your
 container because the userns remapping failed.
-
-This document explains what this error means and how to fix it.
-
-```
-75c416ea735c: Pull complete
-c6ff40b6d658: Pull complete
-a7050fc1f338: Pull complete
-f0ffb5cf6ba9: Pull complete
-be232718519c: Pull complete
-9dea4940377d: Pull complete
-7aa04d3ad875: Pull complete
-24bb62285b16: Pull complete
-56bec22e3559: Pull complete
-failed to register layer: Error processing tar file(exit status 1): container id 1000000 cannot be mapped to a host id
-```
 
 ### What is userns?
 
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/high-uid-error.md
```diff
@@ -8,54 +8,70 @@ order: 21
...
+In the example presented at the top of this document,
+this value is `1000000`.
+Next, start the container and look for the files which receive that invalid value.
+
+One way to do this is to use the `find` command,
+as shown below:
+
+```bash
 # Start a shell inside the container
 $ docker run -it circleci/doc-highid sh
```
kimh: I think `circleci/doc-highid` doesn't make sense if you delete a link to `https://hub.docker.com/r/circleci/doc-highid`.

jekyll/_cci2/high-uid-error.md
```diff
@@ -68,4 +84,14 @@ $ ls -ln file-with-high-id
...
+After locating the offending file,
+change its ownership
+and re-create the image.
+
+**Note:**
+It is possible for an invalid file to be generated and removed
+while a container is building.
+In this case,
+you may have to modify the `RUN` step in the Dockerfile itself.
+Adding `&& chown -R root:root /root` to the problem step
```
kimh: Is it reasonable to assume that users can identify which step to add this command?

jekyll/_cci2/high-uid-error.md
kimh: Maybe we should address https://github.com/circleci/circleci-docs/issues/1639 as well in this PR?

smart-alek: Whoops — yeah, I'll re-add that link.

jekyll/_cci2/high-uid-error.md
smart-alek: This gets into some philosophical territory for me, actually: how deep do we want to go to explain things? Is it _our fault_ that this bug even exists? If it is, then I'd say we need to provide a complete workaround.

PR #2141, opened 2018-03-13
Update tutorials.md

katiecircle:
changing the Android image link back to what it was because it was broken


michelle-luna:
I'm going to close this PR because I reverted the previous one. We should do some testing to see why the build doesn't fail for these types of 404 links and if there is another, better way to do this 'open a new window' behavior.

PR #2142, opened 2018-03-13
Google Container Engine clarifications

### LINKED ISSUE #1670: Setting Up a Job Step to Decode Credentials is confusing
This issue is about <https://circleci.com/docs/2.0/google-container-engine/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/google-container-engine.md>)

The section "Selecting a Base Image" references using `google/cloud-sdk` and the "Setting Up a Job Step to Decode Credentials" section references a custom docker container hosted out of the GCP container registry with auth username/password specified.

I found this quite confusing as in the example I changed `- image: gcr.io/project/image-name` to be `- image: google/cloud-sdk` which fails on build with the error `Error response from daemon: Get https://registry-1.docker.io/v2/google/cloud-sdk/manifests/latest: unauthorized: incorrect username or password`.

After some digging I figured out the issue was that specifying auth fails since `google/cloud-sdk` is a public docker image, but I think this should be drawn out better in the docs.

### LINKED ISSUE #2134: GAE setup instructions are confusing
- "Paste this into a new environment variable’s “Value” field" <- not clear whether "this" is the contents of the JSON key or something else
- not clear which parts of the config.yml are necessary
- I think GCR_CREDS and GOOGLE_AUTH are both used to refer to the same environment variable (GOOGLE_AUTH in most places, GCR_CREDS in the comment in the sample config.yml)
- this document doesn't specify which roles the Service Account needs to have to work properly with CircleCI

This issue is about <https://circleci.com/docs/2.0/google-container-engine/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/google-container-engine.md>)

smart-alek:
Fixes #2134 and fixes #1670.

While working on this document, I realized a couple of things:

1. There is a lot of overlapping information between [this document](https://circleci.com/docs/2.0/google-container-engine/) and the [Authenticating Google Cloud](https://circleci.com/docs/2.0/google-auth/) document.

2. Both documents are mostly focused on steps preceding the *actual deploy* (authentication, environment variables, etc.).

3. The actual deploy instructions seem to be on the [deploy landing page](https://circleci.com/docs/2.0/deployment-integrations/#google-cloud), but these seem to be a bit lean.

---

At the very least, we should consolidate [Authenticating Google Cloud Platform](https://circleci.com/docs/2.0/google-auth/) and [Deploying to Google Container Engine](https://circleci.com/docs/2.0/google-container-engine/).

If we're feeling ambitious, I would pull all relevant Google Cloud setup documentation into a single document that is linked to from the Deploy landing page. The other deploy platforms could benefit from their own spaces as well.

### DIFF
--- a/jekyll/_cci2/google-container-engine.md
+++ b/jekyll/_cci2/google-container-engine.md
@@ -28,14 +28,14 @@ consider using Google's base image.
 You can find this image on DockerHub as [`google/cloud-sdk`](https://hub.docker.com/r/google/cloud-sdk/).
 Otherwise, follow the [SDK installation instructions](https://cloud.google.com/sdk/) for your base image's operating system.
 
-### Authenticate Google Cloud Platform
+### Authenticate `gcloud`
 
 Before you can use the `gcloud` command line tool with CircleCI,
 you must authenticate it.
 To do this,
 follow the instructions in the [Authenticating Google Cloud Platform]({{ site.baseurl }}/2.0/google-auth/) document.
 After completing these steps,
-you should have an environment variable called `GCLOUD_SERVICE_KEY`.
+you should have created an environment variable called `GCLOUD_SERVICE_KEY`.
 Using this particular name is not required,
 but it will be used throughout the examples in this document.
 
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/google-container-engine.md
```diff
@@ -1,75 +1,110 @@
...
+```bash
+gcloud --quiet components update
+gcloud auth activate-service-account --key-file=${HOME}/gcloud-service-key.json
 gcloud --quiet config set project ${GOOGLE_PROJECT_ID}
 gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
 gcloud --quiet container clusters get-credentials ${GOOGLE_CLUSTER_NAME}
 ```
 
-
+Refer to the [Google Cloud](https://circleci.com/docs/2.0/deployment-integrations/#google-cloud) deploy example for further steps.
```
GERey: Feels weird to link people to the rest of deploy section, when the page itself is called "deploying to", but its what the deployment-integrations page is for so w/e

GERey:
LGTM.

PR #2147, opened 2018-03-14
Env vars update

michelle-luna:
This is a clarification of order precedence and addition of couple of other env vars that are loaded but which were not documented. I changed the list around for clarity, then re-ordered the sections to match what I believe to be the correct ordering (but, please check this for accuracy).

### DIFF
--- a/jekyll/_cci2/contexts.md
+++ b/jekyll/_cci2/contexts.md
@@ -54,11 +54,16 @@ To rerun a job and use the context, it **must** be rerun from the Workflows page
 
 ## Environment Variable Usage 
 
-Environment variables are considered in a specific order, as follows:
-1. Project-level environment variables set on the Project Settings page.
-2. Context environment variables (assuming the user has access to the Context).
-3. Environment variables set with the `environment` key of a `run` step.
-4. Environment variables declared inside a shell command in a `run` step, for example `FOO=bar make install`.
+Environment variables are used according to a specific precedence order, as follows:
+
+1. Environment variables declared inside a shell command in a `run` step, for example `FOO=bar make install`.
+2. Environment variables set with the `environment` key for job.
+3. Environment variables set with the `environment` key for a container.
+4. Context environment variables (assuming the user has access to the Context). See the [Contexts]( {{ site.baseurl }}/2.0/contexts/) documentation for instructions.
+5. Project-level environment variables set on the Project Settings page.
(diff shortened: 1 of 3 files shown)


jekyll/_cci2/contexts.md
```diff
@@ -54,11 +54,16 @@ To rerun a job and use the context, it **must** be rerun from the Workflows page
...
-3. Environment variables set with the `environment` key of a `run` step.
-4. Environment variables declared inside a shell command in a `run` step, for example `FOO=bar make install`.
+Environment variables are used according to a specific precedence order, as follows:
+
+1. Environment variables declared inside a shell command in a `run` step, for example `FOO=bar make install`.
+2. Environment variables set with the `environment` key for job.
+3. Environment variables set with the `environment` key for a container.
+4. Context environment variables (assuming the user has access to the Context). See the [Contexts]( {{ site.baseurl }}/2.0/contexts/) documentation for instructions.
+5. Project-level environment variables set on the Project Settings page.
+6. CircleCI environment variables.
```
CharlesBlack89: Aren't all of these CircleCI environmental variables? Please disambiguate :)

marcomorain: We are missing one item in this list. You can set an `environment` on a `run` step: https://circleci.com/docs/2.0/configuration-reference/#run

I beleive this should be number 2 in the list, but I'm not certain.

michelle-luna: thanks! I've clarified that last line for Chris and added a new item 2 per Marc, will wait for review from @AlexanderMann or @notnoopci to confirm the ordering of the list

jekyll/_cci2/contexts.md
```diff
@@ -54,11 +54,16 @@ To rerun a job and use the context, it **must** be rerun from the Workflows page
...
+
+Environment variables declared inside a shell command `run step` will override environment variables declared with the `environment` and `contexts` keys. Global environment variables added on the Contexts page will take precedence over variables added on the Project Settings.
```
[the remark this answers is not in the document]
michelle-luna: thanks, fixed.

michelle-luna:
I'm going to merge this without review from Alex/Mahmood because it is a big improvement over the very confusing and incomplete list we had before and hopefully they will have time to come back and take a look soon.

PR #2111, opened 2018-03-06
Add guide on building snap packages with CircleCI.

felicianotech:
Guide on building Snap packages on CircleCI 2.0.

@smart-alek To be clear, unlike that other PR, this one is a brand new Doc so please feel free to scrutinize the language.

Also, I don't know what "category" this belongs in.

PR #2150, opened 2018-03-14
Update CONTRIBUTING.md with rules about links

smart-alek:
This pull request formalizes several style preferences that surfaced in #2111. At this point, the style guide is starting to outgrow its original home and should be moved into either a separate `STYLE.md` or the Wiki.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,8 @@ For more tips, see GitHub's blog entry on [how to write the perfect pull request
 
 # Style Guide
 
-We also have a short list of guidelines that take precedence over the general guide above. We don't have a lot of rules, but we do try to be consistent with the ones we have.
+We do not have a lot of rules,
+but we do consistently apply them.
 
 ## Markdown
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,8 @@ but we do consistently apply them.
 
 ## Markdown
 
-To keep the emphasis on prose, use [markdown](http://commonmark.org/help/) over plain HTML.
+To keep the emphasis on prose,
+use [markdown](http://commonmark.org/help/), not plain HTML.
 
 ## Bold and Terms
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -4,63 +4,110 @@ Thank you for your interest in contributing to the CircleCI documentation.
...
+Every pull request should have a description
+that explains *why* the change is being made.
+The description adds context
+that is critical for reviewers when giving feedback.
+
 For more tips, see GitHub's blog entry on [how to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request).
 
 # Style Guide
 
-For basic technical writing style, see [Technical writing style](https://en.wikiversity.org/wiki/Technical_writing_style).
```
michelle-luna: let's keep this because it is more general and wide-reaching for folks who are new to technical writing basics (we will not cover all of these style points in our guide)

CONTRIBUTING.md
```diff
@@ -4,63 +4,110 @@ Thank you for your interest in contributing to the CircleCI documentation.
...
+## Links
+
+### Explicit Names
+
+Link text should explicitly name the document or section
+to which it is linking.
+
+For example,
+if you reference a section called "Adding Environment Variables in the App",
+write, "Refer to the 'Adding Environment Variables in the App' section of the Environment Variables document."
```
michelle-luna: might as well show how to make the markdown relative link here.

smart-alek: These are GitHub-specific, right? Before we show _how_, I'd rather know _when_ and _why_ we'd use relative links at all.

michelle-luna: We always use them for internal links in case we decide to change where the `/docs1 directory lives at some point (and probably for other reasons) and we do it like this in Markdown:
`[Adding Environment Variables]({{ site.baseurl }}/2.0/env-vars/#adding-environment-variables)`

smart-alek: Sorry, I'm an idiot. I was thinking of something completely different when you said relative link. Fully agree with this.

PR #2205, opened 2018-04-02
Change template text

eric-hu:
DockerHub limits the tags it shows to the 100 most recent.  For example:

https://hub.docker.com/r/circleci/ruby/tags/

There is no pagination on that page.  The previous template suggesting that link for "all available tags" was incorrect.


felicianotech:
Fair point.

PR #2213, opened 2018-04-04
Revise interpolation workaround

### LINKED ISSUE #1650: $BASH_ENV does not seem to be sourced
This issue is about <https://circleci.com/docs/2.0/env-vars/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/env-vars.md>)

According to the documentation  i should echo commands into $BASH_ENV in order to interpolate variables so i did: 
` - run: echo 'echo Ad nocendum potentes sumus ;export ZAPP_NAME=$CIRCLE_PROJECT_REPONAME HELM_NAME=$(echo ${ZAPP_NAME}|tr "_" "-")' >> $BASH_ENV `
Which should set some env-vars and also echo a threat in Latin every-time an line is executed this sadly does not hapen.

smart-alek:
Fixes #1650

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -86,18 +86,16 @@ CircleCI does not support interpolation
 when defining configuration variables like `working_directory` or `images`.
 All defined values are treated literally.
 
-If you have bash installed,
-you can export the required variable within a command.
+However, it is possible to interpolate a variable within a command
+by setting it for the current shell.
 
 ```yaml
 version: 2
 jobs:
   build:
     steps:
-      - run: echo 'export BAR=foo' >> $BASH_ENV
+      - run: echo 'export VERY_IMPORTANT=$(cat important_value)' >> $BASH_ENV
 ```
-
-Otherwise, interpolated environment variables inside shell commands resolve normally.
 
 ### Declaring Environment Variables for a Job
 
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -92,7 +92,7 @@ you can export the required variable within a command.
 - run:
     name: Export Env Var
     command: |
-      export BAR=foo >> $BASH_ENV
+      echo 'export BAR=foo' >> $BASH_ENV
 ```
 
 ### Declaring Environment Variables for a Job
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -97,6 +97,12 @@ jobs:
       - run: echo 'export VERY_IMPORTANT=$(cat important_value)' >> $BASH_ENV
 ```
 
+**Note**:
+Depending on your shell,
+the actual command may differ from the above example.
+For more information,
+refer to your shell's documentation.
+
 ### Declaring Environment Variables for a Job
 
 To define environment variables for a job, use the `environment` key under the job name in the `jobs` section.
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -82,7 +82,7 @@ jobs:
 
 ### Interpolating Environment Variables 
 
-CircleCI does not support interpolation of environment variables.
+CircleCI does not support interpolation in configuration variables like `working_directory` or `images`.
 All defined values are treated literally.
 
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/env-vars.md
```diff
@@ -85,13 +85,14 @@ jobs:
 CircleCI does not support interpolation of environment variables.
 All defined values are treated literally.
```
ndintenfass: This may be confusing - inside of shell commands env vars do resolve. We don't support interpolation in the configuration variables (eg: working_directory, images, etc.)

jekyll/_cci2/env-vars.md
```diff
@@ -85,13 +85,14 @@ jobs:
...
+If you have bash installed,
+you can export the required variable within a command.
 
 ```yaml
 - run:
+    name: Export Env Var
     command: |
-      export PATH=/go/bin:$PATH
-      go-get ...
+      echo 'export BAR=foo' >> $BASH_ENV
```
ndintenfass: Did the old line not work? Not sure what the point of using the "|" is for a single-line example -- showing adding something to the $PATH seemed more real-world.

jekyll/_cci2/migrating-from-1-2.md
```diff
@@ -134,16 +134,18 @@ jobs:
             DATABASE_URL: postgres://conductor:@localhost:5432/conductor_test
 ```
 
-**Note:** CircleCI 2.0 does not support interpolation of environment variables.
+**Note:**
+CircleCI does not support interpolation of environment variables.
 All defined values are treated literally.
 
-One workaround is to export the required variable within a command.
+If you have bash installed,
```
ndintenfass: Why be so specific to bash in explaining this?

jekyll/_cci2/env-vars.md
smart-alek: Great point. I'll get that clarified.

jekyll/_cci2/migrating-from-1-2.md
smart-alek: This isn't a workaround for people without bash (i.e. alpine-based images). I wanted to head folks off at the pass before becoming invested in a potential workaround.

jekyll/_cci2/env-vars.md
smart-alek: I only updated this because it was the most recent accepted example (based on a thread with Eric in #dx-team). I can see the real-world potential of the earlier example, and the multiline pipe is an oversight on my part. 😬

jekyll/_cci2/migrating-from-1-2.md
gordonsyme: The same approach works for other shells too, but the specifics differ, e.g. `tcsh` uses `setenv FOO=bar` and from what little knowledge I have, doesn't offer anything similar to `BASH_ENV`, but according to the manpage, `tcsh` will _always_ read `~/.tcshrc` so that's basically the same thing.

Armed with knowledge of their own shell and the specifics of our environment, and our config I suspect most people would work this out themselves. Would we be better off making sure that those aspects are very clearly explained?

jekyll/_cci2/env-vars.md
```diff
@@ -82,18 +82,29 @@ jobs:
...
-- run:
-    command: |
-      export PATH=/go/bin:$PATH
-      go-get ...
+version: 2
+jobs:
+  build:
+    steps:
+      - run: echo 'export PATH=/path/to/foo/bin:$PATH' >> $BASH_ENV  # updating PATH
+      - run: echo 'export VERY_IMPORTANT=$(cat important_value)' >> $BASH_ENV  # runtime definition of env var
```
michelle-luna: looking at the react-native config.yml they also add 

 -run: source $BASH_ENV

smart-alek: Hmm, that seems odd to me. I would understand if it were sourced in the same `run` step as the new environment variable. But my understanding is that `$BASH_ENV` is automatically loaded anyway in each `run` step.

Is [this](https://github.com/CircleCI-Public/circleci-demo-react-native/blob/master/.circleci/config.yml) the react-native `config.yml` you meant? I can't find the line you mentioned. :(

PR #2219, opened 2018-04-09
Update xcode version

wtrocki:
Update xcode version for circle CI to let users to use the recent version.

### DIFF
--- a/jekyll/_cci2/ios-tutorial.md
+++ b/jekyll/_cci2/ios-tutorial.md
@@ -38,7 +38,7 @@ For iOS projects, it is possible to run your tests with Fastlane Scan as follows
 jobs:
   build-and-test:
     macos:
-      xcode: "9.0"
+      xcode: "9.2.0"
     steps:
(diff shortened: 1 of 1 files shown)


wtrocki:
Actually 2.0 documentation is missing list of the images here: 
https://circleci.com/docs/1.0/build-image-macos/

Additionally it seems that 9.3 image is available (but I'm not sure if that is officially available image)

michelle-luna:
9.3 is official, see https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions (and know that you are not the only person who has had trouble finding this section)! It looks like we neglected to add the 9.3 support to the changelog, so I will get that done this morning. thanks!

wtrocki:
@michelle-luna Thank you for such quick response! This particular tutorial is really high in google and easy to find in the docs so if we update that to use recent version then it will be easier for people to pick this changes without checking list of available images. It will be nice to link this list from this document as well.

PR #2220, opened 2018-04-09
Document CIRCLE_WORKFLOW_ID in env-vars.md

moredip:
I wanted a way to consistently name artifacts created by one job that could be referenced by another job (i.e. for tagging Docker images). It took me a long time to realize that the undocumented `CIRCLE_WORKFLOW_ID` env var was available to me. 

I don't know if `CIRCLE_WORKFLOW_ID` is an "official" env var, and whether the description is accurate or not.

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -351,3 +351,7 @@ The directory where test timing data can be found.
 `CIRCLE_STAGE`
 
 The job being executed. The default 2.0 job is `build` without using Workflows.
+
+`CIRCLE_WORKFLOW_ID`
+
+A unique identifier for the current workflow instance that the build is running in. This identifier will be the same for every job that runs within this instance of the workflow.


michelle-luna:
Thank you! I've added Gordon as a reviewer to get the answer on whether this is official and to look at the description.

jekyll/_cci2/env-vars.md
```diff
@@ -351,3 +351,7 @@ The directory where test timing data can be found.
 `CIRCLE_STAGE`
 
 The job being executed. The default 2.0 job is `build` without using Workflows.
+
+`CIRCLE_WORKFLOW_ID`
+
+A unique identifier for the current workflow instance that the build is running in. This identifier will be the same for every job that runs within this instance of the workflow.
```
gordonsyme: Yes, I think it is OK to document this one and make it official.

> A unique identifier for the workflow run that the job belongs to. This identifier will be the same for every job that belongs to this workflow run.

would be more consistent with the language we use when talking about workflows, @michelle-luna is that correct?

moredip: Super! Thanks for making it so easy for me to contribute.

PR #2237, opened 2018-04-11
Revise "Scheduled Workflow" documentation

smart-alek:
Resolves [CIRCLE-10041](https://circleci.atlassian.net/secure/RapidBoard.jspa?rapidView=67&modal=detail&selectedIssue=CIRCLE-10041) and makes conceptual overview consistent with Kyle's new video.


jekyll/_cci2/workflows.md
```diff
@@ -172,21 +172,24 @@ Following is a screenshot of the Approval dialog box that appears when you click
 
 ## Scheduling a Workflow
 
-Workflows that are resource-intensive or that generate reports may be run on a schedule rather than on every commit. This feature is configured by adding a scheduled trigger to the configuration of the workflow.
```
michelle-luna: No need to remove this information, please add back

jekyll/_cci2/workflows.md
```diff
@@ -172,21 +172,24 @@ Following is a screenshot of the Approval dialog box that appears when you click
...
 ## Scheduling a Workflow
 
-Workflows that are resource-intensive or that generate reports may be run on a schedule rather than on every commit. This feature is configured by adding a scheduled trigger to the configuration of the workflow. 
+It can be inefficient and expensive
+to run a workflow for every commit for every branch.
+Instead,
+you can schedule a workflow
+to run at a certain time for specific branches.
 
-Configure a workflow to run on a set schedule by using the `triggers:` key with the `schedule` option. The `triggers` key is **only** added under your `workflow` key. This feature enables you to schedule a workflow run by using `cron` syntax to represent Coordinated Universal Time (UTC) for specified branches.
```
michelle-luna: no need to remove, add back

jekyll/_cci2/workflows.md
```diff
@@ -172,21 +172,24 @@ Following is a screenshot of the Approval dialog box that appears when you click
...
-Workflows that are resource-intensive or that generate reports may be run on a schedule rather than on every commit. This feature is configured by adding a scheduled trigger to the configuration of the workflow. 
+It can be inefficient and expensive
+to run a workflow for every commit for every branch.
+Instead,
+you can schedule a workflow
+to run at a certain time for specific branches.
 
-Configure a workflow to run on a set schedule by using the `triggers:` key with the `schedule` option. The `triggers` key is **only** added under your `workflow` key. This feature enables you to schedule a workflow run by using `cron` syntax to represent Coordinated Universal Time (UTC) for specified branches. 
-
-## Video: Schedule Builds to Test and Deploy Automatically
```
michelle-luna: add back the heading

jekyll/_cci2/workflows.md
```diff
@@ -172,21 +172,24 @@ Following is a screenshot of the Approval dialog box that appears when you click
...
+For a quick overview,
+watch the following video.
 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/FCiMD6Gq34M" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen></iframe>
 
-### Nightly Example
```
michelle-luna: add back this heading and section

jekyll/_cci2/workflows.md
smart-alek: This information is incorrect and misleading:

- **Any** workflow can be scheduled — not just ones that are resource-intensive or that generate reports.
- It's not exactly right to call it a "scheduled trigger". In this case, the defined `schedule` **is** the trigger.

jekyll/_cci2/workflows.md
smart-alek: Much of this information has only been moved, not removed:

"Configure a workflow to run on a set schedule by using the `triggers` key with the `schedule` option." --> Moved to lines 188-190.
"The `triggers` key is **only** added under your `workflow` key. --> Removed. Better explained in the linked Configuration Reference. Also not technically correct because the "workflow" key will be a named key.
"This feature enables you to schedule a workflow run by using `cron` syntax to represent Coordinated Universal Time (UTC) for specified branches." --> Moved to lines 220-226. Not entirely correct to call it `cron` syntax. Instead, the value of the `cron` key should be a valid crontab entry.

smart-alek:
@michelle-luna thanks for the review! I added back the headings, but I'm not comfortable making those other changes without more details on why the edits are bad. Thanks!

jekyll/_cci2/workflows.md
michelle-luna: I disagree, please add it back, this ticket was to add a note about a required key, not to go through and re-do this work (sorry, we just have too much work to do). Thank you!

jekyll/_cci2/workflows.md
michelle-luna: Sorry again, but please add this content and link back, I'm not going to revisit all of these decisions again. Your additions are great and not in conflict with what was already there! Thanks!

PR #2256, opened 2018-04-16
Update high-availability.md

iynere:
- add `uuid-ossp` info for `federations` database

quicken & wework have both pointed this out to us


michelle-luna:
adding Marcus as reviewer as well (he may have missed this original PR)

michelle-luna:
My preference is to keep it as-is rather than a blanket note up top.

PR #2258, opened 2018-04-17
Add note about `setup_circle_ci`

justinswart:
I spent some time debugging my builds on CircleCI after migrating to 2.0 until I discovered that it was necessary to add this command to `before_all`. I think that this can be more explicit in the documentation.

Or at least I think the documentation should explain that `setup_circle_ci` is an actual Fastlane command and not just an example of a command one might add to their `Fastfile`.

Thanks 🙂


marcomorain:
LGTM, thanks

PR #2272, opened 2018-04-19
Prefer YAML over YML

yhirano55:
### Summary

With uppercase, it's better to use YAML instead of YML.


michelle-luna:
Hi, thanks for this PR! We have a lot of users who get stuck because they create a config.yaml file (which doesn't work), so that is why we use YML in this introductory document.

yhirano55:
Thanks for the comment. I understood.

PR #2274, opened 2018-04-19
New database doc

### LINKED ISSUE #2257: CircleCI 2.0 PostgreSQL Doc Needs to be Rewritten
This issue is about <https://circleci.com/docs/2.0/postgres-config/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/postgres-config.md>)

This doc looks like it was copied over from 1.0. While the configuration samples have been ported over to 2.0, the concepts don't make sense anymore.

For example, the section `Configuring Passwordless Login` doesn't make sense anymore as that's not how the Docker version of PostgreSQL (at least the Docker Library version) would be configured. Even if you could, tools would need to be installed in the executing image which isn't mentioned.

This should be rewritten with CircleCI 2.0 in mind.

michelle-luna:
Fixes #2257


levlaz:
@michelle-luna this looks great to me, thank you!

PR #2275, opened 2018-04-20
Migration landing page

michelle-luna:


iynere:
i think the cloud-to-server docs info is really valuable but shouldn't live under the 'migration' header but rather just as a page in the Server Administrators docs—i think 'migration' should really specifically target 1.0-2.0 migration & EOL stuff

iynere:
@keybits hey can you take a look at this real quick & make sure the way we're talking about the 2.0 config generator seems accurate ?

michelle-luna:
Awesome, thank you! I will push this now and we can revisit the cloud->server migration content and where to put it in the admin section in future.

PR #2284, opened 2018-04-24
Increase required instance type size

heug:


michelle-luna:
thank you! just adding Dani as a reviewer as well.

PR #2286, opened 2018-04-26
Update caching.md: "Using a Lock File" section

gfranco93:
Recommend lockfile checksums instead of files as cache keys.

### DIFF
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -245,7 +245,7 @@ Listing multiple keys for restoring a cache increases the odds of a partial cach
 
 ### Using a Lock File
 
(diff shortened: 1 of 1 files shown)


smart-alek:
Hey @gfranco93! Thanks for submitting this PR.

So, from your change, you're arguing that the checksums rather than the lockfiles themselves are what should be used as cache keys, right?

I've attached one of our engineers (@eric-hu) to confirm that this is indeed what we want to recommend. If so, I'll pop in and clean up the copy.

Nice Guybrush avatar, also. ;)

gfranco93:
Yes, exactly. Thanks!

eric-hu:
This is a good clarification @gfranco93 , thank you for the PR!

PR #2294, opened 2018-04-27
update arch doc text for nomad client

michelle-luna:


iynere:
looks ok to me—adding @GERey, i feel like he will have thoughts on language here

michelle-luna:
George says yes.

PR #2300, opened 2018-04-30
add example of base64 encoding/decoding

smart-alek:
Per conversation with @eric-hu, follow-up code example for recent PR re: multiline env vars #2273.


eric-hu:
It might be a bit more obvious to users to spell out what goes where.  How about something like this?

```bash
$ echo "foobar" | base64
Zm9vYmFyCg==

# store the above value in env var UI
$ echo $MYVAR
Zm9vYmFyCg==

+$  echo $MYVAR | base64 --decode | docker login -u my_docker_user --password-stdin
Login Succeeded
```

It may also be good to mention that different command line programs will take credentials differently.

PR #2301, opened 2018-05-01
One-liner to fix a broken link

danburke:


smart-alek:
Thanks for catching that @danburke!

PR #2311, opened 2018-05-03
Fix up server docs

michelle-luna:


jekyll/_cci2/aws.md
```diff
@@ -105,7 +105,9 @@ Have available the following information and policies before starting the Previe
...
 2. Run `make init` to init `terraform.tfvars` file (your previous `terraform.tfvars` if any, will be backed up in the same directory).
-3. Fill `terraform.tfvars` with appropriate values.
+3. Fill `terraform.tfvars` with appropriate AWS values for section 1. 
+4. Specify a `circle_secret_passphrase` in section 2, replacing `...` with alpha numeric characters. Passprhase cannot be empty.
+5. Specify the instance size for your Nomad Clients. By default, the value specified in the `terraform.tfvars` file for Nomad Clients that get deployed are `m4.xlarge` instances which support three clients per server.  To increase the number of clients per server to support mutiple parallel builds, for example, modify section 2 of the `enterprise-setup/terraform.tfvars` file to specify a larger instance type for the `nomad_client_instance_type`. Using an `m4.2xlarge` instance with 8 CPUs and 32GB RAM is best practice for Nomad Clients. **Note:** The `builder_instance_type` is only used for 1.0 and is disabled by default in section 3.
```
iynere: 'clients per server' this is not the right language.... each **instance** is a nomad client. we are talking about number of CircelCI jobs (ie, nomad jobs, under the hood) that each **client** is capable of handling

jekyll/_cci2/aws.md
```diff
@@ -107,7 +107,7 @@ Have available the following information and policies before starting the Previe
...
 4. Specify a `circle_secret_passphrase` in section 2, replacing `...` with alpha numeric characters. Passprhase cannot be empty.
-5. Specify the instance size for your Nomad Clients. By default, the value specified in the `terraform.tfvars` file for Nomad Clients that get deployed are `m4.xlarge` instances which support three clients per server.  To increase the number of clients per server to support mutiple parallel builds, for example, modify section 2 of the `enterprise-setup/terraform.tfvars` file to specify a larger instance type for the `nomad_client_instance_type`. Using an `m4.2xlarge` instance with 8 CPUs and 32GB RAM is best practice for Nomad Clients. **Note:** The `builder_instance_type` is only used for 1.0 and is disabled by default in section 3. 
+5. Specify the instance type for your Nomad Clients—defaults to `m4.2xlarge`. To increase the number of CircleCI jobs that each Nomad Client can run, use a larger instance type (refer to AWS' [Amazon EC2 Instance Types](https://aws.amazon.com/ec2/instance-types) guide for details). **Note:** The `builder_instance_type` is only used for 1.0 and is disabled by default in section 3.
```
iynere: we can't say for certain how many jobs a particularly sized nomad client is capable of handling—b/c job resources are configurable. i think it's best to leave that information out, unless someone feels like it's worth stating how many default-sized (ie, `resource_class: medium`) jobs an `m4.2xlarge` nomad client is capable of running in-parallel...

michelle-luna: Thank you! I agree with removing 'three clients per server', but I think using the '-defaults to' is not the best style, because it assumes people know which line and section this is in the tfvars file (Ryan was clear that people have no idea which line is which or what to change, so I will add that info back so it is totally clear what line to change (nomad_client_instance_type) with the best practice info as well.)

PR #2318, opened 2018-05-07
Do not install the CLI tool in a Docker container

ryanwohara:
Installing the local CLI in a container will not work.


smart-alek:
Looks like an excellent clarification. Thanks for the catch @ryanwohara!

PR #2319, opened 2018-05-07

iynere:
rewording / reframing


iynere:
> what we say currently is ambiguous & confusing
> & not fully true
> we don’t actually say what someone would actually use DLC for
> so just adding a note to say what it DOESN’T do isn’t particularly useful

PR #2320, opened 2018-05-08
Remove the link to TypeScript/.circleci/config.yml

sryabkov:
It looks like Microsoft switched away from using Circle CI in the TypeScript project: there is no .circleci/config.yml in the HEAD of master, so the link was dead.


sryabkov:
👍

PR #2334, opened 2018-05-15
There is no such thing as 'cutting a release'

GabLeRoux:
I'm not a native english speaker, but I think the word "creating" was expected here.


smart-alek:
@GabLeRoux Actually there **is** [such a thing](https://english.stackexchange.com/questions/405791/origin-of-phrase-cut-a-release), but I do think that "creating a release" is more clear. :)

PR #2338, opened 2018-05-16
add name parameter to both steps

### LINKED ISSUE #2296: Key `name` for `save_cache` and `restore_cache`
Per reference, key `name` is only available in `run` declaration. And neither [save_cache](https://circleci.com/docs/2.0/configuration-reference/#save_cache) nor [restore_cache](https://circleci.com/docs/2.0/configuration-reference/#restore_cache) support it. But actually [we can use `name`](https://github.com/yarnpkg/yarn/blob/master/.circleci/config.yml#L22) in these two to define its title. It is by accident and may be removed, or it is a feature not documented yet?

smart-alek:
Fixes #2296 and resolves [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-10354).

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -558,7 +558,7 @@ Key | Required | Type | Description
 ----|-----------|------|------------
 paths | Y | List | List of directories which should be added to the cache
 key | Y | String | Unique identifier for this cache
-name | N | String | Title of the step to be shown in the CircleCI UI (default: `save_cache`)
+name | N | String | Title of the step to be shown in the CircleCI UI (default: "Saving Cache")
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -558,6 +558,7 @@ Key | Required | Type | Description
 ----|-----------|------|------------
 paths | Y | List | List of directories which should be added to the cache
 key | Y | String | Unique identifier for this cache
+name | N | String | Title of the step to be shown in the CircleCI UI (default: `save_cache`)
```
michelle-luna: I think the default appears as Restoring Cache and Saving Cache in the UI, so suggest to change defaults to those.

PR #2339, opened 2018-05-16
Adding see also links

michelle-luna:


michelle-luna:
thanks for these comments, I've made all of your changes.

PR #2343, opened 2018-05-17
Update google-auth.md

Gogoro:
Entering `$GCLOUD_SERVICE_KEY` in the admin panel throws an error, I guess it should be `GCLOUD_SERVICE_KEY`?


smart-alek:
@Gogoro that's correct! Adding the `$` is a bash thing. Thanks for helping us out! :)

PR #2345, opened 2018-05-17
update CIRCLE_INTERNAL_TASK_DATA value and location

sararead:
Removed references to CIRCLE_INTERNAL_TASK_DATA. This env var seems to have been removed in this commit: https://github.com/circleci/picard-services/commit/14195162baca1243426ad0fb8fb0c93eda07f884

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -373,10 +373,6 @@ The username or organization name of the project being tested, i.e. “foo” in
 
 The repository name of the project being tested, i.e. “bar” in circleci.com/gh/foo/bar/123.
 
-`CIRCLE_INTERNAL_TASK_DATA`
-
-The directory where test timing data can be found.
-
 `CIRCLE_STAGE`
 
 The job being executed. The default 2.0 job is `build` without using Workflows.
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -373,6 +373,10 @@ The username or organization name of the project being tested, i.e. “foo” in
 
 The repository name of the project being tested, i.e. “bar” in circleci.com/gh/foo/bar/123.
 
+`CIRCLE_INTERNAL_TASK_DATA`	
+
+The directory where test timing data can be found.
+
 `CIRCLE_STAGE`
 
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/parallelism-faster-jobs.md
```diff
@@ -140,7 +140,7 @@ use the `--split-by` flag with the `filesize` split type.
 
 #### Splitting by Timings
 
-CircleCI automatically saves timing data from previous successful runs to a default location (`$CIRCLE_INTERNAL_TASK_DATA/circle-test-results`).
+CircleCI automatically saves timing data from previous successful runs to a default location.
```
smart-alek: If the default location isn't `CIRCLE_INTERNAL_TASK_DATA`, where is it? 🤔

sararead:
@smart-alek I made a couple more changes to this! This env var should be present, but the location of where it was saving changed. I updated the docs to reflect this.

jekyll/_cci2/env-vars.md
```diff
@@ -373,7 +373,7 @@ The username or organization name of the project being tested, i.e. “foo” in
 
 The repository name of the project being tested, i.e. “bar” in circleci.com/gh/foo/bar/123.
 
-`CIRCLE_INTERNAL_TASK_DATA`
+`CIRCLE_INTERNAL_TASK_DATA`
```
smart-alek: Extra whitespace?

eric-hu:
Wait, I don't think we want to publish what the value of that env var is.  Its actual location doesn't matter and could change.  Instead, I think we just want to tell users that their test files, if there are any, are in the foldername inside `$CIRCLE_INTERNAL_TASK_DATA`.  So they should be able to, say, `ls -la $CIRCLE_INTERNAL_TASK_DATA` to see if anything is there.

eric-hu:
I *think* the old documentation is incorrect in that it says `$CIRCLE_INTERNAL_TASK_DATA/circle-test-results`, but I'm not sure the subfolder `circle-test-results` should be in there.

smart-alek:
@eric-hu who could we tap to get some answers on this?

sararead:
Go ahead and update it!

PR #2341, opened 2018-05-16
Add ssh deploy example

smart-alek:
Fixes #2332.

PR #2359, opened 2018-05-23
prevent redundant builds on non-deploy branches

smart-alek:
Follow-up to [this comment](https://github.com/circleci/circleci-docs/pull/2341#discussion_r189397775) on the SSH Deploy guide.

### DIFF
--- a/jekyll/_cci2/deployment-integrations.md
+++ b/jekyll/_cci2/deployment-integrations.md
@@ -391,7 +391,7 @@ workflows:
   version: 2
   build-and-deploy:
     jobs:
-      - build
+      - build:
         filters:
           branches:
             only: master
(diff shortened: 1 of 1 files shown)


smart-alek:
@eric-hu iffy about the whole change itself? How so?

(I'm not tied to pushing this; just grew out of a user's suggestion).

eric-hu:
If the deployment branch is `sequential-branch-filter`, building that branch will always be redundant if it's a protected branch.  In that case, `sequential-branch-filter` only merges commits from master that passed on CI.  If `sequential-branch-filter` isn't a protected branch, it's possible its tip includes a branch that wasn't green on master.

This doesn't seem like a _likely_ scenario, so I don't feel too strongly about it.  Just somewhat iffy.

smart-alek:
@eric-hu got it. I think since there **is** an edge case here, I'd be more comfortable recommending configuration that applies more generally. In light of that, I'm going to close this PR.

PR #2362, opened 2018-05-24
add warning to use environment variables in scripts

### LINKED ISSUE #2314: Add warning to use Environment Variables in scripts
This issue is about <https://circleci.com/docs/2.0/env-vars/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/env-vars.md>). 

I would suggest to add a warning that when you call a script defined in a command in .circleci/config.yml that then the content of the Environment Variables will be logged in clear text. For critical values like passwords or API Keys only the command in .circleci/config.yml should be used. Then the Environment Variables is only shown with it's variable name.

smart-alek:
Fixes #2314 and improves document flow and consistency.

Highlights include:
- Reordered sections. Sections listed in the same order as Overview
- "Setting" > "Declaring" / "Using" / "Adding". Corrected inconsistent usage.
- Created a follow-up to [reformat this section](https://circleci.atlassian.net/browse/CIRCLE-11426) with tables.

@ndintenfass and @marcomorain for technical accuracy
@michelle-luna for language edits

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -100,18 +100,7 @@ in the same way that `docker` does.
 
 To set global environment variables that may be shared across projects, use the Settings > Contexts page of the CircleCI application. See the [Contexts]( {{ site.baseurl }}/2.0/contexts/) documentation for instructions.
 
-## Setting Environment Variables in `.circleci/config.yml`
-
-It is also possible to set environment variables at the **job**, **container**, and **step** levels,
-in order of increasing specificity.
-
-1. Environment variables set at the **job** level
-are overwritten by...
-2. ...environment variables set at the **container** level,
-which are overwritten by...
-3. ...environment variables set at the **step** level.
-
-### Setting Job-Level Environment Variables
+## Setting Job-Level Environment Variables
 
 To set environment variables at the job level,
 use the [`environment` key]({{ site.baseurl }}/2.0/configuration-reference/#job_name).
@@ -126,7 +115,7 @@ jobs:
       FOO: "bar"
 ```
 
-### Setting Container-Level Environment Variables
+## Setting Container-Level Environment Variables
 
 To set environment variables at the container level,
 use the [`environment` key]({{ site.baseurl }}/2.0/configuration-reference/#docker--machine--macosexecutor).
@@ -159,7 +148,7 @@ jobs:
       - image: circleci/postgres:9.6
 ```
 
-### Setting Step-Level Environment Variables
+## Setting Step-Level Environment Variables
 
 To set environment variables at the step level,
 use the [`environment` key]({{ site.baseurl }}/2.0/configuration-reference/#run)
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -94,55 +94,6 @@ Environment variables set within configuration are fully visible in build output
 If you are adding secrets or sensitive data,
 set these at the [project level](#adding-project-level-environment-variables) instead.
 
-### Setting an Environment Variable for a Step
-
-Use the environment key inside a run step to set variables for a single command shell as shown in the following example:
-
-```yaml
-version: 2
-jobs:
-  build:
-    docker:
-      - image: smaant/lein-flyway:2.7.1-4.0.3
-    steps:
-      - checkout
-      - run:
-          name: Run migrations
-          command: sql/docker-entrypoint.sh sql
-          # Environment variable for a single command shell
-          environment:
-            DATABASE_URL: postgres://conductor:@localhost:5432/conductor_test
-```
-
-### Interpolating Environment Variables 
-
-CircleCI does not support interpolation
-when defining configuration variables like `working_directory` or `images`.
-All defined values are treated literally.
-
-However, it is possible to interpolate a variable within a command
-by setting it for the current shell.
-
-```yaml
-version: 2
-jobs:
-  build:
-    steps:
-      - run:
-          name: Update PATH and Define Environment Variable at Runtime
-          command: |
-            echo 'export PATH=/path/to/foo/bin:$PATH' >> $BASH_ENV
-            echo 'export VERY_IMPORTANT=$(cat important_value)' >> $BASH_ENV
-            source $BASH_ENV
-```
-
-**Note**:
-Depending on your shell,
-you may have to append the new variable to a shell startup file
-like `~/.tcshrc` or `~/.zshrc`.
-For more information,
-refer to your shell's documentation on setting environment variables.
-
 ### Setting Environment Variables for a Job
 
 To define environment variables for a job, use the `environment` key under the job name in the `jobs` section.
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/env-vars.md
```diff
@@ -16,19 +16,39 @@ This document describes using environment variables in CircleCI in the following
...
+7. Environment variables [set within shell commands](#setting-command-level-environment-variables).
+Set in your `.circleci/config.yml` file.
+
+CircleCI also loads environment variables in this order.
+If an environment variable is set at multiple scope levels,
+CircleCI uses the variable set at the **lowest level**.
+
+**Warning**:
+Environment variables set within configuration (levels 4-7)
+are **fully visible** in build output.
```
marcomorain: I would change `in build output` to `in build output and on the build page`.

smart-alek: Is it more correct to say `in built output on the build page`? Where are all the possible places for build output?

marcomorain: You can grab build output from the API too.

jekyll/_cci2/env-vars.md
```diff
@@ -16,19 +16,39 @@ This document describes using environment variables in CircleCI in the following
...
+If you are adding secrets or sensitive data,
+set these within the CircleCI application (levels 2-3).
```
marcomorain: I would change

> If you are adding secrets or sensitive data, set these within the CircleCI application (levels 2-3).

> Do not add secret or sensitive data anywhere inside config.yml. Put secret and sensitive data into contexts (+link) or projects settings (+link). The full text of `config.yml` is *visible* to anyone who has access to your project on CircleCI.

jekyll/_cci2/env-vars.md
```diff
@@ -16,19 +16,40 @@ This document describes using environment variables in CircleCI in the following
 
 ## Overview
 
-Environment variables are used according to a specific precedence order, as follows:
```
michelle-luna: Please leave this section as-is, we had it general to specific before and changed it to specific to general in March based on the feedback from customers on Contexts. (https://github.com/circleci/circleci-docs/commit/34af95df586700293547a7aba0843da65ef3737f#diff-c9b824d50d2de38d360f6856babb1c4f)

jekyll/_cci2/env-vars.md
```diff
@@ -80,21 +101,68 @@ Login Succeeded
 Not all command-line programs take credentials
 in the same way that `docker` does.
 
-## Adding Global Environment Variables
+## Setting Context-Level Environment Variables
 
-To add global environment variables that may be shared across projects, use the Settings > Contexts page of the CircleCI application. See the [Contexts]( {{ site.baseurl }}/2.0/contexts/) documentation for instructions.
```
michelle-luna: Don't remove this information please.

jekyll/_cci2/env-vars.md
```diff
@@ -16,19 +16,40 @@ This document describes using environment variables in CircleCI in the following
...
+4. [Job-level](#setting-job-level-environment-variables) environment variables.
+Set in your `.circleci/config.yml` file.
+5. [Container-level](#setting-container-level-environment-variables) environment variables.
+Set in your `.circleci/config.yml` file.
+6. [Step-level](#setting-step-level-environment-variables) environment variables.
+Set in your `.circleci/config.yml` file.
+7. Environment variables [set within shell commands](#setting-command-level-environment-variables).
+Set in your `.circleci/config.yml` file.
+
+CircleCI also loads environment variables in this order.
```
michelle-luna: No, they are not used in the order from lowest level to highest as you have it here. This is what we determined in the March revision and clarified.

jekyll/_cci2/env-vars.md
```diff
@@ -16,19 +16,40 @@ This document describes using environment variables in CircleCI in the following
...
+If an environment variable is set at multiple scope levels,
+CircleCI uses the variable set at the **lowest level**.
+
+**Warning**:
```
michelle-luna: Great, this warning is the fix for the issue that was filed, thanks for adding this. The rest of this PR I would rather not include as you guessed. We don't use terms this way anywhere else, we use the words project, job, context and do not hyphenate them with -level.

jekyll/_cci2/env-vars.md
```diff
@@ -80,21 +101,68 @@ Login Succeeded
 Not all command-line programs take credentials
 in the same way that `docker` does.
 
-## Adding Global Environment Variables
+## Setting Context-Level Environment Variables
```
michelle-luna: I think Context-Level is awkward and not the best for translation either, so I would prefer Setting Environment Variables for a Context/ with a Context. But, and was reviewed with the context documentation previously, so it should stay as-is really.

jekyll/_cci2/env-vars.md
smart-alek: @marcomorain reviewed this new PR for technical accuracy and approved it. There _might_ be different behavior for the special, built-in environment variables, but the rest of this load order (general to specific) is correct. Environment variables at the configuration level will overwrite variables with the same name at a higher level.

jekyll/_cci2/env-vars.md
smart-alek: Hmm, this linked diff doesn't point to a general --> specific Overview. I can definitely see the argument for a specific --> general; that's how [Wikipedia](https://en.wikipedia.org/wiki/Scope_(computer_science)#Levels_of_scope) does it.

Regardless of the order used, though, I think the ordering of the sections in the document should match the precedence order used in the Overview.

jekyll/_cci2/env-vars.md
smart-alek: Which information specifically? I think the only difference in content is the signposting for "Settings > Contexts".

jekyll/_cci2/env-vars.md
smart-alek: Yeah, I agree that the hyphenated combo is awkward. I do like using the word "level" since it's a common term in discussions of [scope](https://en.wikipedia.org/wiki/Scope_(computer_science)#Levels_of_scope).

What about "Setting Environment Variables at the Context Level"?

jekyll/_cci2/env-vars.md
smart-alek: Could you clarify which terms you mean? I'm still using **project**, **job**, and **context**. I agree that the hyphenation is a bit odd.

jekyll/_cci2/env-vars.md
michelle-luna: Where did you get this information though? I think this is not what we agreed before when this came up in customer feedback and you are changing it now, but I'm not sure why.

jekyll/_cci2/env-vars.md
michelle-luna: I'm am saying don't use project-level, job-level, or context-level.

jekyll/_cci2/env-vars.md
michelle-luna: We don't use level, we just say context. 'in a context' or 'on the contexts page'

jekyll/_cci2/env-vars.md
```diff
@@ -18,98 +18,28 @@ This document describes using environment variables in CircleCI in the following
 
 Environment variables are used according to a specific precedence order, as follows:
 
-1. Environment variables declared inside a shell command in a `run` step, for example `FOO=bar make install`.
-2. Environment variables declared with the `environment` key for a `run` step.
-3. Environment variables set with the `environment` key for a job.
-4. Environment variables set with the `environment` key for a container.
+1. Environment variables declared [inside a shell command](#setting-an-environment-variable-in-a-shell-command) in a `run` step, for example `FOO=bar make install`.
+2. Environment variables declared with the `environment` key [for a `run` step](#setting-an-environment-variable-in-a-step).
+3. Environment variables set with the `environment` key [for a container](#setting-an-environment-variable-in-a-container).
```
michelle-luna: Job should be 3
Container should be 4

michelle-luna: and then the job and container sections should follow that order too

PR #2376, opened 2018-05-31
Document yaml multi merge

smart-alek:
New document for writing YAML + documentation of map multi-merging for DRYer configuration files.

Resolves [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-5832).

### DIFF
--- a/jekyll/_cci2/writing-yaml.md
+++ b/jekyll/_cci2/writing-yaml.md
@@ -14,8 +14,9 @@ a data serialization language used to configure CircleCI.
 ## Overview
 
 YAML is a [human-friendly data serialization standard](http://yaml.org/) for all programming languages.
-It is a strict superset of [JSON](https://www.json.org/), another data serialization language.
-This means that it can do everything that JSON can... and more.
+It is a strict superset of [JSON](https://www.json.org/),
+another data serialization language,
+which means it has all of the features of JSON.
 
 CircleCI configuration is stored in a single YAML file,
 located at `~/.circleci/config.yml`,
--- a/jekyll/_cci2/writing-yaml.md
+++ b/jekyll/_cci2/writing-yaml.md
@@ -33,9 +33,24 @@ and consists of one or more key-value pairs.
 key: value
 ```
 
-### Scalar Types
+You can set another key-value pair as a value
+by indenting the nested key.
 
-### Collection Types
+```yaml
+key:
+  another_key: "another value"
+```
+
+Keys and values are not restricted to [scalars](https://softwareengineering.stackexchange.com/questions/238033/what-does-it-mean-when-data-is-scalar).
+You can also map a scalar to a sequence.
+
+```yaml
+scalar:
+  - here
+  - is
+  - a
+  - sequence
+```
 
 ### Default Values and Map Merging
 
--- a/jekyll/_cci2/writing-yaml.md
+++ b/jekyll/_cci2/writing-yaml.md
@@ -29,10 +29,8 @@ it is important to understand the basics of YAML formatting.
 
 ## How to Write YAML
 
-The basic structure of a YAML file is a hash map
+The basic structure of a YAML file is a [hash map](https://en.wikipedia.org/wiki/Hash_table)
 and consists of one or more key-value pairs.
-For a description of hash maps,
-see the [Wikipedia page on hash tables](https://en.wikipedia.org/wiki/Hash_table).
 
 ```yaml
 key: value
--- a/jekyll/_cci2/writing-yaml.md
+++ b/jekyll/_cci2/writing-yaml.md
@@ -9,7 +9,7 @@ order: 20
 *[Basics]({{ site.baseurl }}/2.0/basics/) > Writing YAML*
 
 This document gives an overview of YAML,
-a data serialization language used in CircleCI configuration files.
+a data serialization language used to configure CircleCI.
 
 ## Overview
 
--- a/jekyll/_cci2/writing-yaml.md
+++ b/jekyll/_cci2/writing-yaml.md
@@ -8,8 +8,8 @@ order: 20
 
 *[Basics]({{ site.baseurl }}/2.0/basics/) > Writing YAML*
 
-This document gives an overview of the YAML data serialization language,
-which is used in CircleCI configuration files.
+This document gives an overview of YAML,
+a data serialization language used in CircleCI configuration files.
 
 ## Overview
 
--- a/jekyll/_cci2/writing-yaml.md
+++ b/jekyll/_cci2/writing-yaml.md
@@ -12,6 +12,9 @@ order: 20
 This document describes the most important features of YAML
 for use in CircleCI configuration.
 
+* TOC
+{:toc}
+
 ## Overview
 
 [YAML](http://yaml.org) is a human-friendly data serialization standard for all programming languages.
--- a/jekyll/_cci2/writing-yaml.md
+++ b/jekyll/_cci2/writing-yaml.md
@@ -64,10 +64,11 @@ You can also map a scalar to a sequence.
 
 ```yaml
 scalar:
-  - here
-  - is
-  - a
-  - sequence
+  - never
+  - gonna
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,166 @@
+---
+layout: classic-docs
+title: "Writing YAML"
+short-title: "Writing YAML"
+description: "How to Write YAML on CircleCI"
+order: 20
```
michelle-luna: there should be a category, see template in wiki

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,166 @@
...
+---
+
+*[Basics]({{ site.baseurl }}/2.0/basics/) > Writing YAML*
+
+This document gives an overview of YAML,
```
michelle-luna: gives...-> describes basic information about YAML and how to use anchors to prevent repeated YAML stanzas in your CircleCI configuration. (gives is too informal and should at least be 'gives you', but the goal of this article is to help people with long configuration files so saying that first is probably helpful.)

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,166 @@
...
+a data serialization language used to configure CircleCI.
+
```
michelle-luna: TOC, see wiki template for new document

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+For a more complex example,
+see [this gist](https://gist.github.com/bowsersenior/979804).
+
+## See More
+
+While YAML has several other features,
+the examples above should be enough
+to write an robust, functional `config.yml`.
```
michelle-luna: robust, functional -> to get you started with YAML and help you prevent long, repetitive CircleCI configuration files.

jekyll/_cci2/writing-yaml.md
smart-alek: 👍 to the wording choice re: "gives".

I think I see the actual **goal** of the document differently: isn't this more about giving people 90% of the YAML formatting they need to write a `config.yml`? Only one section actually deals with configuration files being long, and that would be covered by the table of contents.

Something that still isn't clear to me is the difference between the **description** and **overview** of a document. Could you help me with some broad descriptions of both of those, and I'll get them added to the Wiki?

jekyll/_cci2/writing-yaml.md
smart-alek: I added a note to the New Doc Template about where to find the available categories.

jekyll/_cci2/writing-yaml.md
smart-alek: 👍 to removing "robust, functional"; that's definitely not detailed enough.

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+This means it can do everything JSON can... and more.
+
+CircleCI configuration is stored in a single YAML file located at `~/.circleci/config.yml`,
+where `~` is the root of your project's directory.
+Since most of your work with CircleCI occurs in this file,
+it is important to understand the basics of YAML formatting.
+
+## How to Write YAML
+
+The basic structure of a YAML file is a [map](https://en.wikipedia.org/wiki/Map_(higher-order_function))
```
eric-hu: That link should refer to a hashmap article: https://en.wikipedia.org/wiki/Hash_table

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+use the `|` character,
+followed by any number of lines.
+This is especially useful for lengthy commands.
+
+```yaml
+haiku: |
+  Consider me
+  As one who loved poetry
+  And persimmons.
+```
```
eric-hu: There are multiple options for multi-line strings.  It might be worth spelling this out?

The above parses to (in json):

```
{
  "haiku": "Consider me\nAs one who loved poetry\nAnd persimmons."
}
```

An alternative multi-line string syntax is `>`:

```
haiku: >
  Consider me
  As one who loved poetry
  And persimmons.
```

Which compiles to the following json:

```

Output

{
  "haiku": "Consider me As one who loved poetry And persimmons."
}
```

I'm not sure if presenting both options would be too complex for our docs.  If we only want to present one, I think `>` is better since you can actually replace a long string with a `>` multiline.

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+Quotes are not necessary
+when using multiline strings.
+
+### Sequences
+
+Keys and values are not restricted to [scalars](https://softwareengineering.stackexchange.com/questions/238033/what-does-it-mean-when-data-is-scalar).
+You can also map a scalar to a sequence.
+
+```yaml
+scalar:
```
eric-hu: :thismotivatesme:

jekyll/_cci2/writing-yaml.md
smart-alek: Very nice callout — I wasn't aware of that alternative! :)

I think presenting both would probably be too much for the scope of this document. And I agree on sticking with `>`; that's more of a "pure" multiline-string since it's not injecting newlines.

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+This means it can do everything JSON can... and more.
+
+CircleCI configuration is stored in a single YAML file located at `~/.circleci/config.yml`,
+where `~` is the root of your project's directory.
+Since most of your work with CircleCI occurs in this file,
+it is important to understand the basics of YAML formatting.
+
+## How to Write YAML
+
+The basic structure of a YAML file is a [map](https://en.wikipedia.org/wiki/Hash_table)
```
michelle-luna: oh, this should say 'map, see the [wikipedia page for Hash_table](https://en.wikipedia.org/wiki/Hash_table) for a description.

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+  And persimmons.
+```
+
+**Note**:
+Quotes are not necessary
+when using multiline strings.
+
+### Sequences
+
+Keys and values are not restricted to [scalars](https://softwareengineering.stackexchange.com/questions/238033/what-does-it-mean-when-data-is-scalar).
```
michelle-luna: 'scalars, see the following [stackexchange post](https://softwareengineering.stackexchange.com/questions/238033/what-does-it-mean-when-data-is-scalar) about scalar data for details.

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+**Note**:
+Remember to properly indent a key-value pair
+when it is the value of an item in a sequence.
+
+### Anchors and Aliases
+
+To [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) up your `config.yml`,
+you can use anchors and aliases.
```
michelle-luna: remove 'you can' to make this more active. I guess the link to wikipedia is ok this way here, but I'd rather be obvious about where a link takes the user when possible.

jekyll/_cci2/writing-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+song:
+  - &name Al
+  - You
+  - can
+  - call
+  - me
+  - *name
+```
+
+When this list is read by a YAML parser,
```
michelle-luna: this list-> the above list

jekyll/_cci2/writing-yaml.md
smart-alek: And both "YAML" and "JSON" above do not have explanations about where they're going. For consistency's sake, should they also have an extra sentence?

PR #2377, opened 2018-05-31
Add nofollow to internal app links

smart-alek:
Makes all internal application links nofollow.

Resolves [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-7186).

### DIFF
--- a/jekyll/_cci2/nomad.md
+++ b/jekyll/_cci2/nomad.md
@@ -63,20 +63,9 @@ Complete the following steps to get logs from the allocation of the specified jo
 
 1. Get the logs from the allocation with `nomad logs -stderr <allocation-id>`
 
-<!---
-## Scaling the Nomad Cluster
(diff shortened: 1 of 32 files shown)


jekyll/_cci2/nomad.md
```diff
@@ -63,21 +63,10 @@ Complete the following steps to get logs from the allocation of the specified jo
 
 1. Get the logs from the allocation with `nomad logs -stderr <allocation-id>`
 
-<!---
```
michelle-luna: I'd rather keep these comments.

smart-alek: Can you help me understand why?

This information was commented out because it became outdated or inaccurate. Even after we have non-AWS installations, it's likely that this documentation would *still* be inaccurate.

If we're worried about losing this information, we really shouldn't be — GitHub tracks every change. And if this is critical, it should probably be tracked somewhere outside of the document itself.

PR #2379, opened 2018-06-01
Make the name of the triggered job consistent

sryabkov:
__WHY this PR was created__ There is an inconsistency within the `api-job-trigger.md` documentation page. The top part of the page talks about calling the `deploy_production` job while `config.yml` example at the bottom part of the page defines the `deploy_docker` job. 

__WHAT this PR changes__ This PR sets the name of the job at the top of the page to `deploy_docker` to match the `config.yml` definition


sryabkov:
👍

PR #2382, opened 2018-06-01
Update caching.md

iynere:
in advance of a more wide-ranging rework of our caching best practices, we should _definitely_ have cache counters in all of our examples, IMHO


iynere:
omg yes that would be amazing 🙏 @smart-alek

smart-alek:
@iynere ha, good! I actually meant in this PR, but if that's something you don't have bandwidth for, then I'm happy to make an issue for it and pop that up the priority list. :)

smart-alek:
@eddiewebb @keybits do either of you have further caching thoughts before I merge this?

PR #2388, opened 2018-06-04
high-availability.md - noted need for gold support

eddiewebb:
HA install support is a Gold-level feature.


eddiewebb:
CLosing, it sounds like Jane made ask of @michelle-luna through another means.

PR #2389, opened 2018-06-04
Use proper product name of CircleCI Server.

felicianotech:


michelle-luna:
Hi Ricardo,

Server isn't a product name actually, it is a deployment type (docs, PM, and mktg all use these in the same way as far as I understand). We have one product called CircleCI with two deployment options: on your servers or private cloud and or managed by us in our cloud. It doesn't mean we can't improve this doc in some way, just not in this exact way that you have edited it.

felicianotech:
That's strange. Okay thanks.

Issue #2383, opened 2018-06-01
Important info is missing in api-job-trigger.md


sryabkov:
This issue is about <https://circleci.com/docs/2.0/api-job-trigger/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/api-job-trigger.md>)

__Why this issue was opened__ The following was not clear to me until I did some experimenting. Documenting this on the page might save other people time.

Workflows cannot be triggered using the API, and the page does talk about this. However, even though my `config.yml` had a workflow definition, I was still able to add a new job to `config.yml`, and trigger it using the API. In my case, the workflow did not reference the new job, and I am not sure what would have happened if it did and if there were any downstream jobs. 

There is one important caveat. In our jobs referenced by the workflow, we were using contexts (as in https://circleci.com/docs/2.0/contexts/#creating-and-using-a-context). Environment variables defined in the context were not available when the new job was triggered using the API. The workaround we used was to re-define the environment variable on the project level instead of globally in the context. Because there is a very good UI for copying environment variables between projects, this is a very acceptable workaround.

smart-alek:
Hey @sryabkov, thanks for reaching out!

I'm still a little unsure of what information you'd like more of (or what was unclear). Do you want an explicit statement about being able to trigger a non-Workflows job? Or a more prominent statement about not being able to trigger Workflows with the API?

Or is it more to do with the second paragraph, where you discuss the workaround for sharing environment variables with contexts/projects?

Definitely want to help here — I just need a bit of confirmation on what specifically you found lacking, so we can get it right. :) Thanks for your help!

sryabkov:
@smart-alek Thanks and let me try to clarify:

- It was clear that it is not possible to trigger a _workflow_ using the API.
- It was NOT sufficiently clear that it is possible to trigger a _job_ using the API if that job was defined in `config.xml` that had the `workflows` section. __Suggestion:__ add the `workflows` section to the sample `config.yml` at the bottom of the page to illustrate that this scenario is possible.
- It was NOT clear at all that the job that is triggered via the API won't have access to environment variables defined in the referenced context, even if the job is referenced in the workflow

For example, to the last point, while experimenting, I tried this hack
```
workflows:
  version: 2
  build:
    jobs:
      - build:
          context: org-global
      - build_demo:
          requires:
            - build
          filters:
            branches:
              ignore:
                - /feature[\/_\.-].+/
          context: org-global
      - build_and_publish_demo_container:
[...]

PR #2390, opened 2018-06-05
Add notes about jobs being triggered using the API

sryabkov:
This commit/PR contains specific language addressing issues raised in #2383

### DIFF
--- a/jekyll/_cci2/api-job-trigger.md
+++ b/jekyll/_cci2/api-job-trigger.md
@@ -23,6 +23,8 @@ A few notes about this example:
 
 - `CIRCLE_API_TOKEN` should be the API token from your project's settings page.
 - `<vcs-type>/<org>/<repo>` should be replaced with your own VCS and org/repo names; at time of writing, `<vcs-types>` must be either `github` or `bitbucket`.
+- The job triggered using the API won't have access to the environment variables set up in a CicleCI context. 
+- The `config.yml` file where the job that is triggered using the API is defined may contain a workflow definition. The workflow does not have to reference this job.
 
 
 ## Conditionally Running Jobs With the API


sryabkov:
@smart-alek FYI ^^^

smart-alek:
@sryabkov thanks for submitting this. Your last note in #2383 did indeed help clear some stuff up.

I've been chatting with one of our product managers @KunalJain about the original problem you're trying to solve. Our concern is that we'd rather not document confusing behavior for a workaround. The _best_ solution would be to have first-class support for triggering a workflow.

Can you talk to us a little bit about what you're trying to accomplish by triggering a particular job in a workflow — rather than the workflow itself?

smart-alek:
@sryabkov thanks for taking (even more) time to write this up. Consider me convinced! I'll continue chatting with @KunalJain about a more robust solution for the long-term.

But! In the meantime, let's get this merged; two additional notes are not going to hurt the integrity of this document. ;)

sryabkov:
@smart-alek 👍 Thank you. Let me know if you have any other questions. 

By the way, what's the best way to communicate a suggestion/enhancement request? I have some suggestions on improving the workflows, which crystallized in my head as the result of our discussion.

smart-alek:
@sryabkov absolutely! I've got an incoming PR that cleans up this document a bit more, which I'll link to in the issue you initially created. You can take a look at that to make sure my "improvements" haven't sullied your original intents. 😬 

As for communicating suggestions/enhancements, we've got a public [Ideas board](https://circleci.com/ideas/) where customers can submit and vote on their favorite suggestions.

Thanks again for all the communication!

sryabkov:
@smart-alek 

If you tag me on your upcoming PR, I'll be happy to review it. I found some more confusing and inaccurate things on that page.

I posted the comment below to the `CCI-I-192 Trigger workflows via CircleCI REST API` in the [ideas board](https://circleci.com/ideas/). I didn't find a way to get a direct link (if there is none, it is an oversight, in my opinion).

> At the moment, even if it is not possible to trigger the entire workflow, it is possible to trigger a particular job that is referenced in the workflow. It is possible to re-implement the workflow (or implement a completely different one) by making the job call other jobs using the API. The mechanics are described in the documentation (https://circleci.com/docs/2.0/api-job-trigger/). 
> However, when a job is triggered using the API, context environment variables are not available to the job, even if the job references the context in the workflow definition (which sounds like a bug, if not an implementation one than definitely a requirements/design/user experience one). As a workaround, it is possible to define environment variables on the project level, but that defeats the whole point of context environment variables. I'd like to flag this issue here to make sure that it gets addressed as part of the effort to enable triggering workflows using the API.

smart-alek:
@sryabkov whoops, I actually already made the PR (#2395), but if you found other stuff that needs changing, let's get it fixed in another issue. Thanks for all the time you've spent helping us get a better grip on this. :)

sryabkov:
@smart-alek No problem, and sure, I am glad to help. 

In #2395, you corrected the definition of `CIRCLE_API_TOKEN` at the top of the page, in the [Overview](https://circleci.com/docs/2.0/api-job-trigger/#overview) and correctly stated that it is a Personal Access Token, not a Project-level environment variable. The problem is `CIRCLE_API_TOKEN` is also referenced in the second part of the page, in the [Conditionally Running Jobs With the API](https://circleci.com/docs/2.0/api-job-trigger/#conditionally-running-jobs-with-the-api) section, but over there it means something different. In _that_ section, `CIRCLE_API_TOKEN` _is_ a Project-level environment variable. 

I just created #2407 to address another small issue on that page.

smart-alek:
Ah, yep — fair enough! These are the kinds of inconsistencies I love fixing, so the more 👀, the better! I'll make an issue to straighten out the overloaded variable name. 👍

PR #2400, opened 2018-06-06
Reformat builtin env vars as table

smart-alek:
Tabular data should be tabular!

[JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-11426).


jekyll/_cci2/env-vars.md
```diff
@@ -16,6 +16,11 @@ This document describes using environment variables in CircleCI in the following
 
 ## Overview
 
+CircleCI uses Bash,
```
michelle-luna: Move after the first paragraph as this is more specific about how to fill in the field on the page.

PR #2407, opened 2018-06-07
Update the curl command template

sryabkov:
__Why was this PR made__ To clarify that the API call can be made against any branch, not just `master`

This PR is in reference to this documentation page: https://circleci.com/docs/2.0/api-job-trigger/


sryabkov:
@smart-alek FYI ^^^

PR #2437, opened 2018-06-22
Consolidate contexts document within Env Vars doc

smart-alek:
This is a consolidation of the [Contexts](https://circleci.com/docs/2.0/contexts/) doc into the [Environment Variables](https://circleci.com/docs/2.0/env-vars/) doc.

Similar to #2434, there were a lot of inconsistencies between these two docs. Since the Environment Variables doc details how to use every other kind of environment variable, it makes sense to save users the trip to a separate document. With Contexts incorporated into the main env var doc, users have all the info they need to decide which scope level to set their variables.

JIRA issue here: https://circleci.atlassian.net/browse/CIRCLE-11705


KunalJain:
I think we should keep the context docs since we are planning to add more features to it. 

If we need to consolidate, we can remove the section where we talk about env vars and the execution order under the context doc and point users to the env var section.

smart-alek:
@michelle-luna @KunalJain great, sounds like we're all aligned! I'll bring the Contexts document back and point to the environment variables doc for repeated information.

michelle-luna:
Sorry, I think we should just close this PR. The rewrite removes too many signposts that we added on purpose when we created contexts.

smart-alek:
@michelle-luna @KunalJain updated!

PR #2451, opened 2018-06-28
add pull request template

smart-alek:
In the same vein as #2450, this light template should help contributors with submitting pull requests.


felicianotech:
If we do multiple templates, we can use this path instead: `.github/pull_request_template.md`.

PR #2452, opened 2018-06-28
Update yarn cache to be more reliable

Zertz:
Fallbacks are undesirable since they make `yarn.lock` essentially useless and caching yarn's cache also makes it so `node_modules/`is rebuilt and will also include globally installed packages.


felicianotech:
Thank you @Zertz, this looks great.

For the keys, that was more of a style choose but your way works and might be the direction our recommendations are moving so :ok_hand:.

For caching `node_modules`, that might have actually been a bug so thank you.

PR #2454, opened 2018-06-28
Dlc update

michelle-luna:
CIRCLE-12133


iynere:
@michelle-luna added a bunch more to the example—i think without an actual Dockerfile to look at, it's hard to explain in a clear way—and restructured some other stuff 🙏 🙏

michelle-luna:
this can't be merged as-is, I will work on it again tomorrow

michelle-luna:
I added back the description of the video example (since it uses remote docker it is good to have that example written up as well). I also added back the limitations heading and moved the Enabling section to the top because this has always been at the top for sales/support reasons. I removed the extra italics and quotes that we don't use in our style.

PR #2461, opened 2018-06-29

iynere:
it's confusing to say 'you need to contact us to enable dlc' but also say 'enable dlc by adding this key to your config.yml'—let's leave the 'enabling' language for the message about contacting us, and say 'using dlc' elsewhere

### DIFF
--- a/jekyll/_cci2/docker-layer-caching.md
+++ b/jekyll/_cci2/docker-layer-caching.md
@@ -26,7 +26,7 @@ Docker Layer Caching can be used with both the [`machine` executor]({{ site.base
 
 **Note:** DLC has **no** effect on Docker images used as build containers. That is, containers that are used to _run_ your jobs are specified with the `image` key when using the [`docker` executor]({{ site.baseurl }}/2.0/executor-types/#using-docker) and appear in the Spin up Environment step on your Jobs pages.
 
-## Using DLC
+## Enabling DLC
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/docker-layer-caching.md
```diff
@@ -26,13 +26,13 @@ Docker Layer Caching can be used with both the [`machine` executor]({{ site.base
 
 **Note:** DLC has **no** effect on Docker images used as build containers. That is, containers that are used to _run_ your jobs are specified with the `image` key when using the [`docker` executor]({{ site.baseurl }}/2.0/executor-types/#using-docker) and appear in the Spin up Environment step on your Jobs pages.
 
-## Enabling DLC
```
michelle-luna: Agreed, and this section heading should stay "Enabling DLC" where we talk about support

michelle-luna: rest looks good

iynere: that header is for not just the message about contacting support but also the subsequent two sub-sections about the RDE & machine exec, that's y the first 1 is a ## header & the next two are ### headers..... i'm fine w/ it reading 'enabling DLC' tho so i will make the change

iynere:
& now the build shouldn't fail also lol

Issue #2462, opened 2018-06-29
add-ssh-key documentation is wrong for 2.0


keen99:
This issue is about <https://circleci.com/docs/2.0/add-ssh-key/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/add-ssh-key.md>) and https://circleci.com/docs/2.0/configuration-reference/#add_ssh_keys

see discussion at https://stackoverflow.com/questions/50253917/circleci-2-0-ssh-keys-are-missing-migrating-from-circle-1-0/50295407#50295407

but in reality (other users experience as well as my own), the `- add-ssh-keys` step is required before any UI added keys are placed in the container.  

additionally, the ssh-agent is not setup for you in the environment. however, `.ssh/config` for the keys is generated instead (with hostname restrictions as documented)  [edit: corrected]

smart-alek:
Hey @keen99, thanks for filing an issue!

You've got impeccable timing because my [most recent PR](https://github.com/circleci/circleci-docs/pull/2463) is about exactly this! Take a look if you'd like, otherwise sit tight while we get this reviewed and merged. :)

PR #2463, opened 2018-06-29
Clarify that the add_ssh_keys step is more necessary than you think

smart-alek:
Resolves [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-12038).

Users have been under the impression that all SSH keys are automatically added to containers by the `ssh-agent`. This PR should make it more clear that they have to explicitly use the `add_ssh_keys` special step to do that.


smart-alek:
(Also potential fix for #2462)

PR #2467, opened 2018-07-03
Adding LogDNA into the List of Logging Providers

smusali:
Adding LogDNA into the List of Logging Providers


smart-alek:
Hey @smusali, sorry for the wait on this!

I noticed that you also made an issue about logging (#2468). Are the examples in this PR what you were looking for when you made that issue?

smusali:
@smart-alek, hey, thanks for responding! No - that's separate - issue is about the confusion possibility in docs - this PR is just about adding LogDNA reference for logging.

smart-alek:
Cool! So in this case, it's more about adding another example of a logging provider. I'll merge this and then follow up with a PR of my own to make the list of possible providers a bit more clear.

Let's jump over to the issue to have a discussion about that! :)

PR #2473, opened 2018-07-06
Improved instruction on how to define Regexes for branch filters

bvanderdrift:
On first impression and with no background in defining Regexes before, one might thinkg that you would have to define a Regex as such: `/s[a-zA-Z0-9 ]*/s` with the current documentation. This PR is a suggestion to improve the docs to become more clearer.


smart-alek:
A well-argued and welcome change! Thanks @bvanderdrift! :)

PR #2475, opened 2018-07-06
fix yaml example indentation

tomkuehl:


smart-alek:
Whoops, I definitely did this — thanks for catching it @tomkuehl!

PR #2477, opened 2018-07-09
porting artifactory from 1.0 docs

eddiewebb:
Adds basic config.yml and setup for artifactory.

Adding on 1.0 docs I added links to artifactory's guides for Maven and Gradle tools which simply setup


GERey:
Content wise this looks good to me.

PR #2479, opened 2018-07-10
Base checksum on package-lock.json

trodrigues:
This PR changes the recommendation of basing the cache checksum on package-lock.json rather than package.json.

If a package.json has non fixed versions and these change, we could actually end up with a different `node_modules` which effectively means the cache we have saved will be outdated and it might make builds slower than unnecessary until another change to package.json actually forces an update.

See https://docs.npmjs.com/files/package-lock.json for more details.

I've decided not to include a mention to the differences between package.json and package-lock.json as the whole document seems to be not specific to node, but if a recommendation is being issued it should be a more correct one, even if it's not the purpose of the document to explain the reason of such a recommendation.


smart-alek:
@trodrigues thanks for the changes! This is a great follow-up to a (relatively) recent PR I made that [updated caching best practices](https://github.com/circleci/circleci-docs/pull/2453) for specific dependency managers. In the additions there, we're actually already recommending `package-lock.json` — making the document more consistent with itself is always welcome! :)

PR #2480, opened 2018-07-11
Document required dockerfile label

smart-alek:
This PR adds context on how to enable the Docker `ENTRYPOINT` instruction in a custom Dockerfile. It also addresses a number of other issues with this document, including:

- undocumented: adding a `LABEL` to enable the `ENTRYPOINT` instruction.
- incorrect: most of the required tools can be installed with the `RUN` instruction, **not** the `ADD` instruction.
- unclear: you must create a `Dockerfile` **before** choosing a base image; section headers now represent steps in manually creating a Docker image.
- grammatical errors: some article omissions and assorted language fixes.

Resolves [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-11953).

Original request from @ndintenfass 
Tech review from @ryanwohara 
Writing review from @michelle-luna

### DIFF
--- a/jekyll/_cci2/custom-images.md
+++ b/jekyll/_cci2/custom-images.md
@@ -72,7 +72,7 @@ extend the base image
 by using the [`FROM` instruction](https://docs.docker.com/engine/reference/builder/#from).
 
 ```Dockerfile
-FROM golang:1.8.0
+FROM ubuntu:18.04
 ```
 
 ### Add an Entrypoint
--- a/jekyll/_cci2/custom-images.md
+++ b/jekyll/_cci2/custom-images.md
@@ -67,26 +67,17 @@ For example,
 if you want to use version 3.6 of the [official Alpine image](https://hub.docker.com/_/alpine/),
 the full image name is `alpine:3.6`.
 
-### Creating a Dockerfile
-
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/custom-images.md
```diff
@@ -37,61 +33,89 @@ Watch the following video for a detailed tutorial of customizing Docker images.
...
 ``` Dockerfile
-ADD custom_tool /usr/bin/
+ADD ./workdir/contacts /usr/bin/contacts
+ADD ./db/migrations /migrations
 ```
 
-In this case we copy `custom_tool` into the `/usr/bin/` directory of an image. **Note:** `custom_tool` must be in the same directory as `Dockerfile`. Read more about the [`ADD` command](https://docs.docker.com/engine/reference/builder/#add).
-
-## Building the image
+### Building the image
```
michelle-luna: Building the Image (initial caps for headings)

glesperance:
This is probably not the place for this but it seems this doc is incorrect as using the "com.circleci.preserve-entrypoint=true" LABEL doesn't seem to stop CircleCI from overriding the entry-point. Or perhaps  I am misunderstanding the doc.

PR #2488, opened 2018-07-16
Revise document for better structure and explanation of image tags

### LINKED ISSUE #2009: What is the difference between browsers and stretch-browsers
The list of docker images is helpful, but like, the lack of any sort of description of what the differences are is beguiling.

This issue is about <https://circleci.com/docs/2.0/circleci-images/> (source file: <https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/circleci-images.md>)

smart-alek:
After #2009 was re-opened, I spent some time thinking about why users were still having trouble understanding how our image tags worked. This PR is the result.

Highlights include:
- incorrect: the instructions for copying a SHA referenced old menu options
- inconsistent: document was missing "Overview" section
- unclear: a caveat about image tags being maintained by upstream sources (**not** CircleCI)
- unclear: "Best Practices" section has been split into more appropriate sections (in the previous version, Best Practices were introduced before the kinds of images and tags were discussed).

@felicianotech for technical review.
@michelle-luna for copy feedback.

Resolves #2009 and [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-12356).

### DIFF
--- a/jekyll/_cci2/circleci-images.md
+++ b/jekyll/_cci2/circleci-images.md
@@ -13,7 +13,15 @@ This document provides information about pre-built CircleCI images and a listing
 * TOC
 {:toc}
 
-For convenience, CircleCI maintains several Docker images. These images are extensions of official Docker images and include tools that are especially useful for CI/CD. All of these pre-built images are available in the [CircleCI org on Docker Hub](https://hub.docker.com/r/circleci/). The source code for these images is available at [github.com/circleci/circleci-images](https://github.com/circleci/circleci-images). Dockerfiles for each CircleCI image variant are archived at [github.com/circleci-public/circleci-dockerfiles](https://github.com/circleci-public/circleci-dockerfiles).
+## Overview
+
+For convenience,
+CircleCI maintains several Docker images.
+These images are extensions of official Docker images
+and include tools that are especially useful for CI/CD.
+All of these pre-built images are available in the [CircleCI org on Docker Hub](https://hub.docker.com/r/circleci/).
+The source code for these images is [available on GitHub](https://github.com/circleci/circleci-images).
+Dockerfiles for these images are also [archived on GitHub](https://github.com/circleci-public/circleci-dockerfiles).
 
 ## Best Practices
 
--- a/jekyll/_cci2/circleci-images.md
+++ b/jekyll/_cci2/circleci-images.md
@@ -142,7 +142,7 @@ That is, to prevent unintended changes that come from upstream, instead of using
 For example, add `-jessie` or `-stretch` to the end of each of those containers to ensure you're only using that base OS. You can pin down those images to a specific point version, like `circleci/ruby:2.3.7-jessie`, or you can just specify the OS version, with `circleci/ruby:2.3-jessie`. This is possible for any of the CircleCI images.
 
 It is also possible to specify all the way down to the specific SHA of the image you want to use. Doing so allows you to test specific images for as long as you like before making any changes. To find the value, navigate to on older Build page in the CircleCI app that utilized the image you wish to use and select the drop-down for `Spin up Environment`. Inside you will find the SHA256, for example, 
-`circleci/ruby@sha256:df1808e61a9c32d0ec110960fed213ab2339451ca88941e9be01a03adc98396e`. 
+`circleci/ruby@sha256:df1808e61a9c32d0ec110960fed213ab2339451ca88941e9be01a03adc98396e`.
 
 
 {% assign images = site.data.docker-image-tags | sort %}
--- a/jekyll/_cci2/circleci-images.md
+++ b/jekyll/_cci2/circleci-images.md
@@ -62,7 +62,7 @@ by adding an [image tag](https://docs.docker.com/engine/reference/commandline/ta
 
 For example,
 instead of `circleci/golang`,
-you can specify the version and OS
+specify the version and OS
 by using `circleci/golang:1.8.6-jessie`.
 Because the second image specifies a version and OS,
 it is less likely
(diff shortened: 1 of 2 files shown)


smart-alek:
Hmm. It might also be worth renaming this article to "Using Prebuilt Docker Images". On the [landing page](https://circleci.com/docs/2.0/build/), the document is called "Prebuilt Images" and is described as a "Complete list of prebuilt CircleCI Docker images".

There's actually a lot more going on in this document, though, so that title/description pairing could be made more accurate.

jekyll/_cci2/circleci-images.md
```diff
@@ -47,31 +56,35 @@ CircleCI maintains language images for the following languages:
...
-    <iframe width="560" height="315" src="https://www.youtube.com/embed/PgIwBzXBn7M" frameborder="0" allowfullscreen></iframe>
-</div>
-
 ## Pre-installed Tools
 
 All convenience images have been extended with additional tools.
 
-### APT Packages
-
 The following packages are installed with `apt-get` on every image:
```
felicianotech: Maybe use `via` instead of `with` as it makes more contextual sense and matches what we use for curl right after.

jekyll/_cci2/circleci-images.md
```diff
@@ -93,37 +106,93 @@ The following packages are installed with `apt-get` on every image:
...
 
 - [Docker client](https://docs.docker.com/install/)
 - [Docker Compose](https://docs.docker.com/compose/overview/)
 - [dockerize](https://github.com/jwilder/dockerize)
 - [jq](https://stedolan.github.io/jq/)
 
-## Image Variants
+## Customizing a Convenience Image
+
+You can customize convenience images by adding image tags.
```
felicianotech: I have a few thoughts here.

1. Using an existing image tag isn't 'customizing' an image. It's a common use case for images. Technically, a tag is always used (as `latest` will be used if left out). Most people use a tag.
2. The number 1 reason for using a tag, and ideally the most specific tag possible, is to achieve [Deterministic Builds](https://circleci.com/blog/preserve-build-integrity-prevent-problems-deterministic-builds/).
3. Using a SHA hash, if you're referring to the image digest, it's a tag. The digest ID is considered something different. https://docs.docker.com/engine/reference/commandline/pull/
4. The last option, "You want to modify an image by installing additional tools
+or changing the behavior of the image." would be done by extending an image.

jekyll/_cci2/circleci-images.md
```diff
@@ -93,37 +106,93 @@ The following packages are installed with `apt-get` on every image:
...
+- You want to pin an image to a certain version or operating system (OS).
+- You want to specify a particular instance of an image with a SHA.
+- You want to modify an image by installing additional tools
+or changing the behavior of the image.
 
-CircleCI maintains variants of convenience images. These can be created by adding optional suffixes to the end of image tags.
+### Pinning an Image Version or OS
 
-**Note:** If you choose to use the `latest` tag, the image may change unexpectedly and create surprising results.
+Since convenience images are based on the **latest** versions of upstream images,
```
felicianotech: We might want to be careful with using the term `latest` here. `latest` can be a Docker tag, but we don't just use that tag from upstream. Instead, we base our images off of the most recent built Docker image tags fro upstream.

jekyll/_cci2/circleci-images.md
```diff
@@ -93,37 +106,93 @@ The following packages are installed with `apt-get` on every image:
...
+it is best practice
+to use the most specific image possible.
```
felicianotech: Perhaps bold this statement or make it standout in some form or fashion. That is **important** advice.

jekyll/_cci2/circleci-images.md
```diff
@@ -93,37 +106,93 @@ The following packages are installed with `apt-get` on every image:
...
+CircleCI does not control which tags are used.
+These tags are chosen and maintained by upstream projects.
+Do not assume
+that a given tag has the same meaning across images!
 
-- `-ram`: variants that use the RAM volume to speed up builds
+### Specifying a Past Image Build With a SHA
 
-<hr>
+You can also use a SHA
```
felicianotech: We could use the proper term, Docker image digest, and perhaps link to Docker's Doc on it so that we don't have to explain it here?

jekyll/_cci2/circleci-images.md
```diff
@@ -93,37 +132,100 @@ The following packages are installed with `apt-get` on every image:
...
+to pin an image to a fixed version.
+
+### Using an Image Tag to Pin an Image Version or OS
+
+You can pin aspects of a Docker image
+by adding an [image tag](https://docs.docker.com/engine/reference/commandline/tag/#extended-description).
+
+For example,
+instead of `circleci/golang`,
+you can specify the version and OS
```
michelle-luna: remove you can

jekyll/_cci2/circleci-images.md
```diff
@@ -13,28 +13,43 @@ This document provides information about pre-built CircleCI images and a listing
 * TOC
 {:toc}
 
-For convenience, CircleCI maintains several Docker images. These images are extensions of official Docker images and include tools that are especially useful for CI/CD. All of these pre-built images are available in the [CircleCI org on Docker Hub](https://hub.docker.com/r/circleci/). The source code for these images is available at [github.com/circleci/circleci-images](https://github.com/circleci/circleci-images). Dockerfiles for each CircleCI image variant are archived at [github.com/circleci-public/circleci-dockerfiles](https://github.com/circleci-public/circleci-dockerfiles).
+## Overview
```
michelle-luna: Cool!

smart-alek:
@michelle-luna made all your changes!

PR #2492, opened 2018-07-18
Update yarn.md caching path

igor04:
`node_modules` looks a more convenient place for storing packages than `~/.cache/yarn`


felicianotech:
You're right, which is where Yarn and NPM store them. Yarn doesn't use that as a cache directory though. You can run `yarn cache dir` to see that, unless the default directory has been changed by your build.

PR #2506, opened 2018-07-20
fixes inaccuracies and layout issues in certificates.md

eddiewebb:
# Description
THis updates the certificates.md docs page to update it to work with latest server install, and fix confusing layout.

# Reasons

1. Certbot steps and CircleCI specifi steps were intermingled
1. Referenced "dummy certs" but certificates actually need to be valid or containers will fail to start
1.  Adds an approach for "self-signed" certs for folks that doent want to use certbot.

# Context

### DIFF
--- a/jekyll/_cci2/certificates.md
+++ b/jekyll/_cci2/certificates.md
@@ -7,7 +7,7 @@ categories: [administration]
 order: 55
 ---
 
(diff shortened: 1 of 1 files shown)


GERey:
:shipit:

Issue #2468, opened 2018-07-03
Logging in CircleCI 2.0


smusali:
# Document
Monitoring in CircleCI 2.0 Docs

# Category
- Unclear = confusing or ambiguous language

# Problem Details
What I am looking for is the possibility of logging of CircleCI 2.0 Builds. I have seen the possibility of logging in CircleCI 1.0 and/or Enterprise but the confusing thing is to see the logging of CircleCI 1.0 Builds in CircleCI 2.0 Docs instead of 1.0 Docs. So, is it possible to log CircleCI 2.0 Builds anyhow? Where do the logs go and/or get stored in?

## Unclear
- Having info about logging of CircleCI 1.0 Builds in CircleCI 2.0 Docs instead of CircleCI 1.0 Docs.

smart-alek:
Helloooo again @smusali!

Just to restate the problem you were having: is the issue that you were surprised to see documentation for 1.0 rather than 2.0? And you'd like a more 2.0-specific document on how to deal with logging?

smart-alek:
Adding @michelle-luna as well (she made the original document and might have some context that I don't.)

smusali:
@smart-alek @michelle-luna , so sorry for late response! Logging is available for only 1.0 and enterprise, right? So, documentation how to use logging should be, respectively, in 1.0 Docs and Enterprise Docs - there shouldn't be anything about logging in 2.0 Docs - but I see the documentation about logging for 1.0 inside 2.0 Docs - that initially created confusion till I saw the part saying `CircleCI 1.0 Builders store logs` but it is inside [this](https://circleci.com/docs/2.0/monitoring/#logging) which is for 2.0, right?

Besides this, when are you planning to provide logging for 2.0?

smusali:
Thanks a lot, @smart-alek!

I am looking forward to hearing what @GERey says :)

GERey:
@smusali Thanks for your patience! We introduced monitoring in 2.13 as something that gets piped to cloudwatch. Its an initial MVP that doesn't include nomad/builder logs so it doesn't have parity with what we had in 1.0 yet. We're working on getting that out as soon as possible and will update our changelogs with the metrics as we release them. 

[...]

PR #2519, opened 2018-07-24
Updated monitoring docs

GERey:
# Description
Changed the docs to better represent what is 1.0 monitoring and 2.0 monitoring.

# Reasons
It was unclear what was 2.0 and 1.0 specific.


GERey:
The objective of this PR is really to specify how the 2.0 way of monitoring works as opposed to how it used to work in 1.0. All of the 1.0 builder stuff still works so I left it in there. It might be worth just culling that. 

There is one more confusing thing and that is the distinction between 1.x release in server and a 2.x release in server. It might be helpful to clear that up ( these metrics and monitoring was only introduced in 2.13 anyways so it might be helpful to say that as well ) .

smart-alek:
Also, does this fix #2468 ?

PR #2534, opened 2018-07-27
Document env vars security

### LINKED ISSUE #2496: Environment variables could use more text in the "Security Features" doc
# Missing Information
How are environment variables stored?  Can any CircleCI employee see them in plain text?  Or are they stored encrypted with a key for each client, of some kind?  Are you relying on S3 / RDS encryption?  If so, what cipher are you using?  

# Already Looked
https://circleci.com/docs/2.0/security/

# Potential Locations
I would expect to find a section for Environment variables here: https://circleci.com/docs/2.0/security/
There is a single sentence today.

smart-alek:
# Description
This PR adds information about environment variable security to the Security document.

# Reasons
In #2496, a user requested more information about how environment variables are secured.

# Context
Fixes #2496 and resolves [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-12631).


jekyll/_cci2/security.md
```diff
@@ -18,7 +18,7 @@ Security is our top priority at CircleCI, we are proactive and we act on securit
...
+The nature of CircleCI is that our software has access to your code and whatever data that code interacts with. All jobs on CircleCI run in a sandbox (specifically, a Docker container or on-demand VM container) that stands alone from all other builds and is not accessible from the Internet or from your own network. The build container pulls code via git over SSH. Your particular test suite or job configurations may call out to external services or integration points within your network, and the response from such calls will be pulled into your jobs and used by your code at your discretion. After a job is complete the container that ran the job is destroyed and rebuilt. All environment variables are stored using [Hashicorp Vault](https://www.vaultproject.io/). Environment variables are encrypted using AES256-GCM96 and are unavailable to CircleCI employees.
```
tito-villalobos: s/stored/encrypted/

Single word change but important for accuracy.

Detailed explanation for internal understanding:
Vault does secrets management in a number of different ways.  It can store secrets directly, be an encryption service, integrate with aws security features, generate one-time-passwords, etc.  We are looking to leverage different features of Vault for different CCI features.

For user environment variables, we use the Vault encryption service (they call it the "transit secret store").  It's technically possible to "store customer environment variables in vault", but we chose this approach because Postgres is better at scaling for large volumes of data.

Thus, we "encrypt environment variables using vault" and store the resulting ciphertext in postgres.

tadwhitaker: @tito-villalobos Is that Postgres database in GPC or AWS?

tadwhitaker:
LGTM

jekyll/_cci2/env-vars.md
```diff
@@ -42,6 +42,8 @@ Do not add secrets or keys inside the `.circleci/config.yml` file.
 The full text of `config.yml` is visible
 to developers with access to your project on CircleCI.
 Store secrets or keys in [project](#setting-an-environment-variable-in-a-project) or [context]({{ site.baseurl }}/2.0/contexts/) settings in the CircleCI app.
+For more information,
+see the ["Encryption" section]({{ site.baseurl }}/2.0/security/#encryption) of the "Security" document.
```
michelle-luna: Remove quotes on Encryption

PR #2538, opened 2018-07-30
remove incorrect run step from elixir sample config

### LINKED ISSUE #2535: Elixir caching
# Location
https://circleci.com/docs/2.0/language-elixir/

# Explanation
The sample configuration is restoring these caches for the build cache
```yaml
- restore_cache:  # restores saved build cache
    keys:
      - v1-build-cache-{{ .Branch }}
      - v1-build-cache
```

but it's saving these:
```yaml
- save_cache: # don't forget to save a *build* cache, too
    key: v1-mix-cache-{{ checksum "mix.lock" }}
    paths: "_build"
- save_cache: # don't forget to save a *build* cache, too
    key: v1-build-cache-{{ .Branch }}
    paths: "_build"
- save_cache: # and one more build cache for good measure
    key: v1-build-cache
    paths: "_build"
```

I'm not sure if the correct fix is to remove the `checksum` `save_cache` call or to add a `restore_cache` key for `v1-mix-cache-{{ checksum "mix.lock" }}`

smart-alek:
# Description
Removes an unnecessary `save_cache` step from the Elixir example `config.yml`

# Reasons
The configuration example in the docs is both incorrect and doesn't match the actual project code.

# Context
Fixes #2535


tielur:
@smart-alek I believe line 45 also needs to be updated to:
`v1-mix-cache-{{ .Branch }}-{{ checksum "mix.lock" }}`

right?

tielur:
Or maybe you don't include the branch at all? Having a cache for a mix based on the mix.exs checksum is probably good enough.

zzak:
@tielur @smart-alek Thanks for your help on this!

My process has always been to get a PR on the [demo app](https://github.com/CircleCI-Public/circleci-demo-elixir-phoenix) and let that project build in CI to prove the config works as suggested. Then commit it to docs after that gets merged.

I'm not opinionated about how best to cache elixir projects, so deferring that to people (hopefully doc contributors :pray: :bow:) to make that call.

neo35:
Thank you ;)

Zachary Scott <notifications@github.com> schrieb am Di., 31. Juli 2018,
09:56:

smart-alek:
Hey @tielur, you are correct! Here's what I think we should do: for consistency's sake, we'll make this PR match the demo app that @zzak linked to. If you feel strongly about not needing the branch in the cache restoration step, let's move that conversation over to the demo application.

While you're right that a cache based on the checksum is _probably_ enough, we generally encourage cascading fallback caching for all the other languages. This is another case where seeing consistent examples of a configuration pattern helps out a ton.

Thanks for participating and providing input!

PR #2544, opened 2018-08-01
WIP: Update sidebar.html

michelle-luna:
# Description
This WIP PR updates the sidebar navigation to include multiple, nested levels with icons.

# Reasons
Docs redesign

# Context


felicianotech:
Some UX notes:

- Typically with a collapsing menu of this kind, the arrow/triangle that is clicked should flip or rotate when the menu state is changed. For example, the point is typically up with a menu is expanded, and down with the menu is collapsed.
- Every menu starts off expanded in the sidebar. This makes the ability to collapse the menu moot. Collapsable menus give the benefit of allowing a user to see all top-level items without having to scroll or do some other type of action. Right now, the user would have to scroll or start collapsing multiple menus in order to see the entire top-level. They're likely to not collapse the menus at all.
- The Welcome Menu. This can go either way, but perhaps this menu should be the only one that is expanded by default. Gives more visibility to the pages people need to get started.
- It would be very useful to have the current page being viewed highlighted in the sidebar menu.

UI note:

- the triangle trigger for each menu isn't aligned, seems to high. I assume this is known though. Mentioning it just in case.

PR #2566, opened 2018-08-06
Update non-aws.md

michelle-luna:
# Description
remove preview and unhide this doc 

# Reasons
CSE/Sales selling as supported

# Context


iynere:
@mqarty / @michelle-luna has there been confirmation from @guinaut / jane that this is what we want to do ?

PR #2567, opened 2018-08-06
[CIRCLE-10366] Docs new nav

trevor-circleci:
@digitalmaster

This is the initial pull request for this ticket. More changes are left to be made, namely:
- adding collapsible subnav items
- filling in the actual nav items to the left

### Why all the bespoke CSS values?
Lots of clean up needs to happen on this site; I've made an effort to use best practices in this PR, but more work needs to be done on the larger site to enable that. Hopefully, I'll get the opportunity to address these issues in further tickets.
### The commented out code
Leaving that in the `sidebar.html` file temporarily, as a reference to pull the sidebar links/structure from... yes this is the only place that information exists. 

# Description
Creating a new sidenav/sidebar to service the docs site.

# Reasons
The old nav needed beautification and functionality improvements, as well as structural improvements for the sake of code-management.

# Context
https://circleci.atlassian.net/browse/CIRCLE-10366


jekyll/assets/js/sidebar.js
```diff
@@ -0,0 +1,45 @@
...
+  if (localStorage.sidenavActive) {
+    var element = sidebar.querySelector('[data-section=' + localStorage.sidenavActive + ']');
+    if (element.classList.contains('closed')) {
+      element.classList.remove('closed');
+    }
+  }
+
+  window.addEventListener('scroll', function () {
+    // if footer is in frame, removed fixed style (otherwise add it, if it doesn't exist)
+    if ((footer.getBoundingClientRect().top - window.innerHeight) <= 0) {
```
rosston: I'm concerned about running this within a scroll event callback. I believe both `getBoundingClientRect` and `innerHeight` will cause a paint/render, which means that the page will have to paint/render on each individual scroll event (of which there are many), causing performance issues. Maybe the easiest solution is just to [`debounce`](https://lodash.com/docs/4.17.4#debounce) this callback function so it only gets called after the user stops scrolling?

rosston: Yeah, maybe we can pair on this today. I've run it locally, and I think you're right that it may not be a concern, but it's worth chatting about, I think.

rosston:
I checked this on my iPhone SE, and the "menu" button present on the current version (which shows the nav) is no longer showing up. As far as I can tell, there's no way to get to the nav anymore on small screens. Is that intentional/okay?

jekyll/_data/sidenav.yml
```diff
@@ -0,0 +1,217 @@
...
+        - name: Collecting Test Metadata
+          link: /2.0/collect-test-data/
+    - name: Triggers
+      link: /2.0/triggers/
+      children: 
+        - name: Skip and Cancel Builds
+          link: /2.0/skip-build
+        - name: Using the API to Trigger Jobs
+          link: /2.0/api-job-trigger
+        - name: Orchestrating Workflows
```
michelle-luna: Can we change Orchestrating Workflows  to 'Using Workflows to Schedule Jobs' please (per the usability study feedback)

jekyll/_data/sidenav.yml
```diff
@@ -0,0 +1,217 @@
...
+  icon: docs/start.svg
+  children: 
+    - name: Introduction
+      link: /2.0/getting-started/
+      children: 
+        - name: Sign Up & Try CircleCI
+          link: /2.0/first-steps/
+        - name: Hello World
+          link: /2.0/hello-world/
+    - name: Faq
```
michelle-luna: this appears as bold, but it should just be all upper-case, no bold font

trevor-circleci:
@millerpeterson I've addressed point #1 - just an unnecessary slash in the links as entered in the data file. 

For number 3: the reason for that is because it's a flat file system (but not an spa), and as far as I know there's no way to load the variable state of the nav with the dom on this system -- it has to happen after load, hence the re-expansion. Does that make sense? Is there a way I'm missing?

Aside from that, due to the desire to get this out, I'm going to merge this once the checks pass.

millerpeterson:
I'd have to look in more detail at how the nav is working, but off the top of my head one way to get around the re-opening issue would be to start with it invisible, and then have the JS control what gets initially shown (rather than have it rendered, and then getting the JS to set up the nav before it is displayed. Or I wonder if there is a way to get this set up statically via Jekyll at build time?

PR #2596, opened 2018-08-15
Iynere resource class patch

iynere:
# Description
it's not widely known that `resource_class` also, by proxy, allows users to configure additional/lesser disk space per job—let's add this info

# Reasons
see context here : https://circleci.slack.com/archives/C0XUN26MT/p1534197436000039

(summary: customer asked if there was a way to get more disk space for a job. we replied, no, there isn't, but try bumping `resource_class`. they reported that using `resource_class: xlarge` resulted in their job having access to nearly 600gb of available disk space, instead of the usual 100gb or so.


iynere:
closing this based on @DanToml's feedback : https://circleci.slack.com/archives/C0KBNJGRH/p1534361518000100

PR #2607, opened 2018-08-20
Remove auto-cancel caveat

ndintenfass:
Auto-cancel is now functional for workflows when you turn on build processing


ndintenfass:
@michelle-luna I'm going to go ahead and self-merge given how minor this is since it's blocking a marketing email.

ndintenfass:
@michelle-luna oh, you already approved ;)

PR #2612, opened 2018-08-21
Update vm-service.md

iynere:
add note that VM service requires EC2 instance types that support AWS' elastic network adapter

# Description
VM service will fail with instances that don't support enhanced networking / elastic network adapter—add a note so users are aware


jekyll/_cci2/vm-service.md
```diff
@@ -23,6 +23,8 @@ To configure VM service, it is best practice to select the AWS EC2 option in the
 
 If you do not provide a custom [Amazon Machine Image](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html) (AMI) for VM service, `machine` executor and remote Docker jobs on Server will run using the same machine image that we provide by default on Cloud: an Ubuntu 14.04 image with Docker version `17.03.0-ce` and docker-compose version `1.9.0`, along with a selection of common languages, tools, and frameworks. See the [`picard-vm-image` branch of our image-builder repository](https://github.com/circleci/image-builder/tree/picard-vm-image/circleci-provision-scripts) for details.
 
+***Note:** the EC2 instance type that you choose for VM Service must support the AWS Elastic Network Adapter. See AWS' [Enhanced Networking on Linux - Amazon Elastic Compute Cloud](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html) for a list of supported instance types.*
+
```
mqarty: Please add some statement our why, e.g. failures.

GERey: I was fairly certain it was due to the default AMI not supporting enhanced networking. They should be able to use it if they customize their ami with the drivers needed to support those instances.

iynere: @GERey enhanced networking (more specifically, the Elastic Network Adapter) is available on the instance-type level, not the AMI level... the default AMI definitely supports what it needs to, otherwise no one would be able to use VM service. the issue here is that folks don't know that certain instance types simply won't work with VM service due to what I've added in this PR.

mqarty: Can that be called out?

iynere: @mqarty sorry, can _what_ be called out ?

PR #2619, opened 2018-08-22
Migrate OOM doc and blog post

smart-alek:
# Description
Migrates @iynere's blog post to docs and replaces the /oom URL from 1.0 docs.

# Reasons
Migration project. Resolves [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-12506).


iynere:
thanks for doing this !!!!! @smart-alek

jekyll/_cci2/language-android.md
```diff
@@ -251,3 +251,8 @@ and may also increase memory usage.
 ## Deployment
 
 See the [Deploy]({{ site.baseurl }}/2.0/deployment-integrations/) document for examples of deploy target configurations.
+
+## Troubleshooting
+
+See the [Debugging Java OOM errors]({{ site.baseurl }}/2.0/oom/) document
```
iynere: should the link be `/oom` if it's java-specific ? we might want a more general oom page in the future (if we don't already have one)—i'd suggest `/java-oom`

jekyll/_cci2/oom.md
```diff
@@ -0,0 +1,130 @@
...
+title: "Avoiding and Debugging Java Memory Errors"
+description: "How to avoid and debug Java memory errors on CircleCI."
+---
+
+How to avoid and debug Java memory errors on CircleCI.
+
+## Overview
+
+The [Java Virtual Machine](https://en.wikipedia.org/wiki/Java_virtual_machine) (JVM) is a framework
+used for Android and Java development.
```
iynere: not exactly......

if you're looking for a short, 1-sentence explanation of what the JVM is, how about something like : 

> The Java Virtual Machine (JVM) provides a portable execution environment for various types of Java-based applications.

(adapted from https://javaworld.com/article/3272244/core-java/what-is-the-jvm-introducing-the-java-virtual-machine.html which seems really good)

smart-alek: Much more accurate, thanks!

jekyll/_cci2/language-android.md
smart-alek: Ah, great point — not only for this case but all other cases where we're migrating something from 1.0 Land.

PR #2643, opened 2018-08-28
[CIRCLE-12374] fixing highlight issue; adding subnav indent

trevor-circleci:
# Description
- Fixing issue with active section not displaying properly on sidenav.
- Added indentation to sub-items in sidenav/updated styling
- fixed issue where subnav positioning wasn't initialized until scroll

# Reasons
https://circleci.atlassian.net/browse/CIRCLE-12374

# To run docs:
- pull the `circleci-docs` repo.
- cd into the `jekyll` directory
- run `jekyll serve`
- go to `localhost:4000/docs/`

# Before
screen shot 2018-08-28 at 10 12 41 am

# After
screen shot 2018-08-28 at 10 13 56 am


digitalmaster:
@trevor-circleci Could you update PR title to include Jira ticket number `[CIRCLE-XXXX] - ***`. This links PR back to Jira ticket which helps us track updates from the ticket and helps reviewer gather additional context.

jekyll/_includes/sidebar.html
```diff
@@ -2,8 +2,10 @@
 <div class="sidebar-wrapper">
 	<nav class="sidebar fixed">
 		{% assign sidenav = site.data.sidenav %}
+		{% assign current_url = page.url | slice: 1, page.url.size %}
 
 		{% for main-item in sidenav %}
+		  {% assign current_url = page.url | slice: 1, page.url.size %}
```
JuanCaicedo: Does this need to be defined in here? It seems like the def on line 5 should be sufficient

jekyll/_sass/_sidebar.scss
```diff
@@ -66,13 +66,13 @@
...
     font-size: $font-size-micro;
-    font-weight: 100;
+    font-weight: 600;
     margin: 10px 0;
     letter-spacing: .5px;
   }
   .subnav-item {
     font-weight: 100;
-    margin: 5px 0;
+    margin: 5px 0 5px $spacing-medium-horizontal;
```
JuanCaicedo: I like seeing these in variables! Should `5px` be as well?

jekyll/assets/js/sidebar.js
```diff
@@ -45,7 +45,7 @@
 
 
 
-  window.addEventListener('scroll', function () {
+  function setSidebar () {
```
JuanCaicedo: I like seeing this in a named function! Makes the intent much clearer 👍

jekyll/_includes/sidebar.html
```diff
@@ -12,7 +14,7 @@
 						{% endif %}
 						
 						{% if main-item.link %}
-							<a class="{% if page.url == main-item.link %} active {% endif %}" href="{{ site.baseurl }}/{{ main-item.link }}#section={{ main-item.name | slugify }}" data-proofer-ignore>{{ main-item.name}}</a>
+							<a class="{% if current_url == main-item.link %} active {% endif %}" href="{{ site.baseurl }}/{{ main-item.link }}#section={{ main-item.name | slugify }}" data-proofer-ignore>{{ main-item.name}}</a>
```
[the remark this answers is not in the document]
trevor-circleci: The problem here is that each one of these would require a different `isActive` declaration, since the item being compared to is unique to the scope of each loop. 

I could still add it to improve the semantics, but it wouldn't be much of a space/flow improvement.

jekyll/_sass/_sidebar.scss
trevor-circleci: the problem is that there are lots of bespoke values in this design/layout; not being designed with a specific grid in place means spacing values aren't very clean/reusable. 

So I can create variables here, but they would be just to improve semantics, and I'm not sure how valuable that would be. For this instance, where we are using 5px a lot, I'll add a variable, though.

PR #2644, opened 2018-08-28
Remove (maybe replace?) `killall` step in example.

dav:
# Description

Removes part of the example circle.yml code for Sauce Labs config.

# Reasons

This command isn't actually available in the CircleCI environment. Not sure what should be done instead but the example doesn't actually work in the current system.

See screenshot: https://d.pr/i/xcYDjR


dav:
This is what I ended up using.
```
      - run:
          name: Shut Down Sauce Connect Tunnel
          command: |
            kill -9 `cat /tmp/sc_client.pid`
```

michelle-luna:
@dav Great, thanks for the follow-up, much appreciated!

PR #2652, opened 2018-08-29
[CIRCLE-13347] adding archive in preparation for supporting redirects of 1.0 pages

trevor-circleci:
# Description
A first step in the docs redirect work. See title. 
Leaving the 1.0 collection for now, but it will be removed once the redirects are in place.

# Reasons
https://circleci.atlassian.net/browse/CIRCLE-13347
Old doc pages will need to be accessible somewhere, even after being removed from the nav/redirected from.


michelle-luna:
Hi,
We really don't want these pages to be navigable, this product is EOL and we want all links and traffic to go to 2.0, these shouldn't be easy to find or use (they just have to exist).

Thanks!
Michelle

PR #2656, opened 2018-08-30
Update sign up copy

### LINKED ISSUE #2651: Left nav moves off screen when scrolling on sign up page
As soon as I scroll in Chrome on this page https://circleci.com/docs/2.0/first-steps/ the left nav flys out of view (up)

screen shot 2018-08-29 at 11 50 08 am

(Also the instructions say Sign Up button - but there is no signup button other than the link in the instructions)

smart-alek:
# Description
Removes reference to nonexistent signup button.

# Reasons
Fixes #2651


smart-alek:
I prefer the directness of "Go to". "Visit" sounds like a hotel booking. XD

michelle-luna:
If you haven't signed up before, there is a button, that is the reason for this language (it is for a person who hasn't signed up yet).

smart-alek:
@michelle-luna I'm not seeing the button you're talking about. When I open the linked page in an incognito window, this is what I see:

image

felicianotech:
It wouldn't be that page. Open the Doc in Incognito mode or any other Marketing site page.

smart-alek:
@felicianotech @michelle-luna I'm actually noticing a few more inconsistencies that are outside the scope of this PR:

- Neither of the buttons are "Start" buttons. Both begin with "Sign Up"
- Logging in does not change the behavior of this page, but the document mentions logging in by clicking the buttons.

Since this document is the first one in the Introduction, we should probably fix these sooner rather than later. For now, I'll just make tickets so we can capture the work.

michelle-luna:
The button appears when you are on the docs page, so when you are in the docs and you haven't signed up , there is a big green Sign Up button in the upper right. You don't need more tickets for this work. I agree with changing the button names to match whatever they are on the page.

PR #2671, opened 2018-09-04
Clarify machine executor section

### LINKED ISSUE #2604: Missing details of machine image
# Location
- https://circleci.com/docs/2.0/executor-types/#using-machine

# Explanation
- Does the "OS to use for builds" radio box under Build Settings - Build Environment also affect the machine executor or just docker images?
- What type of virtual machine is the machine executor?
- Custom images - not clear that you can't use these in the hosted environment.

Again, I'm happy to start a PR on this for feedback and review.

smart-alek:
# Description
Cleans up copy in the executor types document with minor edits and more examples.

# Reasons
Fixes #2604 and resolves [this JIRA issue](https://circleci.atlassian.net/browse/CIRCLE-13250).

### DIFF
--- a/jekyll/_cci2/executor-types.md
+++ b/jekyll/_cci2/executor-types.md
@@ -80,7 +80,8 @@ CPUs | Processor                 | RAM | HD
 2    | Intel(R) Xeon(R) @ 2.3GHz | 8GB | 100GB
 {: class="table table-striped"}
 
-**Note**: Use of `machine` may require additional fees in a future pricing update. 
+**Note**:
+Using `machine` may require additional fees in a future pricing update.
 
 To use the machine executor with the default `machine` image, set the `machine` key to `true` in `.circleci/config.yml`:
 
--- a/jekyll/_cci2/executor-types.md
+++ b/jekyll/_cci2/executor-types.md
@@ -72,7 +72,8 @@ More details on the Docker Executor are available in the [Configuring CircleCI](
 
 ## Using Machine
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/executor-types.md
```diff
@@ -64,43 +72,69 @@ More details on the Docker Executor are available in the [Configuring CircleCI](
...
-      image: circleci/classic:201708-01
-``` 
-
-Following are the available machine images:
+      image: circleci/classic:2017-01  # pins image to specific version using YYYY-MM format
+```
 
-* `circleci/classic:latest` is the default image. Changes to this will be announced at least one week before they go live.
-* `circleci/classic:edge` receives the latest updates and will be upgraded at short notice.
-* `circleci/classic:[year-month]` This lets you pin the image version to prevent breaking changes. Refer to the [Configuring CircleCI](https://circleci.com/docs/2.0/configuration-reference/#machine) document for versions.
```
michelle-luna: Not sure why we should remove link to the config ref here? Otherwise these minor changes are fine with me, thanks!

smart-alek: @michelle-luna good question. The configuration reference doesn't give any information about versions. If we're going to link to the Configuration Reference anywhere, I think it might make more sense at the top of this `machine` key section.

smart-alek:
@KyleTryon I am entirely happy to kill that word if it's not true or misleading. Feels like "Linux virtual machine" should work just fine.

PR #2693, opened 2018-09-12
Config reuse draft

michelle-luna:
# Description
Add 2.1 config bundling info to core docs

# Reasons
It is possible for anyone to use these keys, so we should have them in core docs now


jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
+---
+layout: classic-docs
+title: "Bundling Config"
```
ndintenfass: We should probably change this to "Reusable Config" or "Reusing Config" -- "Bundling" implies bundler for many people, but more importantly bundling is really an orbs thing rather than the commands/executors/parameters

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+order: 1
+---
+
+
+This document describes how to version your [.circleci/config.yml]({{ site.baseurl }}/2.0/configuration-reference/) file and get started with bundling config.
+
+* TOC
+{:toc}
+
+## Getting Started with Config Bundling
```
ndintenfass: This feels heavy -- we don't talk about how to start projects for other config features. Is the plan that this is the only place we'll talk about these new config elements? If they aren't visible on config reference they end up on an island that seems odd.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+{:.no_toc}
+
+1. Add your project on the Add Projects page if it is a new project. For an existing Project, go to the Project Settings and enable [Build Processing]({{ site.baseurl }}/2.0/build-processing/) with the radio button.
+
+2. Install the CircleCI-Public CLI by following the [Using the CircleCI CLI]({{ site.baseurl }}/2.0/local-cli/) documentation.
```
ndintenfass: This seems orthogonal to the reuse features -- I suggest we start encouraging anyone doing config to install the CLI to use `config validate`, but that's not unique to this particular set of features.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+
+## Built-in Commands
+
+CircleCI has several built-in commands available to all [circleci.com](http://circleci.com) customers and available by default in CircleCI server installations. Examples of built-in commands are:
+
+  * `checkout`
+  * `setup_remote_docker`
+  * `save_to_workspace`
+
+**Note:** Built-in commands are implicitly in the empty scope and are thus syntactically equivalent to primitives such as `run`). This _may_ change in future versions of configuration but is true to maintain compatibility with version `2` configuration.
```
ndintenfass: I think we can remove this -- it's not quite true. However, it IS possible to override the built-in commands, so we should probably say that here instead.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+          command: aws s3 sync << parameters.from >> << parameters.to >><<# parameters.overwrite >> --delete<</ parameters.overwrite >>"
+```
+--->
+
+Defining a command called "s3sync" is invoked in a 2.1 `.circleci/config.yml` file as:
+
+```yaml
+version: 2.1
+
+orbs:
```
ndintenfass: This example needs to be changed to not use orbs, yes?

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+  s3tools: circleci/s3@1
+
+workflows:
+  build-test-deploy:
+    jobs:
+      - deploy2s3:
+        steps:
+          - s3tools/s3sync:
```
ndintenfass: This example may be more complex than we want for commands -- this show using a command from an orb and passing it steps, but this section seems to be just about commands.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+              from: .
+              to: "s3://mybucket_uri"
+              overwrite: true
+```
+
+If the same command were declared locally inside `config.yml` it would look something like:
```
ndintenfass: I think we could start after this line and then delete the example above -- that way it's just about commands local to config.yml

jekyll/_cci2/bundling-config.md
eric-hu: I've found `circleci config process` to be very helpful in writing reusable config. It does what `circleci config validate` does, but also performs the substitution of references to commands, executors, and jobs.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+A command can have the following immediate children keys as a map:
+
+- **description:** (optional) A string that describes the purpose of the command, used for generating documentation.
+- **parameters:** (optional) A map of parameter keys, each of which adheres to the `parameter` spec.
+- **steps:** (required) A sequence of steps run inside the calling job of the command.
+
+### Defining Your Commands
+{:.no_toc}
+
+Commands are declared under the `commands` key of a `config.yml` or `orb.yml` file. For instance, to make a command called `sayhello` it might look like:
```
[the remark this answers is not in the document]
eric-hu: @michelle-luna FYI we did prototype with `orb.yml` at some point, but we ultimately dropped that from orb design.

jekyll/_cci2/bundling-config.md
eric-hu: I _think_ I agree with Nathan here, but I'd like to take a step back and make sure I understand the purpose of this doc.

The work on "Orbs" added 3 groups of concepts into config:

A. "inline" jobs, executors and commands (which can define and take parameters)
B. "inline" orbs
C. registered orbs

(A) is mostly about re-using config within the same config.yml. (C) is for re-using config across projects, potentially across teams and companies with little collaboration. (B) is mostly to help people get to (C), but both (B) and (C) bundle or package up config that you _could_ express only in terms of (A).

With that said, is this doc meant to cover (A), (B), and (C) altogether? I think it's possible to introduce this topic to users starting from (A) only. That would be simpler to read through and digest, but do let me know if that level of separation is out of scope.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+{:.no_toc}
+
+A parameter can have the following keys as immediate children:
+
+| Key Name    | Description                                                                                   | Default value |
+|-------------|-----------------------------------------------------------------------------------------------|---------------|
+| description | Optional. Used to generate documentation for your orb.                                        | N/A           |
+| type        | Required. Currently "string", "boolean", "enum", and "steps" are supported                           | N/A           |
+| default     | The default value for the parameter. If not present, the parameter is implied to be required. | N/A           |
+
```
ndintenfass: There's a nuance here for the `enum` type -- might be worth saying something like "Some parameter types take extra keys." or something like that.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+      destination:
+        description: destination directory
+        type: string
+        default: docs
+    steps:
+      - cp *.md << destination >>
+```
+
+Strings should be quoted if they would otherwise represent another type (such as boolean or number) or if they contain characters that have special meaning in yaml. In all other instances, quotes are optional.
+
```
ndintenfass: We should probably add here that "Empty strings are treated as a falsy value in evaluation of `when` clauses, and all other strings are treated as truthy"

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+* true: `y` `yes` `true` `on`
+* false: `n` `no` `false` `off`
+
+Capitalized and uppercase versions of the above values are also valid.
+
+#### Steps
+{:.no_toc}
+
+Steps are used when you have a job or command that needs to mix predefined and user-defined steps. When passed in to a command or job invocation, the steps passed as parameters are always defined as an array, even if only one step is provided.
```
ndintenfass: Is "array" the word we use elsewhere? In YAML-land I see "sequence" used. As long as this is consistent across docs I don't think it matters which we use here, though.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+  - run: make deps
+  - run: echo "The dependencies are installed"
+  - run: echo "And now I'm going to run the tests"
+  - run: make test
+```
+
+#### Enum Parameter
+{:.no_toc}
+
+Use the `enum` parameter to declare the target operating system for a binary.
```
ndintenfass: This is describing the particular example below -- somehow that example ended up as the description, but it's not accurate. Something more like: "Use the `enum` parameter type when you want to enforce that the value must be one from a specific set of string values"

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+
+Jobs are sets of steps and the environments they should be executed within.
+
+### Job Naming and Organization
+{:.no_toc}
+
+Jobs are defined in your build configuration. Job names are defined
+in a map under the `jobs` key in configuration.
+
+Like most elements, jobs can contain an optional but highly recommended `description`.
```
ndintenfass: This is a bit interesting to think through -- for orbs we definitely want people to put in descriptions, but for normal config it's not nearly so useful.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+  hello-orb: somenamespace/hello-orb@volatile
+workflows:
+  build:
+    jobs:
+      - hello-orb/sayhello:
+          saywhat: Everyone
+```
+--->
+
+### Parameter Scope
```
ndintenfass: This should probably get moved up to the parameters section

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+      # Uses explicitly defined "sayhello"
+      - saygoodbye:
+          requires:
+            - SayHelloChad
+```
+
+### Pre and Post Steps
+{:.no_toc}
+
+**Note:** The keys `pre-steps` and `post-steps` in jobs are available in configuration version 2.1 and later_
```
ndintenfass: The extra "_" at the end of this line may be a typo

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+
+Every job accepts two special arguments: `pre-steps` and `post-steps`.
```
ndintenfass: This should probably be "Every job invocation can optionally accept"

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+
+All jobs accept two special arguments of type `steps`: `pre-steps` and
+`post-steps`.
+
+If an orb user invokes a job with one or both of these arguments,
+the job will run the steps in `pre-steps` first, before any other steps run, and
+then it will run the steps in `post-steps` last, after any other steps run.
+
+Pre- and post- steps allow users to be execute steps in a given job's environment
+without modifying the orb. This is useful, for example, when a user imports a job
```
ndintenfass: Use of the word "orb" here isn't necessary -- replace with "job"

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+before job execution. Pre- and post- steps allow the user to make these additions
+without modifying the imported job.
+
+A `steps` parameter can be used for a similar purpose, passing steps into a job,
+but it requires that the job be modified with an execution site for the parameter.
+
+#### Example of Using Pre- and Post-Steps
+{:.no_toc}
+
+An orb `foo` might define a job:
```
ndintenfass: Remove "orb `foo`" here

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+    machine: true
+    steps:
+      - checkout
+      - run:
+          command: echo "building"
+      - run:
+          command: echo "testing"
+```
+
+Then an orb user could use the job as follows:
```
ndintenfass: "orb"

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+```yaml
+# config.yml
+version: 2.1
+orbs:
```
ndintenfass: I thought we were planning to not include orbs yet? If so, let's not have examples with them as it might confuse the concepts here.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+Conditional steps allow the definition of steps that only run if a condition is
+met. 
+
+<!---
+For example, an orb could define a command that runs a set of steps *if* the
+orb's user invokes it with `myorb/foo: { dostuff: true }`, but not
+`myorb/foo: { dostuff: false }`.
+-->
+
+These conditions are checked before a workflow is actually run. That
```
ndintenfass: For now we could go further and say that "conditional steps currently only use parameter values as inputs"

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+          condition: << parameters.preinstall-foo >>
+          steps:
+            - run: echo "don't preinstall"
+```
+
+```yaml
+# inside config.yml
+version: 2.1
+
+orbs:
```
ndintenfass: see above comment about using orbs in these examples

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+- `resource_class`
+
+A simple example of using an executor:
+
+```yaml
+version: 2.1
+executors:
+  my-executor:
+    docker:
+      - image: circleci/ruby:volatile
```
ndintenfass: change "volatile" to "2.5.1-node-browsers" here -- volatile is our convention in orbs but not in our docker images

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+  test:
+    executor: lein_exec
+    environment:
+      TESTS: unit
+    steps:
+      - checkout
+      - run: echo "how are ya?"
+```
+
+You can also refer to executors from other orbs. Users of an orb can invoke its executors. For example, `foo-orb` could define the `bar` executor:
```
ndintenfass: "orbs"

ndintenfass: (this applies to the next few snippets too -- probably doesn't belong here)

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+
+1. Add your project on the Add Projects page if it is a new project. For an existing Project, go to the Project Settings and enable [Build Processing]({{ site.baseurl }}/2.0/build-processing/) with the radio button.
+
+2. Install the CircleCI-Public CLI by following the [Using the CircleCI CLI]({{ site.baseurl }}/2.0/local-cli/) documentation.
+
+3. Change the `version` key to 2.1 in your `.circleci/config.yml` file and commit the changes to test your build. Ensure that your project build succeeds with the new processing before adding any new 2.1 keys to your config.
+
+4. Run builds with your new configuration by pushing to your GitHub or Bitbucket repo that has been added in a project in CircleCI. You will trigger builds as before, but if you look at the configuration on your Jobs page you will see that it has been run using the new processing service. 
+
+After your build is running successfully with build processing enabled `.circleci/config.yml` file version 2.1, it is possible to add new keys to bundle config stanzas and run the same job more than once with different parameters (re-use jobs).
```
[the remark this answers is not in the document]
michelle-luna: thanks, fixed

jekyll/_cci2/bundling-config.md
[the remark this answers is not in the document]
michelle-luna: thanks, removed.

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, removed and replaced with 'it is possible to override built-in commands with a custom command.

jekyll/_cci2/bundling-config.md
michelle-luna: right, thanks, I moved the comment and update the intro to the inline command, please check it again

jekyll/_cci2/bundling-config.md
michelle-luna: yep, removed

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, fixed.

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, I added (takes extra keys) after enum

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+    parameters:
+      destination:
+        description: destination directory
+        type: string
+        default: docs
+    steps:
+      - cp *.md << destination >>
+```
+
+Strings should be quoted if they would otherwise represent another type (such as boolean or number) or if they contain characters that have special meaning in yaml. In all other instances, quotes are optional.
```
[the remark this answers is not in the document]
michelle-luna: thanks, added 'particularly for the colon character' after yaml.

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, added

jekyll/_cci2/bundling-config.md
michelle-luna: I don't think we use either, actually, so we can decide here and now (for now I will leave it as-is)

jekyll/_cci2/bundling-config.md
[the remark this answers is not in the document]
michelle-luna: fixed, thanks

michelle-luna: added, thanks

jekyll/_cci2/bundling-config.md
michelle-luna: moved, thanks

jekyll/_cci2/bundling-config.md
michelle-luna: fixed, thanks.

jekyll/_cci2/bundling-config.md
michelle-luna: changed, thanks

jekyll/_cci2/bundling-config.md
michelle-luna: replaced, thanks

jekyll/_cci2/bundling-config.md
michelle-luna: removed

jekyll/_cci2/bundling-config.md
michelle-luna: removed

jekyll/_cci2/bundling-config.md
ndintenfass: If we're choosing I think I'd pick sequence because that's how the YAML spec talks about it.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+  * `checkout`
+  * `setup_remote_docker`
+  * `save_to_workspace`
+
+**Note:** It is possible to override the built-in commands with a custom command.
+
+## Examples
+
+<!---
+The following is a an example of part of an "s3tools" orb defining a command called "s3sync":
```
eric-hu: Lines 102-144 can be deleted. 102-144 is an example of defining a command within an orb, while 145-174 is an example of defining a command inline in a config file. They define the same command, but the first way is out of scope for this document.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+      - deploy2s3:
+        steps:
+          - s3tools/s3sync:
+              from: .
+              to: "s3://mybucket_uri"
+              overwrite: true
+```
+--->
+
+Defining a command called `s3sync` locally inside `config.yml` it would look something like:
```
eric-hu: I think this section can be moved down to the `Using the 'parameters' Declaration` section and replace the "foo" command example. Both examples use parameters and they're right next to each other. However, the s3sync example is closer to a real-world usage.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+    parameters:
+      destination:
+        description: destination directory
+        type: string
+        default: docs
+    steps:
+      - cp *.md << destination >>
+```
+
+Strings should be quoted if they would otherwise represent another type (such as boolean or number) or if they contain characters that have special meaning in yaml, particularly for the colon character. In all other instances, quotes are optional. Empty strings are treated as a falsy value in evaluation of `when` clauses, and all other strings are treated as truthy.
```
eric-hu: `and all other strings are treated as truthy.` seems misleading. I think this is true, but some unquoted strings can be coerced into booleans, like `yes`, `on`, and `off`. I think this phrase should be removed.

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+  - run: make deps
+  - run: echo "The dependencies are installed"
+  - run: echo "And now I'm going to run the tests"
+  - run: make test
+```
+
+#### Enum Parameter
+{:.no_toc}
+
+The `enum` parameter may be a lits of any values. Use the `enum` parameter type when you want to enforce that the value must be one from a specific set of string values. The following example uses the `enum` parameter to declare the target operating system for a binary.
```
eric-hu: lits => list

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+      - run: echo "Hello << parameters.saywhat >>"
+
+workflows:
+  build:
+    jobs:
+      - sayhello:
+          saywhat: Everyone
+```
+
+<!---
```
eric-hu: This `Jobs defined in an orb` section can be removed, since the focus of this doc is on reusable config excluding orbs.

jekyll/_cci2/bundling-config.md
eric-hu: To make this example work without orbs:
- the above and below yaml blocks can be merged into 1
  - the comment `Contents of the orb myorb in namespace mynamespace` can be deleted
- under `workflow:`, `myorb/myjob` => `myjob`
- the `orbs` section can be deleted

jekyll/_cci2/bundling-config.md
ndintenfass: We don't allow unquoted booleans as the values of a `string` type (which we should probably mention) -- in `when` and `unless` we count all non-empty strings as truthy, including the string "false" and "no" and "off"

eric-hu: Ah that's an interesting subtlety. In that case, my original comment in this thread can be disregarded. I agree with Nathan's point: using an unquoted string value that YAML interprets as a boolean will result in a type error.

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, changed to sequence in both places

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, added.

jekyll/_cci2/bundling-config.md
michelle-luna: cool, thanks I merged these

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, fixed

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, commented these out

jekyll/_cci2/bundling-config.md
```diff
@@ -0,0 +1,905 @@
...
+   shell: /bin/bash            # From executor
+   working_directory: ~/tests  # From job
+```
+
+### Using Parameters in Executors
+{:.no_toc}
+
+If you'd like to use parameters in executors, define the parameters under the given executor. When you invoke the executor, pass the keys of the parameters as a map of keys under the `executor:` declaration, each of which has the value of the parameter that you would like to pass in.
+
+Parameters in executors can be of the type `string` or `boolean`. Default values can be provided with the optional `default` key.
```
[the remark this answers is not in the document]
michelle-luna: thanks, added

jekyll/_cci2/bundling-config.md
michelle-luna: yes, these are commented out in the markdown, they won't appear in the doc (and should not show up in the preview, let me know if you can see them)

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, replaced foo example with s3sync example

jekyll/_cci2/bundling-config.md
michelle-luna: thanks, I added this last sentence to the paragraph.

jekyll/_cci2/bundling-config.md
michelle-luna: thanks fixed

jekyll/_cci2/bundling-config.md
michelle-luna: yep, this is commented out

PR #2703, opened 2018-09-14
Note that this is for CircleCI 1.0 with config.yml

amcvitty:
I got weird error messages while trying to follow these instructions, and realised its because it's expecting a 1.0 config file. 

So to avoid others hitting the same issue, I'm suggesting to highlight it. 

(as an aside - our projects never ran in 1.0 and all the "upgrade now" stuff is more confusing than helpful for me)

### DIFF
--- a/jekyll/_cci2/examples.md
+++ b/jekyll/_cci2/examples.md
@@ -27,6 +27,8 @@ curl --user ${CIRCLE_TOKEN}: \
         https://circleci.com/api/v1.1/project/<source, eg. github>/<user name>/<project name>/tree/<branch name>
 ```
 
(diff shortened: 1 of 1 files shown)


michelle-luna:
Hi,

Thanks for this feedback, config.yml is 2.0 (not 1.0--1.0 uses circle.yml) so I wonder what weird errors you ran into when you tried this? This content is based off a video created by our support team that you can watch here if you are interested: https://www.youtube.com/watch?v=HB5DehCufG0

amcvitty:
I'm not sure what was going on, I'm clearly mistaken. Sorry!

PR #2706, opened 2018-09-14
Filters are not allowed in job definitions in 2.1

cayennes:
# Description
Call out 2.1 config along with workflows as an exception to being able to filter jobs in the job definition as opposed to the job within the workflow definition

# Reasons
2.1 config syntax doesn't support this because everything is considered to be a workflow even if there is not one explicitly defined.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -63,7 +63,7 @@ steps | Y | List | A list of [steps](#steps) to be performed
 working_directory | N | String | In which directory to run the steps. Default: `~/project` (where `project` is a literal string, not the name of your specific project). Processes run during the job can use the `$CIRCLE_WORKING_DIRECTORY` environment variable to refer to this directory. NOTE: Paths written in your YAML configuration file will _not_ be expanded; if your `store_test_results.path` is `$CIRCLE_WORKING_DIRECTORY/tests`, then CircleCI will attempt to store the `test` subdirectory of the directory literally named `$CIRCLE_WORKING_DIRECTORY`, dollar sign `$` and all.
 parallelism | N | Integer | Number of parallel instances of this job to run (default: 1)
 environment | N | Map | A map of environment variable names and values.
-branches | N | Map | A map defining rules for whitelisting/blacklisting execution of specific branches for a single job that is **not** in a workflow (default: all whitelisted). See [Workflows](#workflows) for configuring branch execution for jobs in a workflow.
+branches | N | Map | A map defining rules for whitelisting/blacklisting execution of specific branches for a single job that is **not** in a workflow or a 2.1 config (default: all whitelisted). See [Workflows](#workflows) for configuring branch execution for jobs in a workflow or 2.1 config.
 resource_class | N | String | Amount of CPU and RAM allocated to each container in a job. (Only available with the `docker` executor) **Note:** A paid account is required to access this feature. Customers on paid plans can request access by [opening a support ticket](https://support.circleci.com/hc/en-us/requests/new).
 {: class="table table-striped"}
 
@@ -269,7 +269,7 @@ jobs:
 
 #### **`branches`**
 
-Defines rules for whitelisting/blacklisting execution of some branches if Workflows are **not** configured. If you are using Workflows, job-level branches will be ignored and must be configured in the Workflows section of your `config.yml` file. See the [workflows](#workflows) section for details. The job-level `branch` key takes a map:
+Defines rules for whitelisting/blacklisting execution of some branches if Workflows are **not** configured and you are using 2.0 (not 2.1) config. If you are using Workflows, job-level branches will be ignored and must be configured in the Workflows section of your `config.yml` file. If you are using 2.1 config, you will need to add a workflow in order to use filtering. See the [workflows](#workflows) section for details. The job-level `branch` key takes a map:
 
 Key | Required | Type | Description
 ----|-----------|------|------------


michelle-luna:
@eric-hu please have a look at these changes for 2.1 I added in and totally made up the new when section. thanks!

jekyll/_cci2/configuration-reference.md
```diff
@@ -27,11 +27,68 @@ You can see a complete `config.yml` in our [full example](#full-example).
...
+Executors define the environment in which the steps of a job will be run, allowing you to reuse a single executor definition across multiple jobs.
+
+Key | Required | Type | Description
+----|-----------|------|------------
+docker | Y <sup>(1)</sup> | List | Options for [docker executor](#docker)
+machine | Y <sup>(1)</sup> | Map | Options for [machine executor](#machine)
+macos | Y <sup>(1)</sup> | Map | Options for [macOS executor](#macos)
+shell | N | String | Shell to use for execution command in all steps. Can be overridden by `shell` in each step (default: See [Default Shell Options](#default-shell-options))
+working_directory | N | String | In which directory to run the steps.
+resource_class | N | String | Amount of CPU and RAM allocated to each container in a job. (Only available with the `docker` executor) **Note:** A paid account is required to access this feature. Customers on paid plans can request access by [opening a support ticket](https://support.circleci.com/hc/en-us/requests/new).
```
eric-hu: This table is missing `environment` as a configurable field on `executors`.

jekyll/_cci2/configuration-reference.md
```diff
@@ -517,6 +574,16 @@ steps:
       when: on_fail
 ```
 
+##### **`when`** (requires version: 2.1)
```
eric-hu: @michelle-luna Can you remove this section? I think it's out of scope for Cayenne's pull request, and needs a more explanation (a `when` step can go in commands and jobs) and a filled out example.  Feel free to open this as another PR and I'll work with you to create examples and explanation.

cayennes: I'm not sure if it was the right thing to do, but I did move that card to done without waiting for this to be merged.

I think it does make sense to document all the 2.1 stuff at once.

cayennes: Thanks both of you for working on this!

eric-hu: @michelle-luna in that case, you can disregard my earlier comment about removing this section from this PR. This section still doesn't quite fit under steps/when. I'll follow up this comment with a couple of examples.

eric-hu: @michelle-luna 

Here's an example of using "when" and "unless" in a job:

```yaml
$ circleci config process asdf.yml
version: 2.1
jobs:
  job_with_optional_custom_checkout-1:
    machine: true
    steps:
    - run:
        command: echo "my custom checkout"
  job_with_optional_custom_checkout-2:
    machine: true
    steps:
    - checkout
workflows:
  build-test-deploy:
    jobs:
    - job_with_optional_custom_checkout-1
    - job_with_optional_custom_checkout-2
  version: 2

# Original config.yml file:
# version: 2.1
#
# jobs:
#   job_with_optional_custom_checkout:
#     parameters:
#       custom_checkout:
#         type: string
#         default: \"\"
#     machine: true
#     steps:
#       - when:
#           condition: <<parameters.custom_checkout>>
#           steps:
#             - run: echo \"my custom checkout\"
#       - unless:
#           condition: <<parameters.custom_checkout>>
#           steps:
#             - checkout
#
# version: 2.1
# workflows:
#   build-test-deploy:
#     jobs:
#       - job_with_optional_custom_checkout:
#           custom_checkout: \"any non-empty string is truthy\"
#       - job_with_optional_custom_checkout
```

And an example of using `when` in a command:

```yaml
$ circleci config process asdf.yml
version: 2.1
jobs:
  simple_job:
    machine: true
    steps:
    - checkout
    - run:
        command: echo "my custom checkout"
workflows:
  build-test-deploy:
    jobs:
    - simple_job
  version: 2

# Original config.yml file:
# version: 2.1
#
# commands:
#   my_custom_checkout:
#     parameters:
#       custom_checkout:
#         type: string
#         default: \"\"
#     steps:
#       - when:
#           condition: <<parameters.custom_checkout>>
#           steps:
#             - run: echo \"my custom checkout\"
#       - unless:
#           condition: <<parameters.custom_checkout>>
#           steps:
#             - checkout
#
#
# jobs:
#   simple_job:
#     machine: true
#     steps:
#       - my_custom_checkout
#       - my_custom_checkout:
#           custom_checkout: \"a non-empty string\"
#
# version: 2.1
# workflows:
#   build-test-deploy:
#     jobs:
#       - simple_job
```

This should be in a separate section from the existing `when` that goes inside of `run`.  They take different arguments, are used for different purposes, and users will have to think about data at different points in their build's lifecycle: 
- customizing a step to run on fail, on pass, or always (determined while the build is running)
- customizing commands and job configuration to run on custom conditions (determined at config-compile time)

 The shape is identical between commands and jobs, so their documentation can be shared.

jekyll/_cci2/configuration-reference.md
```diff
@@ -517,6 +574,16 @@ steps:
...
+
+A conditional step consists of a step with the key `when` or `unless`. Under the `when` key are the subkeys `condition` and `steps`. See the Conditional Steps section of the Reusing Config document for details.
+
+Key | Required | Type | Description
+----|-----------|------|------------
+condition | Y | String | A parameter value
+<step_type> |	Y |	Map or String |	A configuration map for the step or some string whose semantics are defined by the step.
```
eric-hu: `<step_type>` should be `steps`

jekyll/_cci2/configuration-reference.md
```diff
@@ -27,11 +27,68 @@ You can see a complete `config.yml` in our [full example](#full-example).
 
 Key | Required | Type | Description
 ----|-----------|------|------------
-version | Y | String | Should currently be `2`
+version | Y | String | `2` or `2.1` See the Reusing Config doc for an overview of new 2.1 keys available to simplify your `.circleci/config.yml` file, reuse, and paramaterize jobs.
```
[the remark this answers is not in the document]
michelle-luna: thanks, fixed

jekyll/_cci2/configuration-reference.md
michelle-luna: thanks! added

jekyll/_cci2/configuration-reference.md
michelle-luna: Thanks, it is in a different section from the when attribute. I qualified the name of this section as 'The `when` Step' to be more clear. I also added one example (this section will point to the complete section in the Reusing Config document for the additional information about conditional steps.

jekyll/_cci2/configuration-reference.md
michelle-luna: fixed, thanks!

jekyll/_cci2/configuration-reference.md
eric-hu: Just checked this section out in the docs. Looks good 👍

michelle-luna: thanks @eric-hu! I will add in the live links today in conjunction with doing the changelog

PR #2719, opened 2018-09-20
Circle 13442 add webpack build

JuanCaicedo:
# Description
- Add a basic webpack build so that we can include other modules
- Create html snippet to load built js file
- Update circle config so that new assets are persisted

# Reasons
- We want to have more visibility into errors thrown from this part of the site

# Considerations 
- ~We considered the fact that browsers will cache files by name. This would cause it so that the users receive a cached version of our js files, meaning they could be running outdated code~.
    - ~To get around that, we wanted to have a file hash in the url~
    - ~It seemed difficult to get Jekyll to include a dynamic file name~
    - ~An html snippet produced by Webpack, to be included by Jekyll, seemed like the best path forward 😄~
- We found out that Jekyll does filename hashing in order to avoid this problem, so we don't need to further solve it on our end.

# How to test 
- `$ npm run webpack-dev`
- Start server
- Go to <http://localhost:4000/docs/>
- Open up the console
- run `Cookies`, it should be defined 👍

# Follow ups
- This does not include webpack in our development build process. And immediate follow up would be to set up sometime of task automation (rake or npm) that runs webpack in watch mode and that will get kicked off as part of running the server


.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "useBuiltIns": "entry"
```
JuanCaicedo: I'm not really sure what we should use here in babelrc, I think I just took an example from their site

jekyll/_layouts/classic-docs.html
```diff
@@ -28,6 +28,7 @@
 	</div><!-- .main-body -->
 	{% include global-footer.html %}
 	<script src="{{ "/assets/js/vendor.min.js" | prepend: site.baseurl }}" type="text/javascript"></script>
+  {% include app.html %}
```
JuanCaicedo: I'm not sure in what other files we need to include this. Maybe @trevor-circleci  would know?

package.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "This is the public repository for <https://circleci.com/docs/>, a static website generated by [Jekyll](https://jekyllrb.com/). If you find any errors in our docs or have suggestions, please follow our [Contributing Guide](CONTRIBUTING.md) to submit an issue or pull request.",
+  "main": "index.js",
+  "scripts": {
+    "webpack-prod": "NODE_ENV=production webpack -p"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/circleci/circleci-docs.git"
+  },
+  "license": "EPL-1.0",
```
JuanCaicedo: Is this correct? I just copied this line from somewhere 🙈

JuanCaicedo:
I think @erichaberkorn and @trevor-circleci might have some helpful feedback here 😀

package.json
erichaberkorn: I believe this is supposed to be [MIT](https://github.com/circleci/circleci-docs/blob/master/LICENSE)

jekyll/_layouts/classic-docs.html
trevor-circleci: @JuanCaicedo What's happening here? Where is app.html coming from? 

Note: this may be fixed in later commits.

webpack.config.js
```diff
@@ -1,18 +1,28 @@
...
     app: './src-js/app.js'
   },
   output: {
-    path: path.join(__dirname, 'jekyll/assets/js/'),
-    filename: '[name].min.js'
+    // We need the flexibility to build a js file to one place and an html file
+    // in another, which are in different places in the jekyll directory tree.
+    // We set a fairly high level path and empty default path so that we can get
+    // this flexibility.
+    path: path.join(__dirname, 'jekyll/'),
```
trevor-circleci: I'm worried by this. Whenever I've used webpack, it was much more specific and controlled. I think it would be better to have a specific folder to house our assets and then clean up the site over time so that the js assets all live there.

webpack.config.js
```diff
@@ -1,18 +1,28 @@
...
   plugins: [
-    new webpack.ProvidePlugin({
-      React: 'react',
-      ReactDOM: 'react-dom'
+    new HtmlWebpackPlugin({
+      filename: '_includes/app.html',
+      templateContent: ({ htmlWebpackPlugin }) =>
+        `<script src="{{ "/${
+          htmlWebpackPlugin.files.chunks.app.entry
+        }" | prepend: site.baseurl }}" type="text/javascript"></script>`,
```
trevor-circleci: not sure what's happening here or why it's necessary. 

Looks like we're iterating over plugin files to include as JS. Is this the standard way to do things? Why have this template at all? We should just be including the generated webpack JS once in the default template, and including all other js files into the webpack... unless i'm missing something of course.

jekyll/_layouts/classic-docs.html
trevor-circleci: I'm not sure where app.html is coming from. I don't see it in the includes folder. How did you generate this?

JuanCaicedo:
@trevor-circleci and I discussed this in pair, which is why he approved it ✅

A couple follow ups from our conversation:
  - This PR could have been split into two parts: one to setup the webpack build and one to be sure the webpack build gets deployed through our circleci config

webpack.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   output: {
     path: path.join(__dirname, 'jekyll/assets/js'),
     publicPath: '',
-    filename: 'assets/js/dist/[name]-[contenthash].min.js'
```
trevor-circleci: just a note, if we find it too difficult to configure hashing with jekyll plugins, we still might want to use this bit. But that's an aside, just in case, for later.

PR #2720, opened 2018-09-20
[CIRCLE-10363] Implement search

trevor-circleci:
# Description
Swapping out old algolia search for new instantsearch.

# Reasons
Old search was inadequate. 

# Notes
In the future, for more custom functionality, we may want to consider switching to another search framework provided by algolia (it will be more work to implement, but would provide much more functionality/flexibility): https://community.algolia.com/algoliasearch-helper-js/


erichaberkorn:
* It might be nice to provide some sort of indicator to the user of where the end of the search box is
image
At first I thought there was a bug when the clear button showed up in the middle of the screen
image

* On mobile "No results" shows up at the edge of the screen without padding 
image
I think it might look better on both desktop and mobile if "No results" was displayed directly under "Search results for..." like when actual search results exist
image

image

* When there is a search in the query params like http://localhost:4000/docs/?query=using%20workflows refreshing the page flashes the docs home page and then the instantsearch results show up shortly after. Is it possible to show a loading indicator instead of the home page while this is happening?

jekyll/assets/js/instantSearch.js
```diff
@@ -6,4 +6,66 @@ var search = instantsearch({
   routing: true
 });
 
+// initialize SearchBox
+search.addWidget(
+  instantsearch.widgets.searchBox({
```
erichaberkorn: I think we should consider adding the `queryHook` option to here and [here](https://github.com/circleci/circleci-docs/pull/2720/files#diff-50bd272b9e398cba0799cac3f16c7b3dR22) so we can throttle the results. It looks like a request is being made to algolia on every keypress.

jekyll/assets/js/instantSearch.js
```diff
@@ -6,4 +6,66 @@ var search = instantsearch({
...
+      item: '<a href="/docs{{ url }}"><div class="result-item-wrap"><strong>{{ title }}</strong><br><p>{{ content }}</p></div>'
+    }
+  })
+);
+
 search.start();
+
+(function () {
+  // insert search results
+  var searchResetButton = document.querySelector("#search-box .ais-search-box--reset");
```
erichaberkorn: Yay for no jQuery

erichaberkorn:
This is looking very good. The changes I mentioned are mostly nice to haves and I already think searching is better on this branch than master. If time is a factor, the only change I really think we need to make before merging this is the throttling one. Let me know if you would like to pair on any of these issues.

jekyll/assets/js/instantSearch.js
trevor-circleci: Actually, for the instant search results loading with each keystroke, I think the current behavior is preferable. If it becomes a performance issue, maybe we can revisit this?

erichaberkorn: I'm worried that we are going to run into this if we have a few people using this at the same time: https://community.algolia.com/places/pricing.html#rate-limits

It's also considered best practices to throttle requests like this rather than making them on every keystroke.

PR #2729, opened 2018-09-21
Note around DLC usage limits

appplemac:
# Description

We want to avoid confusion in case a customer hits DLC usage limits. So adding a note that states that there _is_ a limit.

### DIFF
--- a/jekyll/_cci2/docker-layer-caching.md
+++ b/jekyll/_cci2/docker-layer-caching.md
@@ -44,6 +44,8 @@ jobs:
      - run: docker build .
 ``` 
 
+**Note:** We enforce limits for the DLC usage to guarantee the stability of our platform. 
+
 ## Enabling DLC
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/docker-layer-caching.md
```diff
@@ -44,6 +44,18 @@ jobs:
...
+
+One DLC volume can only be attached to one `machine` or Remote Docker job at a time. If one DLC volume exists but two jobs that request DLC are launched, CircleCI will create a new DLC volume and attach it to the second job. From that point on the project will have two DLC volumes associated with it. This applies to parallel jobs as well: if two `machine` jobs are run in parallel, they will get different DLC volumes.
+
+Depending on which jobs the volumes are used in, they might end up with different layers saved on them. The volumes that are used less frequently might have older layers saved on them.
+
+The DLC volumes are deleted after 14 days of not been used in a job.
+
+CircleCI will create a maximum of 50 DLC volumes per project, so a maximum of 50 concurrent `machine` or Remote Docker jobs per project can have access to DLC. This takes into account the parallelism of the jobs.
```
DanToml: Maybe add an explicit example of e.g 25 jobs with parallel:2?

PR #2740, opened 2018-09-24
fix documentation for conditional steps

johnswanson:
The `steps` key in a conditional step is just a list of steps--even more
accurate than "Sequence" would be "List of Steps", but I figured I'd
keep things consistent with e.g. the docs for 2.1 `commands`, where
`steps` is described in a similar way, and given the type `Sequence`.


eric-hu:
LGTM

PR #2744, opened 2018-09-24
Circle 13442 add rollbar

JuanCaicedo:
# What
- Load Rollbar through html snippet on all pages coming from this repo
- Add Rollbar env into jekyll config
- Include and init Rollbar 
 
# Rationale
- We want to gain more visibility into errors thrown on the docs site

# Considerations
- This went through a lot of change following https://github.com/circleci/circleci.com/pull/1865. Ultimately we would like to load Rollbar through webpack like our other configs, but this doesn't seem like it will work right now

# How to test
- Open <https://rollbar.com/circle/circleci-docs/>. This is where new errors will end up.
- `$ npm install`
- `$ npm run webpack-watch` (Note that after making js changes, you have to wait for jekyll to pick up those changes and rebuild, around 20s)
- `$ cd jekyll && jekyll serve -Iw`
- Go to <http://localhost:4000/docs/>
- Open the console and type in 
    ```
    window.onerror("Testing circleci.com rollbar", window.location.href)
    ```
- You should see one an error show up in their ui
- Make the following changes in the src code
    ```diff
    diff --git a/jekyll/assets/js/site.min.js b/jekyll/assets/js/site.min.js
    index f233e25f..4936c1b7 100644
    --- a/jekyll/assets/js/site.min.js
    +++ b/jekyll/assets/js/site.min.js
    @@ -22,4 +22,11 @@
      };
      var DAYS_PER_MINUTE = 1 / 24 / 60;
[...]

PR #2745, opened 2018-09-24
moving instantsearch into webpack; adding lodash; filtering searches …

trevor-circleci:
…through lodash throttling

# Description
see title.

# Reasons
Concern over query limits being reached with instantsearch/algolia when sending a search query with each keystroke, led to a decision to throttle search results using lodash..


src-js/app.js
```diff
@@ -1,3 +1,77 @@
 import * as Cookie from 'js-cookie';
+import * as lodash from 'lodash.throttle';
```
JuanCaicedo: Could we rename this to `throttle` since that's all we're using here?
lodash is a pretty popular library, so I find it confusing to see `lodash(function (query, searchFunction) { searchFunction(query); }, 1000)`

src-js/app.js
```diff
@@ -1,3 +1,77 @@
...
 
 window.Cookie = Cookie;
+
+// Instant search initialization
```
JuanCaicedo: Could we put the instantsearch stuff into a separate file and import it in here? I think https://github.com/circleci/circleci-docs/pull/2744 shows a good path forward for doing that

src-js/app.js
```diff
@@ -1,3 +1,77 @@
...
+
+// initialize SearchBox
+search.addWidget(
+  instantsearch.widgets.searchBox({
+    container: '#search-box',
+    cssClasses: {
+      input: 'instantsearch-search'
+    },
+    placeholder: 'Search Documentation',
+    queryHook: lodash(function (query, searchFunction) { searchFunction(query); }, 1000)
```
JuanCaicedo: I think we can now use arrow functions since this file will be transpiled, so this could be written as 
```js
  queryHook: lodash((query, searchFunction) => searchFunction(query), 1000)
```

JuanCaicedo: It's probably worth documenting that we're adding this only as a way to do throttling 👍

src-js/app.js
```diff
@@ -1,3 +1,77 @@
...
+    templates: {
+      empty: 'No results',
+      item: '<a href="/docs{{ url }}"><div class="result-item-wrap"><strong>{{ title }}</strong><br><p>{{ content }}</p></div>'
+    }
+  })
+);
+
+search.start();
+
+(function () {
```
JuanCaicedo: Why is this in an IIFE? I could be wrong here or be missing something about their lib, the impression I had is that IIFEs are good for making containing values so they don't get leaked onto the global scope. I think that's something we don't need to worry about since Webpack will put all of our code into modules. FWIW, I think it does that with IIFEs 😄

src-js/app.js
```diff
@@ -1,3 +1,77 @@
...
+  // insert search results
```
JuanCaicedo: I'm a little confused about what's going on in this part of the code 🙈 Maybe we can pair on it

src-js/app.js
```diff
@@ -1,3 +1,77 @@
...
+  var searchResetButton = document.querySelector("#search-box .ais-search-box--reset");
+  var searchBox = document.querySelector("input.instantsearch-search");
+  var template = document.querySelector("#hits-template");
+  var resultContainer = document.querySelector('.main-body');
```
JuanCaicedo: I think this one is unused?

src-js/app.js
trevor-circleci: sure

src-js/app.js
trevor-circleci: sure

jekyll/_config.yml
```diff
@@ -14,7 +14,7 @@ plugins:
   - jekyll-sitemap
 
 algolia:
-  application_id: 06F2EYGKB6
+  application_id: U0RXNGRK45
```
Peeja: Is there an account name of some sort we can mention in a comment here to keep track?

src-js/instantsearch.js
```diff
@@ -0,0 +1,76 @@
...
+
+  // initialize SearchBox
+  search.addWidget(
+    instantsearch.widgets.searchBox({
+      container: '#search-box',
+      cssClasses: {
+        input: 'instantsearch-search'
+      },
+      placeholder: 'Search Documentation',
+      queryHook: throttle(function (query, searchFunction) { searchFunction(query); }, 1000) // method to throttle search requests
```
Peeja: We want to `debounce` not `throttle`, no? `throttle` should mean that the, if the user types too fast, the final search will be dropped. `debounce` means that the early searches will be dropped, and searching will wait until the input settles.

src-js/instantsearch.js
```diff
@@ -0,0 +1,76 @@
...
+  // initialize SearchBox
+  search.addWidget(
+    instantsearch.widgets.searchBox({
+      container: '#mobile-search-box',
+      cssClasses: {
+        input: 'instantsearch-search'
+      },
+      placeholder: 'Search Documentation'
+    })
+  );
```
Peeja: Do we need to have two separate `searchBox`es?

src-js/instantsearch.js
JuanCaicedo: After reading [this article](https://css-tricks.com/the-difference-between-throttling-and-debouncing/), I think @Peeja is right on this 😄

jekyll/_config.yml
trevor-circleci: not exactly; multiple accounts can access the same information. I don't think a single account makes sense in that case, does it?

src-js/instantsearch.js
trevor-circleci: Yep, looks like it.

src-js/instantsearch.js
trevor-circleci: yeah, due to positioning in the dom, I think it's the simplest solution -- but let me know if you want to sync, or think there's a better way.

trevor-circleci:
@Peeja Did this last PR address your concerns?

src-js/instantsearch.js
JuanCaicedo: @trevor-circleci Could you explain a little more what you mean by positioning in the dom?

This is what I'm seeing at <http://localhost:4000/docs>, and they look like they're directly adjacent siblings 😀

image

jekyll/assets/css/footer.css
```diff
@@ -2,6 +2,8 @@
  * Footer
  * */
 .footer {
+  position: relative;
```
JuanCaicedo: Are there other styles that are making us require this? I tried removing it and didn't see many changes 😄 From what I read on [css tricks](https://css-tricks.com/almanac/properties/p/position/), I _think_ `relative` is the same as the default `static`, except it lets you do positioning with `top`/`left` etc.

In my experience, I've only used `position: relative` when I'm also messing around with `position: absolute`

JuanCaicedo: Are there other styles that are making us require this? I tried removing it and didn't see many changes 😄 From what I read on [css tricks](https://css-tricks.com/almanac/properties/p/position/), I _think_ `relative` is the same as the default `static`, except it lets you do positioning with `top`/`left` etc.

In my experience, I've only used `position: relative` when I'm also messing around with `position: absolute`

jekyll/_sass/_instantsearch.scss
```diff
@@ -0,0 +1,74 @@
+.base-input-style {
+  border-radius: 0;
+  padding: 1.25em 1em 1em 2.8em;
+  width: 100%;
+  border: none;
```
JuanCaicedo: Does this get a border by default, maybe from their code? I'm a little confused to see this and `border-radius: 0`, since those are the defaults 😆

jekyll/_sass/_instantsearch.scss
```diff
@@ -0,0 +1,74 @@
...
+  border-bottom: 1px solid $gray-200;
+  position: fixed;
+  width: 100%;
+  background: white;
+  z-index: 2;
+
+  #search-box {
+    max-width: 1200px;
+    margin: 0 auto;
+    display: block;
```
JuanCaicedo: I think this is the default?

src-js/instantsearch.js
JuanCaicedo: @trevor-circleci Could you explain a little more what you mean by positioning in the dom?

From inspecting those elements, all I could see is that we have some styles on `.main-searchbar` that we don't want to be nested under?

JuanCaicedo: I'm concerned about this as well 😅 For one, I think we're rendering and instantiating two separate searchboxes. This is apparent if you apply this diff
```diff
diff --git a/jekyll/_sass/_instantsearch.scss b/jekyll/_sass/_instantsearch.scss
index 2a283247..ea4275a6 100644
--- a/jekyll/_sass/_instantsearch.scss
+++ b/jekyll/_sass/_instantsearch.scss
@@ -8,7 +8,7 @@
 .main-searchbar {
   border-top: 1px solid $gray-200;
   border-bottom: 1px solid $gray-200;
-  position: fixed;
+  /* position: fixed; */
   width: 100%;
   background: white;
   z-index: 2;
```

docs-search

jekyll/assets/css/footer.css
trevor-circleci: The reason I did this was actually for z-indexing reasons. If the element is positioned relative, it responds to a z-index position, which initial positioning (static) does not allow for. 

I needed it to respond to z-index position so that it sits on top of the sidenav (necessary to prevent ugliness on scrolling).

src-js/instantsearch.js
```diff
@@ -24,7 +24,7 @@ export function init () {
         input: 'instantsearch-search'
       },
       placeholder: 'Search Documentation',
-      queryHook: debounce(function (query, searchFunction) { searchFunction(query); }, 500, { 'leading': true, 'trailing': true }) // method to throttle search requests
+      queryHook: debounce(function (query, searchFunction) { searchFunction(query); setTimeout(renderResults, 100); }, 500, { 'leading': true, 'trailing': true, 'maxWait': 1000 }) // method to throttle search requests
```
JuanCaicedo: There's probably a concern here that I'm not thinking of 🙈But why do we use `setTimeout` here?

jekyll/_sass/_instantsearch.scss
```diff
@@ -0,0 +1,74 @@
+.base-input-style {
+  border-radius: 0;
+  padding: 1.25em 1em 1em 2.8em;
```
JuanCaicedo: Do we use `em` units elsewhere in this project? I might be the only one, but I get confused each time I see it 🙈 If possible, I think `px` would be clearer

jekyll/_sass/_instantsearch.scss
JuanCaicedo: Does this get a border by default, maybe from their code? I'm a little confused to see this and `border-radius: 0`, since those are the defaults 😆

src-js/instantsearch.js
```diff
@@ -0,0 +1,85 @@
+import * as debounce from 'lodash.debounce';
+
+// Instant search initialization
+export function init () {
+  var search = instantsearch({
```
JuanCaicedo: How do we load instantsearch? Maybe we could include it in the webpack build 😀

src-js/instantsearch.js
```diff
@@ -0,0 +1,85 @@
...
+
+  // initialize SearchBox
+  search.addWidget(
+    instantsearch.widgets.searchBox({
+      container: '#search-box',
+      cssClasses: {
+        input: 'instantsearch-search'
+      },
+      placeholder: 'Search Documentation',
+      queryHook: debounce(function (query, searchFunction) { searchFunction(query); setTimeout(renderResults, 100); }, 500, { 'leading': true, 'trailing': true, 'maxWait': 1000 }) // method to throttle search requests
```
JuanCaicedo: There's probably a concern here that I'm not thinking of 🙈But why do we use `setTimeout` here?

jekyll/assets/js/sidebar.js
```diff
@@ -1,89 +1,91 @@
...
-  });
-
-  function setSidebar () {
-    // if footer is in frame, removed fixed style (otherwise add it, if it doesn't exist)
-    if ((footer.getBoundingClientRect().top - window.innerHeight) <= 0 && footer.getBoundingClientRect().top >= window.innerHeight) {
-      if (sidebar.classList.contains('fixed')) {
-        sidebar.classList.remove('fixed');
+    // Show/hide mobile sidebar
+    mobileSidebarCurrent.addEventListener('click', function () {
+      if(mobileSidebarDisplay.classList.contains('hidden')) {
```
JuanCaicedo: Could we use the `hidden` class for the template instead of `display: none`? It would be nice to be consistent between those, but I'm not sure if instantsearch does it's own thing 😆

src-js/instantsearch.js
```diff
@@ -0,0 +1,85 @@
...
+// Instant search initialization
+export function init () {
+  var search = instantsearch({
+    appId: 'U0RXNGRK45',
+    apiKey: 'dd49f5b81d238d474b49645a4daed322', // search-only API Key; safe for front-end code
+    indexName: 'documentation',
+    routing: true
+  });
+
+  // adding conditions to filter search
```
JuanCaicedo: Could we extract each of these into a function and then call each of them from the init body? That would put the info of these comments into a name 😀

src-js/instantsearch.js
```diff
@@ -0,0 +1,85 @@
...
+      resultDisplay.appendChild(results);
+      window.scrollTo(0, 0); 
+
+      pageBody.style.display = "none";
+      resultDisplay.style.display = "block";
+    } else {
+      resultDisplay.appendChild(stateHolder);
+
+      pageBody.style.display = "flex";
+      resultDisplay.style.display = "none";
```
JuanCaicedo: Is this from their docs? Otherwise it would be nice to have some other strategy for doing this that doesn't involve manually setting inline styles, which can be confusing

src-js/instantsearch.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+  };
+
+  searchBox.addEventListener('keyup', renderResults);
+
+  form.addEventListener('submit', function (e) {
+    e.preventDefault();
+    renderResults();
+  });
```
JuanCaicedo: Do we need to render results here? I feel like it should be handled already by keyup, though maybe there is a case I'm not thinking of

src-js/instantsearch.js
```diff
@@ -0,0 +1,85 @@
...
+
+  window.addEventListener('load', renderResults);
+  searchResetButton.addEventListener('click', function () {
```
JuanCaicedo: I wonder if we can handle this by using a `change` event https://developer.mozilla.org/en-US/docs/Web/Events/change

JuanCaicedo: That is, unless the reset functionality comes from instant search and this is something they say to do in the docs

PR #2774, opened 2018-10-02
Update local CLI docs

dnephin:
# Description
Update the local CLI docs

# Reasons
I noticed a few issues with this doc. We still reference the old local cli install instructions and local build was mentioned in two places.

This change moved all the install options into one section and moves the `local execute`instructions all together.

### DIFF
--- a/jekyll/_cci2/local-cli.md
+++ b/jekyll/_cci2/local-cli.md
@@ -15,16 +15,20 @@ This document provides instructions for installing and using the CircleCI CLI.
 ## Overview
 {:.no_toc}
 
-The `circleci` commands enable you to reproduce the CircleCI environment locally and run jobs as if they were running on the hosted application for more efficient debugging and configuration in the initial setup phase. 
+The `circleci` command-line tool enables you to reproduce the CircleCI environment locally and run jobs as if they were running on the hosted application for more efficient debugging and configuration in the initial setup phase.
 
 You can also run `circleci` commands in your [`.circleci/config.yml`]({{ site.baseurl }}/2.0/configuration-reference/) file for jobs that use the primary container image. This is particularly useful for globbing or splitting tests among containers.
 
-**Note:** There are currently two versions of the CLI that you may install. The Local CLI available at <https://circle-downloads.s3.amazonaws.com/releases/build_agent_wrapper/circleci> is an older version that does not yet support Workflows.
-The Local CLI version available at <https://github.com/CircleCI-Public/circleci-cli> has a different set of options, some of which are still in active development, and will be the replacement for the old CLI.
+**Note:** An older version of the circleci cli has been replaced by a new version available at <https://github.com/CircleCI-Public/circleci-cli>. See upgrading and installation instructions below.
 
-## Updating to the CircleCI-Public CLI
+## Install
 
-The CLI upgrade provides additional functionality through a totally new CLI developed as a [CircleCI-Public open source project](https://github.com/CircleCI-Public/circleci-cli). If you already have installed the circle-downloads CLI previously, run the following commands to update and switch:
+There are a few installation options.
+
+### Updating to the new CLI
+{:.no_toc}
+
+The CLI upgrade provides additional functionality through a totally new CLI developed as a [CircleCI-Public open source project](https://github.com/CircleCI-Public/circleci-cli). If you have the old CLI installed, run the following commands to update and switch to the new CLI:
 
 ```
 circleci update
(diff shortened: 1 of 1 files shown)


michelle-luna:
thanks! you can preview them here https://github.com/circleci/circleci-docs/blob/231ffa6fde4215be27a4813845b10d2529a7712c/jekyll/_cci2/local-cli.md

jekyll/_cci2/local-cli.md
```diff
@@ -15,16 +15,20 @@ This document provides instructions for installing and using the CircleCI CLI.
 ## Overview
 {:.no_toc}
 
-The `circleci` commands enable you to reproduce the CircleCI environment locally and run jobs as if they were running on the hosted application for more efficient debugging and configuration in the initial setup phase. 
+The `circleci` command-line tool enables you to reproduce the CircleCI environment locally and run jobs as if they were running on the hosted application for more efficient debugging and configuration in the initial setup phase.
```
felicianotech: Command-Line Interface   ?

jekyll/_cci2/local-cli.md
```diff
@@ -46,129 +51,76 @@ The CLI, `circleci`, is downloaded to the `/usr/local/bin` directory.
...
-followed by the name of the job.
+The CLI allows you to run a single job from the CircleCI on your desktop using docker. 
 
-```bash
-$ circleci build --job test
 ```
+$ circleci local execute --job JOB_NAME
+```
+
+**Note:** If your _config.yml_ uses Workflows and has a job named `build`, only the `build` job will run. It is not possible to run an entire workflow using the CLI.
```
eric-hu: Is this still true if a user passes `--job JOB_NAME`? The way it's phrased now implies that this flag has no effect when there's a `build` job and a `workflows` stanza.

jekyll/_cci2/local-cli.md
```diff
@@ -46,129 +51,76 @@ The CLI, `circleci`, is downloaded to the `/usr/local/bin` directory.
...
+2. Run the following command:
 
 ```
-$ circleci config validate -c .circleci/config.yml
+$ circleci config validate
+Config file at .circleci/config.yml is valid
 ```
-The config validate command checks your local config.yml file for syntax errors.
-
-***Note:*** Local jobs don’t cache dependencies. You may want to comment out dependency sections if you are testing YAML syntax. Or, as in this example, use the `-c` flag to specify a local `config.yml` file that doesn’t pull in large dependencies.
```
eric-hu: Is it incorrect that local jobs don't preserve state across executions?

jekyll/_cci2/local-cli.md
```diff
@@ -46,129 +51,76 @@ The CLI, `circleci`, is downloaded to the `/usr/local/bin` directory.
...
+## Validate A CircleCI config
 
-```
-brew install circleci
-```
-
-**Note:** This will install Docker from Homebrew even if Docker for Mac is installed.
+To ensure that the CLI is configured correctly you can use it to validate a CircleCI config file.
 
+1. Change to the root directory of your CIrcleCI project repository.
```
eric-hu: CIrcleCI -> CircleCI

jekyll/_cci2/local-cli.md
ndintenfass: +1 on Ricardo's suggestion, though I'd go lower case

jekyll/_cci2/local-cli.md
```diff
@@ -15,16 +15,20 @@ This document provides instructions for installing and using the CircleCI CLI.
...
+The `circleci` command-line tool enables you to reproduce the CircleCI environment locally and run jobs as if they were running on the hosted application for more efficient debugging and configuration in the initial setup phase.
 
 You can also run `circleci` commands in your [`.circleci/config.yml`]({{ site.baseurl }}/2.0/configuration-reference/) file for jobs that use the primary container image. This is particularly useful for globbing or splitting tests among containers.
 
-**Note:** There are currently two versions of the CLI that you may install. The Local CLI available at <https://circle-downloads.s3.amazonaws.com/releases/build_agent_wrapper/circleci> is an older version that does not yet support Workflows.
-The Local CLI version available at <https://github.com/CircleCI-Public/circleci-cli> has a different set of options, some of which are still in active development, and will be the replacement for the old CLI.
+**Note:** An older version of the circleci cli has been replaced by a new version available at <https://github.com/CircleCI-Public/circleci-cli>. See upgrading and installation instructions below.
```
ndintenfass: We should probably put a date on this. Or say something like "If you installed the CLI prior to October 2018 you may need to do an extra one-time step to switch over to the newer one."

dnephin:
Opened #2791 with the suggested changes

PR #2787, opened 2018-10-04
Add name description to workflows-jobs section

eric-hu:
# Description
Add name description to workflows-jobs section

# Reasons
I'm not sure if we missed documenting this or we want to document it later. I see we have parameterized jobs documented, and this goes hand in hand with that.


eric-hu:
Thanks for the merge! An example would be a good addition. I'll try to make time for that next week.

PR #2793, opened 2018-10-06
Install via homebrew when Docker for Mac installed

jesstelford:
# Description
Install via homebrew when Docker for Mac installed

# Reasons
I didn't want to have competing versions of Docker installed, but enjoy the convenience of using Homebrew to install circleci-cli. Hopefully this helps future folks also 👍


jesstelford:
:tada:

PR #2796, opened 2018-10-08
Update Build by Project link to #new-project-build

jaysonesmith:
# Description & Reason
Old link pointed to #projects and now goes to the exact place needed.


jaysonesmith:
Absolutely and thanks for the merge!

PR #2797, opened 2018-10-08
Fix SSH wait timeout

kimh:
We wait for 10 mins until users ssh or cancel the job as you can see in the screenshot.

screen shot 2018-10-09 at 7 38 58


kimh:
I also check the value in code.

PR #2826, opened 2018-10-19
Update circleci-images.md

iynere:
clear up some confusion around how particular version of particular packages get installed in particular variants of particular circleci images

# Description

changes made based on feedback here: https://github.com/circleci/circleci-images/issues/128

# Reasons

currently, this page links to the debian stretch package directory for all packages installed in images via apt-get. however, this is misleading, as not all variants of all images are based on debian stretch, and thus the versions included via the links are not always the actual versions of those packages that we install.

our images are based on upstream images that change over time, including changing the base linux distro and version that they extend, so rather than trying to link to the exact package in the exact package directory for each variant of every image we build (which would be impossible...), let's just explain how this all happens. from there, it's relatively easy for folks to figure out why, for example, git version 2.1.4 is installed in their `circleci/node:8` image.


iynere:
ahhh, thank you for catching that @wchargin

PR #2830, opened 2018-10-22
Remove updating Trusty versions json

kimh:
# Description

Remove updating Trusty versions json

# Reasons

We don't need to update the file since we sunset 1.0


mitch2323:
what is this

PR #2846, opened 2018-10-25
Document the new URL for the CLI tool

marcomorain:
Updating the docs to be in line with the latest README.md - https://github.com/CircleCI-Public/circleci-cli/commit/1dca9bca890201f12c6229d28aa9987a989a7eb4


jekyll/_cci2/local-cli.md
```diff
@@ -45,13 +45,14 @@ This command may prompt you for `sudo` if your user doesn't have write permissio
 If you haven't already installed `circleci` on your machine, run the following command to install the CircleCI CLI:
 
 ```
-curl https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh \
-	--fail --show-error | bash
+curl -fLSs https://circle.ci/cli | bash
 ```
 
-The CLI, `circleci`, is downloaded to the `/usr/local/bin` directory.
+By default, the `circleci` app will be installed to the `/usr/local/bin` directory. If you do not have write permissions to `/usr/local/bin`, you may need to run the above command with `sudo`. Alternatively, you can install to an alternate location by defining the `DESTDIR` environment variable when invoking `bash`:
```
michelle-luna: Why would we call this an app?

michelle-luna: If we call it circleci CLI, I'm fine with this (we call the frontend UI the 'CircleCI app') and the installer page says it is installing the "circleci CLI tool" so we should just be consistent there. thanks!

marcomorain:
Thanks for fixing this up @ndintenfass

PR #2852, opened 2018-10-29
Update CLI install command.

felicianotech:
We have a new "pretty" install URL for the CLI. Let's use it.


felicianotech:
Nevermind. This is already addressed in #2846.

PR #2854, opened 2018-10-29
Add tip regarding reserved chars in cache-key

### LINKED ISSUE #2822: Cache keys have disallowed characters
# Missing Information
Some characters will break the ability to save/restore cache if they exist in the character key. Using a key like `https://google.com` would result in this error. More complex URLs may also commonly contain `:, ?, &, =, /, #`
A list of non-safe characters should be included, CircleCI's validation tool could also help catch this or an error at runtime.

# Already Looked
https://circleci.com/docs/2.0/caching/

# Potential Locations
https://circleci.com/docs/2.0/caching/

teesloane:
# Description
fixes #2822 

I managed to break one of my own builds by using the `/` char in the key. It was a bit of a strange one to try and recreate, however. Usage of the `:` did not break my build, and at times the use of `//` still worked.

### DIFF
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -154,6 +154,12 @@ For example, you may want to clear the cache in the following scenarios by incre
 * Language version change, for example, you change ruby 2.3 to 2.4
 * Dependencies are removed from your project
 
+<div class="alert alert-info" role="alert">
+  <b>Tip:</b> Beware using special or reserved characters in your cache key (ex:
+  <code class="highlighter-rouge">:, ?, &, =, /, #</code>), as they may cause issues with your build. Generally,
+  consisider using keys within [a-z][A-Z] in your cache key prefix.
+</div>
+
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/caching.md
```diff
@@ -154,6 +154,12 @@ For example, you may want to clear the cache in the following scenarios by incre
 * Language version change, for example, you change ruby 2.3 to 2.4
 * Dependencies are removed from your project
 
+<div class="alert alert-info" role="alert">
+  <b>Tip:</b> Beware using special or reserved characters in your cache key (ex:
+  <code class="highlighter-rouge">:, ?, &, =, /, #</code>), as they may cause issues with your build. Generally,
+  consisider using keys within [a-z][A-Z] in your cache key prefix.
```
michelle-luna: Beware -> Beware when
consisider -> consider

teesloane: Good catch, thanks!

PR #2859, opened 2018-10-30
Add executor and integer parameter types and rearrange a bit

ndintenfass:
# Description
Add executor and integer parameter types and rearrange a bit to clarify that scoping is not part of the types section

# Reasons
New types available and moved scoping out of the Types section.

### DIFF
--- a/jekyll/_cci2/reusing-config.md
+++ b/jekyll/_cci2/reusing-config.md
@@ -345,53 +345,20 @@ A parameter can have the following keys as immediate children:
 | Key Name    | Description                                                                                   | Default value |
 |-------------|-----------------------------------------------------------------------------------------------|---------------|
 | description | Optional. Used to generate documentation for your orb.                                        | N/A           |
-| type        | Required. Currently "string", "boolean", "enum" (takes extra keys), and "steps" are supported                           | N/A           |
+| type        | Required. See Parameter Types below for details.                           | N/A           |
 | default     | The default value for the parameter. If not present, the parameter is implied to be required. | N/A           |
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/reusing-config.md
```diff
@@ -487,70 +454,89 @@ The `enum` parameter may be a list of any values. Use the `enum` parameter type
...
+      - test:
+          e:
+            name: xenial
+            some-value: foobar
 ```
+#### Integer Parameter
+{:.no_toc}
 
-The above would resolve to the following: 
+The parameter type `integer` is use to pass a numeric integer value. The following example using the `integer` type to populate the value of `parallelism` in a job.
```
johnswanson: Typo, `following example
```suggestion
The parameter type `integer` is use to pass a numeric integer value. The following example uses the `integer` type to populate the value of `parallelism` in a job.
```

jekyll/_cci2/reusing-config.md
```diff
@@ -625,6 +611,93 @@ workflows:
...
+    environment:
+      MYPRECIOUS: "myspecialvalue"
+```
+
+
+
+### The Scope of Parameters
+{:.no_toc}
+
+Parameters are in-scope only within the job or command that defined them. If you want a job or command to pass its parameters to a command it invokes, they must be passed explicitly. Command, job, executor, and parameter names can only contain lowercase letters a-z, digits, and _ and -, and must start with a letter.
```
johnswanson: Re: rules about names, true now, but I thought we were in the process of changing this?

PR #2872, opened 2018-11-02
corrected ports in ssh -L example

PhilippGrashoff:
3000 and 8080 were mixed in text below command, now fits command and text above command


likon39:
Danke

PR #2877, opened 2018-11-06
Specify notification origin (job vs workflow)

teesloane:
Add extra details regarding notification origins (slack and email originate from workflows, while the rest is from jobs).


amirabbaska:
hi

amirabbaska:
okkkkk

PR #2886, opened 2018-11-07
Fixed two typos in the Using-Orbs Example stanzas

guinaut:
# Description
There are two typos in the Using Orbs example stanzas.  This PR addresses both

# Reasons
Item 1: A "v" was between the Orb name and version in "hello: circleci/hello-build@0.0.5"
Item 2: The "sayhello" command was missing the local namespace "hello/sayhello"


guinaut:
I've modified the "sayhello" command to clearly reference a non-existent Orb.  Hopefully this will clue the reader in that the command is a capability but not a "real" option to use.  We should consider making it real.

PR #2907, opened 2018-11-12
Add --frozen-lockfile and also cache node_modules

dtinth:
# Description
In Yarn config guide, add `--frozen-lockfile` and also cache `node_modules`.

# Reasons
- By using `--frozen-lockfile` this ensures that `yarn.lock` must be kept consistent with `package.json`. Otherwise, it will fail the build. This behavior is similar to running `bundle install` on CI/CD environment which fails the build if the lockfile is not kept up-to-date with dependency specification.

- Adding `node_modules` to the cache means that `yarn` can skip copying files from its cache into `node_modules` and exit with “Up-to-date”.


dtinth:
Updated as requested :)

felicianotech:
Great thank you @dtinth.

@michelle-luna The step "Update Algolia Index" ran and failed due to a missing key. This step shouldn't be running on PRs anyway right?

michelle-luna:
yes, there is an open ticket to fix the algolia problem, thanks for your patience!

felicianotech:
Ah okay. Well the build failure here is related the the Algolia step and not this PR so this LGTM.

PR #2908, opened 2018-11-12
[CIRCLE-14872] Don’t assume search results have content or title

millerpeterson:
JIRA ticket: [CIRCLE-14872](https://circleci.atlassian.net/browse/CIRCLE-14872)

# Overview
Looks like there are some items in our index that have no content. Our code for formatting the results is expecting some content to be there, and is throwing exceptions when it isn't, causing issues like the one described in this ticket.

# How to Test
Spin up the app (don't forget to run webpack :wink:). Search for "Language Guides". There should be no console errors, and you should see an entry with just the title called "Language Guides". Search for "Go cace" - your first result should be the go language guide.

### DIFF
--- a/src-js/instantsearch.js
+++ b/src-js/instantsearch.js
@@ -1,10 +1,13 @@
 import * as debounce from 'lodash.debounce';
+import * as get from 'lodash.get';
 
 const formatResultSnippet = (snippet) => {
-  const titleTag = `<strong>${snippet.title.value}</strong>`;
+  const title = get(snippet, ['title', 'value'], '(untitled)');
+  const content = get(snippet, ['content', 'value'], '');
+  const titleTag = `<strong>${title}</strong>`;
   const contextTag = snippet.headings ?
     `<p class="context">${snippet.headings.map(h => h.value).join(' > ')}</p>` : '';
-  const contentTag = `<p class="content">${snippet.content.value}</p>`;
+  const contentTag = `<p class="content">${content}</p>`;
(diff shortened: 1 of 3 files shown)


src-js/instantsearch.js
```diff
@@ -1,10 +1,13 @@
 import * as debounce from 'lodash.debounce';
+import * as get from 'lodash.get';
 
 const formatResultSnippet = (snippet) => {
-  const titleTag = `<strong>${snippet.title.value}</strong>`;
+  const title = get(snippet, ['title', 'value'], '(untitled)');
```
millerpeterson: I don't know of any documents where there would be no title, but just in case.

src-js/instantsearch.js
```diff
@@ -1,10 +1,13 @@
...
+  const content = get(snippet, ['content', 'value'], '');
+  const titleTag = `<strong>${title}</strong>`;
   const contextTag = snippet.headings ?
     `<p class="context">${snippet.headings.map(h => h.value).join(' > ')}</p>` : '';
-  const contentTag = `<p class="content">${snippet.content.value}</p>`;
+  const contentTag = `<p class="content">${content}</p>`;
```
trevor-circleci: @millerpeterson

What does the content render as here when there is no content?

millerpeterson: Just as nothing - it would be an empty <p>. You can see this by searching for "Language Guides" - that one is just a title.

PR #2915, opened 2018-11-13
fix: Fix JavaScript and Node.js spelling

potato4d:
# Description

1. `s/Javascript/JavaScript/ig`
2. `s/NodeJS/Node.js/ig`

# Reasons

JavaScript / Node.js are each correct notation.


potato4d:
> The jekyll-algolia plugin could not connect to your application ID using the API 

:cry:

potato4d:
@michelle-luna 

If I send a Pull Request, an error seems to occur.
This is because there is no API Key corresponding to hard-coded Algolia Application ID.

Could you please execute Rerun with workflows with you?

michelle-luna:
thanks for this patch and for your patience while we fixed up the algolia auto-indexing implementation!

PR #2919, opened 2018-11-14
WIP: Update env-vars.md

iynere:
# Description
remove `CIRCLE_COMPARE_URL`

# Reasons
we no longer publish the compare url once 2.1 config processing is turned on:

https://circleci.atlassian.net/browse/CIRCLE-14245

we want folks to all move to 2.1 eventually, so we should stop documenting this env var, which will no longer work


ndintenfass:
We are working on a solution, so let’s wait to make any major change. Also, it’s not just 2.1, but any build that runs through build processing.

PR #2929, opened 2018-11-16
Fix typo in JVM heap size docs regarding test-results container

samnrubin:
# Description
This is quick fix updating the JVM heap size documentation to refer to the correct container

# Reasons
When running the command given it actually errors out due to the container actually being called `test-results` – figured I'd clean it up for the next person


michelle-luna:
thank you!

PR #2936, opened 2018-11-19
fix and update version number of hello-build orb

kurumai:
# Description
fix and update orb version number `.0.0.5` -> `0.0.7`

# Reasons
No dot(.) is required for the first part of version num.


kimh:
LGTM

PR #2938, opened 2018-11-20
Start localizing into Japanese

kimh:
Adding Japanese translation

This change is the base for localizing docs into Japanese.

Here are a few design considerations:

- All translated pages live under https://circleci.com/docs/ja
- Duplication must be avoided as much as possible
- English and Japanese versions can be maintained separately

A few notes about implementation

- We create a new collection `cci2_ja` that lives under `/ja/2.0`
- We add `lang:` metadata to specify language with default set in config.yml so that we don't need to add the metadata manually for all pages.
- We run `shim-translation.sh` only in CI. This allows us to keep untranslated pages for Japanese while avoiding having two same pages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,9 @@ jobs:
             mkdir -p /tmp/workspace/js
             mv /tmp/workspace/js/ jekyll/assets/js/dist/
       - run:
+          name: Shim untranslated Japanese pages
+          command: ./scripts/shim-translation.sh jekyll/_cci2 jekyll/_cci2_ja
+      - run:
           name: Build the Jekyll site
           command: bundle exec jekyll build --config jekyll/_config.yml,jekyll/_config_production.yml --source jekyll --destination jekyll/_site/docs/ 2>&1 | tee $JOB_RESULTS_PATH/build-results.txt
       - run:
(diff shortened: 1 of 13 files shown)


Peeja:
> English and Japanese versions can be maintained separately

Aren't we [using Crowdin](https://github.com/circleci/circleci-docs/blob/88290213dcdc55a1831c4c01ef14e652044468af/crowdin.yml) to keep them in sync automatically? Are we abandoning that?

Peeja:
> English and Japanese versions can be maintained separately

Aren't we [using Crowdin](https://github.com/circleci/circleci-docs/blob/88290213dcdc55a1831c4c01ef14e652044468af/crowdin.yml) to keep them in sync automatically? Are we abandoning that? (/cc @michelle-luna)

michelle-luna:
@Peeja We are still using crowdin, from what I understand, but the first pass of this localization was from a snapshot, then we will use the integration to keep in sync (I believe that is the plan).

@kimh I can build the site and the welcome/index page and menu all look great. But, I get YAML exception when I  docker-compose up (after running the shim script), .gitkeep is also missing, not sure if that is related. 

``` Error: YAML Exception reading /srv/jekyll/_cci2_ja/about-circleci.md: (<unknown>): mapping values are not allowed in this context at line 2 column 26
jekyll_1  |              Error: YAML Exception reading /srv/jekyll/_cci2_ja/about-circleci.md-e: (<unknown>): mapping values are not allowed in this context at line 2 column 26```

kimh:
I'm merging this now and test in production.

kimh:
Looking good in production.

PR #2940, opened 2018-11-20
Update `create` step to include all required params.

teesloane:
# Description
requested update:
> Under orb tutorial step 5, we need to also include `<vcs-type> <org-name>`. So it should be `circleci namespace create <name> <vcs-type> <org-name>`

note: I just used `ndintenfass_org` as an example org-name. I couldn't tell if there was one used in the gif above the doc.


thomasr8:
Looks good, but I am wondering if it is better to abstract the org name value. Although most developers will know better, I could see some developer mistakenly using the 'ndintenfass' as an org name in their implementation.

Thoughts?

PR #2957, opened 2018-11-27
Add legacy note + links to new CLI

### LINKED ISSUE #2948: unclear what the page 'Testing Config Files Locally' is doing
# Location
https://circleci.com/docs/2.0/examples/#section=configuration

# Explanation
It's really unclear on this page what is going on.  what am I testing?  what will happen when I run the shell script?  Some more background and context plus maybe an example would be helpful.

teesloane:
fixes #2948


jekyll/_cci2/examples.md
```diff
@@ -4,6 +4,13 @@ title: "Testing Config Files Locally"
 description: "Example CircleCI 2.0 Configuration Files"
 ---
 
+
+<div class="alert alert-info" role="alert">
+<b>Note:</b> This document describes a legacy means of testing your config file locally.
```
michelle-luna: I would say 'alternate' means instead of legacy here, and also link from the CLI document to this one for those who need an alternate method like this one.

PR #2959, opened 2018-11-28
[CIRCLE-15078] Commit app.min.js for now

millerpeterson:
JIRA ticket: [CIRCLE-15078](https://circleci.atlassian.net/browse/CIRCLE-15078)

# Overview
Since we've added webpack to the docs site, you need to run it at least once for the Jekyll site to build (see the error in the JIRA ticket for what happens when you don't). Most people on the docs team seem to run the site through Docker, which currently does not run webpack. We should fix the Docker config so that it runs webpack, but for the time being this PR just commits the webpack ouput bundle (app.min.js) so that the site will at least build with the current Docker setup. This shouldn't matter for the production assets as the build process will overwrite the committed version.

# How to Test
Check out this branch, build it. There site should build successfully.


Peeja:
_(Redirecting from Slack:)_

Is there a reason it’s hard to run webpack in the Docker setup? I’m not clear why we’re doing a short-term fix first.

PR #2973, opened 2018-12-04
Update comment

flatfisher:
# Description
Added a comment about not encode on Base64.

# Reasons
Usually, Google service account encode to base64 and set it as an environment variable.
However, it is different when using GCR's custom image. I added a comment so as not to misunderstand. Please let me know if you have other desirable comments.
https://circleci.com/docs/2.0/google-auth/#authorizing-the-google-cloud-sdk


Yafieparfume88:
JSON service account you created

PR #2974, opened 2018-12-04
Localize about-circleci page

kimh:
Localizing `/2.0/about-circleci/` into Japanese.


kimh:
The actual localization has been done by localization agent and this PR is created out of my review, so I'll be merging this page.

PR #2975, opened 2018-12-04
Localize add-ssh-key page

kimh:


kimh:
The actual localization has been done by localization agent and this PR is created out of my review, so I'll be merging this page.

PR #2978, opened 2018-12-04

iynere:
- missing `- ` in `persist_to_workspace` in configuration-reference.md
- TK


michelle-luna:
thanks @iynere the extra white space is from the raw tags inside the code block, they should be like this instead (but can't be removed) 
{% raw %}
```yaml
    steps:
      - restore_cache:
         keys:
           - m2-{{ checksum "pom.xml" }}
           - m2- # used if checksum fails
```
{% endraw %}

PR #2989, opened 2018-12-14
Fix: inline orb example missing `orb` stanza.

teesloane:
# Description
[relevant discussion](https://discuss.circleci.com/t/unable-to-find-orb-with-inline-orb/27145)

### DIFF
--- a/jekyll/_cci2/creating-orbs.md
+++ b/jekyll/_cci2/creating-orbs.md
@@ -108,8 +108,9 @@ To write inline orbs, you need to place the orb elements under that orb's key in
 description: # The purpose of this orb
 
 orbs:
-  codecov: circleci/codecov-clojure@0.0.4
   my-orb:
+    orbs:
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/creating-orbs.md
```diff
@@ -108,8 +108,9 @@ To write inline orbs, you need to place the orb elements under that orb's key in
 description: # The purpose of this orb
```
ndintenfass: I'm not sure why this is here -- it should probably be removed.

michelle-luna:
This example came from the SDK docs (it is the same in that page), we assumed these were tested in the beta and that errors would be found during the reviews https://github.com/CircleCI-Public/config-preview-sdk/blob/master/docs/inline-orbs.md. We should test each and every example snippet, but the docs team hasn't done so for every orbs snippet to be sure.

PR #2996, opened 2018-12-18
Clarify available docker versions.

teesloane:
# Description
Explicitly link to docker versions.

# Reasons
Users are not easily finding what docker versions our platform supports.


jekyll/_cci2/building-docker-images.md
```diff
@@ -111,30 +111,14 @@ Let’s break down what’s happening during this build’s execution:
 
 ## Docker Version
 
-If your job requires a specific docker image, you can set it as a `version` attribute:
```
teesloane: I was guessing that `image` was a typo, or does Docker refer to it's `versions` also as `images`?

Issue #3007, opened 2018-12-20
Document what an "orb reference" is.


alexanderkjeldaas:
# Missing Information
https://circleci.com/docs/2.0/configuration-reference/#orbs-requires-version-21

"orb reference" is undocumented.  Can I refer to a local file with an orb definition?  What is the legal syntax?

PR #3020, opened 2019-01-04

thomasr8:
Added note in Orbs section to better define what an orb reference & fully qualified orb reference is. This is based on customer feedback.

# Description
We received a comment from a customer who was did not understand what an "orb reference" was, so content was added to better define what an "orb reference" and "fully qualified orb reference" means in the context of orbs.

# Reasons


thomasr8:
This pull request fixes #3007

PR #3056, opened 2019-01-15
Updates to the Language Guide: Haskell page

onslaughtq:
# Description
The Language Guide: Haskell page has been updated to use
epoch when creating a cache, in order to work around potential
issues caused by ignoring the version of stack used in the docker
image and by ignoring the resolver used in stack.yaml.

Lastly, the sample configuration at the top of the page was
updated to cache ".stack", in addition to ".stack-work".

# Reasons
I made these changes, because after following the original recommendation of using package.yaml's checksum, a fellow teammate pointed out that we would have a bad cache if we updated the resolver in stack.yaml.  Further discussion led us to discover that even if we considered stack.yaml in the cache key, we would still potentially have issues if we used a newer stack image from docker.


teesloane:
Hi there!

Thank you for submitting this PR. I would recommend some other options before resorting to using the `epoch` key, as it might defeat the purpose of caching..

**Custom Keys**

I would recommend including a custom key inlined with a checksum of, well, _something_ (whether that be the package.yml, which if I remember correctly is auto generated, the `stack.yml` file or a generated `cabal` file). This way, you can bust the cache whenever you want, manually. 

Example:

```
    steps:
      - restore_cache:
          keys:
            # Find a cache corresponding to this specific package-lock.json checksum
            # when this file is changed, this key will fail
            - v1-npm-deps-{{ checksum "package-lock.json" }}
            - v1-npm-deps-{{ checksum "package.json" }}
```

**Cache by specificity**

Because you can have multiple cache keys, CircleCI will try and restore cache in an order of specificity, from most specific to most broad. We have more details of this in our [docs](https://circleci.com/docs/2.0/caching/#restoring-cache).

**Using Epoch**

When you use `epoch` you are going to be generating a new cache _every build_, and so, the likelihood of making use of the cache is quite small.

I'm going to close this PR, but again, thank you for submitting it and in doing so, contributing to our documentation. Please re-open it if you have any questions or further suggestions!

onslaughtq:
Thanks for the feedback.  

Based on my understanding if we use `epoch` when saving the cache, and have a restore_cache step that leaves off `epoch`, we will still be making use of the most recent cache that matches more broadly.
In other words, saving a cache with key of `my-cache-{{epoch}}` will satisfy a key of `my-cache`.  So your point about the likelihood of making use of the cache is not correct.  We will generate a new cache every build, and the cache from the immediate last successful build will be used on the next build.

Regarding using package.yaml, that file is not auto-generated. It is a replacement for a project's .cabal file.  If a .cabal file is missing, the stack tool will use hpack to generate a .cabal file, which is the more universally accepted format for Haskell packages still. 

My point with this pull request is that the documentation is wrong in suggesting that *just* package.yaml be used, because it can't be for projects that use stack. You have to at least specify package.yaml AND stack.yaml, but even those may not be sufficient, which is why `epoch` is better for this case.

teesloane:
Hi @onslaughtq ,

Thank you for your suggestions and corrections. You are correct, I was mistaken. I was assuming that you were using the epoch to restore the key as well (which really _wouldn't_ make sense). 

As per our docs on caching, we try not to recommend using epoch (especially in our language guides) as most cases, manually busting the cache and using the language's respective lock/deps file is sufficient for the majority of users. I will update the docs to make use of the package _and_ the stack file (or, feel free to update your PR and we can merge it!)

onslaughtq:
@teesloane I've pushed a new commit to my branch that removes epoch in favor of using the checksum of package.yaml and stack.yaml. Do you want me to open a new PR, or will you reopen this one?

teesloane:
Lovely! Thank you for making those changes and contributing to our docs (and for helping me learn a bit more about Haskell's ecosytem along the way! 😸 ) Will merge this when CI passes!

PR #3143, opened 2019-02-20
Tweak A Document About The API Endpoint To Delete SSH Key

yuya-takeyama:
# Description

Added some information about the endpoint to delete SSH key.

# Reasons

* The endpoint is not listed in the summary of endpoints
* The example command lacks URL


michelle-luna:
@thomasr8 can you add this to the new API reference in a separate PR please? thank you!

PR #3149, opened 2019-02-20
Synchronize Clojure sample config and walkthrough

MatthewDarling:
# Description
Removed a `~/.lein` cache that was shown in the Clojure config walkthrough, but [removed from the sample config mid-June 2017](https://github.com/circleci/circleci-docs/commit/ed3bd149d182b72355e5901a60dc131b56281135#diff-3d61f9f0003cbef5fb7993bb9fdca4c8).

# Reasons
Noticed the inconsistency, and assumed the sample config was the source of truth. Happy to sync the other way if desired :)


michelle-luna:
thank you!

PR #3160, opened 2019-02-22
Rosie trial installation changes

rosieyohannan:
# Description
I ran through the trial installation and updated the doc for accuracy and clarity. There are still a couple of areas that need addressing: The first step under "CircleCI Configuration" needs clarifying as the docs text and the UI text appear to contradict each other, or at least aren't quite clear. Also there are more sections on the web app settings page than those mentioned in (what is in my version) step 3.

# Reasons
When running through the trial installation I found some inaccuracies between what I encountered and what was documented. Also, I made some formatting changes to keep the steps in the CircleCI configuration section consistent: bullet points for sections on a page, rather than new numbered steps when staying on the same page.


michelle-luna:
thank you!

PR #3171, opened 2019-02-27
CTA for feedback/comments in footer.

### LINKED ISSUE #3170: Add Language Around Feedback and Comments to the footer
In the footer of most doc pages, we mention how a reader could submit a change or report a problem via a GitHub Issue.

For that last point, I think we should add that they should also do this to submit feedback/comments on docs. Not just reporting an issue.

teesloane:
closes #3170


michelle-luna:
thank you!

PR #3172, opened 2019-02-27
Add info for legacy local cli on server.

### LINKED ISSUE #3164: add back old CLI info for server
# Location
We should add a note to the CLI doc about limitation for server (it doesn't work for server customers yet). This should be at the top of the document and link to a section at the bottom about how to install the old CLI.

# Explanation
This is confusing for server customers

teesloane:
closes #3164.

I'm not totally sure that the archived repo this links too is actually the right thing to put in here; is the legacy local-cli still accessibly for installation manually (at least, onto server?)

From looking through the [shell script](https://github.com/circleci/local-cli/blob/master/circleci.sh) in the [archived repo](https://github.com/circleci/local-cli), it seems like the shell script pulls down something called `picard` (I don't have access to that repo, so I'm not sure what it is).

@felicianotech - I noticed you committed last on the legacy cli to update the readme.md. Any thoughts on this?


felicianotech:
I wouldn't point to my old repo. Those instructions wouldn't work because it was only fo the brew/snap installs and those only support the new CLI now.

The old command that install the old CLI via cURL, that is probably what needs to be used. I'd ask the @circleci/developer-experience.

jekyll/_cci2/local-cli.md
```diff
@@ -312,6 +314,10 @@ Further, not all commands may work on your local machine as they do online. For
 
 For security reasons, encrypted environment variables configured in the UI will not be imported into local builds. As an alternative, you can specify env vars to the CLI with the `-e` flag. See the output of `circleci help build` for more information. If you have multiple environment variables, you must use the flag for each variable, for example, `circleci build -e VAR1=FOO -e VAR2=BAR`.
 
+## Using the CLI on CircleCI Server
+
+Currently, only the legacy CircleCI CLI is available to run on server
+installations of CircleCI. The legacy CircleCI CLI is available [here](https://github.com/circleci/local-cli).
```
michelle-luna: This is what we have for installation instructions before: 
### macOS and other Linux Distros

1. Install and configure Docker by using the [docker installation instructions](https://docs.docker.com/engine/installation/).

2. To install the CLI, run the following command:

```
$ curl -o /usr/local/bin/circleci https://circle-downloads.s3.amazonaws.com/releases/build_agent_wrapper/circleci && chmod +x /usr/local/bin/circleci
```

The CLI, `circleci`, is downloaded to the `/usr/local/bin` directory. If you do not have write permissions for `/usr/local/bin`, you might need to run the above commands with `sudo`. The CLI automatically checks for updates and will prompt you if one is available.

teesloane: Added it!

PR #3173, opened 2019-02-28
Example custom node installation alongside non-node containers

### LINKED ISSUE #3049: Include information about how to install specific Node versions in CircleCI Docker images
# Missing Information
A customer had an excellent suggestion while troubleshooting an nvm slowdown relative to iojs.org being down.

It would be great if the CircleCI documentation was updated to include instructions on how to install a specific version of Node in the various `<language>-node` images. There are many blog posts in this space, but we do not appear to have this information in our docs yet.

This is especially important as, at least for the Ruby images, the version of Node installed by default is 8.x, which is behind the current Node LTS release of 10.x.

# Already Looked
Searched in CircleCI docs, e.g. https://circleci.com/docs/?query=node%20version. Nothing appears relevant from the titles and snippets displayed.

# Potential Locations
_Tell us where you would expect to find this information.
This can be either an existing document,
or an entirely new document._
https://circleci.com/docs/2.0/circleci-images/#best-practices
or https://circleci.com/docs/2.0/faq/#what-is-the-best-practice-for-specifying-image-versions
could be options. It also might go best in a new document.

teesloane:
closes #3049


jekyll/_cci2/circleci-images.md
```diff
@@ -50,6 +45,30 @@ to pin an image to a version or operating system (OS).
 - Use a Docker image ID
 to pin an image to a fixed version.
 
+**NOTE:** For docker images that have `-node` (Node.js) included in the container, the latest LTS
```
felicianotech: Docker should be capitalized and `-node` is part of a tag, which forms the image name. The container aspect doesn't exists until it's run.

Maybe something like this?

```
**NOTE:** For Node.js variant Docker images (tags that end in `-node`) the LTS release of Node.js is pre-installed.
```

teesloane: Makes sense! Done.

michelle-luna:
thank you!

PR #3187, opened 2019-03-05
Reorder orb publishing steps

danielcompton:
## Description

As I was reading this doc page I found it a little bit confusing to be validating my Orb in step 4 "When you have finished authoring your orb" and then in step 5 to be designing my orb, and then in step 6 creating a new orb using an inline template. Does it seem like these three steps should be reversed: starting initially with an inline template, then designing the orb, then validating it?

## Reasons

I can't validate my orb config if I haven't created it yet?

If this was to be merged, the diagram of the Orb Publishing Process would also need to be updated.


drazisil:
@michelle-luna This looks good, but the image will need to be changed to reflect the change in step order as well.

Issue #3169, opened 2019-02-26
Getting Started doesn't Get Started.


alexfornuto:
# Location
https://circleci.com/docs/2.0/hello-world/

# Details
I have a CircleCI account, use it frequently for work projects. Followed [this guide](https://circleci.com/docs/2.0/hello-world/) to get a [new project](https://github.com/alexfornuto/terminus-installer/) going using CircleCi

I added the new `.circleci/config.yml` file exactly as described in step 2  of "Use the Hello-Build Orb" to a branch and pushed. No builds. I merged to master and staged more changes in a new branch, this time following step 1 of "Echo Hello World with a `build` Job` and still no jobs. 

As the docs suggest I went to my Projects page to try to **Add Projects**. At this point the docs mention a button called **Build Project**, but the button is **Set Up Project**.

# Location
https://circleci.com/docs/2.0/circleci-images/#php

# Details
Latest Tags aren't listed as they are with other images.

teesloane:
Hi there!

Thanks for filing an issue. You've mentioned a few different things and I have a question or two. 

> I have a CircleCI account, use it frequently for work projects. Followed this guide to get a new project going using CircleCi.

Your link to "[this guide](https://github.com/alexfornuto/terminus-installer/)" brought me to your repo `terminus-installer`. Did you mean to refer to one of our documents? 

Further, you mentioned an improperly named button. Thank you — I will put up a PR to change the docs to change.

You also listed that we don't have any tags visible for PHP; I'm not sure why that is. I'll look into that and get back to you.

alexfornuto:
How embarrassing! My apologies, I was multi-tasking when I wrote this. I've updated the issue to point to [this guide](https://circleci.com/docs/2.0/hello-world/).

teesloane:
Not a problem! Thank you for patience in having us get back to you. I've since put up a PR to amend the incorrect copy reference in our docs and am researching missing docker tags.

Cheers!

felicianotech:
I have a PR for the tags in the reference above ^.

PR #3188, opened 2019-03-05
Improve Docker tag list.

felicianotech:
## Initial Problem

Tags for images such as the PHP image currently aren't displaying on CircleCI Docs ([ref](https://circleci.com/docs/2.0/circleci-images/#php)) while images such as Rust have a limited number of shown tags ([ref](https://circleci.com/docs/2.0/circleci-images/#rust)).

## Fixes

1. This PR fixes the problem above. Images such as PHP, Python, and Rust will show numerous tags.
2. Bumped up the number of tags shown from 50 to 99.
3. Tag list went from 2 columns to 3 because of item number 2.
4. Variant tags are removed *before* calculating the number of tags shown.
5.  A JSON file of our image tags is now available off of this site. Once this PR is merged, it'll be hosted at https://circleci.com/docs/2.0/docker-image-tags.json
6. Instead of linking to Docker Hub, which limits the number of tags show, we now link to the JSON file in number 5.
7. Added a resource section that simply links to the image's GitHub and Dockerfile source.

---

There's a lot more I want to do with regard to displaying images/tags but this should hold us over until the new images are available.


felicianotech:
Related: #3169

felicianotech:
Cool. We're just waiting on @eric-hu?

felicianotech:
Both reviewers approved. Merging.

PR #3189, opened 2019-03-05
Add note about Snapcraft login.

felicianotech:
Adds a note on the expiration time of Snapcraft login files.

Also fixes the Markdown for the ordered list.


michelle-luna:
thank you!

PR #3197, opened 2019-03-07
Add another parallel-testing library (php)

### LINKED ISSUE #3196: Add reference to previousnext/phpunit-finder for globbing phpunit tests based on phpunit.xml
# Missing Information
We wrote a library for globbing phpunit tests for circleci based on file paths defined in your phpunit.xml - https://packagist.org/packages/previousnext/phpunit-finder

# Already Looked

# Potential Locations
Perhaps it would be good to link to this library from the running tests in parallel documentation.

teesloane:
closes #3196 

originally, we were linking to just one third party app (knapsack), but now we have a new one! I figured we might as well make it into a list now:

**Before**

image

---

**After**
image


michelle-luna:
thank you!

PR #3249, opened 2019-03-23
add missing parameters key for type examples

pimlie:
# Description
Add missing parameters key for parameter type examples because otherwise the examples wont work

# Reasons
Because I used the original example for a boolean parameter which then failed to run my jobs as intended


michelle-luna:
thank you!

PR #3257, opened 2019-03-26
Making namespace permissions more clear

aarongajewski:
# Description
I moved the phrase "Only organization owners can create namespaces" to its own "Note" section near the top of the page when discussing the creation of namespaces. 

# Reasons
A customer brought this up and was very confused as he went to create a namespace and learned he was not able to because of his permissions level.

We did have that statement already in the docs but it was buried. Rose and I passed right over it when trying to confirm for the customer. We can keep it in the original section as well as a duplicate but I deleted it for now and was hoping whoever reviews this may have a recommendation.


michelle-luna:
thank you!

PR #3292, opened 2019-04-12
Update _projects.md

thomasr8:
Added section on using the "shallow" parameter to improve performance.

# Description

# Reasons


thomasr8:
@teesloane Will you take a look at this content and make sure that it adheres to the new styles and formatting you created for this guide? Also, will you review to see if this seems to be the right place for this content?

glenjamin:
I think ideally we shouldn’t specify the full example of the non-shallow response. I’ll see if others have an opinion on this.

teesloane:
Just for the sake of not confusing the user I would say don't include the `non-shallow response`

@thomasr8 If we do end up doing the above, let me know as the mid-pane and right-pane on the slate guide don't line up:

Screen Shot 2019-04-12 at 3 35 53 PM


Regardless, here is the link to view the changes: https://44829-48750547-gh.circle-artifacts.com/0/circleci-docs/api/index.html#recent-builds-for-a-single-project

glenjamin:
Could you add the note about performance on the "All projects" section too? If you'd prefer not to repeat then that one is the most expensive for us to handle, so it would be preferrable to treat that one as the primary call.

Relatedly, it would be good to recommend that people prefer the per-project endpoint, as that is likely to be more performant for them, and less expensive for us to calculate.

I also like the idea of only including the shallow example, and leaving the non-shallow one implicit - we might even change the default to be shallow soon.

thomasr8:
@teesloane and @glenjamin Made the recommended changes. When you have time, will you please review to see if there is anything else that might need to be fixed in the content?

src-api/source/includes/_projects.md
```diff
@@ -50,6 +53,8 @@ curl https://circleci.com/api/v1.1/projects?circle-token=:token
 
 Returns an array of all projects you are currently following on CircleCI, with build information organized by branch.
 
+**Note**: When making an API request for Project information, you may experience performance lag and a decrease in overall performance while the request is being processed by the server. To improve performance, CircleCI recommends you pass the `shallow` parameter in your request.
```
glenjamin: Could you put this section next to the "recent builds for all projects" endpoint? That's the one that really affects us.

src-api/source/includes/_projects.md
```diff
@@ -223,6 +230,72 @@ curl https://circleci.com/api/v1.1/project/:vcs-type/:username/:project/tree/:br
 limit | The number of builds to return. Maximum 100, defaults to 30.
 offset | The API returns builds starting from this offset, defaults to 0.
 filter | Restricts which builds are returned. Set to "completed", "successful", "failed", "running", or defaults to no filter.
+shallow | An optional boolean value that may be sent to improve overall performance if set to 'true.'
+
+### Improving Performance In Recent Build Requests Using the `Shallow` Parameter
+
+When making API requests for information about recent builds, you may experience performance lag and a decrease in overall performance while the request is being processed by the server. To improve performance, CircleCI recommends you pass the `shallow` parameter in your request.
+
+The example below shows a user request for recent build information. Notice that when the user passes the `shallow` parameter, a limited set of information is returned, thereby improving response time and minimizing performance lag.
```
teesloane:
```suggestion
The example to the right shows a user request for recent build information. Notice that when the user passes the `shallow` parameter, a limited set of information is returned, thereby improving response time and minimizing performance lag. 
```

src-api/source/includes/_projects.md
```diff
@@ -223,6 +230,72 @@ curl https://circleci.com/api/v1.1/project/:vcs-type/:username/:project/tree/:br
...
+shallow | An optional boolean value that may be sent to improve overall performance if set to 'true.'
+
+### Improving Performance In Recent Build Requests Using the `Shallow` Parameter
+
+When making API requests for information about recent builds, you may experience performance lag and a decrease in overall performance while the request is being processed by the server. To improve performance, CircleCI recommends you pass the `shallow` parameter in your request.
+
+The example below shows a user request for recent build information. Notice that when the user passes the `shallow` parameter, a limited set of information is returned, thereby improving response time and minimizing performance lag. 
+
+
+#### Sample Request Using the `Shallow` Parameter
```
teesloane: So, we actually want to move this in-line with the code examples by using the markdown quote block (`>`). You can see us doing it on line [188](https://github.com/circleci/circleci-docs/pull/3292/files#diff-960fbf6be43efb049b04a2a2ea6c46e6R188)

This ends up interspersing raw text in-between code examples in the right pane, as seen below.

image

The markdown source code to create the above formatting looks like:

image

---


So we still need to add a new curl example that uses `shallow=true`. Adding something like the below should work:


```
curl https://circleci.com/api/v1.1/recent-builds?limit=1&shallow=true

> Sample Request Using the `Shallow` Parameter
```

thomasr8: Fixed

src-api/source/includes/_projects.md
thomasr8: Fixed

src-api/source/includes/_projects.md
thomasr8: Fixed

thomasr8:
Thanks @michelle-luna !

PR #3294, opened 2019-04-12
Orbs/create-orb

mvxt:
# Description
Moved some of the items in Notes to before the 1st step, added / removed some things

# Reasons
- Made the mistake of creating namespace / orb before seeing important notes. Really important to know these things before user just dives in and haphazardly claims a namespace. Most users will read top to bottom, so I don't think we should put information they should know after the first step.
- Removed the bit about "org you specified should already be added to CircleCI CLI". What does that mean? You have to add a personal token to the CLI, but there was nothing about adding an org to anything.

Feedback welcome.


iynere:
@michelle-luna there's some typos / inaccuracies here, i'll push a quick commit & pr

PR #3295, opened 2019-04-12
Adding a workaround of "OUT OF MEMORY" for Haskell

smasuda:
# Description
Adding a workaround of "OUT OF MEMORY" for Haskell.

# Reasons
I believe this is a common issue and worthwhile noting here.

### DIFF
--- a/jekyll/_cci2/language-haskell.md
+++ b/jekyll/_cci2/language-haskell.md
@@ -159,7 +159,7 @@ Excellent! You are now setup on CircleCI with a Haskell app.
 
 ## Common Trouble Shooting
 
-The command `stack test` may fail with an out of memory error. Consider adding the `-j1` flag to the stack test command 
+The command `stack test` may fail with an out of memory error. Consider adding the `-j1` flag to the `stack test` command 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/language-haskell.md
```diff
@@ -157,6 +157,18 @@ Finally, we can take the built executable and store it as an artifact.
 
 Excellent! You are now setup on CircleCI with a Haskell app.
 
+## Common Trouble Shooting
+
+If you see `stack test` fails due to `Process exited with code: ExitFailure (-9) (THIS MAY INDICATE OUT OF MEMORY)`, consider adding `-j1` option like the following. This will avoid the build 
+using multiple cores hence reduce its memory usage, while it may lead a longer
+test time.
```
teesloane: Do you mind rewording this a bit; maybe something like:

> The command `stack test` may fail with an _out of memory_ error. Consider adding the `-j1` flag to  the `stack test` command as seen below (Note: this will reduce test execution to one core, decreasing memory usage as well, but may also increase your test execution time).

smasuda: Thanks for the feedback.  Updated as per suggestion.

michelle-luna:
thank you!

PR #3305, opened 2019-04-16
Minor update

KunalJain:
- Updated to make it clear that anyone can add a context but only admins can restrict them
- Changed `workflows` to `job` to reduce confusion.


michelle-luna:
thanks!

PR #3307, opened 2019-04-16
Update reusing-config.md

### LINKED ISSUE #3276: What's allowed for "condition"?
# Location
https://circleci.com/docs/2.0/reusing-config/

# Explanation
The document has examples of the form
```yml
- when
    condition: << parameters.preinstall-foo >>
```
but it doesn't describe what the allowed values of `condition` are. Is
```yml
- when
    condition: "[ ! -f ./some/file.txt ]"
```
valid? Will it only run the `when` block's `steps` when `some/file.txt` does not exist?

thomasr8:
Rewrote content in Defining Conditional Steps to make the examples more clear. This fixes #3276

# Description

# Reasons


jekyll/_cci2/reusing-config.md
```diff
@@ -940,4 +898,6 @@ workflows:
           preinstall-foo: true
 ```
 
+**Note** Both `condition` and `unless` accept the values `true` or `false`. The value is provided by the value `true` or `false` using parameters (first example), or by the result of a Bash function (second example). Also, please note that`when` only runs when its value is `true`, whereas `unless` only runs when it's value is `false`.
```
michelle-luna:
```suggestion
**Note** Both `condition` and `unless` accept the values `true` or `false`. The value is provided by the value `true` or `false` using parameters (first example), or by the result of a Bash function (second example). Also, please note that `when` only runs when its value is `true`, whereas `unless` only runs when it's value is `false`.
```

mikegee:
This PR introduces a reference to a "second example" using a Bash function. But, that function isn't present in the documentation. I think the author was referring to the second example in the fixed issue. This was confusing for me.

thomasr8:
@mikegee Yes, I intended to reference the second example in the fixed issue. I will reopen a PR and make this more clear. Thank you for the comment!

PR #3308, opened 2019-04-17
Indicate ability to run multiple DBs in docker.

teesloane:
Pertinent [discussion](https://circleci.slack.com/archives/C0KBNJGRH/p1555514865018100)

### DIFF
--- a/jekyll/_cci2/executor-types.md
+++ b/jekyll/_cci2/executor-types.md
@@ -34,7 +34,7 @@ Virtual Environment | `docker` | `machine`
  Build Docker images | Yes <sup>(1)</sup> | Yes
  Full control over job environment | No | Yes
  Full root access | No | Yes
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/executor-types.md
```diff
@@ -34,7 +34,7 @@ Virtual Environment | `docker` | `machine`
  Build Docker images | Yes <sup>(1)</sup> | Yes
  Full control over job environment | No | Yes
  Full root access | No | Yes
- Run multiple databases | No | Yes
+ Run multiple databases | Yes | Yes
```
michelle-luna: we should add a note about the limitation/resource constraint for performance (maybe add an asterisk here in the table and note it below).

teesloane: can do!

PR #3309, opened 2019-04-17
Update aws.md

GERey:
# Description

# Reasons


michelle-luna:
@rosieyohannan this should go in the other repo I think

rosieyohannan:
My understanding is this change is for the 'Trial Install', and the same steps need adding to the install guide in server-documentation

rosieyohannan:
Oh sorry I see this change is to the 2.16 install doc... I thought this only existed in the server-documentation repo? Anyway @GERey is it right to say this new step is for people who are using github enterprise? so I can say if using github it can be skipped? or does everyone on Server use Github Enterprise?

PR #3318, opened 2019-04-19
UUID and UID are different things

mikegee:
I think we want UID here. https://en.wikipedia.org/wiki/User_identifier


drazisil:
Good catch, thanks @mikegee

PR #3323, opened 2019-04-24
Remove whitespace preceeding backticks

kimh:
Probably whitespace in front of code blocks is valid but our translation tool doesn't handle it, so we need to sanitize them. The problem with the translation tool is that the tool doesn't recognize it as code blocks and remove backticks from translated strings.


michelle-luna:
Hi Kim, for the instances inside of steps, this may break the step numbering, so we should test this, will take a look later today, sorry for the delay

kimh:
@michelle-luna Ping

michelle-luna:
Have you tested this locally to see if the step numbering breaks? I think this breaks all the step numbering (all steps will be step 1 or restart to step one after this change).

felicianotech:
It does break numbering.

kimh:
@felicianotech where do we use step numbering?

michelle-luna:
closing for now, re-open if you need it

PR #3338, opened 2019-05-03
Fix misleading statement about pipelines

johnswanson:
We stated that you can "safely go back by disabling" pipelines. This is
not actually true--once you enable pipelines there's no going back.

# Description
- Add a note about how pipelines can't be turned off
- Remove existing comment that "you may safely go back by disabling the radio button for the pipelines feature"
- Add tiny clarification to Limitations section.

# Reasons
We've seen multiple people get confused about this, which is reasonable because our documentation is completely off here.


michelle-luna:
Hi, adding @ndintenfass as a reviewer, build processing was possible to turn off, so I'm curious about how this changed.

ndintenfass:
This is slightly more complicated -- the pricing team made it so that IF you are on a usage-based plan (eg: Performance Plan) and not yet on ~build processing~ pipelines, we then hide the option, making it impossible to self-serve to turn pipelines back off. However, if you are NOT on a usage-based plan you should should still be able to toggle this on and and off. I'm personally a fan of not doing the hiding in the UI, but until that's fixed I agree we should clarify the language on this page, but the nuances might need to be expressed differently than this PR.

ndintenfass:
It's also worth noting that support can use hubot to switch this off, so we might want to mention that here that a support ticket is the path to fix this if people get into a dead end.

PR #3375, opened 2019-05-15
Update machine config ref section.

felicianotech:
Update `machine` executor section of the Configuration Reference. Trying to make it clearer which images are available and to encourage their use. Some images are old, especially the default, and will be going away at some point.


jekyll/_cci2/configuration-reference.md
```diff
@@ -275,31 +275,35 @@ The [machine executor]({{ site.baseurl }}/2.0/executor-types) is configured by u
...
 {: class="table table-striped"}
 
-As a shorthand, you can set the `machine` key to `true`.
 
 Example:
 
-``` YAML
+```yaml
+version: 2.1
+workflows:
```
ndintenfass: There's no need to add this workflow in 2.1 config - it will get added automatically

PR #3413, opened 2019-05-31
Better cache restoring for haskell

smasuda:
# Description
Change the cache keys to save and restore to increase the chance of the cache being reused.

# Reasons
When you use `stack`,  `package.yaml` is updated every time you add, remove , or change the version of your application. Because of this, using `package.yaml` as the cache restore key will decrease the change of using cache. For example, if you add a new dependent package in `package.yaml`,   CircleCI will start building all the libraries from the scratch since it can't find the previous cache.

On the other hand, `stack.yaml` won't change much unless you change the version of LTS to use, so that we can add a secondary cache restore key using `stack.yaml`, to reuse the previous caches even when `package.yaml` is changed.

### DIFF
--- a/jekyll/_cci2/language-haskell.md
+++ b/jekyll/_cci2/language-haskell.md
@@ -115,13 +115,14 @@ compiler as specified in the `stack.yaml` config.
       - restore_cache:
           name: Restore Cached Dependencies
           keys:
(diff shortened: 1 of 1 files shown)


teesloane:
This makes sense, for contributing to our docs! Putting this in the queue to merge!

PR #3429, opened 2019-06-08
fix closing quotes from smart to straight

heavy-stone:
# Description
fix closing quotes from smart to straight

# Reasons
There are smart quotes instead of straight quotes.

Thank you for useful translation!
I found smart quotes and I couldn't run those codes.
I would be glad if you would confirm this!
Thanks!


felicianotech:
Thank you @heavy-stone. :smiley:

PR #3441, opened 2019-06-12
Simplify export and login instructions

bcardiff:
# Description
Simplify export and login instructions without explicitly creating files.
Bonus: use a token that can push only to a specific channel.

# Reasons
Less hardcoded stuff that will work if the snapcraft tool is run in other than `~/project`.

### DIFF
--- a/jekyll/_cci2/build-publish-snap-packages.md
+++ b/jekyll/_cci2/build-publish-snap-packages.md
@@ -64,25 +64,23 @@ snapcraft login
 snapcraft export-login --snaps my-snap-name --channels edge - | xsel --clipboard
 ```
 
-> Note: the exported token can only upload this snap to the channel you specified above (in this example, edge).
+1. Create a Snapcraft token on your local machine that we upload to CircleCI. Assuming your local machine already has these tools installed and you are logged in to the Snapcraft Store (`snapcraft login`), we use the command `snapcraft export-login --snaps my-snap-name --channels edge -` to generate a token that can only upload `my-snap-name` to the channel `edge`. As we don't want this token visible to the public or stored in the Git repository, we will store it in a [private environment variable](https://circleci.com/docs/2.0/env-vars/#adding-environment-variables-in-the-app) called `$SNAPCRAFT_TOKEN`.
 
-1. Create a Snapcraft "login file" on your local machine that we upload to CircleCI. Assuming your local machine already has these tools installed and you are logged in to the Snapcraft Store (`snapcraft login`), we use the command `snapcraft export-login snapcraft.login` to generate a login file called `snapcraft.login`. As we don't want this file visible to the public or stored in the Git repository, we will base64 encode this file and store it in a [private environment variable](https://circleci.com/docs/2.0/env-vars/#adding-environment-variables-in-the-app) called `$SNAPCRAFT_LOGIN_FILE`.
-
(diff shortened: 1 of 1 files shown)


bcardiff:
@felicianotech I also change the env variable to `$SNAPCRAFT_TOKEN` since it seems more appropriate.

felicianotech:
Hi @bcardiff,

Thank you. So This morning I went through these steps myself and I found a few issues.

1. Personally I don't think limiting the token to a specific release channel is useful in our environment. Ideally the CI process is publishing to multiple channels. Maybe `master` to edge and tagged releases to stable. The same could be said for snap name when using CircleCI Context. I publish over 5 snaps via CircleCI for example.
2. `snapcraft export-login` requires user input, so we can't pass it directly to `xsel`. This seems to be a change to Snapcraft though because I don't remember it doing that before. I'm running Snapcraft v3.6.
3. Passing the token to `snapcraft login --with -` doesn't seem to work for me unless the file/token is base64 encoded.

Are you not seeing these issues? Do you have a working build with these changes that I can look at?

My info:

Snapcraft v3.6
Snap v2.39
OS local: Ubuntu 19.04
OS CircleCI: Ubuntu 16.04 based Snapcraft Docker Image

bcardiff:
Hi @felicianotech 

1. I'm ok if you prefer to leave it as before regarding the channel restriction. In our use case the automated build will always push to edge. Tagged releases will be tagged with grade stable and they are manually pushed to stable once the other artifacts of the release are ready. Having the least required scope in the CI seems appropriate. But is definitely your call what you want to advertise in the docs.

2. You are right. The xsel should not be used. I only used the first part of the command since I'm on mac. Would it be better to give the complete output to guide the user? This is what I actually use.

```sh
$ snapcraft export-login --snaps my-snap-name --channels edge,edge/* -
Enter your Ubuntu One e-mail address and password.
If you do not have an Ubuntu One account, you can create one at https://snapcraft.io/account
Email: jdoe@example.org
Password: *******

Exported login starts on next line:
W5y... (stripped) ...BeK

Login successfully exported and printed above. This can now be used with

    echo "<login>" | snapcraft login --with -

to log in to this account with no password and have these
capabilities:

snaps:       ['my-snap-name']
channels:    ['edge', 'edge/*']
permissions: ['package_access', 'package_manage', 'package_push', 'package_register', 'package_release', 'package_update']
expires:     2020-06-17T17:01:10.405759
```

3. As shown in the previous output the exported login `W5y... (stripped) ...BeK` should be able to be used with the `snapcraft login --with -`. We are using it already in our workflow. Maybe the `xsel` issue added some extra information in the value set in the `$SNAPCRAFT_TOKEN`.

---

I'm happy to keep iterating, but if this PR is more a nudge to bring the snap instructions up to date with a different approach & phrasing I'm also happy.

felicianotech:
Hey @bcardiff,

Thank you for this PR but for now, I'm going to close this. As determined in my post over in the Snapcraft Forum (https://forum.snapcraft.io/t/snapcraft-docker-image-update-broke-python-plugin/11976?u=felicianotech), I need to rework my Snapcraft Docker images for CircleCI, which this guide is based on.

Once I get those images reworked (I believe before August is over), I'll come back to this doc and give it a complete overhaul as I  might be making a CircleCI Orb for Snapcraft as well.

If you want to track that work, you can do so here: https://github.com/cibuilds/snapcraft/issues/8

likon39:
Ok.


Den fre 9 aug. 2019 18:34Ricardo N Feliciano <notifications@github.com>
skrev:

bcardiff:
Thanks for the heads up @felicianotech

PR #3468, opened 2019-06-21
Add additional certificate installation process

demophoon:
# Description
This commit adds in an additional, CLI only, variant of installing the
CircleCI Server SSL certificate.

# Reasons
This was something that I came across while deploying CircleCI Server which made renewing certificates a much easier process to automate with Lets Encrypt and may be useful for other users to be made aware of.


rosieyohannan:
This looks good! I'll get it into the next issue of the PDFs too :-).

PR #3470, opened 2019-06-22
Fix examples in reusing-config.md

uraway:
# Description
Added `run` keyword to some examples

# Reasons
examples cannot be processed


teesloane:
@uraway Good catch, thanks for contributing!

PR #3482, opened 2019-06-25
update sidenav to make landing pages more obvious

michelle-luna:
# Description
this moves each landing page underneath the ALL CAPS heading 

# Reasons
5 folks said they never clicked the ALL CAPS headings (so they missed this content for a long time)


gtback:
I didn't know those were clickable until I saw this :-)

Issue #3497, opened 2019-07-02
Android images have a list of packages already installed


kevin-lot:
# Locations
https://circleci.com/docs/2.0/circleci-images/#android

# Details
After [verification](https://hub.docker.com/r/circleci/android/dockerfile), Android images have all this packages already installed.

```
With the exception of Android images, all images include the following packages, installed via apt-get:

bzip2
ca-certificates
curl
git
gnupg
gzip
locales
mercurial
net-tools
netcat
openssh-client
parallel
sudo
tar
unzip
wget
xvfb
zip
```

teesloane:
Thanks for raising this issue @kevin-lot — we are looking into confirming we can update our docs as you've suggested! Thanks for bringing this to our attention.

PR #3507, opened 2019-07-05
Update: Android also installs additional tools.

teesloane:
Android docker images seem to also include the tools that other convenience images have, as noted in #3497 . @iynere do you mind confirming this?

### DIFF
--- a/jekyll/_cci2/circleci-images.md
+++ b/jekyll/_cci2/circleci-images.md
@@ -210,7 +210,7 @@ use the `circleci/postgres:9.5-postgis-ram` image.
 
 ## Pre-installed Tools
 
-All convenience images have been extended with additional tools, installed with `apt-get`
+All convenience images have been extended with additional tools, installed with `apt-get`:
 
 - `bzip2`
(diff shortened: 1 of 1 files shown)


iynere:
@teesloane probably, but not the exact same set of tooling. i think the reason that text is there is that we use different build logic for our android images. so while the toolset is likely largely the same, it's not identical. i defer to @felicianotech though. i think, basically, though, in general, it'd be fine to remove this caveat.

felicianotech:
@iynere You're the one that made this change to Docs in the first place. You don't remember why?

I just tested and these tools are indeed installed in the Android image. I'm not 100% sure if they're install the same way.

iynere:
@felicianotech lol well then it's probably for this reason:

> i think the reason that text is there is that we use different build logic for our android images. so while the toolset is likely largely the same, it's not identical.

teesloane:
It sounds safe to remove the note. Thanks @iynere and @felicianotech.

michelle-luna:
fixes #3497

PR #3512, opened 2019-07-08
Add note to compress if uploading many artifacts

nakkurt:
# Description
Adds a note to docs that compressing artifacts and uploading them as a single file will significantly speed up artifact uploading step if there are hundreds of artifacts.

# Reasons
This note isn't currently surfaced in _docs_ and is only visible in 'known issues' section of support site.
This PR helps folks understand this sooner and prevents them from experiencing slow artifact uploads.

# Background
If there are hundreds of small artifacts then uploading them separately [seems to be significantly slower](https://support.circleci.com/hc/en-us/articles/360024275534?input_string=store_artifacts+step) than uploading a single, compressed file with all these artifacts.

### DIFF
--- a/jekyll/_cci2/artifacts.md
+++ b/jekyll/_cci2/artifacts.md
@@ -74,7 +74,9 @@ jobs:
           path: /tmp/artifacts
 ```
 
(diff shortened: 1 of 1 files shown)


michelle-luna:
thank you!

PR #3519, opened 2019-07-09
typo - megabits to megabytes

jeffgreenca:
# Description
Change Mb -> MB

# Reasons
I really hope the recommendation is 500MB max cache and not 500Mb (~62MB) max cache.


rosieyohannan:
Good spot! Thanks @jeffgreenca :-)

PR #3525, opened 2019-07-11
haskell: Fix examples producing garbage characters in logs

nh2:
# Description

Fixes output like this:

image

# Reasons

See the added comment, and the upstream source https://github.com/commercialhaskell/stack/issues/843 on this problem.

CircleCI cannot interactively delete characters from the output like a terminal can,
so stack needs to be told that CircleCI is not a terminal.


teesloane:
Wow, I didn't know this was possible. Thank you for contributing to our docs, we really appreciate it!!

Issue #3536, opened 2019-07-15
mysql-client is not available no longer


nagaakihoshi:
# Locations
https://circleci.com/docs/2.0/postgres-config/#example-mysql-project

```
sudo apt-get install mysql-client
```

# Details

Current result

```
Package mysql-client is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'mysql-client' has no installation candidate
```

I should set another package 'default-mysql-client' according to official guide.

https://packages.debian.org/stable/database/

So rewrite the line as...

```
sudo apt-get install default-mysql-client
```

nagaakihoshi:
Create a pull request as well.

https://github.com/circleci/circleci-docs/pull/3537

teesloane:
Ah! Thank you for catching this. It looks like it's related to our [image updates](https://discuss.circleci.com/t/docker-convenience-images-updates-july-2019/31239/5). Thanks for putting up a PR as well. I will go ahead and approve that.

teesloane:
Thanks again! Closed by #3546

PR #3537, opened 2019-07-15
mysql-client is not available no longer

nagaakihoshi:
Refer: https://github.com/circleci/circleci-docs/issues/3536

# Description
mysql-client is not available no longer. So change the package from `mysql-client` to `default-mysql-client` .

# Reasons
https://github.com/circleci/circleci-docs/issues/3536

### DIFF
--- a/jekyll/_cci2/postgres-config.md
+++ b/jekyll/_cci2/postgres-config.md
@@ -245,7 +245,7 @@ jobs:
       - run:
           name: Install MySQL CLI; Import dummy data; run an example query
           command: |
-            sudo apt-get install mysql-client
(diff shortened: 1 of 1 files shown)


teesloane:
I'm just looking into why this is failing, but otherwise it's good to merge. Thank you for contributing to our docs!

teesloane:
Hi there @nagaakihoshi - Do you mind pulling master into your branch and re-pushing? We pushed an update that was causing our deps CI step to pass, so we should be good now!

nagaakihoshi:
I re-created a new pull request.

https://github.com/circleci/circleci-docs/pull/3546

so I close this.

PR #3543, opened 2019-07-16
Update executor statements to include windows.

teesloane:
Did a pass on the docs to add "Windows" wherever we are talking about executor types.

Should be merged after #3503


michelle-luna:
Thank you! Maybe there should be something in the circleci-images.md doc? Also, where should we add the info about this windows orb, if anywhere? And can we add the the info from the windows preview docs to the faq.md? We must also say everywhere that this is not supported on installed versions of CircleCI.

teesloane:
That makes sense! I will do that. Also, we can't technically merge this until the windows guide is live as as the new executor types listed will often link to the `hello-world-windows` - so until that link is real our HTML proofer will fail on a non-existing link.

teesloane:
This is good for re-review — @appplemac  - if you have a chance, can you do a quick once over on this?

PR #3546, opened 2019-07-18
mysql-client is not available no longer

nagaakihoshi:
Refer: #3536

Description
mysql-client is not available no longer. So change the package from mysql-client to default-mysql-client .

Reasons
#3536


teesloane:
Thank you!

PR #3573, opened 2019-07-25
Remove the deprecated Xcode 9.1 and 10.2.0 images

appplemac:
Leaving both images in the doc as customers migrating to other versions might want to compare dependencies between the images. Will remove once we are past the EOL period.


marcomorain:
LGTM

PR #3580, opened 2019-07-28
Update sidenav.yml

michelle-luna:
# Description
This updates the sidenav for the orb docs structure changes. This will be broken and once we confirm this the the correct set of links and merge all the new files, we can make an new PR for it.

# Reasons
Moving around the content  and adding a bunch of files, so we need to surface those in the navigation.


jekyll/_data/sidenav.yml
```diff
@@ -71,21 +71,39 @@ en:
...
+        - name: Set up the CircleCI CLI
+          link: 2.0/orb-author-cli/
+        - name: Author an Orb
+          link: 2.0/orb-author/
+        - name: Test and Publish Your Orb
+          link: 2.0/orb-author-validate-publish/
+    - name: Using Orbs
+      link:      
+      children:
+        - name: Configure Your Platform to Work With Orbs
```
ndintenfass: "Your Platform" -- do we use that term elsewhere to refer to project config? Suggest you reserve "platform" for other use.

teesloane:
This new organization makes sense to me 👍

PR #3587, opened 2019-07-31
Update/faq

mvxt:
# Description
Added similar paragraph about payment plan piggybacking to usage based plans section.

# Reasons
In the FAQ we only specify piggybacking orgs / payment plans underneath the "container plans" section and not the "usage based plans" even though we can do it for both. Customer + a few internal people confused.


teesloane:
Thanks for catching this updating our docs @mvxt ! Merging!

PR #3604, opened 2019-08-06
(wip) Use widdershins+slate to compile v2.x of API

teesloane:
**Overview**
- Add updates to how our api docs are compiled.
- API 1.x is still compiled by manually written files
- 2.x is compiled from an openapi json spec.

[Here is an example output](https://47288-48750547-gh.circle-artifacts.com/0/circleci-docs/api/v2/index.html#circleci-api-v2) from the new build process of how the compiled docs would look.

---
**Details**
The changes in this Pr:

- installs a new tool, [widdershins](https://github.com/Mermade/widdershins), which allows us to /generate a slate documentation guide directly from an openapi spec/. There are a lot of tools that can do this, we do not have to use widdershins. @thomasr8 researched and found it as it can create a [slate](https://github.com/lord/slate)-compatible document, which is what our [1.x documentation uses.](https://circleci.com/docs/api/#section=reference)
- We then use slate to, again, rebuild our documentation, but this time (!) it is fully automated; including being given a host of language-usage examples. This file documentation, as per the config changed in this pr, is dropped into a `v2` folder. This means that our documentation for v2 would be available at `https://circleci.com/docs/api/v2/#section=reference` (however, this can be changed aswell)

Other than that, it appears that there may be some additional details that we can add to the spec that helps group the documents better in the compiled output's sidebar.

To do this, it appears we will need to use OpenAPI's concept of ["tags"](https://swagger.io/docs/specification/grouping-operations-with-tags/), but I don't know if this is something that can be added to the inline documentation such as found [here](https://github.com/circleci/circle/blob/9557d7c50730e017ae0b8b644c1478bd11c8dc59/src/circle/http/api/v2/workflows.clj#L158-L166)

---

**Blocked**:

This is currently failing on CI because a few links in the [open api spec](https://circleci.com/api/v2/openapi.json) is referring to internal spec definitions that don't seem to exist. Please see [this failing job](https://circleci.com/gh/circleci/circleci-docs/47288#artifacts/containers/0) ; ie, our HTML proofer is finding some broken links that get compiled from the spec.


sulami:
The example output seems to be slightly confused about body parameters

teesloane:
Yes, it appears we'll have to go through some of the docstrings with a bit of care to line them up. Some of the comments make for good "sidebar material" while others are not:

Screen Shot 2019-08-07 at 9 54 19 AM

thomasr8:
@teesloane and @sulami : I am going to be working on the docstring descriptions this week in the API spec. I will add you both to the PR as reviewers so we can make sure that it displays correctly in Widdershins.

gordonsyme:
We can add tags to the OpenAPI specs, that's "just" a matter of deciding on which routes to group and the values we want to group them with.

The tool we are using lets us add any key to the `:open-api` map as long as it's prefixed with the `:swagger/` namespace. So tags can be added using the key `:swagger/tags` and will be rendered as the key `tags` when generating the OpenAPI spec.

The link failures look like bugs on our side not matching operation IDs correctly, again, ought to be easy to correct.

teesloane:
Thanks @gordonsyme and @thomasr8 !

teesloane:
This is building succesfully now. The 2.x api documentation will be visible at `https://circleci.com/docs/api/v2` when we merge. 

Current state of generated docs is [visible here](https://47915-48750547-gh.circle-artifacts.com/0/circleci-docs/api/v2/index.html#returns-a-pipeline-by-id)

teesloane:
This is building succesfully now. The 2.x api documentation will be visible at `https://circleci.com/docs/api/v2` when we merge. 

Current state of generated docs is [visible here](https://47915-48750547-gh.circle-artifacts.com/0/circleci-docs/api/v2/index.html)

ndintenfass:
Can we merge this now and add the link as part of the stability announcement? Would be nice to start seeing the docs live as we tweak the various names/titles/descriptions

sulami:
What's up with the `»` that are before field names?

ndintenfass:
@sulami's comment should be fixable in the underlying schema, yes? @teesloane @michelle-luna can we go ahead and start having these live at a URL but unlinked prior to the announcement? It will be handy to be able to have a stable place to view changes to the underlying schema as we get ready for the stability blog post.

teesloane:
Yep, I think we're good to merge.

Issue #3563, opened 2019-07-22
List Code with Line Number


jarvar:
# Missing Information
While I find the content easily understandable, I find myself shuttling to code and back to the explanation of the written code. I think it would be clearer without having to count lines of code to pinpoint what is being explained. Just a thought.

teesloane:
Hi @jarvar , thanks for raising this issue. I'm not sure if I understand correctly. Do you mean that it would be easier if we had line numbers in our code block examples? If you could provide an example document which we could improve that might help as well.

Thanks!

teesloane:
Closing due to inactivity.

likon39:
Code# 5794


Den fre 9 aug. 2019 17:08Tyler <notifications@github.com> skrev:

likon39:
need your help can you or do you know someone who can fix driving license
would be very kind of you thank you


Den fre 9 aug. 2019 17:12chann Likone <likon39@gmail.com> skrev:

PR #3639, opened 2019-08-09
add line numbers to large codeblocks

rosieyohannan:
# Description
Adding line numbers per code block, rather than globally. It is possible but the generated output looks bad:

Screenshot 2019-08-09 at 12 42 03

Needs further investigation.

# Reasons
A customer has reported that it's hard to follow code walk-throughs when the blocks are long, without line numbers.


rosieyohannan:
I've realised the way this looks is due to the code block being formatted as a table when line numbers are enabled, so I made them disappear by changing them to the same colour as the background in CSS. Just need to see if I can reduce the space at the top and bottom of the block now, but should have something usable soon...

Screenshot 2019-08-12 at 20 30 41

teesloane:
This looks great!

rosieyohannan:
Done!

Screenshot 2019-08-12 at 21 31 30

rosieyohannan:
I Just need to add this to the rest of the guided code blocks and then ready to merge.

rosieyohannan:
fixes #3563

PR #3646, opened 2019-08-13
replace getting started links to top of each section in sidenav

rosieyohannan:
# Description
Since the change made recently to add the getting started info fro each section into its own section in the sidenav, I realised the capitalised section headings are still clickable, and they all take you back to the welcome page. This PR changes this so each capitalised heading links to the getting started page for that section.


rosieyohannan:
Haha, yeah I tried to get it so those links weren’t clickable but couldn’t find a way.

PR #3647, opened 2019-08-13
Guide: Getting Started with the CircleCI CLI

teesloane:
A document providing an overview of getting started with the CircleCI CLI. This guide walks a user through how they would create a project on CircleCI and use the CLI along the way.


jekyll/_cci2/local-cli-getting-started.md
```diff
@@ -0,0 +1,168 @@
...
+
+Try running the "build" job locally:
+
+```sh
+circleci local execute
+```
+
+This will pull down the docker image you have specified, in this case `circleci/ruby::2.4.2-jessie-node` and run the job. This may take a bit of time depending on the size of the docker image you are using.
+
+You should quite a bit of a text in your terminal. The last few lines of output should look similar to this:
```
michelle-luna:
```suggestion
You should see quite a bit of a text in your terminal. The last few lines of output should look similar to this:
```

PR #3661, opened 2019-08-20
fix code block formatting

rosieyohannan:
# Description
There were two changes to this code at around the same time that led to a formatting error - upping config version and including line numbers in the code blocks. This PR fixes the error.


michelle-luna:
ty!

PR #3685, opened 2019-08-30
update JavaScript page to use modern versions

afeld:
...of the CircleCI config, NodeJS/NPM, and MongoDB. Most importantly, now that NPM generates a `package-lock.json`, that should be used for the dependency cache key, as suggested on [the caching page](https://circleci.com/docs/2.0/caching/).

### DIFF
--- a/jekyll/_cci2/language-javascript.md
+++ b/jekyll/_cci2/language-javascript.md
@@ -14,12 +14,12 @@ This document provides a walkthrough of the [`.circleci/config.yml`]({{ site.bas
 
 ## Quickstart: Demo JavaScript Node.js Reference Project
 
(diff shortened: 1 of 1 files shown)


michelle-luna:
Thank you! However, this change would depend on a change to the deo app config.yml here (we wait to make the docs changes until we have made the change in the demo app) https://github.com/CircleCI-Public/circleci-demo-javascript-express/blob/master/.circleci/config.yml

teesloane:
Hiya @afeld ! Thanks for making this PR. I've updated the dependencies of our [Javascript project](https://github.com/CircleCI-Public/circleci-demo-javascript-express/pull/17), so we should be good to merge this upon a final review.

PR #3688, opened 2019-08-30
Update Gems with needed security fix.

felicianotech:
Update Gems, including Nokogiri which GitHub was complaining about.

@rosieyohannan After that [PR I opened](https://github.com/algolia/html-extractor/pull/2) for upstream, their gem was updated so now we can update.


michelle-luna:
thank you!

PR #3689, opened 2019-08-30
Clarify cache key templates.

teesloane:
More survey results are indicating some confusion with caching. This PR tries to clear up how templates work and how it's possible to interpolate variables and parameters in them.

### DIFF
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -195,7 +195,7 @@ A cache-key is a _user-defined_ string that corresponds to a data cache. A cache
 {% raw %}myapp-{{ checksum "package-lock.json" }}{% endraw %}
 ```
 
-will output a unique string to represent this key. Here, the example is using a [checksum](https://en.wikipedia.org/wiki/Checksum) to create a unique string that represents the contents of a `package-lock.json` file.
+The above example will output a unique string to represent this key. Here, the example is using a [checksum](https://en.wikipedia.org/wiki/Checksum) to create a unique string that represents the contents of a `package-lock.json` file.
 
 So, for the sake of demonstration, the above example might put out a string that looks like so:
 
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -228,11 +228,12 @@ Template | Description
 {% raw %}`{{ arch }}`{% endraw %} | Captures OS and CPU (architecture, family, model) information. Useful when caching compiled binaries that depend on OS and CPU architecture, for example, `darwin-amd64-6_58` versus `linux-amd64-6_62`. See [supported CPU architectures]({{ site.baseurl }}/2.0/faq/#which-cpu-architectures-does-circleci-support).
 {: class="table table-striped"}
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/caching.md
```diff
@@ -189,6 +189,22 @@ neither `save_cache` nor `restore_cache` support globbing for the `paths` key.
...
 
+A cache-key is a _user-defined_ string that corresponds to a data cache. A cache-key can be created by interpolating **dynamic values** — these are called **templates**. Anything you see in a cache-key between curly braces is a template. Consider the following example:
+
+```sh
+{% raw %}myapp-{{ checksum "package-lock.json" }}{% endraw %}
+```
+
+The above example will output a unique string to represent this key. Here, the example is using a [checksum](https://en.wikipedia.org/wiki/Checksum) to create a unique string that represents the contents of a `package-lock.json` file.
+
+So, for the sake of demonstration, the above example might put out a string that looks like so:
```
michelle-luna:
```suggestion
The example may output a string that looks like the following:
```

michelle-luna: just making this a bit simpler for translation ('for the sake of' is colloquial and harder to translate)

jekyll/_cci2/caching.md
```diff
@@ -214,6 +230,10 @@ Template | Description
 
 **Note:** When defining a unique identifier for the cache, be careful about overusing template keys that are highly specific such as {% raw %}`{{ epoch }}`{% endraw %}. If you use less specific template keys such as {% raw %}`{{ .Branch }}`{% endraw %} or {% raw %}`{{ checksum "filename" }}`{% endraw %}, you’ll increase the odds of the cache being used. But, there are tradeoffs as described in the following section.
```
michelle-luna: maybe this should be a section called Notes on Using Keys and Templates with a bulleted list at this point.

teesloane: Good call. Done!

jekyll/_cci2/caching.md
```diff
@@ -189,6 +189,22 @@ neither `save_cache` nor `restore_cache` support globbing for the `paths` key.
...
+```sh
+{% raw %}myapp-{{ checksum "package-lock.json" }}{% endraw %}
+```
+
+The above example will output a unique string to represent this key. Here, the example is using a [checksum](https://en.wikipedia.org/wiki/Checksum) to create a unique string that represents the contents of a `package-lock.json` file.
+
+The example may output a string that looks like the following:
+
+```sh
+{% raw %}myapp-dlkj3oadlkfgjadflkje3ijaoi43oi49odlkja$lakdjfg#lakdj4Djlae<etc>{% endraw %}
```
shuuji3: [Caching Dependencies - CircleCI](https://circleci.com/docs/2.0/caching/) says:
> A base64 encoded SHA256 hash of the given filename’s contents, so that a new cache key is generated if the file changes.

So I think the characters such as `$` and `#` should not appear in the result of the checksum.
In my job result page, I can see the string something like this: `myapp-+V9MbGBceJh_zOWQI6ga3gDPGB26mbhWJxUYmiEidlQ=`

teesloane: Ha! Good point @shuuji3 - I was going for "just imagine a garbled-string to be the checksum" — I should have used something more realistic to a checksum. I'll make an update and remove the bogus chars.

PR #3690, opened 2019-08-30
Clarify how delimit a regular expression for filtration.

aterweele:
# Description
say `slashes` instead of `'/s'`.

# Reasons
`'/s'`can make it seem as if a regular expression is written like `/sv.*/s`.


aterweele:
(force-push because I just noticed a typo in the commit message)

PR #3705, opened 2019-09-04
Delete (unnused) image that causes dirty checkout problems on macOS

stig:
This is due to macOS filesystem being case insensitive, and there's a different file in the same directory that differs only in case. I kept the capitalised one as that is referenced from two different templates, while this one is not.

Here's what I saw:

    $ git checkout jekyll/assets/img/docs/Orbs-Registry.png ; git status

    On branch master
    Your branch is up to date with 'origin/master'.

    Changes not staged for commit:
      (use "git add <file>..." to update what will be committed)
      (use "git checkout -- <file>..." to discard changes in working directory)

    	modified:   jekyll/assets/img/docs/orbs-registry.png

    no changes added to commit (use "git add" and/or "git commit -a")

Thanks to Rosie for helping me figure out what the issue was!


teesloane:
Ah! So glad to be rid of this hanging file! Thanks @stig !

PR #3720, opened 2019-09-11
Update ecs-ecr.md

thomasr8:
Fixed AWS-ECS version to 0.0.10

# Description

# Reasons


felicianotech:
@thomasr8 What's the intention here? If you're trying to update the orbs to their latest versions in the examples, this isn't the latest version.

I'd also warn against keeping docs like this always up to date with the latest version as it can be hard to keep up. One way to help would be to drop the patch release integer. This will keep the version correct longer, and will always point users to the latest security/bug fix release.

If I missed the point, sorry, please explain.

PR #3679, opened 2019-08-29
Update Docs footer in line with website

rosieyohannan:
https://circleci.atlassian.net/browse/CIRCLE-15711?atlOrigin=eyJpIjoiZTBlNjljMmMzNWFmNDdjYWFkZjc0YmEyOTZlNjIyNmIiLCJwIjoiaiJ9

PR #3727, opened 2019-09-13
Add side nav language picker.

teesloane:
This branched off of #3679 , and so we should wait on that to merge (there will be merge conflicts).

---

Kapture 2019-09-13 at 11 24 06

I used the native <select> but can switch it over to something custom (like the language picker at the bottom of the footer on the main site) but that will take a bit more time.


src-js/lang.js
```diff
@@ -0,0 +1,54 @@
...
+
+function reloadWithNewLocale(langCode) {
+  window.location.href = "https://circleci.com/" + languages[langCode].url
+}
+
+// Sets the sidebar language picker to the currently selected language
+function handleSetLanguageOnLoad() {
+  var currentLang = window.currentLang; // current lang is set in _includes/sidebar.html
+  var langSelect = getElById("sidebarLangSelect");
+  for(var i, j = 0; i = langSelect.options[j]; j++) {
```
trevor-circleci: just a note: there may be a slightly more efficient way to do this with the `Array.forEach()` function.

teesloane:
Closing this to incorporate changes into #3679 instead. Thanks for the review @trevor-circleci .

PR #3734, opened 2019-09-16
add margin to the bottom of codeblocks

rosieyohannan:
# Description
I keep noticing formatting looks off across the whole of docs due to paragraphs having a 10px margin-bottom and codeblocks having no bottom margin - the code blocks look too close to the text below them. Not sure if it's always been like this 🤷🏻‍♀️:
Screenshot 2019-09-16 at 20 16 51

This PR adds 10px to make things more even:
Screenshot 2019-09-16 at 20 16 44


thomasr8:
@rosieyohannan This change looks good. I always disliked the spacing between code blocks and text, and this seems like a good fix for this. Great stuff!

rosieyohannan:
Thanks @thomasr8 ... urrr should have said there is a change in there for the orbs image too because for some reason the image wasn't there in my PR and was making the build fail. Very strange!

teesloane:
Looks good @rosieyohannan !

PR #3739, opened 2019-09-18
[WIP]2.18/resource classes

rosieyohannan:
# Description
Changes to explain how to customize resource classes in Server from v2.18


jekyll/_cci2/vm-service.adoc
```diff
@@ -21,9 +21,11 @@ image::vm-service.png[Configuring VM Service on CircleCI Server]
 
 == Configuration
 
-To configure VM service, it is best practice to select the AWS EC2 option in the Management Console Settings, which will allow CircleCI to run remote Docker and `machine` executor jobs using dedicated EC2 instances.
+CAUTION: Any changes to management console settings requires downtime while the CircleCI application restarts.
```
rosieyohannan:
```suggestion
CAUTION: Any changes to management console settings require downtime while the CircleCI application restarts.
```

jekyll/_cci2/customizations.adoc
```diff
@@ -134,3 +134,80 @@ export CIRCLE__OUTER__LOGIN_BANNER_MESSAGE="<insert-your-message-here>
...
+sudo mkdir /etc/circleconfig/picard-dispatcher
+```
+. Run the following:
++
+```shell
+sudo vim /etc/circleconfig/picard-dispatcher/resource-definitions.edn
+```
+. Add your required customizations to the file, then save and exit vim with `:wq` - see below for options and formatting.
++
+WARNING: It is important to use the IDs listen in the example below for the `machine` resources classes. IDs can, however, be changed for Docker resource classes.
```
GERey: I'm wondering if we should actually phrase this to just tell people not to change it. I can see how the machine parts would be confusing but realistically I'd rather not put the burden on support against seeing what weird thing people would change in the machine stanzas.

jekyll/_cci2/vm-service.adoc
```diff
@@ -57,18 +64,15 @@ and see https://github.com/circleci/image-builder/blob/picard-vm-image/provision
...
-To enable the use of these custom images for your installation, you will need to provide the individual AMI IDs in the Settings accessible from the Management console (for example, `<your-circleci-id>.com:8800`). Once you have provided the IDs, save your settings and restart the CircleCI application.
-
-// add screenshpt when available
-// .Custom VM Service Images
-// image::vm-service-custom.png[Custom VM Service Images]
+To enable the use of these custom images for your installation, you will need to provide the individual AMI IDs in the Settings accessible from the Management console (for example, `<your-circleci-id>.com:8800`).
 
-CAUTION: Any changes to management console settings requires downtime while the CircleCI application is restarted.
+== Defining Instance Types
+// Here I need to tell people to enable the new machine large resource, and then enter instance types to use for default and large - but how does instance type correspond to the CPU/RAM stuff we have defined in picard-dispatcher for this release? Are "Instance type" and "CPU/RAM requirements" totally separate in this context?
```
GERey: yeah the answer is nuanced but instance type and cpu/ram are prettying much separate. We do technical have this tiny docker container that helps manage the usage of those vm's ( which is what is being modifying in the machine parts ) but realistically I don't think we have to tell people here how to enable those as long as we have it elsewhere. The machine Picard dispatcher shouldn't be changed at all for machine types.

jekyll/_cci2/customizations.adoc
rosieyohannan: Yeah agreed

rosieyohannan: done

jekyll/_cci2/vm-service.adoc
rosieyohannan: >  The machine Picard dispatcher shouldn't be changed at all for machine types.

But isn't that what's going on with the 'customize your resource classes' new feature in this release? they have the option to edit CPU and RAM for the default and large machine types? As well as for Docker?

rosieyohannan: But anyway, in this section, I will just say they need to provide instance types for default use, and for use when `resource_class: large` is used

rosieyohannan:
I'm going to merge this into the main v2.18 branch to start getting everything together :-)

PR #3743, opened 2019-09-19
Add a User Security "Checklist" to Security document.

teesloane:
We've been discussing / have had a request for providing user's with a checklist of how they can properly ensure they are using the platform securely. I've sourced a few points (thanks @glenjamin  and @mvxt) and am putting this up for consideration. All feedback / input appreciated.

### DIFF
--- a/jekyll/_cci2/security.md
+++ b/jekyll/_cci2/security.md
@@ -112,7 +112,7 @@ If you are getting started with CircleCI there are some things you can ask your 
   build needs and rotate secrets regularly.
   - It is important to rotate secrets regularly in your organization, especially as team members come and go. 
   - Rotating secrets regularly means your secrets are only active for a certain amount of time, helping to reduce possible risks if keys are compromised.
-  - Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build. Consider carefully adjudicating the role and permission systems of other platforms you use outside of CircleCI; for example, when using something such as IAM permissions on AWS, you should ensure that you properly assign authorization levels on a per user basis.
+  - Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build. Consider carefully adjudicating the role and permission systems of other platforms you use outside of CircleCI; for example, when using something such as IAM permissions on AWS, or Github's [Machine User](https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users) feature. 
 - Sometimes user misuse of certain tools might accidentally print secrets to stdout which will land in your logs. Please be aware of:
   - running `env` or `printenv` which will print all your environment variables to stdout.
   - literally printing secrets in your codebase or in your shell with `echo`.
--- a/jekyll/_cci2/security.md
+++ b/jekyll/_cci2/security.md
@@ -113,6 +113,10 @@ If you are getting started with CircleCI there are some things you can ask your 
   - It is important to rotate secrets regularly in your organization, especially as team members come and go. 
   - Rotating secrets regularly means your secrets are only active for a certain amount of time, helping to reduce possible risks if keys are compromised.
   - Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build. Consider carefully adjudicating the role and permission systems of other platforms you use outside of CircleCI; for example, when using something such as IAM permissions on AWS, you should ensure that you properly assign authorization levels on a per user basis.
+- Sometimes user misuse of certain tools might accidentally print secrets to stdout which will land in your logs. Please be aware of:
+  - running `env` or `printenv` which will print all your environment variables to stdout.
+  - literally printing secrets in your codebase or in your shell with `echo`.
+  - programs or debugging tools that print secrets on error.
 - Consult your VCS provider's permissions for your organization (if you are in an organizations) and try to follow the [Principle of Least Privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). 
 - Use Restricted Contexts with teams to share environment variables with a select security group. Read through the [contexts]({{ site.baseurl }}/2.0/contexts/#restricting-a-context) document to learn more.
 - Ensure you audit who has access to SSH keys in your organization.
--- a/jekyll/_cci2/security.md
+++ b/jekyll/_cci2/security.md
@@ -110,11 +110,13 @@ If you are getting started with CircleCI there are some things you can ask your 
 
 - Minimise the number of secrets (private keys / environment variables) your
   build needs and rotate secrets regularly.
-- Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build.
+  - It is important to rotate secrets regularly in your organization, especially as team members come and go. 
+  - Rotating secrets regularly means your secrets are only active for a certain amount of time, helping to reduce possible risks if keys are compromised.
+  - Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build. Consider carefully adjudicating the role and permission systems of other platforms you use outside of CircleCI; for example, when using something such as IAM permissions on AWS, you should ensure that you properly assign authorization levels on a per user basis.
 - Consult your VCS provider's permissions for your organization (if you are in an organizations) and try to follow the [Principle of Least Privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). 
 - Use Restricted Contexts with teams to share environment variables with a select security group. Read through the [contexts]({{ site.baseurl }}/2.0/contexts/#restricting-a-context) document to learn more.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/security.md
```diff
@@ -119,6 +104,19 @@ Following are the system events that are logged. See `action` in the Field secti
 - **success:** A flag to indicate if the action was successful.
 - **request:** If this event was triggered by an external request this data will be populated and may be used to connect events that originate from the same external request. The format is a JSON blob containing `id` (the request ID assigned to this request by CircleCI), `ip_address` (the original IP address in IPV4 dotted notation from which the request was made, eg. 127.0.0.1), and `client_trace_id` (the client trace ID header, if present, from the 'X-Client-Trace-Id' HTTP header of the original request).
 
+## Checklist To Using CircleCI Securely as a Customer
+
+If you are getting started with CircleCI there are some things you can ask your team to consider for security best practices as _users_ of CircleCI:
+
+- Minimise the number of secrets (private keys / environment variables) your build needs.
+- Ensure the secrets you _do_ use are of limited scope in their purpose for your build.
```
glenjamin:
```suggestion
- Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build.
```

jekyll/_cci2/security.md
```diff
@@ -119,6 +104,19 @@ Following are the system events that are logged. See `action` in the Field secti
...
+- Consult your VCS provider's permissions for your organization (if you are in an organizations) and try to follow the [Principle of Least Privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). 
+- Use Restricted Contexts with Github Teams to share environment variables with a select security group. Read through the [contexts]({{ site.baseurl }}/2.0/contexts/#restricting-a-context) document to learn more.
```
glenjamin: is this github only?

jekyll/_cci2/security.md
```diff
@@ -119,6 +104,19 @@ Following are the system events that are logged. See `action` in the Field secti
...
+- Ensure you audit who has access to SSH keys in your organization.
```
glenjamin: Can we link off to an external guide that already exists here? @lizbell-ci has been looking into this area a lot lately

jekyll/_cci2/security.md
```diff
@@ -119,6 +104,19 @@ Following are the system events that are logged. See `action` in the Field secti
...
+- Ensure that your team is using Two-Factor Authentication (2FA) with your VCS ([Github 2FA](https://help.github.com/en/articles/securing-your-account-with-two-factor-authentication-2fa), [Bitbucket](https://confluence.atlassian.com/bitbucket/two-step-verification-777023203.html)).
+- If your project is open source and public, please make note of whether or not you want to share your environment variables. On CircleCI, you can change a project's settings to control whether your environment variables can pass on to _forked versions of your repo_. This is **not enabled** by default.
```
glenjamin: Can we link to another page here?

jekyll/_cci2/security.md
```diff
@@ -119,6 +104,19 @@ Following are the system events that are logged. See `action` in the Field secti
 - **success:** A flag to indicate if the action was successful.
 - **request:** If this event was triggered by an external request this data will be populated and may be used to connect events that originate from the same external request. The format is a JSON blob containing `id` (the request ID assigned to this request by CircleCI), `ip_address` (the original IP address in IPV4 dotted notation from which the request was made, eg. 127.0.0.1), and `client_trace_id` (the client trace ID header, if present, from the 'X-Client-Trace-Id' HTTP header of the original request).
 
+## Checklist To Using CircleCI Securely as a Customer
+
+If you are getting started with CircleCI there are some things you can ask your team to consider for security best practices as _users_ of CircleCI:
+
+- Minimise the number of secrets (private keys / environment variables) your build needs.
```
glenjamin: I think it would be good to add another item in here about regularly rotating secrets - although currently this is a bit of a pain to automate if you use contexts, as there's no official API.

jekyll/_cci2/security.md
teesloane: Yes.

jekyll/_cci2/security.md
```diff
@@ -119,6 +104,20 @@ Following are the system events that are logged. See `action` in the Field secti
 - **success:** A flag to indicate if the action was successful.
 - **request:** If this event was triggered by an external request this data will be populated and may be used to connect events that originate from the same external request. The format is a JSON blob containing `id` (the request ID assigned to this request by CircleCI), `ip_address` (the original IP address in IPV4 dotted notation from which the request was made, eg. 127.0.0.1), and `client_trace_id` (the client trace ID header, if present, from the 'X-Client-Trace-Id' HTTP header of the original request).
 
+## Checklist To Using CircleCI Securely as a Customer
+
+If you are getting started with CircleCI there are some things you can ask your team to consider for security best practices as _users_ of CircleCI:
+
+- Minimise the number of secrets (private keys / environment variables) your
```
taxonomic-blackfish: I think this checklist in general needs to be broken into subsections that include the what as well as the why. For example, the reason secrets need to be rotated regularly is because users who are disassociated from an org could have used builds to log/exfiltrate the secrets even if they never had access to them in their raw form. The way this is phrased right now sounds like we're saying that secrets should be rotated because our shoddy security may leak them.

jekyll/_cci2/security.md
```diff
@@ -119,6 +104,20 @@ Following are the system events that are logged. See `action` in the Field secti
...
+  build needs and rotate secrets regularly.
+- Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build.
```
taxonomic-blackfish: Likewise here with the scoping of the builds, I think the smallest amount of further explanation should be included. E.g., machine users on GH, not using personal/root AWS credentials

jekyll/_cci2/security.md
```diff
@@ -119,6 +104,20 @@ Following are the system events that are logged. See `action` in the Field secti
...
+- Consult your VCS provider's permissions for your organization (if you are in an organizations) and try to follow the [Principle of Least Privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). 
+- Use Restricted Contexts with teams to share environment variables with a select security group. Read through the [contexts]({{ site.baseurl }}/2.0/contexts/#restricting-a-context) document to learn more.
+- Ensure you audit who has access to SSH keys in your organization.
+- Ensure that your team is using Two-Factor Authentication (2FA) with your VCS ([Github 2FA](https://help.github.com/en/articles/securing-your-account-with-two-factor-authentication-2fa), [Bitbucket](https://confluence.atlassian.com/bitbucket/two-step-verification-777023203.html)).
```
taxonomic-blackfish: This sounds like general security advice. This needs further explanation to explain that if someone can take over a GH/BB account, they can push code and use that to steal secrets for example.

jekyll/_cci2/security.md
```diff
@@ -119,6 +104,20 @@ Following are the system events that are logged. See `action` in the Field secti
...
+- If your project is open source and public, please make note of whether or not you want to share your environment variables. On CircleCI, you can change a project's settings to control whether your environment variables can pass on to _forked versions of your repo_. This is **not enabled** by default. You can read more about these settings and open source security in our [Open Source Projects document]({{site.baseurl}}/2.0/oss/#security).
```
taxonomic-blackfish: We've also had complaints in the past about the possibility to have non-public FOSS builds which can be accessed via a logged in user and guessing the URL. I don't know if this has been addressed or if this needs to be clarified. I'm also not sure if this is the right place to do so, so I'm kinda just rambling.

jekyll/_cci2/security.md
teesloane: @taxonomic-blackfish Thanks for pointing this out. Can you expand what you mean by "Machine user's on GH?" I'm not sure what security aspect that refers to.

How does this sound:

```
  - Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build. Consider carefully adjudicating the role and permission systems of other platforms you use outside of CircleCI; for example, when using something such as IAM permissions on AWS, you should ensure that you properly assign authorization levels on a per user basis.
```

bluemind777:
```suggestion
- Ensure the secrets you _do_ use are of limited scope - with only enough permissions for the purposes of your build.
```

jekyll/_cci2/security.md
teesloane: Makes sense! I'm on it!

jekyll/_cci2/security.md
taxonomic-blackfish: Machine users: https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users

I think that is an improvement, yes.

Issue #3747, opened 2019-09-22
Resource class keys are incorrect


levlaz:
# Description
A customer reported that the keys in the documentation do not work for extra-large resources. 

For example, the documentation states that 2XL+ is a valid key, but when you try to use CircleCI will show the following error: 

```
Resource class 2XL+ is not a valid resource class. Valid classes: small, medium, medium+, large, xlarge, 2xlarge, 2xlarge+. Default class medium will be used.
```

# Location
https://circleci.com/docs/2.0/configuration-reference/#resource_class

# Steps to Reproduce
Set `resource_class: 2XL+` in a build.

rosieyohannan:
Hi  @levlaz, thanks for pointing this out! We're working on sorting out the issue, but for now I've hidden these resource classes from the config reference. I will update here when the issue is fixed. Thanks! Rosie

mvxt:
This would be addressed in this PR: https://github.com/circleci/circleci-docs/pull/3746

mvxt:
#3746

mvxt:
Fixed in PR: #3746

Deployed to prod: https://circleci.com/docs/2.0/configuration-reference/#resource_class

PR #3755, opened 2019-09-25
hide resource classes

rosieyohannan:
Hide currently-unavailable resource classes from config reference.


rosieyohannan:
Fixes issue #3747

PR #3758, opened 2019-09-25
add diagram to show parallelism across all executors

rosieyohannan:
Add an image showing how parallelism works with different executors. Customer question that prompted this was regarding remote docker, and whether multiple remote docker environments get spun up with parallelism>1

Question: I stuck the blue text at the top in to cover the need for test splitting, but looking at it now this feels a little lazy. Thoughts? Could leave in for now and I will think about a second diagram to actually visually show the test splitting - maybe a sort of zoom in to one of the executors or something :-)


teesloane:
This looks good ! I have an unrelated question, though. This part of theé image:

image

are we including it next to the `windows` executor to encourage that Windows be run through orbs, or do we just want to more generally raise awareness that orbs can play a roll in defining an executor?

Otherwise, I suppose I would keep the text in and maybe save test splitting for another diagram / put it with the test-splitting document. Thanks @rosieyohannan !

thomasr8:
Looks good @rosieyohannan

rosieyohannan:
@teesloane ummm I thought the only way to use the Windows executor was with the orb? That's why I included it there in the image. Maybe that's not correct though. 🤔

Issue #3772, opened 2019-09-29
Use package-lock.json for CircleCI cache checksum


DanielRuf:
# Locations
https://circleci.com/docs/2.0/language-javascript/

# Conflict Description
The checksum should be created from the lockfile, not the package.json as it changes less often.

JoppeDC:
I'll take a look at this if thats okay

teesloane:
Thank you @JoppeDC  and @DanielRuf ! 

Just as a heads up - we have some progress on this in [this pr](https://github.com/circleci/circleci-docs/pull/3685).

It looks like we just need to update the docs that still reference using the `package.json` as the checksum. Thank you for filing / helping!

teesloane:
Added a comment to the PR, if you can check it out @JoppeDC  — almost ready to merge!

PR #3775, opened 2019-09-30
Changed checksum calculation docs to use package-lock instead of package

JoppeDC:
This references issue #3772, it is better to use the package-lock.json for checksum calculations than the package.json.


jekyll/_cci2/examples-intro.md
```diff
@@ -30,12 +30,12 @@ jobs:
           name: Update npm
           command: 'sudo npm install -g npm@latest'
       - restore_cache:
-          key: dependency-cache-{{ checksum "package.json" }}
+          key: dependency-cache-{{ checksum "package-lock.json" }}
```
teesloane: So, technically, this example cannot use `package-lock` because it's < npm v5.x.

Can you either bump up the version of the node/mongo versions to match [this document](https://circleci.com/docs/2.0/language-javascript/) or revert back to `package.json` (hopefully the former).

JoppeDC: Will revert this change. Maybe you can open a seperate hacktoberfest issue for this.

KyleTryon:
I disagree with this. The lock file is generated any time you run NPM install, meaning it could change frequently even if no dependencies have actually been added or removed.

JoppeDC:
The lockfile isn’t supposed to change on an install? Have you verified that?

PR #3782, opened 2019-10-01
Add example for when attribute (separate from when step)

mvxt:
# Description
Added an example for "The `when` Attribute" in the config reference.

# Reasons
The config reference doesn't have an example right now, so it's easy to miss and it's not clear that there's a difference between the `when` attribute on a `run` step vs. an actual `when` step. Links to the two here:

- https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute
- https://circleci.com/docs/2.0/configuration-reference/#the-when-step-requires-version-21


michelle-luna:
This is the place where we have another example, if you want to add a link. https://circleci.com/docs/2.0/reusing-config/#example-1

mvxt:
@michelle-luna I think that example shows the `when` step not the `when` attribute. In the Config Reference we already have an example for the `when` step so I don't think we'll need to link to it additionally.

If things look okay I'll go ahead and merge this PR tomorrow sometime.

jekyll/_cci2/configuration-reference.md
```diff
@@ -607,6 +607,13 @@ step that needs to upload logs or code-coverage data somewhere.
 
 A value of `on_fail` means that the step will run only if one of the preceding steps has failed (returns a non-zero exit code). It is common to use `on_fail` if you want to store some diagnostic data to help debug test failures, or to run custom notifications about the failure, such as sending emails or triggering alerts in chatrooms.
 
+``` YAML
+- run:
+    name: Cleanup cloud
+    command: # Bunch of stuff here
+    when: always # always cleanup whether job passes or fails
+```
```
marcomorain: Let's use a real example here, such as uploading code coverage results. This is something that you want to do regardless of whether the tests pass or fail.

```yaml
- run:
    name: Upload CodeCov.io Data
    command: bash <(curl -s https://codecov.io/bash) -F unittests
    when: always
```

mvxt: Done

mvxt:
If things look good to both of you this PR is ready to merge.

PR #3787, opened 2019-10-02
Optimize/Refactor Docs CI 

teesloane:
I tried to do a pass at cleaning up our `config.yml`. Mostly:

- Moving more things into their own jobs and making better use of workflows
- now the building of our slate documentation (API v1, v2) is it's own job
- PDFs building is a job.
- The above two new jobs and the JS build job are now run in a parallel workflow before running the `build` job. 
- I also tried to clean up some of the larger bash commands by moving them into separate files in the `scripts` directory.
- tried to provide a good open source example that uses several CCI features - orbs, re-useable commands, more comments in the config, etc.

Generally, the builds seem to be running just a bit faster, but at the least this is (hopefully) more readable and easy to navigate / can be a good open source example.

Thanks for any folks who can review / leave feedback (I could probably use an extra set of 👀 just to make sure that I am caching correctly/workflow-ing 💃 correctly).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ orbs:
 
 # Yaml References enable us to DRY our our config by sharing variables across multiple jobs.
 # In this case, we are commonly using the "workspaces" feature to share
-# build artifacts and files across jobs. For example, we build our Javascript
+# build artifacts and files across jobs. For example, we build our Javascript and
 # persist it to a workspace to be made available when the Jekyll site builds.
 references:
   workspace_root: &workspace_root


.circleci/config.yml
```diff
@@ -1,51 +1,130 @@
...
 
+
+# We use orbs to provide some common ruby and node tasks and shorten our config.
+# Learn more about orbs: https://circleci.com/orbs/
+orbs:
+  ruby: circleci/ruby@0.1.2
+  node: circleci/node@1.1.6
+
+
+# Yaml References enable us to DRY our our config by sharing variables across multiple jobs.
```
rosieyohannan: What is DRY? ... does it just mean shorten/simplify?

.circleci/config.yml
```diff
@@ -1,51 +1,130 @@
...
+# In this case, we are commonly using the "workspaces" feature to share
+# build artifacts and files across jobs. For example, we build our Javascript
```
rosieyohannan:
```suggestion
# build artifacts and files across jobs. For example, we build our Javascript and
```

.circleci/config.yml
```diff
@@ -1,51 +1,130 @@
...
             set -exu
             mkdir -p /tmp/workspace/js
             mv jekyll/assets/js/*.bundle.js /tmp/workspace/js/
-      - persist_to_workspace:
+      - persist_to_workspace: # store the built files into the workspace for other jobs.
           root: *workspace_root
           paths:
             - js
+
+  build_server_pdfs: # this job builds server-related pdf documentation, persisting it to the workspace as well.
```
rosieyohannan:
```suggestion
  build_server_pdfs: # this job builds pdf versions of the Server Administration documentation and persists them to the workspace.
```

.circleci/config.yml
```diff
@@ -1,51 +1,130 @@
...
+          command: ./scripts/build_api_docs.sh -v1
+      - run:
+          name: Build API 2.x documentation with Widdershins and Slate
+          command: ./scripts/build_api_docs.sh -v2
+      - persist_to_workspace:
+          root: *workspace_root
+          paths:
+            - api
 
+  # The Main "Build" job. It pulls in assets from previous jobs (the built api docs, pdfs and javascript)
```
rosieyohannan:
```suggestion
  # The main "Build" job. It pulls in assets from previous jobs (the built api docs, pdfs and javascript)
```

.circleci/config.yml
```diff
@@ -56,85 +135,36 @@ jobs:
...
-            - "src-api/vendor/bundle"
       - run:
           name: Build the Jekyll site
           command: bundle exec jekyll build --config jekyll/_config.yml,jekyll/_config_production.yml --source jekyll --destination jekyll/_site/docs/ 2>&1 | tee $JOB_RESULTS_PATH/build-results.txt
-      - run:
-          name: Build PDFs for Server
-          command: |
-            bundle exec asciidoctor-pdf -v
-            ./scripts/build_pdfs.sh
+      ### NOTE: we are ignore some files in the HTML proofer as it fails on pending translated docs.
```
rosieyohannan:
```suggestion
      ### NOTE: we ignore some files in the HTML proofer as it fails on pending translated docs.
```

rosieyohannan:
Thanks for doing this @teesloane, it's really helpful

.circleci/config.yml
teesloane: Woops! DRY stand for "Don't Repeat Yourself" I should also fix that duplicate `our` typo after 😅

rosieyohannan: Ohhhhh

rosieyohannan:
You/we should totally write blog post about doing this :-)

.circleci/config.yml
```diff
@@ -1,51 +1,130 @@
...
             mv jekyll/assets/js/*.bundle.js /tmp/workspace/js/
-      - persist_to_workspace:
+      - persist_to_workspace: # store the built files into the workspace for other jobs.
           root: *workspace_root
           paths:
             - js
+
+  build_server_pdfs: # this job builds pdf versions of the Server Administration documentation and persists them to the workspace.
+    executor:
+      name: ruby/default # we use the same ruby orb + tag across multiple jobs
```
rosieyohannan: Does it make sense to change to use this docker image for this job? https://hub.docker.com/r/asciidoctor/docker-asciidoctor

.circleci/config.yml
```diff
@@ -1,51 +1,130 @@
...
+          command: bundle check --path=vendor/bundle || bundle install --path=vendor/bundle --jobs=4 --retry=3
+      - save_cache:
+          key: circleci-docs-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
+          paths:
+            - "vendor/bundle"
+
+
+jobs:
+  js_build: # this job is responsible for building Javascript assets and making them available for the "build" job
+    executor: # we specify our executor to use the node orb.
```
rosieyohannan: I'm not clear on the syntax here: looking at https://circleci.com/docs/2.0/configuration-reference/#executors-requires-version-21 but can't see how you're defining executors documented?

PR #3796, opened 2019-10-04
[WIP]Create pipelines.md

michelle-luna:
# Description
This is a new page that will be linked from the New Pipelines page in the app via a dialog. 

# Reasons
When the app changes to opt-out sometime coming soon, we want at least one page that describes the new ways users can access job output and view projects, etc.


teesloane:
Here's a screenshot of how this page is coming along (the sidebar looks a little weird just because I'm using a full-page screenshot capture tool): 

screencapture-localhost-4000-docs-2-0-pipelines-2019-10-09-13_49_05

Feedback appreciated!

PR #3802, opened 2019-10-09
Update building-docker-images.md

### LINKED ISSUE #3696: Docker Folder Mounts Should Be Volume Mounts
# Location
https://circleci.com/docs/2.0/building-docker-images/#mounting-folders

# Explanation
The wording on mounts isn't incorrect, but not the typical Docker terminology. When it comes to mounts, especially with the -v flag, Docker refers to them as "volume mounts" not folder mounts. Hence the flag name `-v` which stands for volume. https://docs.docker.com/engine/reference/commandline/run/#mount-volume--v---read-only

Also, this section list a workaround for volume mounts since they don't work in the `docker` executor. Another alternative is to use the `machine` executor where volume mounts do work.

thomasr8:
Clarified content about volume mounts and added a note to specify that the example shown is a workaround. These additions are based on Ricardo's feedback and fixes #3696 .

# Description

# Reasons

### DIFF
--- a/jekyll/_cci2/building-docker-images.md
+++ b/jekyll/_cci2/building-docker-images.md
@@ -27,9 +27,7 @@ jobs:
 
 When `setup_remote_docker` executes, a remote environment will be created, and your current [primary container]({{ site.baseurl }}/2.0/glossary/#primary-container) will be configured to use it. Then, any docker-related commands you use will be safely executed in this new environment.
 
-**Note:** The use of the `setup_remote_docker` key is reserved for configs in
-which your primary executor _is_ a docker container. If your executor is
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/building-docker-images.md
```diff
@@ -256,6 +253,8 @@ Then, the sample CircleCI `.circleci/config.yml` snippets below populate and bac
 ```
 {% endraw %}
 
+**Note:** The example shown above provides a way for you to utilize volume mounts since they don't work in the `docker` executor. An alternative to this approach is to use the `machine` executor where volume mounts do work
```
teesloane:
```suggestion
**Note:** The example shown above provides a way for you to utilize volume mounts since they don't work in the `docker` executor. An alternative to this approach is to use the `machine` executor where volume mounts do work.
```

PR #3805, opened 2019-10-09
bind to localhost

taxonomic-blackfish:
to prevent access to host from LAN


felicianotech:
Why? It's an open source website. There's nothing to hide but with this PR it prevents browser testing from other devices on the network.

michelle-luna:
ooh, thanks Ricardo, I didn't think about that. We can revert if this blocks us from testing.

taxonomic-blackfish:
If that workflow is needed, there would need to be auth of some sort otherwise anyone on the network can access your laptop, and I think that is a greater risk than an edge case of testing workflows.

taxonomic-blackfish:
It's not about accessing the content which is public. It's about accessing the containers which are running as root and have the `$PWD` mounted to the container. Testing is great and fine. Letting random people on the network hit your laptop doesn't need to happen. Push a branch and have them pull it down.

PR #3825, opened 2019-10-15
Update sample config

rosieyohannan:
# Description
Add clear link to config reference after this survey response: "No - no documentation for each config variable"


rosieyohannan:
Questions:

* Should we still be mentioning 1.0 builders here?
* Should we be using `version: 2.1` config across the board yet, even when not required for the example?
* Or, maybe we could have a small example on this page showing reusing config and an orb to show 2.1 config - highlighting the differences

PR #3833, opened 2019-10-17
Refactored instantsearch.js

snavneet770:


teesloane:
I agree with @KyleTryon . Closing this.

PR #3835, opened 2019-10-17
Re-refactor config.yml

teesloane:
So, I broke the api v2 docs with https://github.com/circleci/circleci-docs/pull/3787 -- the problem was that the v2 docs weren't being copied over to the v2/ path in the jekyll site.

Here is a [link](https://50229-48750547-gh.circle-artifacts.com/0/circleci-docs/api/v2/index.html#circleci-api) of this build working.


teesloane:
I have since fixed the broken API compilation and also merged in upstream master changes. Here is the [v2 api build artifact](https://50991-48750547-gh.circle-artifacts.com/0/circleci-docs/api/v2/index.html#circleci-api)

PR #3837, opened 2019-10-18
correct reference to service key environment variable

brandonkirchner:
# Description
Correct invalid reference to cloud service key environment variable.

# Reasons


brandonkirchner:
after reviewing the doc, I realize that the existing way is actually correct :|

PR #3855, opened 2019-10-22
[WIP]Adding information about making branch filters that can ignore case

andreasmarkussen:
NOTE: I NEED TO ADD MORE INFO TO THE PR.
# Description
Adding information about how to make a branch filter that is case in-sensitive. 
I.e. Ignore if e.g. hotfix is written with upper or lower caps.

# Reasons
This caused me some headache to see that global reg exp modifies does not work.


teesloane:
Hi there. Thanks for making a PR. Unfortunately, I'm not sure that this is something that needs to be in the docs. This information seems like it would belong more appropriately in an internal wiki / developer-docs among your team — that is, that your team shares the same protocols  for making branches with the proper casing (ie, "please only make branches that follow the `*-fix` regex pattern so that your branch runs on our CI configuration").

I hope that makes sense. Please let me know if you have any questions. Closing for now.

PR #3862, opened 2019-10-23

### LINKED ISSUE #3858: add custom plans to the doc and fix note
can we get this article updated? https://circleci.com/docs/2.0/docker-layer-caching/ with the following? (also looks like there's a broken Note?)

    Docker Layer Caching is only available on the Performance and Custom usage plans.

(edited)
circleci.comcircleci.com
Enabling Docker Layer Caching - CircleCI
How to reuse unchanged cache layers in images you build to reduce overall run time
rosejen  1 day ago
would it also make sense to add something about pricing here? we could just link to the pricing page for details. but it may be good to be as transparent as possible

thomasr8:
Updated content to add that DLC is available to both Performance and Custom plans. This fixes #3858

# Description

# Reasons

### DIFF
--- a/jekyll/_cci2/docker-layer-caching.md
+++ b/jekyll/_cci2/docker-layer-caching.md
@@ -8,7 +8,7 @@ order: 70
 ---
 
 Docker Layer Caching (DLC) can reduce Docker image build times on CircleCI. DLC is available on
-the [Performance and Custom](https://circleci.com/pricing/) usage plans, and on installations of [CircleCI Server](https://circleci.com/enterprise/). This document provides an overview of DLC in the following sections:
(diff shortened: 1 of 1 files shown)


thomasr8:
@michelle-luna Fixed the wording in the content to make it more clear that DLC is available in both performance and custom plans. I checked the notes and links on the page and did not see a broken link or problem with the notes. Can you point me to exactly where this issue is on the page, and I will go ahead and fix right away. Thanks :)

rosieyohannan:
I think the problem is with the note below the section you have edited here - it repeats what is said in the paragraph above, ends in a semicolon, and then there's the page TOC, which doesn't make sense. Maybe we could move the bit that you have edited about which plans DLC is available on out of the paragraph and into the note content, and remote the semi colon. I think Rose also mentioned adding a link to the pricing page: https://circleci.com/pricing/ 🤷🏻‍♀️

PR #3879, opened 2019-10-28
Update sidenav.yml

thomasr8:
In Jobs section, added "Viewing Pipelines" after "Status Overview".

# Description

# Reasons

### DIFF
--- a/jekyll/_data/sidenav.yml
+++ b/jekyll/_data/sidenav.yml
@@ -209,7 +209,7 @@ en:
       children:
         - name: Status Overview
           link: 2.0/status/
-          - name: Viewing Pipelines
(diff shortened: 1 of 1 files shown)


jekyll/_data/sidenav.yml
```diff
@@ -209,6 +209,8 @@ en:
       children:
         - name: Status Overview
           link: 2.0/status/
+          - name: Viewing Pipelines
```
teesloane: This is failing because it needs to be de-dented to align with the above and below keys to make yaml happy 😭

thomasr8: Thanks for the help @teesloane. Fixed.

PR #3885, opened 2019-10-30
Install path best practices.

KyleTryon:
# Description
Add best practices on installing binaries via Orbs with parameterized installation paths.

# Reasons
Reduce conflicts with orbs needing to interact with root when in environments where root is not accessible.

### DIFF
--- a/jekyll/_cci2/orbs-best-practices.md
+++ b/jekyll/_cci2/orbs-best-practices.md
@@ -54,6 +54,16 @@ fi
 - Utilize the [“env_var_name” parameter type](https://circleci.com/docs/2.0/reusing-config/#environment-variable-name) to secure API keys, webhook urls or other sensitive data. 
 - [Injecting steps as a parameter](https://circleci.com/docs/2.0/reusing-config/#steps) is a useful way to run user defined steps within a job between your orb-defined steps.Good for if you need to perform an action both before and after user-defined tasks - for instance, you could run user-provided steps between your caching logic inside the command.
 
+**Installing binaries and tools**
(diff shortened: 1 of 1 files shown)


eddiewebb:
@KyleTryon - dod you think it is worth adding an example orb to mimic? I think this model makes sense but does add complexity to the checks for existence, permissions, etc

KyleTryon:
> @KyleTryon - dod you think it is worth adding an example orb to mimic? I think this model makes sense but does add complexity to the checks for existence, permissions, etc

Yes, I was about to use the GCP CLI orb as an example but it's a little different because rather than setting a path, it comes with a script to source. We can leave this open for now and we can make a more common example from another orb.

I'll make a PR to the JQ orb now as it's already close. We'll take a look at that, and if we like that, use it as the example here.

KyleTryon:
I was talking a look at a few orbs (including JQ) to try to update to showcase what we talked about for the install-path. I think we would only showcase one or the other, depending on the installation method for the tool in that orb.

What I mean is, the JQ orb, for instance, we have set to install to `/usr/local/bin` which makes the most sense and doesn't need to be added to the path or anything. And demonstrates basically what we already want from the user.

Other CLIs such as Heroku and the GCP CLI install via a script to a pre-determined path. The GCP orb you need to source the ".inc" file they include: https://github.com/CircleCI-Public/gcp-cli-orb/blob/master/src/commands/install.yml#L18

Heroku on the other hand, installs via a shell script to a pre-determined path. https://github.com/CircleCI-Public/circleci-orbs/blob/staging/src/heroku/orb.yml#L63
`/usr/local/lib/heroku` and `/usr/local/bin/heroku` which also doesnt need root or to be added to the path or anything like that.

So I think im looking for a situation where the user would actually need sudo permissions and to install normally or has the ability to install into a non-root path, which I'm not sure how common that is.

It seems maybe we should just instruct how to add a binary to a `/usr/local/bin` path? And then we could address using apt-* separately?

PR #3888, opened 2019-10-30
improved links in index.md

Shaleengovil25:
### **DESCRIPTION**
added visited color to links.

### **REASON**
it looks better and interactive.


teesloane:
Hi there!

Thank you for making a suggested change to our docs. Unfortunately, we leave all styling to our internal design team as of now. I will bring up this change with them to see if we can make an improvement to how our links are styled according to our design systems.

PR #3891, opened 2019-10-31
Add a note about the Cocoapods 1.8 CDN

BytesGuy:
# Description

Added a note about the Cocoapods 1.8 CDN

# Reasons

This will speed up builds and will be less reliant on our hosting of the specs (there have been cases where AWS outages have caused jobs to fail due to relying on the script)

### DIFF
--- a/jekyll/_cci2/testing-ios.md
+++ b/jekyll/_cci2/testing-ios.md
@@ -70,7 +70,7 @@ example of a minimal config in the
 ### Best Practices
 {:.no_toc}
 
-In addition to the basic setup steps, it is best practice to include
+In addition to the basic setup steps, if you are using Cocoapods 1.7 and lower, it is best practice to include
 downloading CocoaPods specs from the CircleCI mirror (up to 70% faster)
 and linting the Swift code together with the `build-and-test` job:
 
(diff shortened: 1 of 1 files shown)


BytesGuy:
@marcomorain Sounds good, thanks for the feedback! Will action that and update the PR in due course

BytesGuy:
@marcomorain I think this looks a bit more streamlined. Where is the best place to post the old content in Discuss in your opinion?

marcomorain:
I'd put it in https://discuss.circleci.com/c/mobile

jekyll/_cci2/testing-ios.md
```diff
@@ -70,46 +70,18 @@ example of a minimal config in the
 ### Best Practices
 {:.no_toc}
 
-In addition to the basic setup steps, it is best practice to include
-downloading CocoaPods specs from the CircleCI mirror (up to 70% faster)
-and linting the Swift code together with the `build-and-test` job:
+In addition to the basic setup steps, it is best practice to use Cocoapods 1.8 or newer which allows the use of the CDN, rather than having to clone the entire Specs repo. If you are using Cocoapods 1.7 or older, consider upgrading to 1.8 or newer as this change allows for much faster job execution of the `pod install` step.
```
marcomorain:
```suggestion
In addition to the basic setup steps, it is best practice to use Cocoapods 1.8 or newer which allows the use of the CDN, rather than having to clone the entire Specs repo. This will allow you to install pods faster, reducing build times. If you are using Cocoapods 1.7 or older, consider upgrading to 1.8 or newer as this change allows for much faster job execution of the `pod install` step.
```

PR #3898, opened 2019-11-05
added `(preview)` to the autogen categories

jama22:
# Description
added `(preview)` to the autogen categories to make it super clear that these endpoints are in preview


teesloane:
Looks good!

image

PR #3900, opened 2019-11-05
Ensure availability of user management feature is clear

gmemstr:
# Description

Ensures that the availability of user management features is made clearer.

# Reasons

A user was confused about the availability of the user management feature within server and assumed it was also available in cloud. This helps clarify this availability.


gmemstr:
Thanks for the pointer - updated :)

Issue #3749, opened 2019-09-23
What is CircleCI?


jfelixbanguera:
# Missing Information
What is CircleCI? what is it for? what is it not for? Explanation and examples of what it does and how does it run.

# Already Looked
https://circleci.com/docs/
https://circleci.com/docs/2.0/first-steps/
https://circleci.com/docs/2.0/hello-world/
https://circleci.com/integrations/
https://circleci.com/orbs/
https://circleci.com/docs/api/#api-overview
https://circleci.com/open-source/

# Potential Locations
https://circleci.com/docs/

michelle-luna:
Hi,

Thanks for filing this issue, sorry you didn't find this quickly. We have it here in the CI/CD Overview https://circleci.com/docs/2.0/about-circleci/ 
We will add some links from these other documents to make it easier to find as well as the What is CircleCI? heading.

sameshl:
@michelle-luna I would like to take up this task. Could you just elaborate on where I exactly that section needs to be added?

Thanks!

mvxt:
@michelle-luna Do you think this page should be in the "Getting Started" section?

Getting Started > Introduction > What Is CircleCI?

michelle-luna:
works for me! thank you!

rosieyohannan:
I'm going to take this as it's been hanging around a while...

rosieyohannan:
Fixed!

PR #3901, opened 2019-11-05
Improve visibility of CircleCI Overview

rosieyohannan:
fixes issue #3749


rosieyohannan:
I've duplicated the page - not sure that's the best way but looks OK I think!

PR #3909, opened 2019-11-08
adding gist link for org rename script

BoVice:
# Description
I replaced the org rename script codeblock with a link to a gist containing the same script. The gist has all of the doc's webpage unicode removed.

# Reasons
The formatting of the script on the docs page contains some additional unicode that breaks the script when running into the REPL.


rosieyohannan:
@BoVice thanks for this, such a better way to do it! I will do this in future for any large copy paste blocks :-)

PR #3911, opened 2019-11-08
Clarify windows executor is a wrapper for machine

gmemstr:
# Description
Windows is listed as an executor key, which can lead to some confusion. Really, it's a machine image, and the orb is recommended for ease of defining the executor, but this isn't immediately clear.

Note that this isn't necessarily a final edit - I think some more thought may be required for the wording/flow of this executor, but I'd like to get the conversation started.

# Reasons

A customer was confused about the use of `windows` as a top-level executor definition when this doesn't actually work.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -495,8 +495,7 @@ jobs:
       ... // other config
 ```
 
-The Orb is a wrapper around the `machine` executor to make configuration easier. This is the recommended
-way of handling it, but it can be directly defined as well.
+This is the recommended way to set up the Windows executor. It is also possible to directly define a Windows `machine` image as seen below.
 
 ```yaml
 jobs:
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -493,6 +495,20 @@ jobs:
       ... // other config
 ```
 
+The Orb is a wrapper around the `machine` executor to make configuration easier. This is the recommended
+way of handling it, but it can be directly defined as well.
```
teesloane: This makes sense. I would re-word it a little bit; something akin to:

> This is the recommended way to set up the Windows executor. It is also possible to directly define a Windows `machine image` as seen below.

gmemstr: That revised wording does make a bit more sense, implemented!

rosieyohannan:
My understanding is that we want everyone to use the orb - I am adding an alternative version for Server customers without the orb, but this isn't an option for Cloud customers - Can you confirm @appplemac? Closing for now but feel free to reopen.

PR #3913, opened 2019-11-11
correct usage of when in workflows

rosieyohannan:
Update use of `when` in workflows in the config reference

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -1304,22 +1304,17 @@ ignore | N | String, or List of Strings | Either a single tag specifier, or a li
 
 For more information, see the [Executing Workflows For a Git Tag]({{ site.baseurl }}/2.0/workflows/#executing-workflows-for-a-git-tag) section of the Workflows document.
 
-##### **`when` in Workflows**
+##### **Using `when` in Workflows**
 
-## Using 'when' in Workflows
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -1304,13 +1304,11 @@ ignore | N | String, or List of Strings | Either a single tag specifier, or a li
 
 For more information, see the [Executing Workflows For a Git Tag]({{ site.baseurl }}/2.0/workflows/#executing-workflows-for-a-git-tag) section of the Workflows document.
 
-##### **`when` in Workflows**
+##### **Using `when` in Workflows**
 
-## Using 'when' in Workflows
+With CircleCI API v2, you may use a `when` clause (the inverse clause `unless` is also supported) under a workflow declaration with a boolean value to determine whether or not to run that workflow. The most common use of `when` in API v2 is to use a pipeline parameter as the value, allowing an API trigger to pass that parameter to determine which workflows to run.
```
glenjamin: I believe any "truthy" value works, not only booleans. Much to @sulami's chagrin 😁

sulami: That is correct, any type we currently have will do, and will be casted to some form of boolean depending on its value.

rosieyohannan: OK, can you suggest an edit to incorporate this if necessary as not sure I understand 🤪

glenjamin:
```suggestion
With CircleCI API v2, you may use a `when` clause (the inverse clause `unless` is also supported) under a workflow declaration with a truthy or falsy value to determine whether or not to run that workflow. The most common use of `when` in API v2 is to use a pipeline parameter as the value, allowing an API trigger to pass that parameter to determine which workflows to run.
```

rosieyohannan: Thanks!

PR #3917, opened 2019-11-11
Use correct ssh command in docs

JackWink:
`ssh-agent -D` starts a server, `ssh-add -D` deletes known keys, which is what is intended by this section of the docs.

http://man.openbsd.org/ssh-add#D

VS:

http://man.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man1/ssh-agent.1?query=ssh-agent%26sec=1#D


michelle-luna:
thank you! closing since this was fixed in #3920

JackWink:
I was first though :( 

Glad it got fixed

PR #3922, opened 2019-11-12
Multi  Element Tabs

teesloane:
## What:

This PR tries to improve our tabbing abilities to include other kinds of elements. 

## Why:

We are finding that we need to provide documentation on split-versioned content within the same page; most notable, the need to provider content for both Cloud and Server. For example, some of our new UI is present in Cloud but not Server, and our documentation should reflect that.

## Help

I'm not exactly a jquery wizard, so I would gladly welcome all suggestions. Also, there's probably some use cases / edge-cases that I haven't considered. Let me know if you think of something

## Other changes

I have documented this widget and also moved pertinent screenshots of how it works and the WIDGETS.md file into a `/docs` folder. As we continue to improve the functionality and tooling around our docs-repo, we should have a place for pertinent meta-documentation.

## Screens

Here are some gifs of the new tabs in effect:

**Desktop**

tab-desktop

**Mobile**

tabs mobile


felicianotech:
Very cool.

felicianotech:
Same issue ^ in Firefox but without the scrollbar.

teesloane:
@felicianotech -- thanks for actually testing this!! I cannot recreate this; are you on linux or windows? I think non-mac-osx might have scrollbar issues due to the `overflow-x: scroll` for the styling ([this line ](https://github.com/circleci/circleci-docs/pull/3922/files#diff-158e5e70020fac0e3f1e11a76d3e9850R332))

Do you think you could hack on this to try and fix it? I'm not sure how I can recreate it on my end 😓

felicianotech:
> are you on linux or windows?

I tested with Google Chrome Version 78.0.3904.97 (Official Build) (64-bit) on Ubuntu 19.04 (Linux). Since we're dealing with browsers here, theoretically the OS shouldn't matter. Of course, that's not always the case.

felicianotech:
@teesloane I think you're right about `overflow-x: scroll`. My question is, why is it there? I removed that line locally and now the widget looks fine on Chrome for both Ubuntu and macOS, and Firefox for Ubuntu. Is there a test case that didn't work for you until you add that line?

teesloane:
@felicianotech I believe I set it so that in the case of being on mobile the [overflow-x: scroll](https://github.com/circleci/circleci-docs/pull/3922/files#diff-158e5e70020fac0e3f1e11a76d3e9850R332) enables the case in which having multiple tabs on mobile allows you to scroll to the right if there are many tabs (as in the mobile screen gif above) I think perhaps I could apply the overflow only on mobile below, 500~ px, meaning that the scrollbar would only appear on mobile, to let users know they can scroll to see more tabs?

felicianotech:
Can you test not using `overflow-x: scroll` on iPhone (if you have one)? I just tested it on Chrome on Android and scrolling works just fine.

teesloane:
@felicianotech - I tested on the ios simulator, and here's what came out:

**With Overflow-x**

with-overflow-xc

**Without overflow**

without-overflow-xc

felicianotech:
Cool. How did you get access to that simulator?

---

I think I see the first problem here. You're testing the scrolling of the tabs themselves at the top. I tested the scrolling of the content below it.

teesloane:
@felicianotech  - If you have xcode installed (if you have a mac :|) You can use the simulator. I will test on it from time to time when I can't trust the browser or need iOS specific testing.

---
Is the problem that we don't want to show the scrollbar? If not, please let me know, I think we might be on a different page!  If the problem is the scrollbar, I believe that the scrollbar is showing up for the "scrollability" of the tabs (which is what I was testing) based on there being the overflow-scroll property.

felicianotech:
The problem I'm showing in the screenshot I provided is that there's a scrollbar showing even though there's nothing to scroll.

teesloane:
@felicianotech Ah, ok. Sounds good!

felicianotech:
@teesloane I pushed a change. Looks good in my browsers. Mind checking on yours and reviewing my commit?

teesloane:
Looks good @felicianotech ; I removed my console log and fixed a doc string. Other than that, it tested ok on my machine. If you think it's ready to go, can you approve this and we'll merge?

PR #3925, opened 2019-11-13
Be explicit about CIRCLE_COMPARE_URL being absent from v2.1

stig:
----

# Description
`CIRCLE_COMPARE_URL` is not present in v2.1 environments, but we do have an alternative, as outlined in https://discuss.circleci.com/t/is-circle-compare-url-still-a-thing/33366/2 -- we'll be trying to work this more fully into the official documentation, but in the interest of providing clarity early on I think this helps.

# Reasons
We've seen some posts in discuss about this, most recently in https://discuss.circleci.com/t/is-circle-compare-url-still-a-thing/33366 and I thought it would be nice to provide a bit more information.


sibelius:
can we get the full commit range from a branch/pull request using pipeline variable?

related to this https://twitter.com/sseraphini/status/1280274128751276032

PR #3937, opened 2019-11-18
Removing Typo: Redundant "Machine" Word

HamzaIbnFarooq:
# Description
Removing Typo: Redundant use of "Machine" word.

# Reasons
I think it will look good this way.


rosieyohannan:
Good spot, thanks! 🎉

PR #3940, opened 2019-11-19
Fix filepath for CIRCLE_DISPATCHER_RESOURCE_DEF

heug:
path should be prepended by "/etc"

# Description
Add /etc to beginning of filepath for CIRCLE_DISPATCHER_RESOURCE_DEF

# Reasons
Incorrect filepath


MikeyYeahYeah:
Good catch!  Thanks for the update!

PR #3967, opened 2019-11-27
Add a note on LDAP interaction for domain-service

emauton:
# Description
Note that LDAP authentication will fail if `domain-service` is down.

# Reasons
`domain-service` is responsible for LDAP authentication proper, while
`federations-service` binds the resulting LDAP identities to CircleCI
User objects.


rosieyohannan:
Thanks @emauton! ☺️

PR #3972, opened 2019-11-28
Update the docker installation example

glenjamin:
# Description
Update the docker installation example, including moving installing a custom docker version out of the main body of the example.

# Reasons
Improves the security, and also makes it hopefully a little clearer.

Prompted by seeing a tweet suggesting the older method was insecure, which is arguably true.

The example was using very old versions of things.

### DIFF
--- a/jekyll/_cci2/building-docker-images.md
+++ b/jekyll/_cci2/building-docker-images.md
@@ -87,7 +87,7 @@ jobs:
           docker push CircleCI-Public/circleci-demo-docker:$TAG
 ```
 
-If the primary container you're using doesn't already have the docker CLI installed, then [you'll need to install it](https://docs.docker.com/install/#supported-platforms) somehow.
+If the primary container you are using doesn't already have the docker CLI installed, then [you will need to install it](https://docs.docker.com/install/#supported-platforms) somehow.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/building-docker-images.md
```diff
@@ -67,38 +67,35 @@ jobs:
...
       - run: |
           TAG=0.1.$CIRCLE_BUILD_NUM
-          docker build -t   CircleCI-Public/circleci-demo-docker:$TAG .     
-          docker login -u $DOCKER_USER -p $DOCKER_PASS         # (4)
+          docker build -t CircleCI-Public/circleci-demo-docker:$TAG .
+          echo $DOCKER_PASS | docker login -u $DOCKER_USER --password-stdin  # (4)
           docker push CircleCI-Public/circleci-demo-docker:$TAG
 ```
 
+If the primary container you're using doesn't already have the docker CLI installed, then [you'll need to install it](https://docs.docker.com/install/#supported-platforms) somehow.
```
rosieyohannan:
```suggestion
If the primary container you are using doesn't already have the docker CLI installed, then [you will need to install it](https://docs.docker.com/install/#supported-platforms) somehow.
```

PR #3979, opened 2019-12-02
Fix invalid indentation in config example

AntoinePrv:
# Description
I changed the indentation in the YAML example.

# Reasons
It does not seem to be valid YAML (fails `cicreci config check`).


rosieyohannan:
Thanks so much for this!

PR #3981, opened 2019-12-02
Update screenshots for new UI

teesloane:
Screenshots for the following pages have been updated

- [X] [Artifacts](https://circleci.com/docs/2.0/artifacts/) (2/2)
- [X] [Hello World](https://circleci.com/docs/2.0/hello-world/) (1/1)
- [X] [Concepts](https://circleci.com/docs/2.0/concepts) (2/2)
- [X] [Getting Started](https://circleci.com/docs/2.0/getting-started/) (6/6)
- [X] Change "Jobs" ⇢ "Pipelines in `sidenav.yml`
- [X] [Workflows](https://circleci.com/docs/2.0/workflows/) doc (2/2)
- [X] [Status](https://circleci.com/docs/2.0/status/) doc


rosieyohannan:
Looking at the Getting started page makes me wonder if we need two pages - Getting Started and Getting Started on Server ... or something. You've added some great screenshots in to the steps, but as there aren't old UI equivalents I wonder if it might be confusing for Server people. What do you think?

rosieyohannan:
Screenshot 2019-12-17 at 11 45 33

This tabbed box seems to be missing a top. Weird! (getting started SSH section)

This

rosieyohannan:
Screenshot 2019-12-17 at 11 51 30

This screenshot does not appear to fit the text/section - I wonder if you've just updated the image and it's used in several places and now doesn't work here?

rosieyohannan:
Screenshot 2019-12-17 at 11 54 25

Ah yeah the image appears here too. The new version doesn't fit the text - the italics are referring to sections you can see in the old screenshots but they don't appear in the new ones - also - could we have the old screenshot and the new in tabbed blocks here for Server people?

rosieyohannan:
Screenshot 2019-12-17 at 12 08 32

The screenshots here are great, showing a manual build approval, but they don't fit the text - There's no Jobs page, or Workflows tab in the new UI I think. - I believe the X team have been referring to these pages as the "job details page" and "workflow details page" but users will need to know to navigate to these using the breadcrumbs (knowing what the icons mean) or from the pipelines page I think?

rosieyohannan:
@teesloane I think this is a super super hard task you're tackling! The screenshots look great, and I love the use of arrows and numbered steps etc. Maybe we could share the assets you're using so we can standardise things a bit? Also - the shadow behind the screenshots - is that something you're adding on in jekyll / CSS or to the actual image? 🎉🥳

teesloane:
Thanks for the thorough review @rosieyohannan ! I think you are right; there's more to it than I had originally anticipated with these changes. (I also didn't notice that some screenshots are being used in more than one place 😅 ). I will revise this to make sure it uses cloud/server tabs, and revisit the text surround the screenshots.

I think that having a separate getting-started doc for server might be useful as a lot of the step text has to change to reflect the new ui. I think filling the document with tabs would make the markup more complicated to maintain and might be frustrating / confusing for folks who are new to the platform.

teesloane:
The Pr got a bit longer - sorry! I:

- update copy some of the respective screenshots
- organized screenshots into tab blocks between _cloud_ and _server_ (which we should probably rename in another PR)
- 

My editor did some auto formatting on the markdown which cleaned up some of the yaml blocks + removed space. I tried to revert this so as to make the PR less noisy but some might still leak through.

jekyll/_cci2/getting-started.md
```diff
@@ -54,9 +54,13 @@ The `- image: circleci/ruby:2.4.1` text tells CircleCI what Docker image to use
...
 
-4. On the next screen, you're given some options for configuring your project on CircleCI. Leave everything as-is for now and just click the **Start building** button a bit down the page on the right. ![]( {{ site.baseurl }}/assets/img/docs/CircleCI-2.0-setup-project-circle101.png) ![]( {{ site.baseurl }}/assets/img/docs/CircleCI-2.0-start-building.png)
+4. On the following screen, choose a language from the dropdown to get a pre-populated config.yml file with suggested best-practices for your project. If the language you are using is not in the list, click the "Use Hello World" button at the bottom of the list. Copy the YAML configuration that appears and save it as a file called "config.yml" in a folder named `.circleci` in the root of your repository's directory.
+
+![]( {{ site.baseurl }}/assets/img/docs/CircleCI-add-new-project-list-2.png)
```
rosieyohannan: If you put the image ref at the end of line 59 after the text it will appear in line with step 4, and then step 5 will be labeled correctly. Currently step 5 is showing as `1.` for me.

rosieyohannan: Also just checking you wanted the images in this section to only reflect the new UI? I think that's pretty much OK as the actual steps are largely the same - apart from where you say select the language from the dropdown - in the old UI they are buttons.

rosieyohannan: Also comment for lines 135 and 137 - the first image in these steps is new UI and the second is old - maybe change to both be new UI to show the workflows map? (sorry I can't add comments to places in the doc where you haven't already made changes for some reason)

rosieyohannan:
I feel like for the screenshots we should maybe label the tabs New UI and Old UI, rather than Cloud/Server - as customers don't know it as Cloud, and those on Cloud have no idea what Server is. And some of the people wanting the Old UI will be Cloud users anyway.

teesloane:
@rosieyohannan - I might be misconstruing from memory but @michelle-luna Did we talk about the naming of screenshots before the break re: "cloud" and "server" as ok terminology?

rosieyohannan:
Looks great, Tyler, 👏🏻!

PR #3985, opened 2019-12-03
Update server version from 2.17 to 2.18

annapamma:
This page was under the heading for 2.18 server installation

# Description
Keep the server version consistent.

# Reasons
The instructions should be clear about which server version is being installed


rosieyohannan:
Good spot @annapamma! Thanks!

PR #3987, opened 2019-12-03
Add note to convimg doc.

felicianotech:
Added a note on what image tags we won't build.

If anyone has a suggestion for a better heading though, feel free to change it.

### DIFF
--- a/jekyll/_cci2/circleci-images.md
+++ b/jekyll/_cci2/circleci-images.md
@@ -257,7 +257,7 @@ The following packages are installed via `curl` or other means.
 
 1. If an image isn't listed above, it is not available. As the Convenience Image program is revamped, proposals for new images are not currently being accepted.
 1. Old versions of software will not be rebuilt. Once an upstream image stops building the tag for a specific release, say Node.js v8.1.0, then we stop building it too. This means other tools in that image, such as `npm` in this example, will no longer be updated either.
-1. We don't support building preview, beta, or release candidate images tags. On occasion they'll be available but these tags tend to cause our build system for Convenience Images to fail. If you need a non-stable release of of language, we suggest installing it via [an orb](https://circleci.com/orbs/) or a custom Docker image instead.
+1. We don't support building preview, beta, or release candidate images tags. On occasion they'll be available but these tags tend to cause our build system for Convenience Images to fail. If you need a non-stable release of a language, we suggest installing it via [an orb](https://circleci.com/orbs/) or a custom Docker image instead.
 
 
 ## Latest Image Tags by Language
--- a/jekyll/_cci2/circleci-images.md
+++ b/jekyll/_cci2/circleci-images.md
@@ -252,6 +252,14 @@ The following packages are installed via `curl` or other means.
 - [dockerize](https://github.com/jwilder/dockerize)
 - [jq](https://stedolan.github.io/jq/)
 
+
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/circleci-images.md
```diff
@@ -252,6 +252,14 @@ The following packages are installed via `curl` or other means.
 - [dockerize](https://github.com/jwilder/dockerize)
 - [jq](https://stedolan.github.io/jq/)
 
+
+## What Isn't Available
```
rosieyohannan:
```suggestion
## Out of Scope
```

jekyll/_cci2/circleci-images.md
```diff
@@ -252,6 +252,14 @@ The following packages are installed via `curl` or other means.
...
+
+1. If an image isn't listed above, it isn't available. As the Convenience Image program is revamped, proposals for new images aren't currently being accepted.
```
rosieyohannan:
```suggestion
1. If an image isn't listed above, it is not available. As the Convenience Image program is revamped, proposals for new images are not currently being accepted.
```

jekyll/_cci2/circleci-images.md
```diff
@@ -252,6 +252,14 @@ The following packages are installed via `curl` or other means.
...
+1. Old versions of software won't be rebuilt. Once an upstream image stops building the tag for a specific release, say Node.js v8.1.0, then we stop building it too. This means other tools in that image, such as `npm` in this example, will no longer be updated either.
```
rosieyohannan:
```suggestion
1. Old versions of software will not be rebuilt. Once an upstream image stops building the tag for a specific release, say Node.js v8.1.0, then we stop building it too. This means other tools in that image, such as `npm` in this example, will no longer be updated either.
```

jekyll/_cci2/circleci-images.md
```diff
@@ -252,6 +252,14 @@ The following packages are installed via `curl` or other means.
...
+1. We don't support building preview, beta, or release candidate images tags. On occasion they'll be available but these tags tend to cause our build system for Convenience Images to fail. If you need a non-stable release of of language, we suggest installing it via [an orb](https://circleci.com/orbs/) or a custom Docker image instead.
```
rosieyohannan:
```suggestion
1. We don't support building preview, beta, or release candidate images tags. On occasion they'll be available but these tags tend to cause our build system for Convenience Images to fail. If you need a non-stable release or language, we suggest installing it via [an orb](https://circleci.com/orbs/) or a custom Docker image instead.
```

rosieyohannan: Not sure this is what was meant? Just took a guess :-)

jekyll/_cci2/circleci-images.md
rosieyohannan: Maybe? Not sure this is quite right...

felicianotech: Umm. I'm still not sure but I think it's better than mine.

jekyll/_cci2/circleci-images.md
felicianotech: This should be "of a". I'll push a commit.

felicianotech:
@rosieyohannan I added all your suggestions but one. For that one I submitted a new commit with a fix.

PR #3988, opened 2019-12-03
add reusable content methods

rosieyohannan:
Find ways to single source content that is used in multiple locations to help all sections remain up to date.

- [x] Server version as a site variable
- [x] Reusable snippets in _includes

PR #4003, opened 2019-12-10
Create `enable-pipelines` snippet.

teesloane:
# Description
This follows along closely after @rosieyohannan 's https://github.com/circleci/circleci-docs/pull/3988 in which we are using a `snippets` folder for commonly used pieces of the docs. In this case, I am creating a snippet that instructs the user on how to enable pipelines for their project, which occurs (as far as I could grep) at least 5 times in our project

### DIFF
--- a/jekyll/_cci2/orbs-user-config.md
+++ b/jekyll/_cci2/orbs-user-config.md
@@ -17,20 +17,20 @@ The section below describes how you can set your platform version to 2.1 so you 
 
 Setting your platform version to work with orbs is a very simple process, requiring you to perform the following steps:
 
-1) Use CircleCI version 2.1 at the top of your `.circleci/config.yml` file.
+1. Use CircleCI version 2.1 at the top of your `.circleci/config.yml` file.
 
 `version: 2.1`
 
-**Note** If you do not already have Pipelines enabled, you'll need to go to **Project Settings -> Advanced Settings** and turn it on.
+**NOTE:** {% include snippets/enable-pipelines.md %}
 
-2) Add the `orbs` stanza below your version, thereby invoking the orb. The example below shows how you can invoke the `aws-cli` orb.
+2. Add the `orbs` stanza below your version, thereby invoking the orb. The example below shows how you can invoke the `aws-cli` orb.
 
 ```
 orbs:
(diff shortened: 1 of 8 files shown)


jekyll/_cci2/orbs-user-config.md
```diff
@@ -39,17 +39,18 @@ The example below demonstrates how you can invoke the `hello-build` orb in the `
...
 workflows:
-    "Hello Workflow":
-        jobs:
-          - hello/hello-build
+  "Hello Workflow":
+    jobs:
+      - hello/hello-build
 ```
 
 **Note:** If your project was added to CircleCI prior to 2.1, you must enable pipelines to use the `orbs` key.
```
rosieyohannan: Here too?

rosieyohannan: Or maybe you're just replacing it when it's a step not just a note.

teesloane: Thanks for pointing this one out. I actually think we should axe this line because it's implied that we're using 2.1 as it's requested at the top of this file (thus enabling it for orbs as well).

rosieyohannan:
There's sections like this too (first para) where we point people to the enable pipelines doc rather than telling them how to enable pipelines - my preference is to give the steps on the current page so using your new snippet. I think there are a few times we do this - https://circleci.com/docs/2.0/pipeline-variables/#section=configuration

teesloane:
@rosieyohannan - I added the snippet to the `pipeline-variables.md` doc, good catch. Let me know if you think of any other places for this snippet before we merge this in.

rosieyohannan:
@teesloane Found it here too: https://circleci.com/docs/2.0/reusing-config/#getting-started-with-config-reuse My screenshot can probably just be deleted here too and replaced with your new snippet 🎉

teesloane:
@rosieyohannan Good find! Added the snippet / removed your screenshot.

rosieyohannan:
Here's another one with my unnecessary screenshot: https://circleci.com/docs/2.0/configuration-cookbook/#configuring-your-environment-for-the-circleci-platform-and-orbs

PR #4015, opened 2019-12-16
Update Xcode docs with latest manifests

marcomorain:


marcomorain:
The table looks like this when rendered:

image

marcomorain:
image

PR #4023, opened 2019-12-17
Update configuration-cookbook.md

mixmatch:
# Description
Fix references to non-existing heroku orb version (1.0.0 -> 0.0.10)

# Reasons
```
#!/bin/sh -eo pipefail
# Cannot find circleci/heroku@1.0.0 in the orb registry. Check that the namespace, orb name and version are correct.
# 
# -------
# Warning: This configuration was auto-generated to show you the message above.
# Don't rerun this job. Rerunning will have no effect.
false
```


rosieyohannan:
good spot, thanks @mixmatch!

PR #4025, opened 2019-12-18
PyTest: test-reports -> test-results

sirlantis:
Improved consistency: the pytest example was the only one that diverged from the "test-results" nomenclature.


teesloane:
Good catch @sirlantis, thanks!

PR #4026, opened 2019-12-18
change v to lowercase and use Reference not Docs for API guides

rosieyohannan:


jekyll/_data/sidenav.yml
```diff
@@ -274,7 +274,7 @@ en:
           link: 2.0/configuration-reference/
         - name: API Reference
```
rosieyohannan:
```suggestion
        - name: API v1.1 Reference
```

PR #4055, opened 2020-01-09
Fix quotes in description.

felicianotech:
The Docs site doesn't currently support having raw double quotes in a page description. Either we need to remove the quotes completely or use HTML entities (which is what I do in this PR).


teesloane:
It appears that the original error has been repaired it in-branch. Thank you for finding the fix @felicianotech .

PR #4056, opened 2020-01-09
Change machine `Configurable resources (CPU/RAM)` to Yes

annapamma:
It seems this may be out-of-date, as the machine executor can now accept the resource_class key with the options medium and large.

Reference:
https://circleci.com/docs/2.0/configuration-reference/#machine-executor-linux

# Description
In the table comparing the Docker to Machine executor, I changed the value of the cell for Configurable resources (CPU/RAM), Machine from 'No' to 'Yes'.

# Reasons


rosieyohannan:
Good spot @annapamma! Thanks :-)

PR #4057, opened 2020-01-10
Switch to use correct terms for concurrency and parallelism

rosieyohannan:
Concurrency = more that one job running at the same time due to a workflow
Parallelism = setting `parallelism` >1 for a job to optimize a test suite (must inc test splitting).

Parallel/parallelism is often used in docs where concurrent/concurrency is more accurate. This PR is my attempt to fix this.


teesloane:
Nice work! I did a quick grep for `parallel` and think I might have found some more that can be swapped:

> Jobs have a maximum runtime of 5 hours. If your jobs are timing out, consider running some of them in parallel.

- from: [config-reference line 133](https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/configuration-reference.md#L133)

> Workflow configuration with sequenced jobs followed by parallel jobs chained to a final sequenced job. 

- from [example configs line 23](https://github.com/circleci/circleci-docs/blob/fix/concurrency-parallelism/jekyll/_cci2/example-configs.md#L23)

PR #4062, opened 2020-01-14
Add new GPU classes — attempt 2

appplemac:
A hopefully more successful version of circleci/circleci-docs/pull/4027.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -528,13 +528,18 @@ gpu.nvidia.medium\* |   8   | 30  | 1    | Nvidia Tesla T4 | 16
 
 ###### Example Usage
 ```yaml
+version: 2.1
+
 jobs:
   build:
-    machine: true
-    resource_class: gpu.nvidia.small
+    machine:
+      resource_class: gpu.nvidia.small
+      image: ubuntu-1604-cuda-10.1:201909-23
     steps:
       ... // other config
 ```
+
+See the [Available Linux GPU images](#available-linux-gpu-images) section for the full list of available images.
 
 ##### GPU Executor (Windows)
 
@@ -548,6 +553,7 @@ windows.gpu.nvidia.medium\* |   8   | 30  | 1    | Nvidia Tesla T4 | 16
 version: 2.1
 orbs:
   win: circleci/windows@2.3.0
+
 jobs:
   build:
     executor: win/gpu-nvidia
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -312,6 +312,20 @@ CircleCI supports multiple machine images that can be specified in the `image` f
 
 The machine executor supports [Docker Layer Caching]({{ site.baseurl }}/2.0/docker-layer-caching) which is useful when you are building Docker images during your job or Workflow.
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -484,33 +498,72 @@ jobs:
...
+      ... // other config
+```
+
+See the [Available Linux GPU images](#available-linux-gpu-images) section for the full list of available images.
+
+##### GPU Executor (Windows)
+
+Class                       | vCPUs | RAM | GPUs |    GPU model    | GPU Memory (GiB)
+----------------------------|-------|-----|------|-----------------|------------------
+windows.gpu.nvidia.medium\* |   8   | 30  | 1    | Nvidia Tesla T4 | 16
```
teesloane: I'd recommend using the `<sup>` with a number rather than the `\*`  tag as used in [other tables](http://localhost:4000/docs/2.0/configuration-reference/#executors-requires-version-21) across this doc

jekyll/_cci2/configuration-reference.md
```diff
@@ -484,33 +498,72 @@ jobs:
...
+version: 2.1
+orbs:
+  win: circleci/windows@2.3.0
+
+jobs:
+  build:
+    executor: win/gpu-nvidia
+    steps:
+      - checkout
+      - run: echo 'Hello, Nvidia GPU on Windows'
```
teesloane: Is there another way we can demonstrate that Nvidia is being used? Maybe using something like [this command](https://askubuntu.com/a/5420) but on Windows?

jekyll/_cci2/configuration-reference.md
```diff
@@ -484,33 +498,72 @@ jobs:
...
+```yaml
+version: 2.1
+
+jobs:
+  build:
+    machine:
+      resource_class: gpu.nvidia.small
+      image: ubuntu-1604-cuda-10.1:201909-23
+    steps:
+      ... // other config
```
appplemac: Todo: change the steps into "nvidia-smi" and "docker run --gpus all nvidia/cuda:9.0-base nvidia-smi" to show that both docker and non-docker support is availabe

appplemac:
@teesloane Addressed your comments. Removing `\*` was a bit gnarly as it’s in almost all resource class tables. There also isn’t a way to have a `<sup>` with the same number appear twice in the text, right? With the asterisk we could give customers a heads up on what the asterisk means both at the beginning of the section and at the end of the section. Let me know.

teesloane:
I see what you mean @appplemac. In the configuration reference there are several instances of `<sup>1</sup>` and they are respectively addressed following their appearance (more like footnotes rather than endnotes). So yes, it is possible to have multiple `<sup>` tags of the same value; they are not footnote/endnotes proper.

 However, I think we are a bit inconsistent in this document (for example, `## executors` and `### job_name` both use a `<sup>1 </sup>` which is then explained only in the latter occurrence. 

Mostly, I'm aiming for consistency as this is our longest and most popular document, but I'm not sure what the answer is. Generally, until we build jumping footnotes (💡) I would aim to put the marker's explanation not at the end of the doc but as close to the table. 

I'm ok with either solution, though, now that I see your point. I think you can go with your gut, and I'll look into cleaning up footnotes more thoroughly in another PR.

PR #4067, opened 2020-01-15
change orbs ref to reusable config ref

rosieyohannan:
Orbs reference doesn't seem like the right name for this doc. It covers all reusable content. Changing to reflect this and adding a link outside the orbs section to make it easier to find.

### DIFF
--- a/jekyll/_cci2/reusing-config.md
+++ b/jekyll/_cci2/reusing-config.md
@@ -1,13 +1,13 @@
 ---
 layout: classic-docs
-title: "Orbs Reference Guide"
-short-title: "Orbs Reference"
-description: "Reference guide for CircleCI 2.1 orbs"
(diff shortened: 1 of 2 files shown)


teesloane:
This change makes sense to me. Do we do anything to handle redirects for people visiting the old links?

thomasr8:
@rosieyohannan and @teesloane Very good question. Since we link to this document from other places, we will need to either update those links or create re-directs.

rosieyohannan:
@teesloane @thomasr8 my understanding is that as the link is made up of the name of the markdown doc, and I haven't changed that, all links should still work if this PR were to be merged?

rosieyohannan:
I've changed the top level heading but that never becomes part of the URL I don't think

teesloane:
Oh, right. Yes, I believe you are correct.

PR #4069, opened 2020-01-15
Update Xcode Version in all docs

marcomorain:
I'm currently looking at usage of macOS on CircleCI, by version, and
I've found many projects (~100) that have built twice with Xcode 9.3.1 in
the last 3 weeks. I beleive that this is because we have Xcode 9.3.1 as
the version in our 'hello world' example.

I've gone through the docs and updated all macOS snippets to specify
Xcode 11.3 instead.

Other changes:

- I've removed any quotation marks around the Xcode version where they are
  not necessary
- I've changed the config snippets to use config 2.1 where appropriate
- I've removed some references to the cocoapods-specs download script
  that is no longer required with newer versions of CocoaPods.

### DIFF
--- a/jekyll/_cci2/concepts.md
+++ b/jekyll/_cci2/concepts.md
@@ -83,8 +83,8 @@ An image is a packaged system that has the instructions for creating a running c
        image: circleci/classic:201708-01
 #...       
    build3:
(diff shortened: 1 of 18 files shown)


rosieyohannan:
Thanks so much for this @marcomorain!

PR #4075, opened 2020-01-18
Add executor footnote for executor selection limit

chris-mcdermut-vacasa:
# Description
Added a footnote below the executors info table to explain the superscripted `(1)` note which indicates that there is a selection limit of one for choosing an executor

# Reasons

The options `docker` `machine` `macos` and `windows` all have a `(1)` superscript but no close footnote to explain its meaning. Looking at other similar footnotes, it can be deduced that this is supposed to mean that there is a selection limit of exactly 1 for a chosen executor.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -103,7 +103,7 @@ working_directory | N | String | In which directory to run the steps.
 environment | N | Map | A map of environment variable names and values.
 {: class="table table-striped"}
 
-<sup>(1)</sup> exactly one of them should be specified. It is an error to set more than one.
+<sup>(1)</sup> One executor type should be specified per job. If more than one is set you will recieve an error.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -103,6 +103,8 @@ working_directory | N | String | In which directory to run the steps.
 environment | N | Map | A map of environment variable names and values.
 {: class="table table-striped"}
 
+<sup>(1)</sup> exactly one of them should be specified. It is an error to set more than one.
```
rosieyohannan:
```suggestion
<sup>(1)</sup> One executor type should be specified per job. If more than one is set you will recieve an error.
```

rosieyohannan:
Thanks so much for this @chris-mcdermut-vacasa! Just a minor editorial suggestion for the wording :-)

jekyll/_cci2/configuration-reference.md
chris-mcdermut-vacasa: Makes sense, I will update the footnote where I copied this note from as well 👍

chris-mcdermut-vacasa:
@rosieyohannan updated the original footnote where I got this text from

rosieyohannan:
Thanks so much @chris-mcdermut-vacasa!

PR #4084, opened 2020-01-23
[WIP]Download artifacts with api v2

rosieyohannan:
# Description
Add use case steps to api developers guide

### DIFF
--- a/jekyll/_cci2/api-developers-guide.md
+++ b/jekyll/_cci2/api-developers-guide.md
@@ -431,3 +431,73 @@ For a more detailed breakdown of each value returned in this request, please ref
 
 
 ## Download Artifacts
+
+The following section details the steps you need to follow to download artifacts generated when a job is run. We will go through the steps of returning a list of artifacts for a job, and then changing the command we run to download the full set of artifacts. If you are looing for instructions to download the latest artifacts for a pipeline, without needing to specify a job number, see our [API v1.1 guide](https://circleci.com/docs/2.0/artifacts/#downloading-all-artifacts-for-a-build-on-circleci) – keep checking back here as this functionality will be added to API v2 in the future.
+
+### Prerequisites
+
+Before making an API call, make sure you have met the following prerequisites:
+
+* You have set up a GitHub or BitBucket account with a repository to use with CircleCI.
+* You have completed CircleCI onboarding and you have a project setup.
+* You have a [personal API token](https://circleci.com/account/api).
+* You have been [authenticated](#get-authenticated) to make API calls to the server.
+
+### Steps
+
+1. First we will ensure your api token is set as an environment variable. You maybe have already done this during authentication, but if not, run the following substituting your personal api token:
+
+    ```
+    export CIRCLECI_TOKEN={your_api_token}
+    ```
+
+2.  Next we need the job number for the job you want to get artifacts for. You can find job numbers in the UI - either in the breadcrumbs on the Job Details page, or in the URL.
+
+    INSERT SCREENSHOT HIGHLIGHTING JOB NUMBER
+
+3.  Next we can use `curl` to return a list of artifacts for a specific job. Note, you can format JSON responses by piping the `curl` command into the `jq` utility if you have it installed: `curl ... | jq`:
+
+    ```sh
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/api-developers-guide.md
```diff
@@ -348,3 +348,73 @@ job -
...
+    Check the table below for help with formatting your API call correctly
+
+    Parameter | Description
+    --- | ---
+    `VCS` | Version Control System - either `github`/`gh` or `bitbucket`/`bb`
+    `org-name` | Organization name, or your personal username to your VCS
+    `repo-name` | Name of your project repo
+    `job-number` | The number for the job you want to download artifacts from - see step 2
+    
+    <BR>
```
teesloane: We should avoid using html in markdown -- if this is a spacing issue we might be able to solve it a different way.

jekyll/_cci2/api-developers-guide.md
```diff
@@ -348,3 +348,73 @@ job -
...
+    ```
+
+    Check the table below for help with formatting your API call correctly
+
+    Parameter | Description
+    --- | ---
+    `VCS` | Version Control System - either `github`/`gh` or `bitbucket`/`bb`
+    `org-name` | Organization name, or your personal username to your VCS
+    `repo-name` | Name of your project repo
+    `job-number` | The number for the job you want to download artifacts from - see step 2
```
teesloane: Most tables in our docs use a striped class, which you can add, as below:

```
    | Parameter      | Description                                                             |
    | ---------------| ------------------------------------------------------------------------|
    | `VCS`          | Version Control System - either `github`/`gh` or `bitbucket`/`bb`       |
    | `org-name`     | Organization name, or your personal username on your VCS                |
    | `repo-name`    | Name of your project repo                                               |
    | `job-number`   | The number for the job you want to download artifacts from - see step 2 |
```

jekyll/_cci2/api-developers-guide.md
```diff
@@ -348,3 +348,73 @@ job -
...
+    
+    <BR>
+    You should get a list of artifacts back - if the job you chose has artifacts associated with it. Here's and extract from the outut when requesting artifacts for a job that builds these docs:
```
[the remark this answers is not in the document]
rosieyohannan: @teesloane Yeah it looked weird due to the text starting straight after the table. I just stuck this in as a quick fix but I'm wondering why we should avoid HTML in cases like this?

jekyll/_cci2/api-developers-guide.md
rosieyohannan: @teesloane Yeah it looked weird due to the text starting straight after the table. I just stuck this in as a quick fix but I'm wondering why we should avoid HTML in cases like this?

jekyll/_cci2/api-developers-guide.md
rosieyohannan: Ah I think there's quite a bit of different table formatting around docs because I copied this from the config reference I think. I will update

jekyll/_cci2/api-developers-guide.md
teesloane: Generally, it's more that `<BR>` is not an ideal html tag -- it's not [semantic](https://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br) - although you could argue I'm being _pedantic_ ;)

Regardless, it looks like there are only [three ways](https://stackoverflow.com/a/48583859) to create blank lines/empty spaces in markdown — so feel free to go with the `<br/>`? One other alternative I can think of would be to maybe try inserting an `<hr />`  by using `***` or `---` which is semantic markdown.

rosieyohannan: AH ok! Yeah when I stuck <BR> in it was the result of the first stack overflow hit I got for how to add line space in markdown. Will check the other options.

rosieyohannan: Yeah I tried --- but then I get a horizontal rule so not sure. I will format the table correctly and maybe that will solve my problem anyway

rosieyohannan: That worked!

PR #4087, opened 2020-01-24
Clarify usage of machine image key for Server

heug:
# Description
Add additional clarification that it is possible to use other image keys for Server installs

# Reasons
Customer wasn't interpreting this in the right way, so wanted to clarify to avoid confusion

### DIFF
--- a/jekyll/_cci2/executor-types.md
+++ b/jekyll/_cci2/executor-types.md
@@ -120,7 +120,7 @@ jobs:
 
 All images have common language tools preinstalled. Refer to the [specification script for the VM](https://raw.githubusercontent.com/circleci/image-builder/picard-vm-image/provision.sh) for more information.
 
-**Note:** The `image` key is not required on self-hosted installations of CircleCI Server (see example above) but if it is used, it should be set to: `image: default`.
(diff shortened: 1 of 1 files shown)


rosieyohannan:
Hey @heug I'm not sure this is true? Happy to be wrong though of course! Here's a slack convo from when I tried to clear this up a while back: https://circleci.slack.com/archives/CAT9Q7TGW/p1574767654216200 ... Maybe what we need to do is point out that the actual default image that gets used here is changable by the sys admin for the server install? But I didn't think it was the actual image key that would be changed. Let me know what you think :-)

heug:
Gave it another go at clarification, thank you for linking me to that conversation!

rosieyohannan:
@heug Do you have examples of the other image keys that can be used? I was under the impression this was not possible.

heug:
`ubuntu-1604:201903-01` looked to be working, and a customer reported that they are able to use images from a private Docker registry

appplemac:
I believe it’s only possible to use Machine images with Machine executor. Using Docker images won’t work in the `image` key for the Machine executor.

rosieyohannan:
Closing this for now, but we can move the discussion to Slack 😀

PR #4090, opened 2020-01-24
Do not reference irrelevant add-ssh-key page

pmocek:
# Description
Remove reference to the [adding ssh keys document](https://circleci.com/docs/2.0/add-ssh-key) document from the "Enable Your Project to Check Out Additional Private Repositories" section.

# Reasons
The [adding ssh keys document](https://circleci.com/docs/2.0/add-ssh-key) is not relevant to enabling of a project to check out additional private repositories.  That document describes the addition of an SSH key *for deploying to one's server*, not for checking out repositories.


rosieyohannan:
Good spot, thanks for helping us clarify this @pmocek :-)

PR #4096, opened 2020-01-28
Adding iOS deployment documentation

BytesGuy:
# Description
Created a new document detailing Fastlane configurations for deploying to App Store Connect, TestFlight and VS App Center

# Reasons
Our original [iOS deployment docs](https://circleci.com/docs/2.0/testing-ios/#configuring-deployment) were fairly out of date and required revamping. These sample FastLane configurations provide users with an opportunity to do CD from CircleCI with their apps in a totally automated fashion.

Some of the wording and structure will need to be adjusted as this is just an early rough draft, but the bulk of the content is done.

I believe a separate document for this is required as it is a bit lengthy and is mostly focused on Fastlane, though if it would be better as a drop-in replacement to the [current section](https://circleci.com/docs/2.0/testing-ios/#configuring-deployment) then I am happy to change it around 😄 

I would like to add a Firebase sample in before this goes live too as that will cover all the major services.

### DIFF
--- a/jekyll/_cci2/deploying-ios.md
+++ b/jekyll/_cci2/deploying-ios.md
@@ -43,14 +43,22 @@ increment_build_number(
 
 ### Setting up
 
-TODO: Describe creating a new user for ASC and setting Fastlane envars
+To set up Fastlane to automatically upload iOS binaries to App Store Connect and/or TestFlight, a few steps need to be followed to allow Fastlane access to your App Store Connect account. For ease of use and security, it is advisable to create a new user, with limited privileges, in App Store Connect that will only be used for your CI jobs.
 
+1. Head over to the [Users and Access](https://appstoreconnect.apple.com/access/users) section of App Store Connect
+2. Click on the `+` to add a new user
+3. Fill out the form, giving the user a unique email (such as `ci@yourcompany.com`) and select the "App Manager" role along with selecting the app to provide access to.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/deploying-ios.md
```diff
@@ -0,0 +1,206 @@
+---
+layout: classic-docs
+title: Deploying iOS Applications
+short-title: Deploying iOS Applications
+categories: [platforms]
+description: Deploying iOS Applications
+order: /?
```
teesloane: I would just put a `1` here - I believe this is an old key that is no longer used by jekyll and so an arbitrary value is ok; I believe your build is breaking because `/?` is an invalid value.

BytesGuy: Ah thanks Tyler, I forgot to ask this in my opening comment. Will fix that now!

BytesGuy:
@TylerConlee / @marcomorain are we all set to merge this in? I'll update the main testing-ios doc after merging so it points to the new deployment doc 🙂

teesloane:
Yes! I think you pinged the wrong Tyler, by the way, @BytesGuy . Either way, let's SHIIIIIIIIP IT! Thanks for helping us fill out our iOS docs.

🚢 choo! 🚢

BytesGuy:
Whoops! That’s the stupid GitHub autocomplete for you haha 😄

Issue #4102, opened 2020-02-03
Work directory inside GOPATH is not required


AndersonQ:
# Locations
https://circleci.com/docs/2.0/language-go/

```
# directory where steps are run. Path must conform to the Go Workspace requirements
working_directory: /go/src/github.com/CircleCI-Public/circleci-demo-go
```

# Details
The doc page states the need to work inside GOPATH. However since Go 1.11 it is no longer required as it introduces [go modues](https://golang.org/cmd/go/#hdr-Modules__module_versions__and_more). The need to work inside the GOPATH depends upon the version of Go been used, rather than been a CircleCI requirement.

The way it's presented is a bit confusing. The example uses go 1.12 (`image: circleci/golang:1.12`) which do not require to work inside the GOPATH, therefore it might lead to understand that it's a CircleCi requirement.

To set `working_directory` is not required, therefore I suggest to remove the lines above from the docs.
Also the [circleci-demo-go](https://github.com/CircleCI-Public/circleci-demo-go/blob/master/.circleci/config.yml) does not set `working_directory` which makes the doc page inconsistent.

teesloane:
Closed in #4103

PR #4103, opened 2020-02-03
Remove `working_directory` as to work inside GOPATH is not required

AndersonQ:
# Description

The way it's presented is a bit confusing. The example uses go 1.12 (`image: circleci/golang:1.12`) which do not require to work inside the GOPATH, therefore it might lead to understand that it's a CircleCi requirement.

To set `working_directory` is not required, therefore I suggest to remove the lines above from the docs.
Also the [circleci-demo-go](https://github.com/CircleCI-Public/circleci-demo-go/blob/master/.circleci/config.yml) does not set `working_directory` which makes the doc page inconsistent.

# Reasons
Issue  #4102


teesloane:
closes #4102

PR #4108, opened 2020-02-04
update asciidoctor and associated workflows 

rosieyohannan:
Keeping asciidoc up to date so I can make use of new features in PDFs as they become available. Also, work on making better use of workflows for integrating the PDF content.

- [x] up asciidoctor-pdf version
- [x] fix reusable content for pdfs - use asciidoc `include::` rather than Liquid
- [x] only build PDFs for `server/*` branches
- [x] fix server only content display colour for PDF output (blue)


jekyll/_config.yml
```diff
@@ -136,4 +136,4 @@ analytics:
     go_button_clicked: go-button-clicked
 rollbar_env: development
 
-server_version: "2.18.3"
+# server_version: "2.18.4"
```
rosieyohannan: removing this for now as it's not working in the PDFs - I have a fix in mind but won't be ready for Server release :-(

.gitignore
```diff
@@ -11,5 +11,7 @@ jekyll/_includes/app.html
 jekyll/assets/js/*.bundle.js
 run-results/
 node_modules/
+release/tmp/CircleCI-Server-AWS-Installation-Guide.pdf
```
rosieyohannan: I'm still getting these files committed for some reason :/

rosieyohannan: Oh right yes that makes sense - Thanks!

PR #4115, opened 2020-02-05
Provide a link to pipeline variables from CIRCLE_COMPARE_URL env var

jeff-knurek:
# Reasons
When reading this section and finding an alternative, I was quite lost on what the alternative actually was.


rosieyohannan:
Thanks so much for this @jeff-knurek! 🎉

PR #4116, opened 2020-02-05
Clarify description of deploy key with write access

jdblischak:
# Description

I removed the reference to "user key" in the [description](https://circleci.com/docs/2.0/gh-bb-integration/#deployment-keys-and-user-keys) of a deploy key with write access.

# Reasons

In this one place in the documentation, it implies that a deploy key with write access is a "user key". This really confused me. This PR updates this section to be consistent with the rest of the documentation of deploy vs user keys.

Also, I indented the paragraph from PR #2958 to fix the numbering of the markdown list. Currently it is 1-1-2-3 when it should be 1-2-3-4.

cc: @pat-s


rosieyohannan:
Thanks so much for this @jdblischak! 🎉

PR #4117, opened 2020-02-05
Add note about namespace deletion

mvxt:
# Description
Just found out that namespaces can't be deleted. This piece of information is not visible in the orb docs, so I added it.

# Reasons
Information visibility


rosieyohannan:
Good shout @mvxt!

PR #4124, opened 2020-02-06
Static Install / Single-Box Install Minor Doc Updates

mvxt:
# Description
While experimenting w/ Static Install, just made a few updates to the instructions to make the steps more explicit and detailed.

# Reasons
Original directions had some ambiguity which could potentially lead to confusion.


rosieyohannan:
Thanks for this @mvxt!

PR #4126, opened 2020-02-06
fix include path

rosieyohannan:
Include path was incorrect for snippets in asciidoc files, I hadn't noticed due to asciidoctor safe mode being turned on so local staging version is displaying correctly. 

This PR removes the content (non-essential) until I work out a fix.

### DIFF
--- a/jekyll/_cci2/architecture.adoc
+++ b/jekyll/_cci2/architecture.adoc
@@ -8,7 +8,7 @@
 //:circle-warning: image:circle-warning.png[]
 //:circle-failure: image:circle-failure.png[]
 
-include::../_includes/snippets/server-only.adoc[]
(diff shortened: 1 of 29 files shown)


jekyll/_cci2/architecture.adoc
```diff
@@ -8,7 +8,7 @@
 //:circle-warning: image:circle-warning.png[]
 //:circle-failure: image:circle-failure.png[]
 
-include::../_includes/snippets/server-only.adoc[]
+include::../jekyll/_includes/snippets/server-only.adoc[]
```
rosieyohannan:
```suggestion
//include::../jekyll/_includes/snippets/server-only.adoc[]
```

PR #4129, opened 2020-02-07
sudo won't work here

HunterSherms:
went ahead and updated the docker-compose version while I'm in here.

# Description
removed `sudo`

upversioned docker-compose

# Reasons
I copy and pasted this code and it blew up. Trying to protect others from the same fate.


rosieyohannan:
Thanks for this @HunterSherms! Much appreciated :-)

PR #4138, opened 2020-02-12
Update/gpu

mvxt:
# Description
-  Added section about GPUs in Choosing Executor doc.
- Moved Docker-related sections and nested them under `Using Docker`.

# Reasons
Thought it'd be good to mention that we have GPU executors, and having all of the Docker-related information under main Docker heading made more sense. Came from a customer's feedback


mvxt:
Here's what the GPU section looks like:

Screen Shot 2020-02-12 at 3 29 19 PM

The new Table of Contents after moving the Docker sections under Docker:

Screen Shot 2020-02-12 at 3 31 07 PM

PR #4154, opened 2020-02-17
reusable snippets for content that appears multiple times

rosieyohannan:
- [x] machine resources
- [x] language guides

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -459,19 +459,14 @@ jobs:
 
 ##### Machine Executor (Linux)
 
-Class            | vCPUs | RAM
------------------|-------|------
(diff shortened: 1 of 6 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -459,19 +459,14 @@ jobs:
...
-2xlarge          | 16    | 64GB
-{: class="table table-striped"}
+{% include snippets/machine-resource-table.md %}
 
 ###### Example Usage
 ```yaml
 jobs:
   build:
-    machine: true
+    machine:
```
rosieyohannan: Just stuck this change in as spotted it was inconsistent with other examples. This Ubuntu image is, for now, the recommended image to use, rather than the default.

PR #4160, opened 2020-02-18
Convert `:ga` to `:availability :general`

ganezasan:
# Description
I converted `:ga` flag to `:availability :general`. 

# Reasons
If we didn't set `:availability :general`, no one can use that `resouce_class`. There is another option in `:availability`, it's a `:performance-plan`, but server customers don't use that plan.
So, I just wrote `required field` into that description.

> https://github.com/circleci/picard-services/blob/a5ce64b2a63b182e563eec84b610eb25c2e5f1eb/dispatcher/src/picard_dispatcher/task_def_schema.clj

And also, if we use `:ga`, Server can't trigger any builds, and there is no error message.


rosieyohannan:
Thanks so much for this @ganezasan! I'm just double checking the syntax with the release engineering team then I'll merge the change :-)

PR #4168, opened 2020-02-20
Configuration Reference Guide - Slate

thomasr8:
# Description
Initial upload of Slate-formatted Configuration Reference Guide

# Reasons


src-crg/README.md
```diff
@@ -0,0 +1,115 @@
+<p align="center">
+  <img src="https://raw.githubusercontent.com/lord/img/master/logo-slate.png" alt="Slate: API Documentation Generator" width="226">
+  <br>
+  <a href="https://travis-ci.org/lord/slate"><img src="https://travis-ci.org/lord/slate.svg?branch=master" alt="Build Status"></a>
```
rosieyohannan: Maybe don't want to have reference to TravisCI in our docs?

src-crg/source/index.html.md
```diff
@@ -0,0 +1,856 @@
...
+  - shell
+
+search: true
+---
+
+# CircleCI 2.1 Configuration Reference Guide
+
+## Overview
+
+This document is a reference for the CircleCI 2.1 configuration keys used in the `config.yml` file. The presence of a `.circleci/config.yml` file in your CircleCI-authorized repository branch indicates you want to use the 2.1 infrastructure.
```
rosieyohannan: The presence of the file only indicates you want to use CCI 2.x, 2.1 is indicated with the version key

src-crg/source/index.html.md
```diff
@@ -0,0 +1,856 @@
...
+
+You can see a complete `config.yml` example in our [full example](#full-example).
```
rosieyohannan: I don't think the full example is actually included in this doc?

src-crg/source/index.html.md
```diff
@@ -0,0 +1,856 @@
...
+
+**Note:** If you already have a CircleCI 1.0 configuration, the `config.yml` file enables you to test 2.1 builds on a separate branch, leaving any existing configuration in the old `circle.yml` style unaffected and running on the CircleCI 1.0 infrastructure in branches that do not contain `.circleci/config.yml`.
```
rosieyohannan: I would compare 2.1 and 2.0 here if anything, I feel like we don't need to mention 1.0 anymore?

src-crg/source/index.html.md
```diff
@@ -0,0 +1,856 @@
...
+jobs:
+  my-job:
+    executor: my-executor
+    steps:
+      - run: echo outside the executor
+```
+
+See the [Using Parameters in Executors](https://circleci.com/docs/2.0/reusing-config/#using-parameters-in-executors) section of the [Reusing Config]({{ site.baseurl }}/2.0/reusing-config/) document for examples of parameterized executors.
+
+## **`jobs`**
```
rosieyohannan: This is 2.0 config? I thought this slate version of the reference was just for 2.1 only keys?

src-crg/source/index.html.md
```diff
@@ -0,0 +1,856 @@
...
+
+Key | Required | Type | Description
+----|-----------|------|------------
+only | N | List | List of branches that only will be executed
+ignore | N | List | List of branches to ignore
+{: class="table table-striped"}
+
+Both `only` and `ignore` lists can have full names and regular expressions. Regular expressions must match the **entire** string. For example:
+
+``` YAML
```
rosieyohannan:
```suggestion
```yaml
```

src-crg/source/index.html.md
```diff
@@ -0,0 +1,856 @@
...
+  build:
+    branches:
+      only:
+        - master
+        - /rc-.*/
+```
+
+In this case, only "master" branch and branches matching regex "rc-.*" will be executed.
+
+``` YAML
```
rosieyohannan:
```suggestion
```yaml
```

src-crg/source/index.html.md
```diff
@@ -0,0 +1,856 @@
...
+steps | Y | List | A list of [steps](#steps) to be performed
+working_directory | N | String | The directory used to run the steps. Default: `~/project` (where `project` is a literal string, not the name of your specific project). Processes run during the job can use the `$CIRCLE_WORKING_DIRECTORY` environment variable to refer to this directory. **Note:** Paths written in your YAML configuration file will _not_ be expanded; if your `store_test_results.path` is `$CIRCLE_WORKING_DIRECTORY/tests`, then CircleCI will attempt to store the `test` subdirectory of the directory literally named `$CIRCLE_WORKING_DIRECTORY`, dollar sign `$` and all.
+parallelism | N | Integer | Number of parallel instances of this job to run (default: 1)
+environment | N | Map | A map of environment variable names and values.
+branches | N | Map | A map defining rules to allow/block execution of specific branches for a single job that is **not** in a workflow or a 2.1 config (default: all allowed). See [Workflows](#workflows) for configuring branch execution for jobs in a workflow or 2.1 config.
```
rosieyohannan: Setting branch execution for a job isn't available in 2.1 config

src-crg/source/index.html.md
```diff
@@ -0,0 +1,856 @@
...
+```yaml
+version: 2.1
+
+parameters:
+  run_integration_tests:
+    type: boolean
+    default: false
+
+workflows:
+  version: 2
```
rosieyohannan: Workflows version 2 is not required for 2.1 config I think

src-crg/README.md
teesloane: This is the vendored README file that comes with Slate — we will want to replace this entire thing with the docs needed on how to build this component of our docs.

rosieyohannan: Oh right I see

src-crg/source/index.html.md
thomasr8: Fixed

src-crg/source/index.html.md
thomasr8: fixed

src-crg/source/index.html.md
thomasr8: Fixed

src-crg/source/index.html.md
thomasr8: Removed note mentioning 1.0.

src-crg/source/index.html.md
thomasr8: Removed reference to 2.0 sample configuration, since this is a 2.1 document.

src-crg/source/index.html.md
thomasr8: @rosieyohannan You are right. Branch execution is not available in 2.1. Removed from page.

src-crg/source/index.html.md
thomasr8: @rosieyohannan . You are right here. Version 2 is not part of 2.1. Removed from page.

teesloane:
Nice work! This looks nice.

I've annotated a (large) screenshot of the local build (attached below). Stuff with red circles / wobbly lines (I'm bad with the mouse) indicate needed changes - most of them are the same type.

- we don't need to use `{: class="table table-striped"}` - it's a main docs site class
- some of the code blocks are not set to yaml
- we probably don't need to have the "Example:" - it would have normally had text below it in the original config-reference file but now all fenced code blocks will be moved to the third pane.
- we'll want to remove the footer
- change the header icon to CCI 

I think the main challenge with using slate for this document is that there is a lot of interlinking between documents and the configuration reference; slate is outside of jekyll and so variables like `{{site.base_url}}` inside slate will not link internally - which could pose a problem - either we would have to manually write in all links, or find some other solution. Similarly, all documents that would links to a slate configuration-reference would have to be hard-coded which makes things a bit more brittle unfortunately. 
-

thomasr8:
Added CircleCI docs logo

logo

teesloane:
I've added the circle logo up to replace the slate header. 
image

PR #4181, opened 2020-02-25
[CIRCLE-24236] Document artifacts service security

pete-woods:
# Description
Artifacts service may start redirecting non-HTML resources, so we should both document this, and explain the security considerations behind it.

# Reasons
https://circleci.atlassian.net/browse/CIRCLE-24236

### DIFF
--- a/jekyll/_cci2/security.md
+++ b/jekyll/_cci2/security.md
@@ -33,7 +33,7 @@ A few different external services and technology integration points touch Circle
 
 - **Dependency and Source Caches** Most CircleCI customers use S3 or equivalent cloud-based storage inside their private cloud infrastructure (Amazon VPC, etc) to store their dependency and source caches. These storage servers are subject to the normal security parameters of anything stored on such services, meaning in most cases our customers prevent any outside access.
 
-- **Artifacts** It is common to use S3 or similar hosted storage for artifacts. Assuming these resources are secured per your normal policies they are as safe from any outside intrusion as any other data you store there.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/security.md
```diff
@@ -33,7 +33,8 @@ A few different external services and technology integration points touch Circle
 
 - **Dependency and Source Caches** Most CircleCI customers use S3 or equivalent cloud-based storage inside their private cloud infrastructure (Amazon VPC, etc) to store their dependency and source caches. These storage servers are subject to the normal security parameters of anything stored on such services, meaning in most cases our customers prevent any outside access.
 
-- **Artifacts** It is common to use S3 or similar hosted storage for artifacts. Assuming these resources are secured per your normal policies they are as safe from any outside intrusion as any other data you store there.
+- **Artifacts** To help prevent other builds from accessing your browser local storage when viewing artifacts, HTML and XHTML pages are hosted on their own project-specific subdomain of `*.circle-artifacts.com`. Non-HTML artifacts will usually be (`302 FOUND`) redirected to an S3 URL to allow for the highest
+download speed. Because these artifact types are hosted on a single S3 domain, you should avoid entering sensitive data into your browser for these URLs.
```
taxonomic-blackfish: Should be:

> Because these artifact types are hosted on a single S3 domain, artifacts may access your browser local storage on HTML and XHTML pages, and so you should avoid...

The idea being we can't guarantee that stuff they load from S3 won't access local storage.

PR #4185, opened 2020-02-25
Replace wrong JavaScript repo link

david-szabo97:
# Reasons
Right now the JavaScript example points to the React Native example. In my opinion this is wrong and I've a found a React demo made by CircleCI so I guess that should be used instead.

### DIFF
--- a/jekyll/_includes/snippets/language-guides.md
+++ b/jekyll/_includes/snippets/language-guides.md
@@ -65,3 +65,4 @@
 [sample-scala]: https://github.com/ariv3ra/samplescala
 [circleci-demo-windows]: https://github.com/CircleCI-Public/circleci-demo-windows/
 [parallel-compile-circleci]: https://github.com/eddiewebb/parallel-compile-circleci/blob/master/.circleci/config.yml
+[circleci-demo-javascript-react-app]: https://github.com/CircleCI-Public/circleci-demo-javascript-react-app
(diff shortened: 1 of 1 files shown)


teesloane:
Hi there!

Thanks for filing an issue. While it may seem confusing, things are currently in the correct place (although they could use some work 😅)

Our NodeJS Javascript Example points to a repository that uses express, and the "Javascript" one you suggested changing points to React Native because it uses JavaScript as well :). 

And just a heads up: your change resulted in a dead link:

image

- Our markdown parser swaps in the links at the bottom of the doc into the table -- and there isn't any `circleci-demo-javascript-react-app` as you have pasted -- if you have found one we can use, you'll need to include a link to it below for the table's links to work!

Thank you!

david-szabo97:
@teesloane Thanks for the feedback. I wasn't aware of the links on the bottom.

This is the repo I found:
https://github.com/CircleCI-Public/circleci-demo-javascript-react-app

You are totally right though. How about leaving React Native row in the table as-is and just adding this React demo app as a new row?

teesloane:
Hmm! Since we actually have a "React Native" item in the table, why don't we go with your original change — I think you should be able to put the repo you found into the list at the bottom of the snippet and we should be good to go! Do you mind making that change?

david-szabo97:
@teesloane There we go!

PR #4191, opened 2020-03-02
Remove explanation about custom resource_class for machine executor

ganezasan:
# Description
Remove explanation about custom resource_class for machine executor. And add WARNING to share how to change the machine executor resources.

# Reasons
Updating `resource-definitions.edn` is not enough to change the machine and windows resource_class. We define the actual instance types in `app_config.edn` in `vm-service` container that made by replicated management console.

So, if we set custom resource_class for machine executor, the actual build machine isn't changed. 

> https://circleci.atlassian.net/browse/CIRCLE-24763


rosieyohannan:
Thanks for this @ganezasan! I just pushed a change to move the warning up in the doc and add a link. Let me know if it looks OK to you :-)

ganezasan:
@rosieyohannan Thank you for updating it! It looks all good to me :)

PR #4192, opened 2020-03-02
Remove windows from the executors since windows is using orbs

sophiedophie:
# Description
Removed `windows` from the executors list

# Reasons
Windows builds are using orbs now, not using executors


rosieyohannan:
This is a good point, I think we should replace this with the correct method for using the default windows executor from the windows orb as a reusable executor. I guess you just invoke the orb and then use `executor: win/default` when needed 🤔

teesloane:
Hmmm, I'm a bit on the other side of the fence.  While users can/should be using orbs, last I checked it is possible to specifically set the executor key to a windows image.  I think we should leave it in but make a superscript note regarding using orbs?

rosieyohannan:
@teesloane Ah I see, I did not realise this was possible. Totally happy for it to stay in if it does actually work. Maybe worth checking with @appplemac?

appplemac:
This does work but I recommend only mentioning the Orbs approach.

If possible, I’d suggest a note below the table with "looking for windows? it doesn’t use a resource class, click here for info on orbs instead".

sophiedophie:
@appplemac @teesloane @rosieyohannan I've just added the note for users to check the orb! Thanks for your all comments

PR #4194, opened 2020-03-02
add windows resource class details

rosieyohannan:
Add how to set windows resource class using orb parameter


rosieyohannan:
@appplemac I'm not sure about the indenting for `size` here but have asked for the orb registry page to be updated too

PR #4198, opened 2020-03-03
docs: docker executor can build docker images

olizilla:
Remove a note that suggests the machine executor is the only way to build a docker image on circleci... `setup_remote_docker` is a thing now.


rosieyohannan:
Woah good spot, @olizilla! Thanks 😄 🎉

PR #4210, opened 2020-03-09
Update caching instructions for yarn users

kylemh:
# Description
Advise on install command

# Reasons
I've noticed that the out-of-box instructions on caching dependencies using yarn doesn't work.

Example:
https://discuss.circleci.com/t/circle-ci-2-0-yarn-dependencies/30912/3

### DIFF
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -505,7 +505,6 @@ steps:
   - save_cache:
       paths:
         - ~/.cache/yarn
-        - node_modules
       key: yarn-packages-v1-{{ .Branch }}-{{ checksum "yarn.lock" }}
 ```
 
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -508,11 +508,11 @@ steps:
       key: yarn-packages-v1-{{ .Branch }}-{{ checksum "yarn.lock" }}
 ```
 
-We recommend using `yarn --frozen-lockfile --cache-folder=.cache/yarn` for two reasons.
+We recommend using `yarn --frozen-lockfile --cache-folder ~/.cache/yarn` for two reasons.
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/caching.md
```diff
@@ -508,6 +508,12 @@ steps:
       key: yarn-packages-v1-{{ .Branch }}-{{ checksum "yarn.lock" }}
 ```
 
+We recommend using `yarn --frozen-lockfile --cache-folder ~/.cache/yarn` for two reasons.
+
+1) `--frozen-lockfile` ensures that a whole new lockfile is created and it also ensures your lockfile isn't altered. This allows for the checksum to stay relevant and your dependencies should identically match what you use in development.
+
+2) The default cache location depends on OS: https://github.com/yarnpkg/yarn/blob/25d5526c8995443b03c113c44f9c8e9a79c2c074/src/util/user-dirs.js#L26 `--cache-folder ~.cache/yarn` ensures we're explitly matching our cache save location.
```
teesloane: I don't think we need to link directly to the Yarn codebase. Can you remove the link and start the next sentence accordingly?

kylemh:
Additionally, I've done a lot of testing surrounding improving build times for projects that have a 300mb node_modules and a 600mb one.

It seems like caching offers no benefit (and is actually worse in the larger instance) compared to simply passing node_modules around the workspace. I realize this issue is outside the scope of support, but I'm curious if you had thoughts on that.

jekyll/_cci2/caching.md
kylemh: Resolved

teesloane:
@kylemh - Unfortunately, I haven't got much to contribute on that end. This might get more discussion on our [discuss forum](https://discuss.circleci.com/), however. Still, I am curious what you mean by "compared to simply passing node_modules around the workspace" — if you care to expand on that, I would be keen to hear more.

PR #4245, opened 2020-03-18
Clarify updating Cocoapods

BytesGuy:
If a customer is using an older Xcode image, they will need to update Cocoapods if they want to use the CDN. This change provides the command for this.


teesloane:
Awesome, thanks @BytesGuy !

PR #4246, opened 2020-03-18
server known issue 2.19

rosieyohannan:
Add known issue to warn HTTP proxy Server customers not to upgrade until safe

### DIFF
--- a/jekyll/_cci2/v.2.19-overview.adoc
+++ b/jekyll/_cci2/v.2.19-overview.adoc
@@ -35,4 +35,6 @@ CAUTION: If at any time your organization name has been changed, there is a <<up
 
 ## Known Issues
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/v.2.19-overview.adoc
```diff
@@ -35,4 +35,6 @@ CAUTION: If at any time your organization name has been changed, there is a <<up
 
 ## Known Issues
 
+* **If you are using an HTTP proxy for your installation, do not upgrade to v2.19.x until further notice**. We have known issues around job step logging and docker commands customization that make the current 2.19.x releases incompatable with an HTTP proxy setup of CircleCI Server. This issue will be fixed in a future 2.19.x patch release and the 'safe' version will be flagged up clearly for affected customers.
```
rosieyohannan:
```suggestion
* **If you are using an HTTP proxy for your installation, do not upgrade to v2.19.x until further notice**. We have known issues around job step logging and docker commands customization that make the current 2.19.x releases incompatible with an HTTP proxy setup of CircleCI Server. This issue will be fixed in a future 2.19.x patch release and the 'safe' version will be flagged up clearly for affected customers.
```

PR #4273, opened 2020-04-03
fix container env var description

rosieyohannan:
slack convo: https://circleci.slack.com/archives/C0KBNJGRH/p1585757699025400

PR #4276, opened 2020-04-03
Clarify Docker environment variable info

Peeja:
Suggestions on top of #4273.

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -317,20 +317,19 @@ Once created, environment variables are hidden and uneditable in the application
 
 ## Setting an Environment Variable in a Container
 
-Environment variable can also be set for a Docker container. To do this, use the [`environment` key]({{ site.baseurl }}/2.0/configuration-reference/#docker--machine--macos--windows-executor). **Note**: environment variables set in this way are not available to steps run within the container, these env vars are only available for the entrypoint/command run by the container.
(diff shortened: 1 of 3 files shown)


jekyll/_cci2/env-vars.md
```diff
@@ -342,14 +341,13 @@ jobs:
...
       - image: circleci/python:3.6.2-jessie
-      environment:
+        environment:
           FLASK_CONFIG: testing
           TEST_DATABASE_URL: postgresql://ubuntu@localhost/circle_test?sslmode=disable
       - image: circleci/postgres:9.6
-      environment:
+        environment:
           POSTGRES_USER: conductor
           POSTGRES_DB: conductor_test
```
Peeja, line 350: @rosieyohannan Notable: These examples still aren't right. Neither of the images used in the primary containers in these jobs has `LABEL com.circleci.preserve-entrypoint=true`, and neither config specifies an `entrypoint` or a `command` explicitly, so the built-in `ENTRYPOINT` and `COMMAND` will be ignored. That means the specified `environment`s won't have anything useful to apply to. The only one that matters is the last one, `circleci/postgres:9.6`, which isn't a primary container, and which will actually run its `ENTRYPOINT`.

I don't know a good use case for forcing a primary container to run its `ENTRYPOINT` off the top of my head, so we'd need someone to come up with one if we want to demonstrate it.

rosieyohannan: Thanks so much for these clarifications @Peeja, I think I understand now. I'm going to merge your changes into my branch, and then just make this example way more generic - i.e. specify no images - until we have a good working example for this use case 👍 Then I'll come back to you for another review :-)

Peeja: @rosieyohannan Cheers!

PR #4279, opened 2020-04-06
[CIRCLE-25241] Add RAM disk reference documentation

pete-woods:
- Also provide a simple example for how to use it.

# Reasons
https://circleci.atlassian.net/browse/CIRCLE-25241

### DIFF
--- a/jekyll/_cci2/executor-types.md
+++ b/jekyll/_cci2/executor-types.md
@@ -102,7 +102,7 @@ Nearly all of the public images on Docker Hub and Docker Registry are supported 
 
 ### RAM disks
 
-A RAM disk is available at `/mnt/ramdisk` that offers a [temporary file storage paradigm](https://en.wikipedia.org/wiki/Tmpfs), similar to using `/dev/shm`.
+A RAM disk is available at `/mnt/ramdisk` that offers a [temporary file storage paradigm](https://en.wikipedia.org/wiki/Tmpfs), similar to using `/dev/shm`. Using the RAM disk can help speed up your build, provided that the `resource_class` you are using has enough memory to fit the entire contents   of your project (all files checked out from git, dependencies, assets generated etc).
 
 The simplest way to use this RAM disk is to configure the `working_directory` of a job to be `/mnt/ramdisk`:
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/executor-types.md
```diff
@@ -100,6 +100,30 @@ Docker Images may be specified in three ways, by the image name and version tag
 
 Nearly all of the public images on Docker Hub and Docker Registry are supported by default when you specify the `docker:` key in your `config.yml` file. If you want to work with private images/registries, please refer to [Using Private Images]({{ site.baseurl }}/2.0/private-images).
 
+### RAM disks
+
+If you suspect your builds are I/O limited, it can be helpful to use the default Docker RAM disk at `/dev/shm`. For certain builds, this can reduce build times dramatically.
```
teesloane, line 105: Should we be mentioning that it is helpful to use `/dev/shm` but then in the next sentence also say we cannot use it because it's `NOEXEC`?  

I would maybe write something to the effect that "A RAM disk is available at `/mnt/ramdisk` that offers a [temporary file storage paradigm](https://en.wikipedia.org/wiki/Tmpfs) similar to using `/dev/shm`.

pete-woods: Done!

jekyll/_cci2/executor-types.md
```diff
@@ -100,6 +100,28 @@ Docker Images may be specified in three ways, by the image name and version tag
 
 Nearly all of the public images on Docker Hub and Docker Registry are supported by default when you specify the `docker:` key in your `config.yml` file. If you want to work with private images/registries, please refer to [Using Private Images]({{ site.baseurl }}/2.0/private-images).
 
+### RAM disks
+
+A RAM disk is available at `/mnt/ramdisk` that offers a [temporary file storage paradigm](https://en.wikipedia.org/wiki/Tmpfs), similar to using `/dev/shm`.
```
teesloane, line 105:
```suggestion
A RAM disk is available at `/mnt/ramdisk` that offers a [temporary file storage paradigm](https://en.wikipedia.org/wiki/Tmpfs), similar to using `/dev/shm`. Using the RAM disk can help speed up your build, provided that the `resource_class` you are using has enough memory to fit the entire contents   of your project (all files checked out from git, dependencies, assets generated etc).
```

PR #4289, opened 2020-04-09
Added info about inability to trigger 2.1 jobs

the21st:
# Description
Copied the note "You cannot currently trigger jobs that use 2.1 config from the API" that is included here https://circleci.com/docs/2.0/api-job-trigger/

# Reasons
I am having trouble manually triggering a job via the API and now I noticed the note saying this is incompatible with 2.1 configs. If this was included in this page I'd have seen it sooner.


the21st:
@teesloane Didn't know you can trigger a pipeline via the 2.1 api - could you point me to the docs please?

PR #4294, opened 2020-04-14
fix example for env_var_name

willsalz:
# Description
The example for `env_var_name` mislabeled the new file as the original file.

# Reasons
I was super confused.


rosieyohannan:
Ooo good spot, thanks, @willsalz!

PR #4295, opened 2020-04-15
Update test collection docs for Jest

jamescrowley:
# Description
* Remove reference to jest.config.js. This Jest config file is not required, as the example includes the parameters in the CLI arguments.
* Updated example CircleCI config for Jest test data. JEST_JUNIT_OUTPUT_DIR should be a folder, not a file, `store_test_results` was not present as a step, and should be for this config to work (and to be similar to other examples in this document)
* Added caveats for referenced article, as it's now out of date

### DIFF
--- a/jekyll/_cci2/collect-test-data.md
+++ b/jekyll/_cci2/collect-test-data.md
@@ -490,7 +490,7 @@ steps:
       path: ./reports/junit
 ```
 
-For a full walkthrough, refer to this article by Viget: [Using JUnit on CircleCI 2.0 with Jest and ESLint](https://www.viget.com/articles/using-junit-on-circleci-2-0-with-jest-and-eslint).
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/collect-test-data.md
```diff
@@ -470,17 +470,9 @@ A working `.circleci/config.yml` section might look like this:
 #### Jest
 {:.no_toc}
 
-To collect Jest data, first create a Jest config file called `jest.config.js` with the following:
```
chanmya13C, line 473: https://github.com/circleci/circleci-docs.wiki.git

jamescrowley: @chanmya13C that link doesn't work for me, was there something I should do with this? thanks!

teesloane: Sorry, I believe the above comment is spam.

PR #4298, opened 2020-04-15
Migration from TeamCity: Template -> Meta-Runners

dmitry-treskunov:
I think that Orbs are similar to TeamCity meta-runners, but not templates: https://www.jetbrains.com/help/teamcity/working-with-meta-runner.html

Meta-runners are basically XML files that can be shared between different TeamCity servers and used as build steps.

### DIFF
--- a/jekyll/_cci2/migrating-from-teamcity.adoc
+++ b/jekyll/_cci2/migrating-from-teamcity.adoc
@@ -433,9 +433,9 @@ jobs:
 ----
 |===
 
-=== Meta-Runners
+=== Build Templates/Meta-Runners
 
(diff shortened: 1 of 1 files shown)


teesloane:
Hmm. Interesting. From a cursory glance at Team City's docs on Templates it sounds like templates are still kind of similar?

> Build configuration templates allow you to eliminate duplication of build configuration settings. If you want to have several similar (not necessarily identical) build configurations and be able to modify their common settings in one place without having to edit each configuration, create a build configuration template with those settings. Modifying template settings affects all build configurations associated with this template.

Perhaps they may not map one-to-one to Orbs functionality, so maybe we could set the heading/body text to say `Build Templates/Meta-Runners` ?

dmitry-treskunov:
Sounds good

PR #4313, opened 2020-04-21
The pipelines feature is no longer in preview

stig:
All new projects have been using pipelines for a long time, and we're
actively migrating old projects to pipelines.


rosieyohannan:
Thanks so much for this @stig!

PR #4321, opened 2020-04-24
size up h6 so they aren't smaller than main text

rosieyohannan:
This is probably a very clunky way of getting the desired result but.....

h6 now same size at h5, but h6 is generally in italics.

Original report here: https://circleci.slack.com/archives/C0KBNJGRH/p1587401595092400

before:
Screenshot 2020-04-24 at 08 55 34

after:
Screenshot 2020-04-24 at 08 55 47

### DIFF
--- a/jekyll/assets/css/main.scss
+++ b/jekyll/assets/css/main.scss
@@ -419,3 +419,7 @@ pre {
   border: 1px solid grey;
   border-radius: 3px;
 }
+
+h6{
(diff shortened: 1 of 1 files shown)


teesloane:
I think this is fine fix for now! eventually we could scale the sizes from h1 -> h6, starting at the smallest according to body text. This will probably be best left to design, so this is a good patch for now!

PR #4336, opened 2020-04-30
Fix incorrect aws-ecr-orb command name

takp:
# Description

The orb command names are incorrect.

Sources:
- https://github.com/CircleCI-Public/aws-ecr-orb/blob/master/src/commands/build-and-push-image.yml
- https://circleci.com/orbs/registry/orb/circleci/aws-ecr#jobs-build-and-push-image

# Reasons

When I tried to use aws-ecr-orb with `build_and_push_image` command (same as the document), it returned an error: `Cannot find a definition for job named aws-ecr/build_and_push_image`

### DIFF
--- a/jekyll/_cci2/configuration-cookbook.md
+++ b/jekyll/_cci2/configuration-cookbook.md
@@ -90,14 +90,14 @@ orbs:
 workflows:
   build-and-deploy:
     jobs:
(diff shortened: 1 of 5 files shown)


rosieyohannan:
@takp Thanks so much for spotting this error, and taking the time to fix it for us. Much appreciated!! ☺️

PR #4354, opened 2020-05-08
Add: note insights is only available on server.

teesloane:
Insights are only available on our server product now; this warning directs users to check out the api-v2. 

screenshot:
image


rosieyohannan:
Oo I like it, very clear!

PR #4362, opened 2020-05-10
suggest an edit gh-bb-integration.md

ju-hyun-park:
suggest an edit.

either GitHub and Bitbucket -> either GitHub or Bitbucket


rosieyohannan:
Thanks for fixing this @ju-hyun-park! ☺️

PR #4366, opened 2020-05-11
save PDFs as artifacts for easier reviews

rosieyohannan:


.circleci/config.yml
```diff
@@ -48,7 +48,6 @@ commands:
 # Workflows orchestrate a set of jobs to be run;
 # the jobs for this pipeline are # configured below
 workflows:
-  version: 2
```
rosieyohannan, line 51: not required in 2.1 config

.circleci/config.yml
```diff
@@ -149,11 +148,13 @@ jobs:
...
           name: Build PDFs for Server
-          command: ./scripts/build_pdfs.sh
-      - persist_to_workspace:
-          root: *workspace_root
-          paths:
-            - pdfs
+          command: ./scripts/build_pdfs_local.sh
+      - store_artifacts: 
+          path: release/tmp/
+     # - persist_to_workspace:
```
rosieyohannan, line 154: For now I'm manually putting the PDFS in the right location when a release happens. But if each build can generate PDFs as artifacts that will make reviewing easier.

PR #4377, opened 2020-05-13
📗Python language guide update

teesloane:
- imports customer engineering's new python guide.


teesloane:
Update: waiting on a review and the sample app also needs to be setup and run:

image

teesloane:
Not sure how long this will be blocked for, so will close and re-open when new sample-apps are generated.

PR #4379, opened 2020-05-14
fix: typo /_cci2_ja/code-coverage.md

2357gi:
# Description
Changed 2 occurrences of config.yaml into config.yml

# Reasons
Because .yaml is not recognized.


teesloane:
Thanks!

PR #4384, opened 2020-05-15
Minor English fixes

dsinn:
Where correct, use possessive pronoun "its" instead of contraction for "it is"

# Description
Just grammar.

# Reasons
Grammatical correctness.


rosieyohannan:
Thanks for these fixes, @dsinn! 😀

PR #4386, opened 2020-05-18
Fix psycopg2 typo

pkatseas:
Quick typo fix


rosieyohannan:
Good spot, thank you @pkatseas! ☺️

PR #4414, opened 2020-05-26
pins the docker-ce-cli version to prevent api mismatch error

dullyouth:
# Description
Pins the docker-ce-cli version to prevent api mismatch errors

# Reasons

Without pinning the CLI version, the default install uses too new a version of the CLI. This pins the version to prevent an error

Below is the result of the script being executed as defined in the documentation
```
# docker ps
Error response from daemon: client version 1.40 is too new. Maximum supported API version is 1.39
```

```
# dpkg -l |grep docker
ii  docker-ce                        5:18.09.9~3-0~ubuntu-xenial                amd64        Docker: the open-source application container engine
ii  docker-ce-cli                    5:19.03.9~3-0~ubuntu-xenial                amd64        Docker CLI: the open-source application container engine
```

### DIFF
--- a/jekyll/_cci2/update-nomad-clients.adoc
+++ b/jekyll/_cci2/update-nomad-clients.adoc
@@ -149,7 +149,7 @@ curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
 add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
 apt-get install -y "linux-image-$UNAME"
 apt-get update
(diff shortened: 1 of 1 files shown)


felicianotech:
@dullyouth This was due to an error with the Docker v19.03.9 release which has since been fixed. If you're using a CircleCI Convenience Image, it should contain Docker v19.03.8 or older, or v19.03.11 or newer, all of which will work just fine.

dullyouth:
Incorrect. This specific issue was on 18.09.9, which is defined in your nomad user data documentation as the version of docker to install.

michelle-luna:
Thank you, sorry for the long delay and confusion (this doc is for Server, not cloud).

PR #4426, opened 2020-05-29
Make H1's linkable with proper margins. 

teesloane:
Fixes [reported issue](https://circleci.slack.com/archives/C0KBNJGRH/p1590585060233300)

### DIFF
--- a/jekyll/assets/css/main.scss
+++ b/jekyll/assets/css/main.scss
@@ -151,7 +151,7 @@ div.article-container{
 	width: calc(100% - 260px);
 }
 	article{
-    h1,
+		h1,
 		h2,
 		h3,
 		h4,
--- a/jekyll/assets/css/main.scss
+++ b/jekyll/assets/css/main.scss
@@ -151,6 +151,7 @@ div.article-container{
 	width: calc(100% - 260px);
 }
 	article{
+    h1,
 		h2,
 		h3,
 		h4,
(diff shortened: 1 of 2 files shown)


teesloane:
Hey @felicianotech ! This is pretty low-priority so no rush on the review. I'm wondering if you know if there's a reason we don't make our `article h1`'s linkable. This PR make the small link icon appear beside h1 headings, and also enables us to link to h1 anchors and keeping them in view.

felicianotech:
The typical reason is because an HTML page should only have a single `H1`, presenting the title of the page and should be related to the `<title>` tag. The "link" for the `H1` then is the URL of the page you are reading, which is already available in the browser's address bar.

As an example of this, here's four different documentation websites. They all user a hover icon and/or clickable headers to give you links for that section yet all three do it for H2s and below and not the H1:

- [Docker Docs](https://docs.docker.com/develop/)
- [Stripe Docs](https://stripe.com/docs/connect)
- [Vue.js Docs](https://vuejs.org/v2/guide/components-registration.html#Component-Names)
 - {DigitalOcean Docs](https://www.digitalocean.com/docs/kubernetes/how-to/create-clusters/)

teesloane:
@felicianotech That makes a lot of sense. Unfortunately we have some docs that use h1 `#` markdown headers throughout the docs. I will run a grep search and run a fix in a new PR. Thanks for your input! I will close this PR and reference it in the upcoming fix.

PR #4433, opened 2020-06-01
[experimental - draft] Try moving server content to separate page

rosieyohannan:


jekyll/_config.yml
```diff
@@ -90,6 +90,13 @@ collections:
     permalink: /2.0/:path/
     prefix: "/2.0"
     selectorName: "2.0"
+  ccis2:
```
rosieyohannan, line 93: This is where I thought I needed to add the new content to get it to build...

teesloane:
I believe the reason this is failing (it almost works!) is because the duplicated content from `2.0` has an `.index.md` file with this front matter:

```md
---
layout: classic-docs
title: "Welcome to CircleCI Documentation"
description: "Welcome to CircleCI Documentation"
permalink: /2.0/
---
```

where the `permalink` is obstructing this from publishing the proper new url you created in the config.

***

For structure, I would recommend that we create a `_server` folder, that would look something like this (see highlighted section in the left:

image

This way, we _could_ do versioning in the url by duplicated folders (2.19, 2.20, 3.0, etc), and everything would fall into place, rather than having to group by collection as we do now.

rosieyohannan:
@teesloane OK! I moved to content to _server as suggested and changed the permalink in index.md and it's building correctly now! 🎉 

Only problem is there's no side navigation as it seems that requires the language to be set - which highlights another problem with adding the Server content in this way... getting the language switcher working.

SO... `_layouts/classic-docs.html` includes the sidebar, and then `_includes/sidebar.html` adds in the side navigation depending on which language is set.

I had added the Server navigation in `_data/sidenav.yml`, which isn't going to work because we need the language switching to work for Server too.

So do we need separate layout, sidebar and sidenav for this new page maybe? 🤔

teesloane:
@rosieyohannan I was wondering if the sidebar would be a problem. To be honest, we should probably have a totally different sidebar for server docs. I haven't looked into how to do this yet, but I will poke at it this afternoon and see what I can come up with 🤔 .

rosieyohannan:
@teesloane since your last comment I've tried making a separate layout and sidebar, but then we talked somewhere (slack I think) about needing dynamic URL generation for some reason. I will try to find the conversation...

rosieyohannan:
@teesloane https://circleci.slack.com/archives/DH23Y7MTN/p1591294525001800

teesloane:
Thanks for the ping @rosieyohannan. So, I still need to think about this, but the [original post](https://justwriteclick.com/2017/07/30/investigating-jekyll-for-versioned-content/) that we discussed could probably solve our problems. I still need to do some research, but I believe we can avoid having to duplicate our content within the repo, and instead use a proper release cycle based on git tags/stable branches, and then write a script that will build this content  at build time, as suggested in the article.

The challenges with this are:

1. We need to be more strict about release timings, and using git.
2. We have to introduce a change into our build process (our build process will also become longer 😄 )
3. We will have to go through server content and change all the links to use a dynamic variable, set  in config.yml, so that the output content goes to the correct path. 

I will take this for a spin soon, and see how far I can get. Thanks for your hard work on this so far.

rosieyohannan:
@teesloane yippee! This sounds like the right way forward to me, always a niggling feeling that having duplicated in there couldn't be the most efficient way to do this.

Thanks so much for working on this!

rosieyohannan:
@teesloane closing this as it's superseded by badges and a million other ideas!

PR #4444, opened 2020-06-07
Fix example in Logic Statements

uraway:
# Description

Add `condition` key to the logic statement example

# Reasons

Validation passes, but it does not seem to be processed correctly without `condition` key.


sulami:
Hi,

the example is taken from `when` on a workflow, where there is no `condition` key. You're right that `condition` is required for the `when` step. We'll update the example to make that clearer and avoid future confusion.

rosieyohannan:
Fix here: https://github.com/circleci/circleci-docs/pull/4445 ... Thanks for pointing this out, @uraway!

PR #4457, opened 2020-06-15
Add note for the skip build

ganezasan:
# Description
Added a note that `[ci skip]` doesn't cancel the scheduled workflow. 
The message below is the answer from the engineering team.
> If you change the cron schedule and put [ci skip] into the commit message then we never tell workflows-conductor about the updated cron schedule.

# Reasons
We got a question about this, and it's better to write on the document.
> https://circleci.slack.com/archives/C4C9J5KHN/p1591346914466500


roopakv:
From that comment, it also seems like if I change the corn schedule in a PR with commit message `[ci skip]`. That change is never saved or passed on to the conductor service. Which I don't see the docs having reflected.

I only have half the context so I might be wrong :)

PR #4459, opened 2020-06-15
Add an alternative save/restore cache approach

### LINKED ISSUE #4458: Alternative suggestion for Python dependency caching approach
The [current recommended approach](https://circleci.com/docs/2.0/language-python/#cache-dependencies) requires:

1. a set of `sudo chown` commands to be executed in order to restore, which feels heavy-handed. 
2. `pipenv` to be installed, which is handy because it's a great tool, but still not the de factor package manager for Python, I think.

I would like to suggest an alternative approach. See my upcoming PR for details.

dawngerpony:
I wasn't a big fan of the approach to save/restore cache that requires two `sudo chown` commands so I have suggested an alternative approach.

# Description
Add a second paragraph suggesting a different approach to dependency caching.

# Reasons
Fixes #4458

### DIFF
--- a/jekyll/_cci2/language-python.md
+++ b/jekyll/_cci2/language-python.md
@@ -181,6 +181,32 @@ jobs:
 Use the `chown` command
 to grant CircleCI access to dependency locations.
 
+Alternatively, the following example avoids using `pipenv` and installs all dependencies
+into the user's home directory, negating the need for a set of `sudo chown` commands.
+
+{% raw %}
+
+```yaml
+version: 2
+jobs:
+  build:
+    # ...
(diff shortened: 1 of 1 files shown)


teesloane:
Hi there! Thanks for suggesting some changes.  While I agree with your changes, I do have to mention the following:

- if we make this change, we will also have to change the caching section above, which at that point, would be a lot of duplication of content.
- Our python sample application is in the process of being re-written to make use of orbs, so most of this content will be made redundant shortly.

Thanks for taking the time to contribute but for the above reasons I will be closing this PR.

PR #4479, opened 2020-06-22
add SSH key troubleshooting feedback

jkzilla:
# Description
Add the sentence or something similar to - Note that `~/.ssh/config` file is what the `add_ssh_keys` step creates or appends to along with adding the key to ~/.ssh/.

# Reasons

https://circleci.zendesk.com/agent/tickets/74412

Feedback from a customer having issues from trying to use the key before a custom git push bash script - 

"Thank you for letting me know that what the add keys step does is to modify ~/.ssh/config - that's extremely useful information, and I really wish it had been stated explicitly in the documentation I linked above."

### DIFF
--- a/jekyll/_cci2/add-ssh-key.md
+++ b/jekyll/_cci2/add-ssh-key.md
@@ -59,7 +59,7 @@ CircleCI also will not accept OpenSSH's default file format - use `ssh-keygen -m
 Even though all CircleCI jobs use `ssh-agent`
 to automatically sign all added SSH keys,
 you **must** use the `add_ssh_keys` key
-to actually add keys to a container.
+to actually add keys to a container. Note that `~/.ssh/config` file is what the `add_ssh_keys` step creates or appends to along with adding the key to ~/.ssh/.
 
 To add a set of SSH keys to a container,
 use the `add_ssh_keys` [special step]({{ site.baseurl }}/2.0/configuration-reference/#add_ssh_keys)


jkzilla:
Update from customer: 
BTW, here are some things I think https://circleci.com/docs/2.0/add-ssh-key/ in particular needs to be explicit and clear about:

* Can you list multiple fingerprints in a single add_ssh_keys step? (Yes, I believe this is supported)

* What filename the add_ssh_keys step puts the key in. That is, I should be able to know that key would've been placed in ~/.ssh/id_rsa_0d83ea403b755f799d3d40b59a0e1db5 from just reading the documentation, without having to run build jobs with debug output.

* What specifically an add_ssh_keys step does. I believe it does this, but I may still be mistaken about something or missing something:

For each fingerprint,
- Get the key with that fingerprint from this project's settings, and write it to a file in ~/.ssh (explain how the name for this file is generated)
- Append a stanza to ~/.ssh/config that looks like this:

Host <hostname from this key in project settings>
IdentitiesOnly yes <--- if this is configurable, explain exactly how to configure it, here
IdentityFile /home/www-data/.ssh/<KEY_FILENAME>

I think, as of now, that add_ssh_keys does *not* actually explicitly add an identity to the ssh-agent. But either way, this document would explicitly say whether it does or not.

https://circleci.com/docs/2.0/gh-bb-integration/ just has this enigmatic phrase, "When you push to your GitHub repository from a job, CircleCI will use the SSH key you added." That should be changed to a) explicitly say that add_ssh_keys modifies .ssh/config, and b) link to the add_ssh_keys documentation for more detail.

Also, steps 2 and 3 on that list are ambiguous about something else. For github, it says "paste the key you created in step 1", and for circleci it says "add the key you created in step 1". In both case "the key" is ambiguous, since this is a key pair, and in fact these same words mean two different things. For step 2, github, it means the _public key_ from the key pair; for step 3, circleci, it means the _private key_ from the key pair. I understood from context which key to put where, but not everyone will know. Some people trying to follow this documentation will be confused, or paste the wrong keys in and not think about it until it doesn't work. The documentation should be clear about which key to paste into which system.

PR #4480, opened 2020-06-23
Fix SEC-1344

ashafer01:
# Description
Switch semantic priority of `curl|bash` and package managers for local CLI installation as piping curl to bash is generally considered bad.

# Reasons
https://circleci.atlassian.net/browse/SEC-1344


taxonomic-blackfish:
Security says :+1:

marcomorain:
@ashafer01 i've added a commit on top to mention Chocolatey.

LGTM though.

PR #4497, opened 2020-06-29
Update Enable Checks

thomasr8:
Added a note about how users can either trigger a re-run of all checks, or just failed checks. This is based on JIRA ticket Circle-21682.

# Description

# Reasons


rosieyohannan:
Looking at the slack conversation attached to the jira ticket it sounds like the issue is that whether customers choose "rerun failed checks" or "rerun all checks" on the github side, either will lead to a full rerun of all workflows at the CircleCI end. Maybe we could make that more explicit in the note? If you agree that's a correct reading of the issue of course :-)

PR #4498, opened 2020-06-30
edits to reflect current auto-cancel feature status

rosieyohannan:
Now that all Cloud customers have pipelines enabled by default we need to update what we have documented for the auto-cancel builds feature, while keeping in relevant info for Server customers.

### DIFF
--- a/jekyll/_cci2/skip-build.md
+++ b/jekyll/_cci2/skip-build.md
@@ -63,10 +63,13 @@ If you are frequently pushing changes to a branch, you increase the chances of b
 
 To save time, you can configure CircleCI to automatically cancel any queued or running builds when a newer build is triggered on that same branch.
 
-**Note:**
-Your project's default branch will never auto-cancel builds. This feature only applies to non-workflow builds, builds triggered by pushes to GitHub, or workflow builds that use pipelines.
+**Note:** Your project's default branch (usually `master`) will never auto-cancel builds. 
 
-### Steps to Enable Auto-Cancel for New Builds Triggered by Pushes to GitHub without Workflows
+### Auto Cancel for CircleCI Server installations
+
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/skip-build.md
```diff
@@ -59,38 +59,37 @@ If you push multiple commits at once, a single `[ci skip]` or `[skip ci]` will s
...
-**Note:**
-Your project's default branch will never auto-cancel builds. This feature only applies to non-workflow builds, builds triggered by pushes to GitHub, or workflow builds that use pipelines.
+**Note:** Your project's default branch (usually `master`) will never auto-cancel builds. 
 
-### Steps to Enable Auto-Cancel for New Builds Triggered by Pushes to GitHub without Workflows
+### Steps to Enable Auto-Cancel for Pipelines Triggered by Pushes to GitHub or the API
 {:.no_toc}
 
-1. In the CircleCI application, go to your project's settings by clicking the gear icon next to your project.
+**Notes:** It is important to carefully consider the impact of enabling the auto-cancel feature, for example, if you have configured automated deployment jobs on non-default branches. Auto-cancelling workflows requires enabling pipelines.
```
gordonsyme, line 71: ~ I think we can drop `Auto-cancelling workflows requires enabling pipelines.`, everything in dot-com has Pipelines enabled.

If these docs are also used for Server then we can keep it but modify a bit.

rosieyohannan: Ah yes I missed that, thanks!

rosieyohannan, line 71:
```suggestion
**Notes:** It is important to carefully consider the impact of enabling the auto-cancel feature, for example, if you have configured automated deployment jobs on non-default branches.
```

PR #4508, opened 2020-07-02
Anchor to rack < 2.1.0

makotom:
# Description
Downgrading `rack` to 2.0.9 to fix Config Reference Guide.

# Reasons
I don't know the reason but it worked. :P

### DIFF
--- a/src-crg/Gemfile.lock
+++ b/src-crg/Gemfile.lock
@@ -93,7 +93,7 @@ GEM
       activesupport (>= 3.1)
     parallel (1.10.0)
     public_suffix (2.0.5)
-    rack (2.2.3)
+    rack (2.0.9)
     rb-fsevent (0.9.8)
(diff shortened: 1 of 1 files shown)


glenjamin:
What made you think of trying that?

makotom:
> What made you think of trying that?

Because `bundle exec middleman build --verbose --clean` said the exception was raised in `vendor/bundle/ruby/2.6.0/gems/rack-2.2.3/lib/rack/lint.rb`, and I found out that [the version of `rack` was changed recently](https://github.com/circleci/circleci-docs/commit/7e7ed0051592ed9690eda2ae12f6b9052a095e6a#diff-97b6c9a96e3c932bdf76f364cfc29c47).
I had to try virtually one-by-one from 2.2.3 to 2.0.9 though 😹

PR #4510, opened 2020-07-02
Include what broke when docs building breaks (and fail the job)

glenjamin:
# Description
This should help avoid us shipping any broken assets in future.

# Reasons
https://app.circleci.com/pipelines/github/circleci/circleci-docs/24561/workflows/c42fbe58-8b04-483e-b104-03c3ad3d30a0/jobs/61314/parallel-runs/0/steps/0-108

### DIFF
--- a/scripts/build_api_docs.sh
+++ b/scripts/build_api_docs.sh
@@ -1,5 +1,6 @@
-
 #!/bin/bash
+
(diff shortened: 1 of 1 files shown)


rosieyohannan:
Thanks @glenjamin!

PR #4521, opened 2020-07-06
Update another reference to the cache expiry duration

glenjamin:
# Description

As noted elsewhere in the docs, caches now expire after 15 days.

# Reasons

This was causing confusion for a customer in a support ticket


rosieyohannan:
Ooo good spot, thanks @glenjamin!

PR #4534, opened 2020-07-10
Synchronize launch configuration for Nomad clients with enterprise-setup

makotom:
# Description & Reasons
Due to CIRCLE-27149, we updated [a reference launch configuration for Nomad clients](https://github.com/circleci/enterprise-setup/commit/f0fe945e23b4ec601960f034bb95c47cda321589) in [`enterprise-setup`](https://github.com/circleci/enterprise-setup). This PR reflects that update to documentations as well so that customers using CircleCI Server in private subnets on AWS can mitigate the issue that the `Rebuild With SSH` feature is broken.

# Considerations
I've confirmed that both updated launch configuration passed [`realitycheck`](https://github.com/circleci/realitycheck) and another tests for SSH rebuilds.

### DIFF
--- a/jekyll/_cci2/nomad-metrics.adoc
+++ b/jekyll/_cci2/nomad-metrics.adoc
@@ -68,6 +68,8 @@ export http_proxy=""
 export https_proxy=""
 export no_proxy=""
 export aws_instance_metadata_url="http://169.254.169.254"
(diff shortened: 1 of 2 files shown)


rosieyohannan:
Thanks so much for doing this @makotom!

PR #4537, opened 2020-07-10
Update lang from "run" to assume Workflow usage

KyleTryon:
# Description
Workflows are now active on all projects on CircleCI. There is no longer a "Run", which may now be considered a pipeline. One piece of information omitted in this commit that might be worth adding back, if the user does not specify a workflow, and has a job named "build" we will automatically generate a workflow for the build job.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -125,11 +125,7 @@ See the [Using Parameters in Executors](https://circleci.com/docs/2.0/reusing-co
 
 ## **`jobs`**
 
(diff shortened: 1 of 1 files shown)


rosieyohannan:
Thanks @KyleTryon! Do you happen to know if the run concept is obsolete for Server too, or is this a 2.1 thing? Currently we need to be mindful that devs using Server are also using this config reference. Also I have a feeling for 2.1 a workflow is added if there is't one even if there's no job called build.... I might have got the wrong end of the stick on that though, I can check with Gordon or someone.

KyleTryon:
> Thanks @KyleTryon! Do you happen to know if the run concept is obsolete for Server too, or is this a 2.1 thing? Currently we need to be mindful that devs using Server are also using this config reference. Also I have a feeling for 2.1 a workflow is added if there is't one even if there's no job called build.... I might have got the wrong end of the stick on that though, I can check with Gordon or someone.

My server knowledge is not as confident. Would need to verify that piece.

PR #4555, opened 2020-07-15
Update references

KyleTryon:
- Update references from old images and orbs to new versions.
- Modernize examples
- Add matrix testing

### DIFF
--- a/jekyll/_cci2/reusing-config.md
+++ b/jekyll/_cci2/reusing-config.md
@@ -251,7 +251,7 @@ It is also possible to allow an orb to define the executor used by all of its co
 
 The following example declares a Docker executor with a node image. The tag portion of the image string is parameterized with a `version` parameter. A `version` parameter is also included in our `test` job so that it can be passed through the job into the executor when the job is called from a workflow.
 
-When calling the `test` job from the workflow, we can take advantage of [matrix jobs](https://circleci.com/docs/2.0/configuration-reference/#matrix-requires-version-21) to run your job multiple times in parallel with a different set of parameters. In this instance, we may test our node application against many version of Node.js
+When calling the `test` job in the `matrix-tests` workflow, [matrix jobs](https://circleci.com/docs/2.0/configuration-reference/#matrix-requires-version-21) are used to run the job multiple times, concurrently, each with a different set of parameters. The node application is tested against many versions of Node.js
 
 
 ```yaml
--- a/jekyll/_cci2/reusing-config.md
+++ b/jekyll/_cci2/reusing-config.md
@@ -374,7 +374,7 @@ jobs:
 
 Parameters are declared by name under a job, command, or executor. The immediate children of the `parameters` key are a set of keys in a map.
 
-The following example a command named `greeting` is designed with a single parameter named `to`. the `to` parameter is used within the steps to echo _"hello"_ back to the user.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/reusing-config.md
```diff
@@ -246,34 +246,48 @@ jobs:
 
 It is also possible to allow an orb to define the executor used by all of its commands. This allows users to execute the commands of that orb in the execution environment defined by the orb's author.
 
-### Example of Using an Executor Declared in config.yml in Multiple Jobs
+### Example of Using an Executor Declared in `config.yml` with Matrix Jobs.
 {:.no_toc}
 
-The following example declares and invokes an executor in two jobs that need to run in the same Docker image and working directory with a common set of environment variables. Each job has distinct steps, but runs in the same environment.
+The following example declares a docker based executor which contains a node image with the tag portion of the image string parameterized with a `version` parameter. Further, in our `test` job, we similarly included a version parameter so that it may be "passed through" the job into the executor when we call it from a workflow.
```
rosieyohannan, line 252:
```suggestion
The following example declares a Docker executor with a node image. The tag portion of the image string is parameterized with a `version` parameter. A `version` parameter is also included in our `test` job so that it can be passed through the job into the executor when the job is called from a workflow.
```

jekyll/_cci2/reusing-config.md
```diff
@@ -246,34 +246,48 @@ jobs:
...
 
-### Example of Using an Executor Declared in config.yml in Multiple Jobs
+### Example of Using an Executor Declared in `config.yml` with Matrix Jobs.
 {:.no_toc}
 
-The following example declares and invokes an executor in two jobs that need to run in the same Docker image and working directory with a common set of environment variables. Each job has distinct steps, but runs in the same environment.
+The following example declares a Docker executor with a node image. The tag portion of the image string is parameterized with a `version` parameter. A `version` parameter is also included in our `test` job so that it can be passed through the job into the executor when the job is called from a workflow.
+
+When calling the `test` job from the workflow, we can take advantage of [matrix jobs](https://circleci.com/docs/2.0/configuration-reference/#matrix-requires-version-21) to run your job multiple times in parallel with a different set of parameters. In this instance, we may test our node application against many version of Node.js
```
rosieyohannan, line 254:
```suggestion
When calling the `test` job in the workflow, [matrix jobs](https://circleci.com/docs/2.0/configuration-reference/#matrix-requires-version-21) are used to run the job multiple times, concurrently, with a different set of parameters for each. The node application is tested against many versions of Node.js
```

jekyll/_cci2/reusing-config.md
```diff
@@ -322,82 +336,68 @@ jobs:
...
+    steps:
+      - run: echo "Node will not be installed."
 ```
 
 ## Using the `parameters` Declaration
 
 Parameters are declared by name under a job, command, or executor. The immediate children of the `parameters` key are a set of keys in a map.
 
-The following example defines a command called `sync`:
+The following example a command named `greeting` is designed with a single parameter named `to`. the `to` parameter is used within the steps to echo _"hello"_ back to the user.
```
rosieyohannan, line 377:
```suggestion
In the following example, a command named `greeting` is designed with a single parameter named `to`. The `to` parameter is used within the steps to echo _Hello_ back to the user.
```

PR #4558, opened 2020-07-16
contexts.md: add an example of non-bash interpolation

danbst:
# Description

Add an example of context variable interpolation in non-bash places.

# Reasons

This works for us but I haven't found it in documentation. This may also require changes to https://circleci.com/docs/2.0/env-vars/#using-parameters-and-bash-environment section, which says interpolation is not supported.


lewang:
Hi @danbst,  thanks for contributing!  I looked into it and all environment variables are interpolated in the docker image stanza, not only ones from contexts.  Does this clarification PR make it clearer? https://github.com/circleci/circleci-docs/pull/4646/files

Issue #4468, opened 2020-06-17
https://circleci.com/docs/api/ could/should redirect to API v2 rather than v1.1


rosieyohannan:

teesloane:
Closed in #4578

PR #4578, opened 2020-07-22
Add: migrate root api to v1, move v2 to root.

teesloane:
addresses #4468


felicianotech:
Locally there isn't an index page for `/docs/api/`. Is this something that gets created via CircleCI?

teesloane:
@felicianotech - yep, it gets compiled in our build process now, set to display the latest version of the API.

PR #4579, opened 2020-07-22
[CIRCLE-27357] Deprecate putting token in query parameters

cayennes:
# Description
Document that we are deprecating putting the token in query parameters and use basic auth in examples

# Reasons
https://circleci.atlassian.net/browse/CIRCLE-27357

### DIFF
--- a/jekyll/_plugins/api_endpoint.rb
+++ b/jekyll/_plugins/api_endpoint.rb
@@ -45,12 +45,12 @@ module Jekyll
     def curl_params(endpoint)
       params = endpoint['params']
       if params and (endpoint['method'] == 'GET')
-        '&' + params.map {|param| "#{param['name']}=#{param['example']}"}.join('&')
+        '?' + params.map {|param| "#{param['name']}=#{param['example']}"}.join('&')
       end
     end
 
     def api_curl(endpoint)
-      "curl #{curl_args_padded(endpoint)}https://circleci.com#{endpoint['url']}?circle-token=:token#{curl_params(endpoint)}"
+      "curl -u <circle-token>: #{curl_args_padded(endpoint)}https://circleci.com#{endpoint['url']}#{curl_params(endpoint)}"
     end
 
     def params(params)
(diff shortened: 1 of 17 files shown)


jekyll/_plugins/api_endpoint.rb
```diff
@@ -45,12 +45,12 @@ def curl_args_padded(endpoint)
     def curl_params(endpoint)
       params = endpoint['params']
       if params and (endpoint['method'] == 'GET')
-        '&' + params.map {|param| "#{param['name']}=#{param['example']}"}.join('&')
+        '?' + params.map {|param| "#{param['name']}=#{param['example']}"}.join('&')
```
teesloane, line 48: I think this Jekyll plugin is unused, but we might as well make the change :)

jekyll/_plugins/api_endpoint.rb
```diff
@@ -45,12 +45,12 @@ def curl_args_padded(endpoint)
...
       end
     end
 
     def api_curl(endpoint)
-      "curl #{curl_args_padded(endpoint)}https://circleci.com#{endpoint['url']}?circle-token=:token#{curl_params(endpoint)}"
+      "curl -u <circle-token>: #{curl_args_padded(endpoint)}https://circleci.com#{endpoint['url']}#{curl_params(endpoint)}"
```
aengelberg, line 53: Do we prefer customers use this over `-H Circle-Token:...`? Both seem fine, just curious which is considered the ideal way to authenticate, if we have already chosen one.

teesloane: Our API v2 docs, seem to recommend using headers. 

> Authentication
> API Key (api_key_query)
>
> Parameter Name: circle-token, in: query.
> API Key (api_key_header)
>
> Parameter Name: Circle-Token, in: header.
>HTTP Authentication, scheme: basic HTTP basic authentication. The username should be set as the circle-token value, and the password should be left blank. Note that project tokens are currently not supported on API v2.

teesloane:
We've got some dead links in the HTML proofer, but otherwise, I'm happy to merge this (I'm indifferent to whether we use headers or the `-u` flag - but I'm not sure I'm the right person to ask).

aengelberg:
This is an excellent change, thanks for making it happen

PR #4583, opened 2020-07-24
Testing page first draft + adding registry image

KyleTryon:
Initial testing page changes. More will be coming on "unity testing" but likely after the end of this quarter. I think we could write more about Orb tools as well but we aren't quite done with the v10 major change, therefore linking to live updating assets.


jekyll/_cci2/testing-orbs.md
```diff
@@ -1,187 +1,132 @@
...
+          - image: cimg/base:stable
+        steps:
+          - my-orb/my_command:  # Invoking the orb command from the development orb
+              my_parameter: "value"
+          - run:
+              name: Validate my_command
+              command: |
+                if [ "$ITWORKED" != "yes" ]; then
+                echo "error" && exit 1
+                fi
```
rosieyohannan, line 86: If I understand right this should all be the job declaration, not within the orb itself and so aligned a few tabs to the left (including the `jobs` key above)?

jekyll/_cci2/testing-orbs.md
```diff
@@ -1,187 +1,132 @@
...
-      - run: echo "Hello, build!"
-```
-{% endraw %}
+## Integration Testing
 
-2) Validate the orb with `circleci orb validate src/orb.yml`.
+Using [development orbs]({{site.baseurl}}/2.0/orb-concepts/#development-orbs), it is possible to publish changes to a live development version of an orb and use that development version within a live configuration for testing. Development versions of orbs can later be promoted to production orbs.
 
-3) Publish a dev version with `circleci orb publish src/orb.yml namespace/orb@dev:0.0.1`.
+Integrations tests validate the usage of your orb and its parameters in a live environment on CircleCI using sample project code (if required). Integration test jobs should be declared that both use the commands defined within your orb, and validate the correct functioning of those commands. See the following examples for testing a command, and testing a job.
```
rosieyohannan, line 65: "Integration test jobs should be declared **within the live project environment** that both use the commands..." maybe add to clarify where these tests should be written?

jekyll/_cci2/testing-orbs.md
KyleTryon: Lets zoom on this, im not following.

jekyll/_cci2/testing-orbs.md
KyleTryon: ^ Ideally in the end they will be using the orb init command but I can't add much information about that yet.

gmemstr:
Might be worth using something like https://browserframe.com or https://screenshot.rocks to generate the registry screenshot for a cleaner look

KyleTryon:
> Might be worth using something like https://browserframe.com to generate the registry screenshot for a cleaner look

I was looking for something _exactly_ like this. Perfect, thank you.

jekyll/_cci2/testing-orbs.md
rosieyohannan: 👍

rosieyohannan:
no longer required

PR #4588, opened 2020-07-27
Add a requirement to trigger builds with a restricted context

ganezasan:
# Description
Add the requirement to trigger builds with a restricted context

# Reasons
Users need to sign up for CircleCI to trigger builds with a restricted context. Even if they don't sign up for CircleCI, they can trigger builds, but to trigger a build with a restricted context, CircleCI needs to check the accounts permission, so users must sign up for CircleCI.


rosieyohannan:
Thanks @ganezasan!

PR #4602, opened 2020-07-30
add Twitch to social buttons

jerdog:
Using FontAwesome, let's add our Twitch account

# Description
We need to drive more awareness of our Twitch account, and Docs is a great place to start

# Reasons
Twitch. Where the cool kids hangout and laugh at the other kids doing live coding streams

### DIFF
--- a/jekyll/_includes/social-buttons.html
+++ b/jekyll/_includes/social-buttons.html
@@ -12,6 +12,9 @@
     <a href="https://github.com/circleci" target="_blank" class="social-btn-github" title="GitHub">
       <i class="fa fa-github" aria-hidden="true"></i><span class="sr-only">GitHub</span>
     </a>
+    <a href="https://twitch.tv/circleci" target="_blank" class="social-btn-linkedin" title="LinkedIn">
+      <i class="fa fa-twitch" aria-hidden="true"></i><span class="sr-only">Twitch</span>
+    </a>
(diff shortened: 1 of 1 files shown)


rosieyohannan:
Thanks for this @jerdog! Just adding @janedotbiz as reviewer too because I think we're trying to keep the docs footer inline with what we have in outer? I could be wrong though :-)

jerdog:
Thanks. I have requested this change be made with outer as well

jekyll/_includes/social-buttons.html
```diff
@@ -12,8 +12,11 @@
     <a href="https://github.com/circleci" target="_blank" class="social-btn-github" title="GitHub">
       <i class="fa fa-github" aria-hidden="true"></i><span class="sr-only">GitHub</span>
     </a>
+    <a href="https://twitch.tv/circleci" target="_blank" class="social-btn-linkedin" title="LinkedIn">
```
janedotbiz, line 15: Not sure why these links have these classes, as they don't seem to be being used, but in any case I would recommend changing this class to `social-btn-twitch` and `title="Twitch"` so that it matches the content.

jerdog: And I'm not sure why I spaced on that lol

jerdog: @janedotbiz I have fixed that derp.

PR #4612, opened 2020-08-05
Xcode 12b4 & Beta Image Info

BytesGuy:


jekyll/_cci2/testing-ios.md
```diff
@@ -27,23 +27,30 @@ Periodically, we will update the version of macOS each image includes to ensure
 
 We announce the availability of new macOS containers, including Xcode betas, in the [annoucements section of our Discuss site](https://discuss.circleci.com/c/announcements).
 
+### Beta Image Support
+
+We endeavour to make beta Xcode versions available on the macOS executor as soon as we can to allow developers to test their apps ahead of the next stable Xcode release. 
+
+Unlike our stable images (which are frozen and will not change), as beta images are released they are replaced in situ until a GM (stable) image is released, at which point the image is frozen and no longer updated. If you are requesting an image using an Xcode version that is currently in beta, please expect it to change when Apple releases a new Xcode beta with minimal notice. This can include breaking changes in Xcode/associated tooling which are beyond our control.
```
teesloane, line 34: Great section. Only suggestion would be to replace `in situ` with something more common and accessible: `in place` for example, or `as beta images are releases, each release will overwrite the previous release` etc.

PR #4614, opened 2020-08-06
fix typo in instruction

usami:
# Description
Fixing typo in the [Ending a job from within a step](https://circleci.com/docs/2.0/configuration-reference/#ending-a-job-from-within-a-step) section.

# Reasons
I think `by using using` is a typo.


rosieyohannan:
Thank you @usami! ☺️

PR #4617, opened 2020-08-06
update cookbook example to be clearer about pipeline parameter usage

### LINKED ISSUE #4495: Unclear how to use logic in configuration - missing an example or links to further docs
# Location
https://circleci.com/docs/2.0/configuration-cookbook/#using-logic-in-configuration

# Explanation
> The action parameter will default to build on pushes, but you can supply a different value to select a different workflow to run, like report.

The first part (**The action parameter will default to build on pushes**) is clear. 

But what do you mean by "**but you can supply a different value to select a different workflow to run, like report**"? When does this apply? How/when can I supply a different value? Can you give an example? May be I am missing some explanation about pipeline parameters usage?

---

I suggest the marked text in screenshot below to become links to further docs or to provide an example in the same section.
image

rosieyohannan:
Fixes #4495

### DIFF
--- a/jekyll/_cci2/configuration-cookbook.md
+++ b/jekyll/_cci2/configuration-cookbook.md
@@ -504,7 +504,7 @@ workflows:
       - report
 ```
 
-The `action` parameter will default to `build` on pushes. Below is an example of supplying a different value using API v2 Trigger a New Pipeline endpoint and Pipeline Parameters to select a different workflow to run, in this case, `report`. Remember to substitute [`project-slug`({{ site.baseurl }}/2.0/api-developers-guide/#getting-started-with-the-api) for your values.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-cookbook.md
```diff
@@ -504,7 +504,15 @@ workflows:
       - report
 ```
 
-The `action` parameter will default to `build` on pushes, but you can supply a different value to select a different workflow to run, like `report`.
+The `action` parameter will default to `build` on pushes. Below is an example of supplying a different value using API v2 Trigger a New Pipeline endpoint and Pipeline Parameters to select a different workflow to run, in this case, `report`. Remember to substitute [`project-slug`({{ site.baseurl }}/2.0/api-developers-guide/#getting-started-with-the-api) for your values.
```
rosieyohannan, line 507:
```suggestion
The `action` parameter will default to `build` on pushes. Below is an example of supplying a different value using the API v2 Trigger a New Pipeline endpoint and Pipeline Parameters to select a different workflow to run, in this case, `report`. Remember to substitute [`project-slug`({{ site.baseurl }}/2.0/api-developers-guide/#getting-started-with-the-api) for your values.
```

PR #4631, opened 2020-08-12
home page and nav updates

michelle-luna:
# Description
Updates to the home page and navigation based on CPE feedback (to make orbs and images more obvious).

# Reasons
CPE feedback and slightly lower ratings of home page this year compared to last year.

### DIFF
--- a/jekyll/_cci2/index.md
+++ b/jekyll/_cci2/index.md
@@ -5,67 +5,67 @@ description: "Welcome to CircleCI Documentation"
 permalink: /2.0/
 ---
 
-Use the tutorials, samples, how-to, and reference documentation to learn CircleCI.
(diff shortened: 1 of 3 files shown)


teesloane:
This looks great and scales nicely on mobile. I noticed there is a bit of text/links in the upper right corner that might need to be moved? 

Screen Shot 2020-08-17 at 9 58 30 AM

rosieyohannan:
Sorry I've not left review comments yet! I have had a look and it looks great to me. Just wanted to find some time to look again more thoroughly in case there's anything I missed. Will do that today hopefully :-)

rosieyohannan:
> This looks great and scales nicely on mobile. I noticed there is a bit of text/links in the upper right corner that might need to be moved?
> 
> Screen Shot 2020-08-17 at 9 58 30 AM

@teesloane I can't get it to look like that - is that what happens when you shrink your browser?

teesloane:
This view happens on `circleci.com/docs`  but not on `/docs/2.0`

teesloane:
I added a fix for the links not displaying correctly. The sidebar changes are looking good to me.

PR #4633, opened 2020-08-13
Update artifacts.md

sgesin:
# Description
Replaced $build_number with :build_num in curl command to match the placeholder later on in the document.

# Reasons
All placeholders are documented as having a `:` characters as a prefix and the `$build_number` placeholder did not match that pattern.


rosieyohannan:
Thanks for spotting this @sgesin!

rosieyohannan (approved):

PR #4635, opened 2020-08-13
add badges to show product feature availability

rosieyohannan:


rosieyohannan:
Not sure where to put the badges really... And I think I should shrink the Cloud one to fit the text. and align them centrally (vertically) @teesloane  I like your idea of putting them at the top of the TOC on the right maybe 🤔 

Screenshot 2020-08-18 at 18 41 50

teesloane:
This is looking great @rosieyohannan ! I think your insights for what should come next makes sense - we'll apply some styling and we can always move them into the table of contents on shipping it.

rosieyohannan:
Any thoughts @teesloane? I'm going to start the long process of going through the docs to add the versions in I think...

Screenshot 2020-08-19 at 21 41 32

rosieyohannan:
I actually wonder if it would be better to put the buttons to the right of the H1 heading and have the explanation "this document is applicable.." as hover-over text?

Screenshot 2020-08-19 at 21 41 32

teesloane:
> I actually wonder if it would be better to put the buttons to the right of the H1 heading and have the explanation "this document is applicable.." as hover-over text?
> 
> Screenshot 2020-08-19 at 21 41 32

I think that might be something the Design team could better answer. But personally - I like it.

teesloane:
Update: smaller badges, aligned with title, with popover content.

Aug-24-2020 11-26-26

jekyll/_cci2/contexts.md
```diff
@@ -180,8 +183,7 @@ Environment variables are used according to a specific precedence order, as foll
 Environment variables declared inside a shell command `run step`, for example `FOO=bar make install`, will override environment variables declared with the `environment` and `contexts` keys. Environment variables added on the Contexts page will take precedence over variables added on the Project Settings page.
 
 ## Secrets Masking
-
-_Secrets masking is not currently available on self-hosted installations of CircleCI Server_
+{% include feature-cloud.html %}
```
rosieyohannan, line 186: @teesloane Sometimes we need to point out a section on a page that is, for example, Cloud only, even though the full doc is for all versions. I stuck this include in as a 'better' way to do that but I imagine you can devise an actually better way using the in-line-with-title badges and tooltips? If not right now, this might be out of scope for this PR but something to add in after?

teesloane: Hmm. We could make a heading include that generates an h2 with a badge to the right of it, sort of like what the heading does - but I agree — might be best for another PR?

rosieyohannan: Yeah good call, I'll remove this change for now and deal with subsections later :-)

jekyll/_cci2/jobs-steps.md
```diff
@@ -13,6 +13,7 @@ The document provides an overview of orbs, jobs, steps and workflows.
 {:toc}
 
 ## Orbs Overview
+_orbs are not currently available on self-hosted installations of CircleCI Server_
```
teesloane, line 16:
```suggestion
_Orbs are not currently available on self-hosted installations of CircleCI Server._
```

teesloane (changes requested):

teesloane (approved):

PR #4636, opened 2020-08-14
add S3 lifecycle page to docs site and PDF

rosieyohannan:
Original request: https://gist.github.com/makotom/7a110394d4afba9988e548ec96bb854a

### DIFF
--- a/jekyll/_cci2/storage-lifecycle.adoc
+++ b/jekyll/_cci2/storage-lifecycle.adoc
@@ -24,7 +24,7 @@ NOTE: Scroll table to the right to view all content.
 |===
 | **Prefix**
 | **Tag**
-| **Days**
+| **Suggested Number of Days**
 | **Description of files**
 | **Side effects of deletion**
 
--- a/jekyll/_cci2/storage-lifecycle.adoc
+++ b/jekyll/_cci2/storage-lifecycle.adoc
@@ -15,6 +15,8 @@ This guide is intended to help system administrators of CircleCI Server installa
 
 == Lifecycle Policy Configuration for S3 Buckets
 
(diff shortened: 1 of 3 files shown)


jekyll/_cci2/storage-lifecycle.adoc
```diff
@@ -0,0 +1,82 @@
...
+:icons: font
+:toc: macro
+:toc-title:
+:sectanchors:
+
+
+[.serveronly]_This document is intended for system administrators of self-hosted installations of CircleCI Server._
+
+This guide is intended to help system administrators of CircleCI Server installations on AWS better understand how S3 storage is used. This can help to cut compute resource costs and avoid issues for service users when removing resources.
+
```
makotom, line 15: Could you add a emphasized note around here that, at least for 2.19.x, the lifecycle policy below is _not_ configured out-of-box, and is needed to be manually configured?
Actually there are multiple customers who believe that there is a built-in functional garbage-collecting mechanism, for whom we act like MythBusters to tell them that there is not.

jekyll/_cci2/storage-lifecycle.adoc
```diff
@@ -0,0 +1,82 @@
...
+== Lifecycle Policy Configuration for S3 Buckets
+
+NOTE: Scroll table to the right to view all content.
+
+[.table.table-striped]
+[cols=5*, options="header", stripes=even]
+|===
+| **Prefix**
+| **Tag**
+| **Days**
```
makotom, line 25: Related to the comment above, maybe better to emphasize here that the days are kind-of suggestions, and the numbers are totally up to administrators.

PR #4653, opened 2020-08-21
Remove extra bracket in ruby versions link

nbialostosky:
# Description
This PR removes an extra bracket `[` that we had in a link within the docs. It was not allowing the markdown to work and causing this to be displayed in the docs:

Screen Shot 2020-08-21 at 2 20 39 PM

# Reasons
The extra bracket was making the content display as text and not a link, the update will ensure its rendered as a link and references the proper documentation section.

### DIFF
--- a/jekyll/_cci2/deploying-ios.md
+++ b/jekyll/_cci2/deploying-ios.md
@@ -216,7 +216,7 @@ workflows:
       - adhoc
 ```
 
-**Note:** The Firebase plugin may cause errors when run with the macOS system Ruby. It is therefore advisable to [switch to a different ruby version][({{ site.baseurl }}/2.0/testing-ios/#using-custom-ruby-versions)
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/deploying-ios.md
```diff
@@ -216,7 +216,7 @@ workflows:
       - adhoc
 ```
 
-**Note:** The Firebase plugin may cause errors when run with the macOS system Ruby. It is therefore advisable to [switch to a different ruby version][({{ site.baseurl }}/2.0/testing-ios/#using-custom-ruby-versions)
+**Note:** The Firebase plugin may cause errors when run with the macOS system Ruby. It is therefore advisable to [switch to a different ruby version]({{ site.baseurl }}/2.0/testing-ios/#using-custom-ruby-versions)
```
teesloane, line 219:
```suggestion
**Note:** The Firebase plugin may cause errors when run with the macOS system Ruby. It is therefore advisable to [switch to a different ruby version]({{ site.baseurl }}/2.0/testing-ios/#using-ruby)
```
This should fix the failing build due to our html-proofer.

nbialostosky:
@teesloane Thanks for the feedback and change that needed to be made, went ahead and took care of that!

PR #4657, opened 2020-08-25
tjs/swap-redoc - [CIRCLE-28069] Custom x-codeSamples for API v2 GET /context with query params

skimke:
# Description
Writing out manual `x-codeExamples` for `GET /context` docs.

WHY THIS WAY? (https://circleci.atlassian.net/browse/CIRCLE-28069?focusedCommentId=110082)
TL;DR  `x-examples` doesn’t apply well to our case

Tried but failed at adding a bunch of different permutations of `x-examples`, by also referencing other repos. In the end, settled on writing out `x-codeSamples` instead, unfortunately. 

The most frustrating example I couldn’t replicate was: https://docs.pdfgeneratorapi.com/openapi.json#operation/mergeTemplate, see also their [openapi.json](https://docs.pdfgeneratorapi.com/api-docs.json). Maybe their `x-example` values are working for them because in this case, there is a `requestBody` and it’s a POST endpoint.  Maybe they're also manually adding them to `x-codeSamples` 🤷 

Also looked at a few others like [this](https://github.com/Axway-API-Builder-Ext/api-builder-extras/blob/6da90314d5ed569ca50620a601322d9afc10b649/api-builder-plugin-fc-sap-lama/openapi/sap-lama-api.json#L1104) and [this](https://github.com/arunjoh/cmad-nard/blob/3791432ad43faae17939daff5784a13a2cdcb1c0/stackoverflow/apis/src/main/resources/web/swagger-ui/spec/swagger20.json#L491) - ALL of these examples seem to indicate that `x-examples` is only meant for extending a param `”in”: “body”`, which isn’t our case

# Reasons
https://circleci.atlassian.net/browse/CIRCLE-28069
We want to show approximations of what adding query params would be like in code samples

Screen Shot 2020-08-27 at 12 49 37 PM


teesloane:
Wondeful, thank you. I will setup the jq merge and get this integrating shortly.

src-api/openapi-patch.json
```diff
@@ -13,9 +13,7 @@
             },
             "x-examples":{
               "owner-id":{
-                "value":{
-                  "owner-id":"c65b68ef-e73b-4bf2-be9a-7a322a9df150"
-                },
+                "value":"c65b68ef-e73b-4bf2-be9a-7a322a9df150",
```
skimke, line 16: @teesloane this could fix it but I haven't confirmed. Is there a new artifact each build?

teesloane: Yep, if you search for "api/v2/index.html" you should be able to see the output.

skimke:
@teesloane this is good to go now after some local testing

src-api/openapi-patch.json
skimke: Thanks 🙏

teesloane:
Great @skimke . I just need to fix the broken html links from our other docs and then we should be good-ish to merge (might need to make some design changes to get the CircleCI Logo in there, etc

PR #4668, opened 2020-08-28
[CIRCLE-28455] Add docs for multiple contexts

sulami:
# Description
Add documentation for using multiple contexts.

Needs a redo of the image on the contexts docs as well.

Also need someone to write three sentences of Japanese for the contexts docs.

# Reasons
https://circleci.atlassian.net/browse/CIRCLE-28455


jekyll/_cci2/contexts.md
```diff
@@ -62,8 +62,9 @@ If you find you need to rename an org or repo that you have previously hooked up
       my-workflow:
         jobs:
           - run-tests:
-              context: org-global
```
rosieyohannan, line 65: presumably this way is still valid if using a single context? Also - can server customers use the syntax you've added here? Just wondering if we need separate examples...

sulami: The old way is still valid, though I think it might be less confusing if there's only one set of semantics presented.

As discussed on Slack, server customers will have to wait until a handful of services have been updated in server, I assume that's going to happen with 3.0 or 3.1.

rosieyohannan: 👍 Yep I get that Server customers can't use the list of contexts, but can they add a single context as you show here?

sulami: They cannot, at least not until at least orb-service has been updated in server :/

jekyll/_cci2/configuration-reference.md
```diff
@@ -1361,7 +1361,7 @@ Jobs may be configured to use global environment variables set for an organizati
 
 Key | Required | Type | Description
 ----|-----------|------|------------
-context | N | String | The name of the context. The initial default name was `org-global`. Each context name must be unique.
+context | N | String/List | The name of the context(s). The initial default name was `org-global`. Each context name must be unique.
```
rosieyohannan, line 1364:
```suggestion
context | N | String/List | The name of the context(s). The initial default name was `org-global`. Each context name must be unique. If using CircleCI Server, only a single Context per workflow is supported.
```

jekyll/_cci2/configuration-reference.md
```diff
@@ -1361,7 +1361,7 @@ Jobs may be configured to use global environment variables set for an organizati
 
 Key | Required | Type | Description
 ----|-----------|------|------------
-context | N | String | The name of the context. The initial default name was `org-global`. Each context name must be unique.
+context | N | String/List | The name of the context(s). The initial default name was `org-global`. Each context name must be unique. If using CircleCI Server, only a single Context per workflow is supported.
```
rosieyohannan, line 1364: `The initial default name was `org-global`` seems like a weird statement - can we change `was` to `is`?  I realise this isn't in the scope of this PR - just jumps out!

rosieyohannan (approved):

PR #4689, opened 2020-09-07
add support bundle guidance and remove incorrect link

rosieyohannan:


jekyll/_cci2/troubleshooting.adoc
```diff
@@ -7,12 +7,14 @@
...
 
-This document describes an initial set of troubleshooting steps to take if you are having problems with your CircleCI installation on your private server. If your issue is not addressed below, please https://help.replicated.com/docs/native/packaging-an-application/support-bundle/[generate a support bundle] and contact our Support Engineers by https://support.circleci.com/hc/en-us/requests/new[opening a support ticket].
+This document describes an initial set of troubleshooting steps to take if you are having problems with your CircleCI installation on your private server. If your issue is not addressed below, you can generate a support bundle and contact CircleCI Support Engineers by https://support.circleci.com/hc/en-us/requests/new[opening a support ticket].
 
 toc::[]
 
-[discrete]
-=== Debugging Queuing Builds
+=== Generating a Support Bundle
+To download a support bundle, select **Support** from the Managment Console menu bar, and then select **Download Support Bundle**. Often CircleCI support engineers will request support bundles to help diagnose/fix the problem you are experiencing.
```
rosieyohannan, line 15:
```suggestion
To download a support bundle, select **Support** from the Managment Console menu bar, and then select **Download Support Bundle**. CircleCI support engineers will often request support bundles to help diagnose/fix the problem you are experiencing.
```

jekyll/_cci2/troubleshooting.adoc
```diff
@@ -137,13 +138,11 @@ If the nomad client logs contain the following error message typw, check port 85
...
 ```
 
-[discrete]
 === Why is the cache failing to unpack?
 
 If a `restore_cache` step is failing for one of your jobs, it is worth checking the size of the cache - you can view the cache size from the CircleCI Jobs page within the `restore_cache` step. We recommend keeping cache sizes under 500MB – this is our upper limit for corruption checks because above this limit check times would be excessively long. Larger cache sizes are allowed but may cause problems due to a higher chance of decompression issues and corruption during download. To keep cache sizes down, consider splitting into multiple distinct caches.
 
-[discrete]
-=== How do I get round the API service being impacted by a high thread count
+=== How do I get round the API service being impacted by a high thread count?
```
lloydChris, line 145:
```suggestion
=== How do I get around the API service being impacted by a high thread count?
```

lloydChris (approved):
Minor suggestion, but you are the language expert so I will defer to you.

jekyll/_cci2/troubleshooting.adoc
rosieyohannan: Yes you are absolutely right! Thanks :-)

teesloane (approved):

PR #4690, opened 2020-09-08
Clean: remove settings page.

### LINKED ISSUE #4680: Old settings doc
This doc is out of date. I think we could just bin it though? https://circleci.com/docs/2.0/settings/

teesloane:
The settings page is a legacy document that the new UI deprecates.

closes #4680


rosieyohannan (approved):
Yeah I approve! I think we shouldn't be documenting UI pages in this way - they should be discoverable within the UI. This page does not add anything 🎉

Issue #4679, opened 2020-09-03
Check old workflows docs


rosieyohannan:
We have a few very short workflows troubleshooting docs which should probably just be part of the main workflows doc - or part of a troubleshooting guide. Also the content looks to be out of date:

* https://circleci.com/docs/2.0/workflows-waiting-status/
* https://circleci.com/docs/2.0/workflows-not-starting/

teesloane:
Confirmed: these two documents, as well as `workflows-overview.md` all receive less than 0.3 % of monthly traffic. I will migrate their content to workflows.md.

PR #4694, opened 2020-09-10
Clean: remove redundant workflows documents.

teesloane:
The documents removed have less that 0.3% of monthly  traffic. Also, the content already exists in the workflow document.


jekyll/_cci2/workflows.md
```diff
@@ -523,6 +525,13 @@ When you use workflows, you increase your ability to rapidly respond to failures
 
 This section describes common problems and solutions for Workflows.
 
+### Workflow and Subsequent Jobs Do Not Trigger
+
+Oftentimes if you do not see your workflows triggering, a configuration error is
```
rosieyohannan, line 530:
```suggestion
If you do not see your workflows triggering, a common cause is a configuration error
```
```suggestion
Oftentimes if you do not see your workflows triggering, a configuration error is
```

rosieyohannan: Just a suggestion as "oftentimes" isn't (I don't think) in common usage in the UK so thought it might not localize well. Take it or leave it :-D

rosieyohannan (approved):
🗑️ 🎉

teesloane:
Closes #4679

PR #4696, opened 2020-09-10
Remove condition key and add note about workflows

nbialostosky:
# Description
This PR will remove the `condition:` key from the workflow example as that isn't supported/doesn't work and it also adds a note specifically to not include it at the workflow level.

# Reasons
A customer raised this in a ticket and I tested and confirmed the current functionality. These changes will hopefully prevent future confusion.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -1674,7 +1674,7 @@ Truthiness rules are as follows: `false`, `null`, `0`, the empty string, and `Na
 Logic statements always evaluate to a boolean value at the top level, and coerce as necessary. They can be nested in an arbitrary fashion, according to their argument specifications, and to a maximum depth of 100 levels.
 
 **Note:**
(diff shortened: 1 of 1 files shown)


sulami (approved):
Good from my side.

jekyll/_cci2/configuration-reference.md
```diff
@@ -1673,16 +1673,18 @@ Truthiness rules are as follows: `false`, `null`, `0`, the empty string, and `Na
 
 Logic statements always evaluate to a boolean value at the top level, and coerce as necessary. They can be nested in an arbitrary fashion, according to their argument specifications, and to a maximum depth of 100 levels.
 
+**Note:**
+When using logic statements at the workflow level, do not include the `condition:` key. That is only needed for job level logic statements.
```
teesloane, line 1677:
```suggestion
**Note:**
When using logic statements at the workflow level, do not include the `condition:` key (the `condition` key is only needed for `job` level logic statements).
```

teesloane (changes requested):

teesloane (approved):

PR #4714, opened 2020-09-15
Add a note that API v2 isn't supported project token

ganezasan:
# Description
Add a note that API v2 isn't supported project token

# Reasons
Some customers asked this issue via support tickets.


rosieyohannan:
Thank you @ganezasan!

rosieyohannan (approved):

PR #4729, opened 2020-09-17
Typo fix: peak -> peek

tunzor:
# Description
Incorrect use of _peak_ instead of **peek**.
> See the following snippets for a ~peak~ peek into how you can customize..."

# Reasons
Noticed the typo when following the docs, doesn't take away from content, just fixing a homophone issue.


rosieyohannan:
Thank you for spotting this @tunzor!

rosieyohannan (approved):

PR #4732, opened 2020-09-18
Revert "Clean: remove settings page."

rosieyohannan:
Reverts circleci/circleci-docs#4690

@teesloane I propose reverting this change and instead I'll update the page to be an actual intro section to this docs section?

Screenshot 2020-09-18 at 09 43 57

What do you think?


teesloane (approved):

teesloane:
Makes sense! Approved.

PR #4741, opened 2020-09-21
overhaul and update settings intro page

rosieyohannan:
New UI images
Include intro to what else is in the section
~Include Plans info~ on second thoughts I think this needs a whole little section of its own at some point - out of scope for this PR


jekyll/_data/sidenav.yml
```diff
@@ -216,6 +214,8 @@ en:
       children:
         - name: Status Overview
           link: 2.0/status/
+        - name: Pipelines Overview
```
rosieyohannan, line 217: I moved this page from the Projects section to the Pipelines section - seems a better fit?

teesloane: Definitely.

teesloane (approved):

PR #4744, opened 2020-09-23
Add rate limits as motivation to use authenticated docker pulls

lewang:
# Description
Originally we were going to add a new document just for rate limits.  The contents of that document ended up being similar to the existing `private-images`, so I decided to merge the two.

# Reasons
We are specifically mentioning this because Docker Hub intend to enforce rate limits starting November 1, 2020.

https://circleci.atlassian.net/browse/CIRCLE-29510

### DIFF
--- a/jekyll/_cci2/private-images.md
+++ b/jekyll/_cci2/private-images.md
@@ -1,8 +1,8 @@
 ---
 layout: classic-docs
-title: "Using Private Images"
-short-title: "Using Private Images"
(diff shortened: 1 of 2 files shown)


teesloane (approved):

glenjamin:
I think it would be good to call out multiple contexts more explicitly on this page somehow

melyalex (commented):
[]()

PR #4754, opened 2020-09-24
Upgrade InstantSearch.js

tfe:
# Description

This upgrades InstantSearch.js and the Algolia client to current versions, updating the HTML/CSS/JS of our integration to adapt to changes in those libraries while maintaining existing styles and functionality.

# Reasons

[CIRCLE-27855](https://circleci.atlassian.net/browse/CIRCLE-27855) requests an update to Docs search to include results from the Orbs registry, which requires querying another index. Multiple index search is supported only by more recent versions of InstantSearch.js.

### DIFF
--- a/jekyll/_includes/head.html
+++ b/jekyll/_includes/head.html
@@ -31,8 +31,8 @@
 	{% include segment.html %}
 
 	<!-- Instant Search CDN links /-->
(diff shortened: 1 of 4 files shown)


janedotbiz (approved):
The styling isn't exactly the same as it was, but it's changing in #4793 anyway.

PR #4756, opened 2020-09-24
CircleCI Xcode Image Release, Update and Deprecation Policy

BytesGuy:
# Description
Added a new document describing our Xcode Image Release, Update and Deprecation Policy

# Reasons
By ratifying this policy in the docs, we hope to be more transparent with our customers and set reasonable exceptions for Xcode image support. By having this policy in place, it also provides us guidance for capacity planning to allow us to continue to support new image versions


rosieyohannan (approved):
🎉

melyalex (commented):
[]()

PR #4766, opened 2020-09-28
remove orbs doc that are no longer required for the 'using orbs' flow

rosieyohannan:
Both these docs have had 0.04% of user traffic in recent months, have been remove from the side bar navigation and are no longer part of the recommended "learning to use orbs" flow in docs.


KyleTryon (approved):
heck ya!

PR #4769, opened 2020-09-29
Remove vendored lunr 0.7.0 as it is no more used

tnir:
# Description
Remove vendored lunr 0.7.0.

# Reasons

As it is no more used. That library was originally introduced in 93893a57b836ccfe173cf130e87a09bc1dab19b4 in 2016-03, which probably for v1 docs I think, but there has been no update in the form of this.

On API v2 docs, ReDoc relies on lunr (2.3.8 as of 2020-09-29) via NPM.


tnir:
With #4771 and this, lunr 0.x can be completely removed from this project.

teesloane (approved):

PR #4775, opened 2020-09-30
Remove Istio from example configs

tnir:
# Description
Removes Istio from [example configs](https://circleci.com/docs/2.0/example-configs/) since Istio project removed circleci config from the repo on 2019-08-10 to migrate their CI to Prow.

See https://github.com/istio/istio/pull/16183

# Reason
Leaving the link as-is makes us see 404s...


teesloane (approved):
Thanks for the PR! 🚢

Issue #4768, opened 2020-09-28
Configs examples in ordered lists are indented one whitespace character


dsayling:
# Description
Configs examples in ordered lists are indented one whitespace character. (Maybe unordered too?)

# Location
See the config under ordered list 1, https://circleci.com/docs/2.0/getting-started/#using-the-workflows-functionality
If you copy and paste this config, it doesn't work.

# Steps to Reproduce
When you highlight the text, the whitespace character is visible.

rosieyohannan:
Fixed!

PR #4781, opened 2020-10-01
Fix: indentation

teesloane:
A src block has an additional column of white space, making it invalid yaml.


rosieyohannan (approved):

rosieyohannan:
fixes #4768

PR #4793, opened 2020-10-02
Add Orbs registry results to search

tfe:
Per [CIRCLE-27855](https://circleci.atlassian.net/browse/CIRCLE-27855), redesigns the search results page and adds a results category for Orbs from the Orbs registry.

This also depends on and includes changes from #4754. This will be simpler to review if we merge that first and then update this branch from master; otherwise this PR will include changes from both.


jekyll/_sass/_subnav.scss
```diff
@@ -0,0 +1,31 @@
...
+#subnav-placeholder,
+.subnav.component.js-fixed {
+  width: inherit;
+}
+
+.deep-link-anchor {
+  &.subnav-offset {
+    top: -$nav-offset-subnav;
+  }
+}
```
janedotbiz, line 31: The first list item is shown as flush left in the design.

```css
.component.subnav li:first-child {
  margin-left: 0;
}
```

Before:
Screen Shot 2020-10-09 at 11 20 06 AM

After:
Screen Shot 2020-10-09 at 11 21 39 AM

Additionally, the design has everything (search input, first subnav item, and results) all with the same left alignment, but more indented (relative to search icon placement) than what's here.

Design:
Screen Shot 2020-10-09 at 9 42 11 AM

jekyll/_sass/_instantsearch.scss
```diff
@@ -15,25 +15,46 @@
...
+
+    .ais-SearchBox-submit,
+    .ais-SearchBox-reset {
+      position: absolute;
+      top: calc(50% - 20px/2);
 
-      @media screen and (min-width: $screen-sm) {
-        left: 6px;
+      svg {
+        width: 20px;
```
janedotbiz, line 34: The reset icon is 14px x 14px in the design, and also a light grey colour.
Screen Shot 2020-10-09 at 9 39 21 AM
Design:
Screen Shot 2020-10-09 at 11 13 35 AM

jekyll/_sass/_instantsearch.scss
```diff
@@ -46,41 +67,68 @@
...
+    background-color: white;
+    border: 0.5px solid #D8D8D8;
   }
 
-  em {
-    font-style: normal;
-    font-weight: bold;
+  mark {
+    color: #1A649A;
+    background-color: #EEF2F5;
```
janedotbiz, line 110: The design has this as a dark blue (`#19639A`) with a 5% opacity.
Using transparency allows the contrast of the `mark` background colour compared to the result background colour to be more similar between the hover and non-hover state of the result (white vs grey background)

janedotbiz (changes requested):
Is it possible to have the subnav background colour go full width so that the highlighted search result doesn't peak out from under the side of it? I see the underline is not full-width in the design so that might be a bit tricky.
Screen Shot 2020-10-09 at 10 39 20 AM

The very first result is covering up the subnav underline on hover:
Screen Shot 2020-10-09 at 10 34 35 AM

I noticed the highlighted search result doesn't quite line up with the width and alignment of the subnav (underline)
Screen Shot 2020-10-09 at 10 34 15 AM

jekyll/_sass/_subnav.scss
janedotbiz: Oh also, the non-active items in the subnav in the design are bolded, but gray; 
rather than a regular weight and black(ish)

tfe:
OK, I think I've resolved everything except for the width of the subnav bottom border when in fixed position, which I ran out of time on. Let me know if you see a way to fix that; all the things I tried didn't work.

janedotbiz:
Screen Shot 2020-10-14 at 11 56 27 AM

Are you seeing this too? When I first type in a search query, Orbs is highlighted in the menu, even though the Docs results are what are displayed, and the scroll position is slightly off. Once I scroll or click in the subnav things appear to be normal.

tfe:
@janedotbiz I chased down that behavior to the subnav JS not accounting for it being hidden on initial page load, as it is for docs search (all of its position calculations are incorrect because hidden element positions are all 0). Here's a PR to make it robust to that (https://github.com/circleci/circleci.com-shared/pull/2), and I'll follow it up with a PR here that triggers the new event introduced there upon showing results.

janedotbiz (approved):
Thank you!

It'd be nice to address the subnav underline at some point, but I don't think that's worth holding this up.

teesloane:
Thanks @janedotbiz and @tfe! I will merge this in today (and presumably can close #4754 following that).

PR #4810, opened 2020-10-05
Update status-badges.md

korya:
Is there any chance in the world that this process can be automated? I don't see a single reason not to have it automated on the Project Settings the way it used to be before the transition to the new UI.


rosieyohannan:
Hi @korya! Thanks for letting us know this feature is important to you. My understanding is that at the present time the feature is in our UX roadmap but has not been prioritised due to low usage in the past. I have reported your message to the team and you might want to add you thoughts to our ideas board so they remain visible and perhaps others will contribute: https://ideas.circleci.com/. ☺️

korya:
Thanks!

PR #4814, opened 2020-10-06
[CIRCLE-29714] update registry links to devhub

dsayling:
# Description
Mass update the links to the dev hub. Includes a localized update to the links previously not present.

# Reasons
CIRCLE-29714


lokst (commented):
@dsayling Thanks for updating the links! I was wondering if we should not link to the localized pages for now, until we have finished localizations?

dsayling:
Sure thing. We can make a large localization update to the docs when that's complete.

lokst (approved):
Looks good! I also just realized that it might be better to link to https://circleci.com/developer/orbs rather than https://circleci.com/developer/orbs/ to avoid a redirect, but I don't think it's a blocker since it works either way

dsayling:
Oh I didn't think of that. I just replaced the existing path. I'll make that change too, easy peasy.

dsayling:
I just searched for developer/orbs/[^a-z] and removed the trailing slashes. I think that covers the useages.

thomasr8 (approved):
Looks good 👍

PR #4820, opened 2020-10-09
Remove instructions for accessing the old UI

annapamma:
This page contained a note and screenshots for access the old UI with the "Old Experience" button. However, access to the old UI was sunset as of Aug 14.

# Description
This proposed change removes those instructions, to prevent readers from looking for a button they no longer have access to.

# Reasons
Remove outdated reference to old UI for clarity


rosieyohannan:
Ooops good spot @annapamma! thanks :-)

rosieyohannan (approved):

PR #4825, opened 2020-10-12
Remove obsolete note about jobs that use 2.1 config

yaningo:
# Description
Removing note "_Triggering a new job is not currently supported with configurations that specify version: 2.1_" in the API (v1) documentation:

- https://circleci.com/docs/api/v1/#trigger-a-new-job
- https://circleci.com/docs/api/v1/#trigger-a-new-job-with-a-branch

# Reasons
The note is no longer applicable (see [PR #4821](https://github.com/circleci/circleci-docs/pull/4821))


stig (commented):
While this appears to work, I'm not sure it's something we want to encourage: It replaces the workflows section in the config with one that includes just that job. This workflow could fail if the job in question requires a context, or a workspace, or a cache. I think it would be better to encourage the use of a [conditional workflow](https://circleci.com/docs/2.0/pipeline-variables/#conditional-workflows) that you can trigger via the [v2 API](https://circleci.com/docs/api/v2/#operation/triggerPipeline).

teesloane:
@yaningo - is there any follow up on this ? Wondering whether we should move this forward or close the draft.

PR #4832, opened 2020-10-13
Add additional example on tag filters for workflows

nbialostosky:
# Description
This PR adds an additional example of how to utilize tag and branch filtering to run specific workflows.

# Reasons
We had a customer make the suggestion of adding this example to the documentation and I believe it adds some additional clarity to the different ways you can use tags to filter workflows.

### DIFF
--- a/jekyll/_cci2/workflows.md
+++ b/jekyll/_cci2/workflows.md
@@ -447,7 +447,7 @@ workflows:
   staging: # This workflow will only run on 'main' and will not run on tags
     jobs:
       - test:
-          filters: &filters-staging
+          filters: &filters-staging # this yaml anchor is setting these values to "filters-staging"
             branches:
               only: main
             tags:
@@ -456,11 +456,11 @@ workflows:
           requires:
             - build
           filters:
-            <<: *filters-staging
+            <<: *filters-staging # this is calling the previously set yaml anchor
   production: # This workflow will only run on tags (specifically starting with 'v.') and will not run on branches
     jobs:
       - test:
-          filters: &filters-production
+          filters: &filters-production # this yaml anchor is setting these values to "filters-production"
             branches:
               ignore: /.*/
             tags:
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/workflows.md
```diff
@@ -430,6 +430,48 @@ workflows:
               ignore: /.*/
 ```
 
+In the example below, two jobs are defined (`test` and `deploy`) and three workflows utilize those jobs:
+
+- The `build` workflow runs for all branches except `main` and is not run on tags.
```
rosieyohannan, line 435: For this to be true do we need to add the `tags:` `ignore:` keys for the `build` workflow?

nbialostosky: @rosieyohannan great question! You actually don't since we don't automatically run workflows on tags, so by not including it doesn't run on tags. I set up this config example in one of my test accounts, and you can see the three scenarios here:

Screen Shot 2020-10-15 at 8 06 26 AM

rosieyohannan: Ah interesting! Thanks :-)

jekyll/_cci2/workflows.md
```diff
@@ -430,6 +430,48 @@ workflows:
...
+          filters: &filters-staging
+            branches:
+              only: main
+            tags:
+              ignore: /.*/
+      - deploy:
+          requires:
+            - build
+          filters:
+            <<: *filters-staging
```
rosieyohannan, line 459: Can we just add comments in for these lines to say we've using the same 'staging' filters for this job (same for use of alias in production), just to spell it out?

rosieyohannan (approved):
🎉

jekyll/_cci2/workflows.md
rosieyohannan: Maybe we can add a comments to say this explicitly if it's not too wordy?

jekyll/_cci2/workflows.md
nbialostosky: I updated this with commit 6502576 can you check to see if that works?

rosieyohannan: Great, thank you!!

PR #4835, opened 2020-10-14
fix URL and some editorial changes and update UI images

rosieyohannan:
UI image is fixed in a separate PR. The focus of this PR is mainly to make the following URL just an example, rather than a clickable link.

Screenshot 2020-10-23 at 14 57 06


teesloane (approved):
Is this still a [WIP]?  Looks good to go otherwise!

rosieyohannan:
> Is this still a [WIP]? Looks good to go otherwise!

Yeah I just failed to press save when removing the WIP 🙄

PR #4836, opened 2020-10-14
Fix docker orb usage

ruescar:
# Description
Update docker/check usage example

# Reasons
docker/checks takes **env_var_name** not **string**


ruescar:
@teesloane Why did you close this PR?

PR #4844, opened 2020-10-16
List bash as a required tool

J3RN:
# Description
Add bash to the list of required tools for primary images.

# Reasons
I was surprised to see a build using a custom image built on Alpine Linux fail because `bash` wasn't installed. Alpine apparently doesn't include `bash` by default, so I'm hoping to save the next person a few minutes by implying they should check to ensure that `bash` is present in their image.


felicianotech (approved):
LGTM

PR #4850, opened 2020-10-18
Update Urls in single.md and migrating-from-jenkins.md file

kishanjasani:
# Description
Update Urls in `single.md` and `migrating-from-jenkins.md` file.


teesloane (approved):
Thanks!

PR #4852, opened 2020-10-19
Update machine image examples to use Ubuntu 20.04 image

lokst:
# Description
Reference new `ubuntu-2004:202010-01` image in examples

# Reasons
We have an Ubuntu 20.04 machine image available now


felicianotech (approved):
LGTM

PR #4863, opened 2020-10-21
Update caching.md

fernfernfern:
# Description
It looks like some html artifact was added to the cache key? `<etc>` wouldn't be valid b64 output. or maybe it was supposed to represent padded output and became malformed? You can see [examples here](https://en.wikipedia.org/wiki/Base64#Examples). 

# Reasons
it looks a bit odd. I added the output padding, but just removing the `<etc>` would work as well I think


teesloane (approved):
Good catch, thanks @fernfernfern !

PR #4879, opened 2020-10-27
Clarify where steps aren't run due to being killed

BytesGuy:
# Description
Added clarity around the the `when` attribute, `store_artifacts` and  `store_test_results` not running if the job has been **killed** by a cancel request or reaching the global 5 hour timeout.

# Reasons
Support Ticket #80787


teesloane (approved):
Lovely, thanks @BytesGuy !

Issue #4883, opened 2020-10-28
incorrect orb name template in orb-concepts and using-orbs pages.


irajtaghlidi:
# Locations
https://circleci.com/docs/2.0/orb-concepts/#jobs
and
https://circleci.com/docs/2.0/using-orbs/#jobs

# Details
The orb name is incorrect template, it should be start with `<namespace>` and next with `<orb>` name, like:
```
orbs:
    <orb-name>: <namespace>/<orb-name>@1.2.3
```

But, it's in this wrong format now:
```
orbs:
  <orb>: <orb>/<namespace>@x.y #orb version

```

rosieyohannan:
Fixed in #4884

PR #4884, opened 2020-10-28
fix wrong orb name template in orb-concept and using-orbs pages.

irajtaghlidi:
fix some incorrect orb name template in two document pages.

related to #4883


rosieyohannan:
@irajtaghlidi thanks so much for spotting this error, and fixing it! 😀

rosieyohannan (approved):

PR #4890, opened 2020-10-29
Add: Note re: adding users.

teesloane:
See [pertinent conversation](https://circleci.slack.com/archives/C0109328RTR/p1603830259101200).

### DIFF
--- a/jekyll/_cci2/getting-started.md
+++ b/jekyll/_cci2/getting-started.md
@@ -290,7 +290,7 @@ cat <file_name>      # show me the contents of the file <file_name>
 
 ## Collaborating with Teammates
 
-It is easy to bring others into into a workflow such as described above.
+It is easy for teammates and collaborators to view and follow your projects.
 Teammates can make a free CircleCI account at any time to view your pipelines,
 even if they are not committing any code. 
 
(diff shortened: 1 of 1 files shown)


thomasr8 (approved):

jekyll/_cci2/getting-started.md
```diff
@@ -288,6 +288,12 @@ cd <directory_name>  # change directory to the <directory_name> directory
 cat <file_name>      # show me the contents of the file <file_name>
 ```
 
+## Collaborating with Teammates
+
+It is easy to bring others into into a workflow such as described above.
```
rosieyohannan, line 293:
```suggestion
It is easy to bring others into a workflow such as described above.
```

rosieyohannan: Just wondering actually - is "workflow" being used here in the general, rather than CCI sense? if so maybe we can phrase differently? Something like - 'It's simple for others to view and follow your projects'?

teesloane: Yes - _workflow_ is being used in general (poor choice, haha!) I think I better use your suggestion!

Issue #4719, opened 2020-09-15
Concepts/#cache:


thomasr8:
- add a link to clearing cache stepsgetting-started
- add link to academy

PR #4896, opened 2020-11-04
[CIRCLE-30688] parallelism-faster-jobs: fix glob command example

teka23:
#4719 Description
Fix an example where the pattern is not wrapped with quote marks.

# Reasons
https://circleci.atlassian.net/browse/CIRCLE-30688

These are the only instances that I can find without wrapped with quote marks.


rosieyohannan:
Thanks so much for spotting and fixing this @teka23! ☺️

rosieyohannan (approved):

PR #4921, opened 2020-11-13
docker/checks command only accepts env variable name

ganezasan:
# Description

The usage of the `docker/checks` is wrong, and it needs to pass environment variable names into docker-password and docker-username.

I confirmed this example in the following build.
> https://app.circleci.com/pipelines/github/ganezasan/circleci-sandbox/3313/workflows/c61ce437-e083-4f5f-8687-fcabbed10079/jobs/8167

# Reasons
A customer pointed out this issue through a ticket.

### DIFF
--- a/jekyll/_cci2/private-images.md
+++ b/jekyll/_cci2/private-images.md
@@ -78,8 +78,8 @@ jobs:
     machine: true
     steps:
       - docker/check:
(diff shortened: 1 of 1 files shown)


rosieyohannan (approved):
Thank you! 🎉

PR #4941, opened 2020-11-24
Add note on where "sudo" should go

makusu2:
# Description
Specify where the `sudo` should go.

# Reasons
Since this is a piped operation, doing `sudo curl -fLSs https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh | bash` won't work.


teesloane (approved):
Good catch! Thanks @makusu2 !

PR #4951, opened 2020-11-27
Added instruction for copying public key to authorized_hosts

PetersonRyan:
# Description
Added instruction to copy public SSH key to `authorized_hosts` file, allowing SSH access using the private key just generated.

# Reasons
Following this guide myself to add my own server's SSH key to my CircleCI project resulted in frustration as the guide, along with many StackOverflow threads, did not remind me to copy the public key for the server's SSH key pair to its own `authorized_hosts` file.
This step is crucial when adding SSH keys for any server that does not already handle this step automatically (like using GitHub's GUI).


teesloane:
Hi @PetersonRyan - sorry for the late follow up. A few months ago this document was updated with the following:

 `Note: You may need to add the public key to ~/.ssh/authorized_keys in order to add SSH keys.`

There is a small typo in the above currently so I'll patch that in a quick PR, but hopefully this is sufficient to clear up future confusion. We appreciate your pull request nonetheless.

PR #4958, opened 2020-12-03
Move information about tracking available memory to own note

nbialostosky:
# Description
We have a note that combined two concepts:

1. Info about possibly needing to pin the number of CPU count
2. Info about how to find the amount of memory you have been allocated

Combined it reads that the command to find memory helps find information about CPU usage, but in reality, they are separate concepts.

This PR will move the information about finding memory allocation to it's own note so the concepts are split up.

# Reasons
Was working with a customer and was referencing this section and it caused some confusion.


teesloane (approved):
Lovely! Thanks @nbialostosky!

PR #4964, opened 2020-12-08
fix proxy setting

ganezasan:
# Description
Change the scripts for setting proxy

# Reasons
The initial document's script doesn't work for the current server, and we noticed `/etc/circle-installation-customizations` is just a shell script, so to set environment variables, it needs to set `export` before the environment variables.

### DIFF
--- a/jekyll/_cci2/proxy.adoc
+++ b/jekyll/_cci2/proxy.adoc
@@ -71,7 +71,7 @@ EOF
 systemctl restart replicated\*
 ```
 
-NOTE: NOTE: If the proxy requires user authentication, you will need to be added user authentication information into <proxy-ip>.
+NOTE: NOTE: If the proxy requires user authentication, you will need to add user authentication information into <proxy-ip> like below:
 
 ```
 export HTTP_PROXY=http://<userid>:<password>@<proxy-ip>:<port>
--- a/jekyll/_cci2/proxy.adoc
+++ b/jekyll/_cci2/proxy.adoc
@@ -81,7 +81,7 @@ export HTTPS_PROXY=https://<userid>:<password>@<proxy-ip>:<port>
 NOTE: The above is not handled by by our enterprise-setup script and will need to be added to the user data for the Services Machine startup or done manually.
 
 // not quite clear what is meant by this note... what's user data in this context?
-// If the cusotmer automatically apply the above setting, they need to add them into ServiceBox's user data.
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/proxy.adoc
```diff
@@ -60,18 +61,31 @@ If you run in Amazon's EC2 service then you'll need to include `169.254.169.254`
...
+export HTTPS_PROXY=<proxy-ip:port>
+export NO_PROXY=169.254.169.254,<circleci-service-ip>,127.0.0.1,localhost,ghe.example.com
+export JVM_OPTS="-Dhttp.proxyHost=<proxy-ip> -Dhttp.proxyPort=<proxy-port> -Dhttps.proxyHost=<proxy-ip> -Dhttps.proxyPort=<proxy-port> -Dhttp.nonProxyHosts=169.254.169.254|<circleci-service-ip>|127.0.0.1|localhost|ghe.example.com"
+EOF
+) | sudo tee -a /etc/circle-installation-customizations
 
 systemctl restart replicated\*
 ```
 
+NOTE: NOTE: If the proxy requires user authentication, you will need to be added user authentication information into <proxy-ip>.
```
makotom, line 74:
```suggestion
NOTE: NOTE: If the proxy requires user authentication, you will need to add user authentication information into <proxy-ip> like below:
```

jekyll/_cci2/proxy.adoc
```diff
@@ -60,18 +61,31 @@ If you run in Amazon's EC2 service then you'll need to include `169.254.169.254`
...
+
+```
+export HTTP_PROXY=http://<userid>:<password>@<proxy-ip>:<port>
+export HTTPS_PROXY=https://<userid>:<password>@<proxy-ip>:<port>
+```
+
 NOTE: The above is not handled by by our enterprise-setup script and will need to be added to the user data for the Services Machine startup or done manually.
 
 // not quite clear what is meant by this note... what's user data in this context?
+// If the cusotmer automatically apply the above setting, they need to add them into ServiceBox's user data.
```
makotom, line 84:
```suggestion
// If the customer wants to apply the above setting automatically, they need to add them into ServiceBox's user data.
```

makotom (approved):
Added some English word tuning, but in general it looks good to me! Thank you for your work. 😸

jekyll/_cci2/proxy.adoc
ganezasan: Thank you so much!

PR #4968, opened 2020-12-10
fix mistranslation

kurumai:
# Description
Fix mistranslation of Japanese doc

# Reasons
"The SHA1 hash of the last commit of the current build." was mistranslated as "現在のビルドの前回のコミット"


mfunaki (approved):
LGTM

PR #4971, opened 2020-12-11
Indicate which version of Yarn examples are referring to and add Yarn 1.x example

jamsinclair:
# Description
- Make it clear which Yarn version the example is for
- Adds example for caching with Yarn 1.x versions

# Reasons

The fourth  Documentation value states
> 4. Clear: documentation should be clear.
> https://github.com/circleci/circleci-docs/blob/master/docs/CONTRIBUTING.md#circleci-documentation-values

The documentation examples do not make it clear which version of Yarn they are referring to.

Yarn 1.x is still very popular and frequently used in projects. Users of Yarn 1.x may be confused by the current example.


teesloane (approved):
Lovely — thanks @jamsinclair !

PR #4975, opened 2020-12-14
Add default setup_remote_docker version

ganezasan:
# Description

Add 17.09.0-ce version which is the default version of the setup_remote_docker

# Reasons

A customer asked about it


teesloane (approved):
Thanks!

PR #4983, opened 2020-12-17
[CIRCLE-31528] Align search results with search bar and nav

tfe:
This applies a design change to align the left side of search results with the left side of the search bar spyglass and header/nav. Previously it was indented.

I didn't un-indent the contents of the div that was removed in order to keep the diff clean and not muddy the line history, but I'm happy to change that if desired.

Note that using a `row` element without a child `col` element is a misuse of the Bootstrap grid system, but the header and search bar are already doing this so we have to make the same mistake here to get them to align with each other.

### DIFF
--- a/jekyll/_layouts/classic-docs.html
+++ b/jekyll/_layouts/classic-docs.html
@@ -14,7 +14,6 @@
 	<div id="hits-target" style="display:none;">
 		<div class="container">
 			<div class="row">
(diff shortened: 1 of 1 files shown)


janedotbiz (approved):
Thanks for the note about why you left `row`

tfe:
Thanks for the review. It looks like I don't have permissions to merge this, so if you could that would be appreciated.

PR #4985, opened 2020-12-17
Add sans-serif to all font-family entries.

gmemstr:
# Description

Helps smooth out the fallback behavior when Roboto isn't present (e.g Google Fonts down/blocked).

# Reasons

I have Google Fonts blocked (mostly for performance reasons), which leads to the docs falling back to a serif font.


teesloane (approved):

teesloane:
Good idea! Thanks!

PR #5000, opened 2020-12-29
Remove outdated range cron syntax limitation.

MikhailTymchukFT:
# Description
Remove sentence `Range elements within comma-separated lists of elements are also **not** supported.`
I believe this is no longer true.

# Reasons
We use such syntax without any issues.


teesloane (approved):

teesloane:
Hi @MikhailTymchukFT  do you mind sharing the syntax you are using that is working in your workflows so that I might test it?

MikhailTymchukFT:
@teesloane 

We have this cron schedule for a workflow section:
```yml
triggers:
  - schedule:
      cron: "0,5,10,15,20,25,30,35,40,45,50,55 * * * *" # every 5 minutes
```

PR #5004, opened 2021-01-05
Delete symlink workaround once htmlproof passes

kimh:
The existence of symlink made `aws s3 sync` to fail in deploy job https://app.circleci.com/pipelines/github/circleci/circleci-docs/26932/workflows/0dea04d6-07e6-474f-b4cc-d1bc5497bef3/jobs/73385


hanks (approved):
LGTM， add `rm -f` could be better to suppress file not found error if happens.

kimh:
> LGTM， add rm -f could be better to suppress file not found error if happens.

I thought about it and I want the step to be deterministic or fail, so I decided to go without `-f`.

PR #5024, opened 2021-01-14
Correct path to modules folder in Yarn 1 example

heuperman:
# Description
Correct the path to the modules folder in the Yarn 1.x caching example.

# Reasons
The current suggested path to cache in the Yarn 1.x example actually points to the Yarn 2 install cache folder, making this a rather misleading example. Yarn 1.x creates a project level `node_modules` folder when installing, in the same way as NPM does. The path `./node_modules` is much more likely to point to the correct folder.

### DIFF
--- a/jekyll/_cci2/yarn.md
+++ b/jekyll/_cci2/yarn.md
@@ -68,7 +68,7 @@ An example for Yarn 1.x:
           name: Save Yarn Package Cache
           key: yarn-packages-{{ checksum "yarn.lock" }}
           paths:
-            - ~/.cache/yarn
+            - ./node_modules
 #...
 ```
 {% endraw %}


teesloane:
Hi there!  I haven't used Yarn 1 in a little while but -- I  believe in this example we are specifying on [line 66](https://github.com/circleci/circleci-docs/pull/5024/files#diff-d0b151db5c9fa421f98c1ea05a9fdefb64da10eead38a8e8cdac3f8313798e33L66) that the cache folder should be at the user root, as specified with the ([--cache-folder flag](https://classic.yarnpkg.com/en/docs/cli/cache#toc-change-the-cache-path-for-yarn)) - and so we also specify to cache that folder instead of the standard `node_modules` folder.

heuperman:
Ah I missed that, thank you for clarifying!

PR #5040, opened 2021-01-21
Update description of "checkout" step

lalli-flores:
# Description
The description on `checkout` states that if the path already exists in the workspace, then it will _pull_ origin. In reality, the script does a _fetch_ command and it doesn't integrate remote changes into the current branch, if there are any.
Updating the docs to use more accurate wording, as a `git pull` is different from a `git fetch`

# Reasons
I've recently updated my workflow to use `checkout`. Because the description states that the script does a pull, I assumed that meant `gill pull`, however, after my workflow failed and looking at the `checkout` script, it only does a `git fetch`. I'm suggestion an update to the docs to reflect a more accurate wording that mimics Git's commands

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -882,7 +882,7 @@ path | N | String | Checkout directory. Will be interpreted relative to the [`wo
 {: class="table table-striped"}
 
 If `path` already exists and is:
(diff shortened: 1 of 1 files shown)


teesloane (approved):

teesloane:
Thanks for the patch!

PR #5041, opened 2021-01-22
not fixing circle_working_directory

dsayling:
I tried fixing working directory expansion, but it appears to not be worth fixing. Adding this hack back

### DIFF
--- a/jekyll/_cci2/sample-config.md
+++ b/jekyll/_cci2/sample-config.md
@@ -776,6 +776,7 @@ jobs:
           name: Build iperf3
           working_directory: iperf
           command: |
+            CIRCLE_WORKING_DIRECTORY=$(eval "echo $CIRCLE_WORKING_DIRECTORY")
             IPERF3_MAKE_PREFIX=$CIRCLE_WORKING_DIRECTORY/<< parameters.label >>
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/sample-config.md
```diff
@@ -776,6 +776,7 @@ jobs:
           name: Build iperf3
           working_directory: iperf
           command: |
+            CIRCLE_WORKING_DIRECTORY=$(eval "echo $CIRCLE_WORKING_DIRECTORY")
             IPERF3_MAKE_PREFIX=$CIRCLE_WORKING_DIRECTORY/<< parameters.label >>
```
glenjamin, line 780: Would `../<< parameters.label >>` work instead?

dsayling: ya, but I think this here to demonstrate the need to use "eval" to expand the working directory into a usable path.

dsayling: The only other reference is here in discuss: https://discuss.circleci.com/t/circle-working-directory-doesnt-expand/17007

glenjamin: oh hah! i only quickly looked at this and thought it was part of the build config for the docs themselves 🤦

teesloane (approved):

PR #5042, opened 2021-01-22
Remove details about SSH

ndp:
(It seems like this is how you would like feedback, so here goes...)

I read these paragraphs multiple times and can't figure out what they are saying. Well, I think I understand, but I'm not sure why this is a benefit: It seems to imply that if I ssh in, I'll have an interactive shell that will produce different results than my normal build. Why would I want that?

Perhaps they are making one point (and should be a single paragraph)? or two different points... idk. 

I _think_ the benefit of "SSH into builds" is described well without these graphs. 

Or somebody who understands this better can perhaps reword them. 
Thx

# Description

# Reasons

### DIFF
--- a/jekyll/_cci2/features.md
+++ b/jekyll/_cci2/features.md
@@ -20,11 +20,6 @@ Following are the most popular features CircleCI offers that Developers are inte
 
 Often the best way to troubleshoot problems is to [SSH into a job]({{ site.baseurl }}/2.0/ssh-access-jobs/) and inspect things like log files, running processes, and directory paths. CircleCI 2.0 gives you the option to access all jobs via SSH.
 
-When you log in with SSH, you are running an interactive login shell. You are also likely to be running the command on top of the directory where the command failed the first time, so you are not starting a clean run. 
-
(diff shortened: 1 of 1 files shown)


teesloane:
Hi @ndp  - thanks for opening a PR. I agree with you that the points you have removed are a little confusing. I did a bit of research on interactive / non-interactive shells with regard to our internal system so I think we can clear this up. If you are interested we can update your PR with something along the lines so we can merge your work in:

I believe we can go ahead with removing line 23 and update line 25 with something similar to the following:

> Note: When CircleCI runs your pipeline, the [`run` ](https://circleci.com/docs/2.0/configuration-reference/#run) command executes shell commands in a _non-interactive_ shell. When SSH-ing into a build, you are using an _interactive_ shell (see the section on [Invocation](https://linux.die.net/man/1/bash) in the bash manual). An interactive bash shell load a series of startup files (such as `~/.bash_profile`) which may or may not change the outcome of SSH debugging process.

jekyll/_cci2/features.md
```diff
@@ -20,11 +20,6 @@ Following are the most popular features CircleCI offers that Developers are inte
```
morshinnikish, line 20: понятно!я сейчас учусь писать коды(програмирование)

PR #5043, opened 2021-01-23
Drop systemd-run in favour of sudo -u

glenjamin:
# Description
While systemd-run does offer slightly more isolation, the additional
complexity and compatibility concerns make this awkward, and leads
to some users struggling to get up and running

As the user already needs to be careful with leaking state to the
filesystem, it doesn't seem unreasonable to leave the also responsible
for leaking processes

### DIFF
--- a/jekyll/_cci2/runner-installation.adoc
+++ b/jekyll/_cci2/runner-installation.adoc
@@ -193,7 +193,7 @@ sudo chmod 600 /opt/circleci/launch-agent-config.yaml
 
 === Create the circleci user & working directory
 
-These will be used when executing the `build-agent`.
+These will be used when executing the task agent. These commands must be run as a user with permissions to create other users (e.g. `root`).
 
 ```bash
 id -u circleci &>/dev/null || adduser --uid 1500 --disabled-password --gecos GECOS circleci
--- a/jekyll/_cci2/runner-installation.adoc
+++ b/jekyll/_cci2/runner-installation.adoc
@@ -177,7 +177,7 @@ api:
   auth_token: AUTH_TOKEN
 runner:
   name: RUNNER_NAME
-  command_prefix: ["/opt/circleci/launch-task"]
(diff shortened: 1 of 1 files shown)


pete-woods (approved):

mvxt:
@glenjamin Actually, this change (along with a few items noted in JIRA ticket) would also make runner viable for CentOS. The entire reason CentOS support got stuck was because of PolicyKit permissions issues in executing a systemd-unit

bjohnso5:
Followed these steps to get a runner up in a local VM, and all went well with one exception:
* [this section](https://circleci.com/docs/2.0/runner-installation/index.html#create-the-circleci-user-working-directory) it wasn't clear that you'd need to be root to run the steps

The result of the workflow can be found [here](https://app.circleci.com/pipelines/github/bjohnso5/runner-dummy/1/workflows/69de1ba3-345e-4350-b06e-e411d6500c3d/jobs/1). The last job failed, but it was simply a copy of what was in circleci/dummy-runner, so I'm not sure if it was intended to pass or not 🤷 .

glenjamin:
> [this section](https://circleci.com/docs/2.0/runner-installation/index.html#create-the-circleci-user-working-directory) it wasn't clear that you'd need to be root to run the steps

Is there an easy change we can make to resolve this?

If so, lets do that and then merge this?

bjohnso5:
> > [this section](https://circleci.com/docs/2.0/runner-installation/index.html#create-the-circleci-user-working-directory) it wasn't clear that you'd need to be root to run the steps
> 
> Is there an easy change we can make to resolve this?
> 
> If so, lets do that and then merge this?

Just something simple like `These will be used when executing the build-agent. This command must be run as root`?

glenjamin:
Sounds good, are you ok to push a commit with that onto this PR?

jrahme-cci (approved):

PR #5046, opened 2021-01-26
Fix runner config reference heading levels

glenjamin:
# Description
The current heading level was off, leaving this to be nested under macOS.

Also reword `command_prefix` to be a bit clearer.


teesloane:
@thomasr8 do you know if we can merge this?

thomasr8 (approved):

PR #5049, opened 2021-01-27
Add a note that circleci tests split cannot be run locally

nbialostosky:
# Description
This PR will add a note that the `circleci tests split` command cannot be run locally.

# Reasons
We had customers asking if this was possible or not, and at the moment it isn't. We note in the docs steps on adding the CLI locally (since it can be used for other things) but could lead to confusion that the command could be run. 

See comment here for confirmation this command cannot be run locally:
https://github.com/CircleCI-Public/circleci-cli/issues/383#issuecomment-755330662

### DIFF
--- a/jekyll/_cci2/parallelism-faster-jobs.md
+++ b/jekyll/_cci2/parallelism-faster-jobs.md
@@ -46,6 +46,8 @@ CircleCI supports automatic test allocation across your containers. The allocati
 
 To install the CLI locally, see the [Using the CircleCI Local CLI]({{ site.baseurl }}/2.0/local-cli/) document.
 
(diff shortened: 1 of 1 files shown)


glenjamin (commented):
@danmux Did I remember correctly that there's a way to test this out locally in some scenarios?

danmux:
>@danmux Did I remember correctly that there's a way to test this out locally in some scenarios?

Yep there is a way, but it is hacky and the command is hidden in the public cli - so currently i think it has to be considered not available - why it is not I don't know. If others find it useful (even with some limitations) then why not - but that is a product call.

glenjamin (approved):

PR #5050, opened 2021-01-27
runner updates (reference to intel removed)

### LINKED ISSUE #4934: Runner Platform Support is Unclear
Doc: https://circleci.com/docs/2.0/runner-overview/#circleci-runner-limitations

For platform support, it means "Intel +" such as this: `Intel + Ubuntu`

Are we saying then that this wouldn't run on an AMD x86_64 processor? While I'm not 100% positive that it does, I'm fairly sure that it would. Intel doesn't technically own their own 64-bit platform right now so unless we're using Intel specific microcode APIs, amd64 is supported.

thomasr8:
Replaced reference to intel with x86_64

This fixes #4934 

# Description

# Reasons

### DIFF
--- a/jekyll/_cci2/runner-overview.adoc
+++ b/jekyll/_cci2/runner-overview.adoc
@@ -34,7 +34,7 @@ There are two key use cases CircleCI aims to meet with the runner:
 
 == Runner operation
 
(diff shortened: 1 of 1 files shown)


glenjamin (approved):
I think this is repeated on the install phase

PR #5057, opened 2021-01-29
Added caveat for use of Shared Libraries

makotom:
# Description
Added a caveat that we do not support Shared Libraries.

# Reasons
Increasing number of feedbacks refer to the use of Shared Libraries.

### DIFF
--- a/jekyll/_cci2/jenkins-converter.md
+++ b/jekyll/_cci2/jenkins-converter.md
@@ -15,7 +15,7 @@ Currently, the converter only supports declarative Jenkinsfiles. While the numbe
 
 * A limited number of syntaxes and plugins are supported. Jenkinsfiles relying on unsupported syntaxes and plugins cannot be converted. Please manually remove them.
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/jenkins-converter.md
```diff
@@ -15,6 +15,8 @@ Currently, the converter only supports declarative Jenkinsfiles. While the numbe
 
 * A limited number of syntaxes and plugins are supported. Jenkinsfiles relying on unsupported syntaxes and plugins cannot be converted. Please manually remove them.
 
+* Only a single Jenkinsfile is accepted per request. Namely, [Shared Libraries](https://www.jenkins.io/doc/book/pipeline/shared-libraries/) will not be resolved, and resulting `config.yml` can be incomplete. Note that under certain cases the converter does not raise errors even if there are unresolvable Shared Libraries.
```
teesloane, line 18:
```suggestion
* Only a single Jenkinsfile is accepted per request. Namely, [Shared Libraries](https://www.jenkins.io/doc/book/pipeline/shared-libraries/) will not be resolved, and the resulting `config.yml` may be incomplete. Note that under certain cases the converter does not raise errors even if there are unresolvable Shared Libraries.
```

teesloane (changes requested):
Just a small wording change, otherwise good to go.

thomasr8 (approved):

dsayling (approved):

teesloane (approved):

PR #5069, opened 2021-02-04
Add note about range elements for days

nbialostosky:
# Description
This PR adds a disclaimer that setting up a range of days written out is not supported. 

# Reasons
Crontab says this is valid:

https://crontab.guru/#0_5_*_*_Tue-Sat

And while that might be the case for some systems, it isn't valid within a CircleCI config. Instead, the values need to be comma separated digits like: https://crontab.guru/#0_5_*_*_2,3,4,5

### DIFF
--- a/jekyll/_cci2/workflows.md
+++ b/jekyll/_cci2/workflows.md
@@ -276,7 +276,7 @@ A valid `schedule` requires a `cron` key and a `filters` key.
 The value of the `cron` key must be a [valid crontab entry](https://crontab.guru/).
 
 **Note:**
-Cron step syntax (for example, `*/1`, `*/20`) is **not** supported. Range elements within comma-separated lists of elements are also **not** supported. 
+Cron step syntax (for example, `*/1`, `*/20`) is **not** supported. Range elements within comma-separated lists of elements are also **not** supported. In addition, range elements for days (for example, `Tue-Sat`) is **not** supported, use comma seperated digits instead.
 
 The value of the `filters` key must be a map that defines rules for execution on specific branches.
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/workflows.md
```diff
@@ -276,7 +276,7 @@ A valid `schedule` requires a `cron` key and a `filters` key.
 The value of the `cron` key must be a [valid crontab entry](https://crontab.guru/).
 
 **Note:**
-Cron step syntax (for example, `*/1`, `*/20`) is **not** supported. Range elements within comma-separated lists of elements are also **not** supported. 
+Cron step syntax (for example, `*/1`, `*/20`) is **not** supported. Range elements within comma-separated lists of elements are also **not** supported. In addition, range elements for days (for example, `Tue-Sat`) is **not** supported, use comma seperated digits instead.
```
thomasr8, line 279:
```suggestion
Cron step syntax (for example, `*/1`, `*/20`) is **not** supported. Range elements within comma-separated lists of elements are also **not** supported. In addition, range elements for days (for example, `Tue-Sat`) is **not** supported. Use comma-separated digits instead.
```

thomasr8: Just a couple of minor changes needed. Overall looks good.

teesloane:
Just to cross-check, a [user submitted a PR](https://github.com/circleci/circleci-docs/pull/5000) that they are able to use the cron syntax we say is unavailable - is it the case that we don't officially support it but it still works?

nbialostosky:
@teesloane I think we just need to improve the language around that maybe? What the docs are trying to communicate is that you can't use a range element (like `3-5`) within a comma-separated list, i.e. this doesn't work:

```
    triggers:
      - schedule:
          cron: "5 4 * * 1,3-5,6"
```

So you would need to list them all out, i.e.:

```
    triggers:
      - schedule:
          cron: "5 4 * * 1,3,4,5,6"
```

teesloane:
Ah, got it. Thanks @nbialostosky ! I will close that user's PR and update the wording after we merge this!

thomasr8 (approved):

PR #5071, opened 2021-02-05
Update a warning for the customization script to use docker proxy in machine / remote docker

ganezasan:
# Description

The customization is only available after 2.19 and later versions.

# Reasons

A customer failed to configure this due to the old server version.


makotom (approved):
LGTM!

The build is failing for some reasons, and we might need to have it fixed before merging this to the main branch.

ganezasan:
@makotom Thanks! I merged PR!

PR #5074, opened 2021-02-08
Add: clarification on valid cron syntax.

teesloane:
Clarifies and closes #5000.

See discussion in https://github.com/circleci/circleci-docs/pull/5069


nbialostosky (approved):
This is great, I think this is really clear about the whole 'range' situation and what does and doesn't work!

PR #5076, opened 2021-02-08
docker image updated

thomasr8:
Added an updated docker image to the sample configuration.

# Description

# Reasons

### DIFF
--- a/jekyll/_cci2/language-android.md
+++ b/jekyll/_cci2/language-android.md
@@ -52,7 +52,7 @@ jobs:
   build:
     working_directory: ~/code
     docker:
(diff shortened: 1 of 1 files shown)


thomasr8:
@teesloane : If you have a few minutes, would you please review the sample configuration to verify if the steps are still correct? @felicianotech mentioned that these steps may need to be updated based on his understanding.

Thanks 👍

teesloane (approved):

teesloane:
We can bump this; the guide could use a refresh so I'll consult with CPE to see if it's something they will be taking on in their language-guide updates.

PR #5094, opened 2021-02-17
Changed the words master to main

punkdata:
# Description
I replace the word master with main

# Reasons
I saw this word master in our text and considering that the industry is replacing these words with more generic and acceptable terms I suggest we use main as a replacement for master moving forward.


teesloane (approved):
Good catch, thanks @punkdata !

PR #5097, opened 2021-02-19
Add docs on key rotation. 

teesloane:


jekyll/_cci2/managing-api-tokens.md
```diff
@@ -9,75 +9,95 @@ version:
...
+    - _Read Only_ tokens grant read only access to the project's API.
+    - _Admin_ tokens grant read and write access for the project's API.
 
-**Note**:
-API tokens cannot be modifed
-after they have been created.
-The only way to change existing tokens
-is to delete and recreate them.
+**Note**: API tokens cannot be modified after they have been created. The only
+way to change existing tokens is to delete and recreate them.
```
gocamille, line 32: @teesloane -- would we want to say here that `The only way to change existing tokens is to delete and recreate them, also known as token rotation`, to make it clear that the token rotation steps will do this job?

teesloane: Good call. Updated 👍 !

gocamille (approved):
Woot! 🚢

PR #5110, opened 2021-02-26
Language selector now redirects to the same content

teka23:
Make sidebar language selector redirect to the same content.

# Reasons

It used to redirect to the base URL. So, I was not using this selector,
because directly writing `ja` to the URL were much easier.

# Description

This implementation relies on regexp to parse the current full URL to
replace the `docs/ja` or `docs` part to the target slug.

It also redirects to the base URL as a fallback if the URL does not match.

### DIFF
--- a/src-js/lang.js
+++ b/src-js/lang.js
@@ -9,8 +9,8 @@ function getElById(id) {
 }
 
 var languages = {
-  "en": {name: "English", url: "docs/2.0/"},
-  "ja": {name: "日本語", url: "docs/ja/2.0/"}
+  "en": {name: "English", url: "docs/"},
+  "ja": {name: "日本語", url: "docs/ja/"}
 }
 
 // List of elements that have dynamic content based on our selected language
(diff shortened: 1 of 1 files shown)


teesloane:
I checked out the branch and changing the picker didn't seem to reload the page in the changed language.

I am embarrased to say that I started work on this but got distracted for another project. You can see my implementation [live here](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/ts-lang-picker-preview/2.0/migrating-from-jenkins/). I have put up a [draft PR](https://github.com/circleci/circleci-docs/pull/5114) in case we can collaborate; my implementation might be a bit naive and missing something.

teka23:
@teesloane I hardcoded hostname etc here. I guess that is why it is not working. I will fix it

teka23:
@teesloane ♻️ Please

teesloane (approved):
Looks good 👍

teka23:
@teesloane Thank you!

PR #5119, opened 2021-03-02
Clarify regex usage

fernfernfern:
# Description
The current regex example will trigger if a branch name starts with `server/`, but for it to match any string that contains `server/` it would need to be `/.*server\/.*/`.

# Reasons
 Changing the wording will help clarify the outcome.


teesloane (approved):
Thanks @fernfernfern !

PR #5121, opened 2021-03-02
[WEB-95] Search Results for mobile

tfe:
This adds a mobile version of the search result category subnav, as well as some page layout structure improvements to enable that change.

Those changes resolve some long-standing Bootstrap grid misuse in the header and search bar:

1. Search query bar and results container had a row without   accompanying columns (should just be a container sans row). The   query bar CSS had manually implemented (and incomplete for mobile)   margins to compensate, which are now removed.

2. The header logo was absolutely positioned, resulting in it appearing   outside its surrounding column/container (too far left). This commit   removes that, allowing it to be positioned on the left edge of its   container where it will align with other containers on the page,   such as the search query bar and results container.

3. The search results item wrapper had negative margins to allow space   around the hover/highlight background color, however this resulted   in horizontal scrolling on narrow screens because it was larger than   the gutter width. Note: there is still some horizontal scrolling   present on narrow screens being caused by something in the footer.

### DIFF
--- a/jekyll/_sass/_subnav.scss
+++ b/jekyll/_sass/_subnav.scss
@@ -46,3 +46,29 @@ $nav-offset-subnav: $nav-offset + $subnav-offset;
     top: -$nav-offset-subnav;
   }
 }
+
+// Another set of styles specific to this project:
+
+// 1. Override hiding of subnav on mobile
+.component.subnav {
+  @media screen and (max-width: $screen-xs-max) {
+    display: block;
+  }
+}
+#subnav-placeholder {
+  @media screen and (max-width: $screen-xs-max) {
+    display: block;
+  }
+}
+
+// 2. Add style for mobile
+.component.subnav {
+  @media screen and (max-width: $screen-xs-max) {
+    font-size: $font-size-small;
+
+    li {
+      padding-top: 8px;
+      padding-bottom: 8px;
+    }
+  }
+}
--- a/jekyll/assets/css/navbar.css
+++ b/jekyll/assets/css/navbar.css
@@ -26,7 +26,6 @@
   .nav-toggle.two .widescreen a.logo {
     position: absolute;
     top: 18px;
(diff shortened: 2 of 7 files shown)


jekyll/_sass/_subnav.scss
```diff
@@ -46,3 +46,29 @@ $nav-offset-subnav: $nav-offset + $subnav-offset;
...
+    display: block;
+  }
+}
+
+// 2. Add style for mobile
+.component.subnav {
+  @media screen and (max-width: $screen-xs-max) {
+    font-size: $font-size-small;
+
+    li {
```
janedotbiz, line 69: On the smallest screens (under 370px wide) the "Images" tab wraps, reducing the margin from 24px would help
Screen Shot 2021-03-02 at 4 26 48 PM
.

jekyll/assets/css/navbar.css
```diff
@@ -26,7 +26,6 @@
   .nav-toggle.two .widescreen a.logo {
     position: absolute;
     top: 18px;
-    left: 0;
```
janedotbiz, line 29: The left nav is now out of alignment (too far left) with the search bar and logo:
Screen Shot 2021-03-02 at 4 25 54 PM

janedotbiz (changes requested):

tfe:
Something isn't right here in general, because you shouldn't be seeing the mobile version of the sidebar nav. I also can't reproduce the left nav alignment thing.

Can you try clearing the jekyll build cache (`jekyll reset` I think), rebuild, and look again?

tfe:
OK I take some of that back, I am getting the mobile sidebar appearing on desktop now behind the desktop version. I'll get that fixed up.

tfe:
OK, I resolved that issue and the rest should clear up with a cache reset and rebuild on your end.

janedotbiz (approved):

PR #5122, opened 2021-03-03
Correct version of Python in example config

ogii:
# Description
Corrected the version of python specified in the compiled config snippet to match that of the parameter being passed in
3.8 -> 2.7

# Reasons
Having two conflicting version numbers in the config would lead to people getting confused when reading the docs, and perhaps reuse that incorrect config for their builds.


ganezasan (approved):
LGTM 🎉 
It's an obvious mistake 🤣

PR #5125, opened 2021-03-05
Insert a linebreak inbetween numbered footnotes

makotom:
Additional touches in regard to #5124: Inserting a line break for better legibility of footnotes.


kurumai (approved):
LGTM

PR #5138, opened 2021-03-09
[CIRCLE-30998] Touch up contexts documentation

zalimeni:
So excited to have these docs available! 

Noticed a few opportunities for clarification/correcting terminology, and reducing duplication w/ our API docs. Also fixed up some formatting for consistency.

# Description
- Remove unnecessary backslashes
- Remove references to supported languages (those are really examples; we support any valid HTTP client)
- Normalize and clarify terminology
- Make formatting more consistent
- Defer to API docs for response schema/interpretation rather than duplicating it

### DIFF
--- a/jekyll/_cci2/contexts.md
+++ b/jekyll/_cci2/contexts.md
@@ -224,9 +224,9 @@ Environment variables declared inside a shell command `run step`, for example `F
 
 ### Secure Environment Variable Creation, Deletion, and Rotation
 
-This section will walk through interacting with environment variables using the CircleCI CLI or API.
+This section will walk through interacting with context environment variables using the CircleCI CLI or API.
 
-#### Creating Environment	Variables
+#### Creating Environment Variables
 
 ##### Using CircleCI’s CLI
 {:.no_toc}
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/contexts.md
```diff
@@ -237,121 +237,110 @@ to set up your CircleCI command line interface._
...
-
-1. Call the [Add Environment Variable](https://circleci.com/docs/api/v2/#operation/addEnvironmentVariableToContext)
-   endpoint with the appropriate request body in one of our supported languages.
-   For this request, replace the `context-id` and the `env-var-name` with the ID
-   for the context and the new environment variable name. The request body
-   should include a `value` key with a string value.
-   
-2. If the request is successful, the response will include the name of the
-   environment variable that has been updated. An unsuccessful response will
-   include a clear error message explaining the error reported.
```
zalimeni, line 258: IMO it's best to defer to API docs for this, both to keep these shorter and to avoid falling out of sync. I held onto the request schema for ease of use, but we might consider dropping that part too for the same reason.

jekyll/_cci2/contexts.md
```diff
@@ -237,121 +237,110 @@ to set up your CircleCI command line interface._
...
+To create an environment variable using the API, call the [Add Environment Variable](https://circleci.com/docs/api/v2/#operation/addEnvironmentVariableToContext)
+endpoint with the appropriate request body. For this request, replace the 
+`context-id` and the `env-var-name` with the ID for the context and the new 
+environment variable name. The request body should include a `value` key 
+containing the plaintext secret as a string.
 
-#### Deleting Context Environment Variables
+#### Deleting Environment Variables
```
zalimeni, line 257: Some of the titles excluded this, and it seemed unnecessary given the doc focuses on contexts (as opposed to the older/less flexible Project Env Var feature)

jekyll/_cci2/contexts.md
```diff
@@ -237,121 +237,110 @@ to set up your CircleCI command line interface._
 
 To create an environment variable using our CLI, perform the following steps:
 
-1. If you have not already done so, find the right context name that will contain the new environment variable. Execute this command in the CLI: \
+1. If you have not already done so, find the right context name that will contain the new environment variable. Execute this command in the CLI:
 `circleci context list <vcs-type> <org-name>`
-2. Store a new environment variable under that context. Execute this command in the CLI: \
-`circleci context store-secret <vcs-type> <org-name> <context-name> <secret name>`
+2. Store a new environment variable under that context. Execute this command in the CLI:
+`circleci context store-secret <vcs-type> <org-name> <context-name> <env-var-name>`
```
zalimeni, line 243: We used `env-var-name` in multiple places, so it seemed best to be consistent even though the term "secret" is more general. Right now we only support env vars for context contents (i.e. files, etc. are not a feature), so "env var" is maybe the better to lean into for now.

jekyll/_cci2/contexts.md
```diff
@@ -237,121 +237,110 @@ to set up your CircleCI command line interface._
...
 
 Because environment variables can be shared, passed around between employees and
 teams, and exposed inadvertently, it is always good practice to periodically
 rotate secrets. Many organizations automate this process, running a script when
-an employee leaves the company or when a secret has been considered leaked.
+an employee leaves the company or when a secret may have been compromised.
 
 Context environment variables can be rotated using CircleCI’s CLI, or by
-directly accessing our API.
+directly accessing the API.
```
zalimeni, line 294: Just making "our" vs. "the" consistent (IME we use "the" more often in product docs)

jekyll/_cci2/contexts.md
```diff
@@ -237,121 +237,110 @@ to set up your CircleCI command line interface._
...
+and the `env-var-name` with the ID for the context and the environment variable name 
+that should be updated. The request body should include a `value` key containing the 
+plaintext secret as a string.
 
 
 ## Secrets masking
 _Secrets masking is not currently available on self-hosted installations of CircleCI Server_
 
-Contexts hold project secrets or keys that perform crucial functions for your applications. For added security CircleCI performs secret masking on the build output, obscuring the `echo` or `print` output of contexts.
+Contexts hold potentially sensitive secrets that are not intended to be exposed. For added security,
```
zalimeni, line 334: I felt it was more accurate to focus on why we mask here, rather than the (not guaranteed) likelihood that secrets are critical to applications (esp. in CI without CD, that may not be the case)

jekyll/_cci2/contexts.md
```diff
@@ -237,121 +237,110 @@ to set up your CircleCI command line interface._
...
 
-#### Rotating Context Environment Variables
+For this request, replace the `context-id` and the `env-var-name` with the ID 
+   for the context and the environment variable name that should be updated.
 
+#### Rotating Environment Variables
 
-Rotation occurs when an old API token is replaced with a new token. 
+Rotation refers to the process of updating a secret's value without deleting it 
+or changing its name.
```
zalimeni, line 286: We should be careful to use the term "secret" rather than "token" when referring to our env vars; they often contain tokens, but not always, so that could be confusing to users - esp. bc the more common use of that term in our case is to describe CircleCI API tokens and VCS OAuth tokens.

jekyll/_cci2/contexts.md
```diff
@@ -237,121 +237,110 @@ to set up your CircleCI command line interface._
...
+3. Delete the environment variable by executing this command:
+   `circleci context remove-secret <vcs-type> <org-name> <context-name> <env-var-name>`
 
-2.  If the request is successful, the response will include a message object
-    confirming the variable was successfully deleted. An unsuccessful response
-    will include a clear error message explaining the error reported.
+##### Using CircleCI’s API
+{:.no_toc}
 
+To delete an environment variable using the API, call the [Delete Environment Variable](https://circleci.com/docs/api/v2/#operation/addEnvironmentVariableToContext) endpoint.
```
zalimeni, line 278: This diff got really weird given it was mostly reformatting 🤨

jekyll/_cci2/contexts.md
```diff
@@ -237,121 +237,110 @@ to set up your CircleCI command line interface._
...
 
-Contexts hold project secrets or keys that perform crucial functions for your applications. For added security CircleCI performs secret masking on the build output, obscuring the `echo` or `print` output of contexts.
+Contexts hold potentially sensitive secrets that are not intended to be exposed. For added security, 
+CircleCI performs secret masking on the build output, obscuring `echo` or `print` output that 
+contains env var values.
 
 The value of the context will not be masked in the build output if:
 
-* the value of the context is less than 4 characters
-* the value of the context is equal to one of `true`, `True`, `false` or `False`
```
zalimeni, line 352: These are env var values _within_ the context, but in this case, it seemed fine to just remove the prepositional bit rather than updating

teesloane:
Wow, this is great @zalimeni ! Thanks for the clarifications and touch ups; I appreciate you adding line-specific comments for clarity. Looks good to me!

teesloane (approved):

PR #5140, opened 2021-03-09
add analytics eventName query param

joel-thompson:
# Description
Adds the ability to fire an analytic event when passing the `eventName` query param.

# Reasons
Part of the requirements for the ticket below is to have an event fire when clicking on the doc link from github. If we had the ability to fire an event with a query we could use that in the github actions CTA.
[CIRCLE-33157](https://circleci.atlassian.net/browse/CIRCLE-33157)

### DIFF
--- a/jekyll/assets/js/main.js
+++ b/jekyll/assets/js/main.js
@@ -56,7 +56,7 @@ var trackEvent = function (name, properties, options, callback) {
   });
 };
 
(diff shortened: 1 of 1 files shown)


teesloane:
@joel-thompson I just wanted to check in on this Draft PR. Is it ready for review?

PR #5141, opened 2021-03-10
Speed up `html-proofer`  test 

ganezasan:
# Description

- use `parallel` feature in `html-proofer` to speed up test
- use medium+ resouce_class in CircleCI
- export `html-proofer test` and `jekyll build` commands into Rakefile
- pack artifacts before uploading them into CircleCI artifacts

# Reasons

The `html-proofer` test generally takes 4 to 5 mins, so I customized it to use `parallel` feature in  `html-proofer` to speed up this test. Also, I changed circleci.yml to use `medium+` resource_class, and now the `html-proofer` test tasks about 1 min.

> https://app.circleci.com/pipelines/github/circleci/circleci-docs/27419/workflows/0309440a-a3fe-4bc0-8e5e-9667ab5002e6/jobs/75692/parallel-runs/0/steps/0-120

Also, the html-proofer and jekyll build commands were complexed. And it prevents running these commands locally.
So, I exported `html-proofer test` and `jekyll build` commands into Rakefile. 

```
$ bundle exec rake build
$ bundle exec rake test
```

PR #5146, opened 2021-03-12
Speedup jekyll build

ganezasan:
# Description
- bump ruby version to 2.7.2
- add `jekyll-include-cache` plugin
- clean classic-docs.html 
- remove `asciidoctor-pdf` from Gemfile, this is because the package doesn't relate to the Jekyll build

# Reasons

`jekyll build` used to take about 2.5 minutes, but with this change, `jekyll build` is now faster and only takes 1 min.
Also, there is a side-effect of the ruby version change, and `html-proofer` test is also getting faster and tasks about 12 seconds.

> https://app.circleci.com/pipelines/github/circleci/circleci-docs/27439/workflows/f9511119-794e-4a5f-949f-504406e1e918/jobs/75763

After the https://github.com/circleci/circleci-docs/pull/5141 has merged, I'll change the PR draft to normal.


ganezasan:
@teesloane Ruby 3.0 has changed `keyword argument` specification, and `jekyll-assets` isn't applicable to this change.
So we can't upgrade Ruby version to 3.0.0 at this moment.
> https://bugs.ruby-lang.org/issues/14183

Also, this is the latest preview, and as far as I checked, there is no problem.
>  http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/fix-speedup-jekyll-build-preview

Gemfile
```diff
@@ -1,20 +1,23 @@
...
 
-ruby '2.6.3'
+ruby '2.7.2'
 
-gem 'jekyll', "3.8.6"
+gem "jekyll", '~> 4.2.0', github: "jekyll/jekyll"
 gem 'html-proofer'
 gem "rack", ">= 2.0.6"
 gem 'asciidoctor'
-gem 'asciidoctor-pdf', '~> 1.5.3'
```
teesloane, line 9: I believe we need this as it's used by `scripts/build_pdfs` and `scripts/build_pdfs_asciidoc.sh` ?

teesloane (changes requested):

teesloane:
> @teesloane Ruby 3.0 has changed `keyword argument` specification, and `jekyll-assets` isn't applicable to this change.
> So we can't upgrade Ruby version to 3.0.0 at this moment.
> 
> > https://bugs.ruby-lang.org/issues/14183
> 
> Also, this is the latest preview, and as far as I checked, there is no problem.
> 
> > http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/fix-speedup-jekyll-build-preview

Thanks @ganezasan ! I left one comment.

Gemfile
ganezasan: @teesloane At this time, we don't need to use `asciidoctor-pdf` package as a local package.

Previously, the `scripts/build_pdfs` script uses `asciidoctor-pdf` of the local package. However, `scripts/build_pdfs_asciidoc.sh` expected to use `asciidoctor-pdf` of global package, and I updated `scripts/build_pdfs`  as the same way. `scripts/build_pdfs_asciidoc.sh` is expected to run on a circleci job, and the executor is `asciidoctor/docker-asciidoctor`, also  `asciidoctor-pdf` package has installed in its image.

Here is the change in the `scripts/build_pdfs`.
> https://github.com/circleci/circleci-docs/pull/5146/files#diff-a68ac2d75157c9152f1f6c45bc5b054eef76bd1cdc71500efdf56495e75e8397R14

So, to run these scripts locally, you can use the following command.

```
$ docker run -it --rm -v `pwd`:/tmp asciidoctor/docker-asciidoctor:1.1.0 bash -c 'cd /tmp && sh ./scripts/build_pdfs_asciidoc.sh'
```

teesloane (approved):

teesloane:
It looks as though our [re-index search](https://app.circleci.com/pipelines/github/circleci/circleci-docs/27551/workflows/b0217637-e6e3-45be-a062-759067edaff6/jobs/76186) failed, so I have reverted.

PR #5156, opened 2021-03-17
Docs Refresh: Table of contents and new footer

teesloane:
See [DES-592](https://circleci.atlassian.net/browse/DES-592).

[Preview link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/docs-refresh-preview/2.0/about-circleci/?section=getting-started)


tfe:
I can't work on this branch directly (no permissions to the repo), but I did do the work of bringing it up to date with current master over in this PR: #5378.

We can either merge that into this (by approving and merging that PR), or I can create a new PR for the design refresh that includes both this and that.

I did want to raise a question about the TOC approach though:

What was the reason for selecting the Liquid-based TOC plugin? It seems like the only reason that plugin exists and uses the approach of doing all the logic in the template is to work around limitations of running Jekyll on GitHub pages, a limitation which we don't have.

I would strongly suggest looking at something different that keeps the TOC generation logic in the Markdown parser (ideally) or Ruby realm. I found this one, for example: https://github.com/toshimaru/jekyll-toc.

teesloane:
Thanks for getting this branch up-to-date. 

Regarding the approach I took - I made some assumptions and could have done more research - I went with the first library I found that could separate the table of contents from the content  - it looks like using a Jekyll plugin makes more sense though (and means less code).

j2em6h:
With the side nav staying left aligned, the header/top nav can switch to using the `container-fluid` class to replace `container-semi-fluid`

PR #5180, opened 2021-03-23
arm faqs

thomasr8:
# Description
Initial draft of Arm FAQs page.

# Reasons


teesloane:
Content looks good but I'm wondering if we should integrate this with the `[arm-resource](https://circleci.com/docs/2.0/arm-resources/)` doc since there's not too much content either in the resources nor the FAQ yet (also, there is some overlap between the [limitations](https://circleci.com/docs/2.0/arm-resources/#limitations) and this FAQ. Anybody have any thoughts on this?

annabaker:
> Content looks good but I'm wondering if we should integrate this with the `[arm-resource](https://circleci.com/docs/2.0/arm-resources/)` doc since there's not too much content either in the resources nor the FAQ yet (also, there is some overlap between the [limitations](https://circleci.com/docs/2.0/arm-resources/#limitations) and this FAQ. Anybody have any thoughts on this?

@teesloane I agree -- I think it's probably cleaner to consolidate these sections into the same page for now, until they need broken down.

thomasr8:
@teesloane and @annabaker : Agree with both of you; the content is pretty limited at this point, so one page makes sense. @teesloane : would you like me to pull the content into the page you created, or would you prefer to do it?

teesloane:
> @teesloane and @annabaker : Agree with both of you; the content is pretty limited at this point, so one page makes sense. @teesloane : would you like me to pull the content into the page you created, or would you prefer to do it?

If you could that'd be great!

PR #5217, opened 2021-04-02
Clarify that you mean ARM CPUs and not Arms 💪 

ThisIsMissEm:
Unfortunately I received a link to this page on April Fools day, so for a sweet moment I was a little confused, before remembering that ARM is a type of computer CPU, and as it's an acronym it should be capitalised as ARM not written as Arm or arm.

# Description

See above

# Reasons
A link to a GitHub and/or JIRA issue (if applicable). n/a

Explained above, but it's ARM and not Arm.

### DIFF
--- a/jekyll/_cci2/arm-resources.md
+++ b/jekyll/_cci2/arm-resources.md
@@ -1,26 +1,26 @@
 ---
 layout: classic-docs
-title: "Arm resources"
-short-title: "Using Arm resources on CircleCI"
-description: "Using Arm resources on CircleCI"
+title: "ARM resources"
+short-title: "Using ARM resources on CircleCI"
+description: "Using ARM resources on CircleCI"
 version:
 - Cloud
 ---
 
 # Overview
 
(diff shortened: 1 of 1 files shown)


jerdog:
Did not mean to close this...

So a bit of backstory, "ARM" [rebranded themselves](https://en.m.wikipedia.org/wiki/Arm_Ltd.#Name) in 2017 as "Arm" or "arm". While the architecture does tend to be referred to as "ARM" still, we are using "Arm" as these are "Arm compute resource classes." In addition, Arm has taken to referring to their architecture as "[Arm64](https://developer.arm.com/solutions/infrastructure/developer-resources/ci-cd/circleci)". 

I don't know that we need to make a change here, other than _maybe_ to provide a link to the Arm Wikipedia article to allow for the developer to read more about Arm themselves.

thomasr8:
As "Arm" has become the industry standard for referring to the ARM architecture, I am going to close this for now. If we receive feedback that this is still confusing for users, we can make this change in the docs.

PR #5240, opened 2021-04-12
Update badge target to work with new UI

william-richard:
# Description
The branch page is now accessed with a query param instead of a URL path

# Reasons
These instructions are out of date, and create links that 404 if you follow them as written.


teesloane (approved):
Lovely, thanks!

PR #5258, opened 2021-04-20
Removing incorrect info

PK1312:
# Description
The server example for the docker machine executor section uses the image tag, but the image tag is not available on server. Removing.

# Reasons
The information was inaccurate.


teesloane:
Closing as the pipeline is not running. Work forwarded to https://github.com/circleci/circleci-docs/pull/5279

PR #5259, opened 2021-04-20
[WEB-131] Header Design Refresh

tfe:
**Ticket** [WEB-131](https://circleci.atlassian.net/browse/WEB-131)

Updated styles and functionality for header, including search and mobile nav.

The base markup and styles for this come from blog navigation (renamed from `blog-nav` to `global-nav` for use here), which is similar in design. Additional markup and styles are added on to support the unique elements of this design, like search.

Screen Shot 2021-04-20 at 14 51 33-fullpage

## How to test
http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/nav-refresh-preview/


teesloane:
Hi @tfe - Do you mind re-creating this brach with a `-preview` branch suffix (ex: `nav-refresh-preview`); it will auto-create a preview deploy for testing purposes.

src-js/instantsearch.js
```diff
@@ -214,8 +214,10 @@ export function init () {
   });
 
   window.addEventListener('load', renderResults);
-  searchResetButton.addEventListener('click', function () {
+  searchResetButton.addEventListener('click', function (e) {
```
j2em6h, line 217: I couldn't find where the visibility for the reset button was getting set, but the behaviour of it is a bit weird. 

The reset button not only clears the input, but also hides the results. On desktop the input is still visible, but on mobile the whole search experience is closed so if you intended to just clear the input you'd have to reopen search. On the other hand there is no way to close the search on mobile without putting something in the input (which causes the reset X to appear). There is no esc key on mobile. Perhaps there should be 2 buttons, one to clear the input (I'd recommend some small grey text that said clear on the right) and one to close the search, and hide the results. I'll leave a comment on the Jira ticket to see what the design team thinks.

jekyll/_sass/_sidebar.scss
```diff
@@ -15,8 +16,7 @@ $sidebar-spacing-base: 5px;
   height: 92vh;
   padding: 30px 24px 60px 15px;
   position: absolute;
-  bottom: 0;
```
j2em6h, line 18: Absolute position isn't going to have much effect in this context without the `bottom` value. 

On tablet sized screens, whatever was contributing to the proper position on desktop goes away and I'm seeing a gap under the side nav (also the top of the sidenav is under the top nav): 
Screen Shot 2021-04-26 at 11 28 23 AM

jekyll/_sass/_sidebar.scss
```diff
@@ -50,7 +50,6 @@ $sidebar-spacing-base: 5px;
 
   &.fixed {
     position: fixed;
-    top: 118px;
```
j2em6h, line 53: Fixed position won't have an impact without a `top` value.

jekyll/_sass/_sidebar.scss
```diff
@@ -15,8 +16,7 @@ $sidebar-spacing-base: 5px;
   height: 92vh;
```
j2em6h, line 16: This 92vh value is very approximate. 

Setting a `top` value equal to the height of the top nav, and `bottom: 0` instead of height would always be correct. (or `height: calc(100vh - #{$nav-height})` ... except the nav height isn't consistent across screensizes.

jekyll/_sass/_global-nav.scss
```diff
@@ -0,0 +1,409 @@
...
+    font-weight: 300;
+    letter-spacing: 0.34px;
+    line-height: 24px;
+    margin-left: 9px;
+  }
+}
+
+.global-nav--row {
+  &-mobile {
+    @media (max-width: $screen-sm-max) {
```
j2em6h, line 75: `$screen-sm-max` = ` 991px`, so this covers tablet, I'm not sure it makes sense for the "mobile" nav to be used for tablet on Docs, especially since the side-nav is still visible on tablet, and click the the menu button only gets you the "start building" button and the link to circleci.com... and actually covers up the side nav: 

Screen Shot 2021-04-26 at 11 33 06 AM

At 991 pixels at the smallest, I think the button and the link can still fit on screen in the nav itself, with no need for menu icon at this size. The button may need to be made narrower.

jekyll/_sass/_sidebar.scss
```diff
@@ -142,46 +141,18 @@ $sidebar-spacing-base: 5px;
 }
 
 .sidebar-mobile-wrapper {
-  border-top: 1px solid $gray-200;
-  border-bottom: 1px solid $gray-200;
```
j2em6h, line 146: This border-bottom created a visible end to the menu, without it this looks a bit odd with just the white slicing items in half. 
Screen Shot 2021-04-26 at 11 43 32 AM

The border on `.mobile-sidebar`, which presumably replaces this, isn't visible until you've scrolled to the bottom of the menu.

jekyll/_sass/_sidebar.scss
```diff
@@ -142,46 +141,18 @@ $sidebar-spacing-base: 5px;
...
-      float: right;
-      margin-top: 5px;
-      transform: rotate(90deg);
-    }
-  }
+  width: 100%;
 
   .mobile-sidebar {
     @extend .sidebar;
-    padding-right: 16px;
```
j2em6h, line 180: This padding-right I think was preventing the open arrows from overflowing, and creating a horizontal scrollbar.

Screen Shot 2021-04-26 at 11 43 32 AM

Screen Shot 2021-04-26 at 11 43 42 AM

j2em6h (changes requested):
A lot to look at here, this is what I noticed.

j2em6h:
I also just noticed the Jira ticket includes a requirement for the language switcher in the top nav.

jekyll/_sass/_global-nav.scss
```diff
@@ -0,0 +1,409 @@
...
+    text-decoration: none;
+    color: $green;
+  }
+
+  &.active {
+    font-weight: $font-weight-bold;
+  }
+}
+
+.global-nav--external-link {
```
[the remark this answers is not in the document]
tfe: Right, this is another case where this design diverged from established styles (blog navigation in this case), so I elected to preserve the established one.

Do you think we should have this one diverge from that, or update blog to match, or leave it as-is?

jekyll/_sass/_instantsearch.scss
```diff
@@ -60,7 +48,6 @@
   padding: $spacing-small-vertical 0;
   background: #FAFAFA;
   min-height: 500px;
-  margin-top: $spacing-medium-vertical;
```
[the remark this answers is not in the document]
tfe: Looking into this.

jekyll/_includes/hits-target.html
```diff
@@ -1,7 +1,7 @@
 <div id="hits-target" style="display:none;">
   <div class="container-semi-fluid">
     <div id="subnav-placeholder">
-      <div class="component subnav dynamic-fixed" data-top-nav-offset="120">
+      <div class="component subnav dynamic-fixed" data-top-nav-offset="68">
```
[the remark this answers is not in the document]
tfe: Looking into this. I'm going to try making the header element sticky instead of fixed so it just always takes up the correct amount of space in the layout flow.

src-js/instantsearch.js
tfe: I know that the instantsearch JS does some of this on its own in a way we can't influence, so our options may be somewhat limited. We can look into it though.

jekyll/_sass/_global-nav.scss
tfe: I think I ended up with this way because it's how the header behaves on the blog, and with the search bar the header elements were just too wide in total to fit at the tablet breakpoint.

jekyll/_sass/_global-nav.scss
```diff
@@ -0,0 +1,409 @@
...
+// Customizations/extensions of outer blog-nav
+///////////////////////////////////////////////////////////////////////////////
+
+.global-nav--component {
+
+  a.global-nav--lockup-link {
+    color: initial;
+  }
+
+  .btn-primary {
```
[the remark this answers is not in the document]
j2em6h: The only other `btn-` I see on the site outside of the nav is on this page about 1.0, which was sunset in 2018: https://circleci.com/docs/enterprise/single-box/

jekyll/_sass/_global-nav.scss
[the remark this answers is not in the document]
j2em6h: Let's cover this in our meeting with design, but I agree, I think the two should be consistent.

jekyll/_sass/_sidebar.scss
tfe: It does; it uses whatever the natural (statically-positioned) position is, which is what we want in this case.

jekyll/_sass/_sidebar.scss
tfe: This needs to be removed to match the removal of the `top` rule below. The gap is due to the poorly-specified `92vh` height specification you called out elsewhere. All three of these rules (`top`, `bottom`, and `height`) never made sense to all be present, since there's no way for them to all be mutually applied.

jekyll/_sass/_sidebar.scss
tfe: I agree this is a poor way to specify this element, but its existence never bit us before. I'll update it to the rule you suggested. I'd prefer this to specifying the top value trying to match the nav height, because if the height calculation gets off in the future it's less catastrophic than an incorrect top value.

jekyll/_sass/_global-nav.scss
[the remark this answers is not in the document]
j2em6h: DevHub also used the heavy weight, the blog should maybe be the one to change.

jekyll/_sass/_global-nav.scss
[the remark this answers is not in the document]
tfe: OK, I'll update the styles. Can you clarify what is needed for the `text-align` property? It sounds like it should be centered, however the button is `width: auto` so centering would be a no-op.

tfe:
@j2em6h This is close to ready for a re-review. The only outstanding items I'm aware of are:

1. Open question to Andrew about search cancel button on desktop (I hacked something temporary in for now; you'll probably see it).
2. Open question to Andrew about tablet styles (use collapsable nav or some condensed version of the desktop nav).
3. https://github.com/circleci/circleci-docs/pull/5259/files#r623539928

tfe:
@teesloane Could you please update the preview branch with the latest changes?

Aytac998 (approved):

jekyll/_sass/_global-nav.scss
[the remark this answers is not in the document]
j2em6h: I'm seeing a width of 250px, coming from `.btn-primary`, if it should be `auto` instead then you're right, though it is 250px wide in the design.

Screen Shot 2021-04-30 at 10 35 38 AM

Screen Shot 2021-04-30 at 10 36 02 AM

jekyll/_includes/global-nav.html
```diff
@@ -1,161 +1,63 @@
...
+        <div class="global-nav--footer">
+          <a
+            href="{{ '/signup/' | outer_url }}"
+            class="btn-primary visitor-item"
+            target="_blank"
+            data-analytics-action="click-outer-cta"
+          >
+            Start Building for Free
+          </a>
+          <a href="/dashboard" class="btn-primary dashboard-link">
```
teesloane, line 56: I believe this has to be `circleci.com/dahboard` (or `https://app.circleci.com/pipelines` )? as this will link to `/docs/dashboard`; which is what is causing the build to fail with the html-proofer gem.

jekyll/_includes/ja/global-nav.html
```diff
@@ -1,161 +1,63 @@
...
+        <div class="global-nav--footer">
+          <a
+            href="{{ '/signup/' | outer_url }}"
+            class="btn-primary visitor-item"
+            target="_blank"
+            data-analytics-action="click-outer-cta"
+          >
+            無料でビルドを開始する
+          </a>
+          <a href="/dashboard" class="btn-primary dashboard-link">
```
teesloane, line 56: I believe this has to be `circleci.com/dahboard` (or `https://app.circleci.com/pipelines` )? as this will link to `/docs/dashboard`; which is what is causing the build to fail with the html-proofer gem.

teesloane (changes requested):

teesloane:
> @teesloane Could you please update the preview branch with the latest changes?

Until the build passes the preview cannot be deployed. I think I found the cause of the break and have posted the potential fix.

Aytac998 (approved):

jekyll/_includes/global-nav.html
j2em6h: I updated this to `https://app.circleci.com/dashboard` on the preview branch

jekyll/_includes/ja/global-nav.html
j2em6h: I updated this to `https://app.circleci.com/dashboard` on the preview branch

tfe:
@j2em6h I am guessing the issues you saw were due to an outdated or cached CSS file on the preview build, because there weren't any missing changes and it's working fine now.

I pulled in the dashboard link fix from the preview branch which should fix the build here. I need to tweak the desktop search cancel button design per Andrew's most recent feedback, and then this should be ready to go.

j2em6h (approved):
Looks good to me! Thanks Todd.

tfe:
Looks like @teesloane needs to re-review before this can be merged.

teesloane (approved):

PR #5266, opened 2021-04-22
Rename workflows map to workflows graph

erichaberkorn:
# Description
Switch "workflows map" to "workflows graph" everywhere.

# Reasons
Before this commit both workflows graph and workflows map were being used. Since graph is technically correct, we will use that terminology.


gabmass (approved):
🎉

teesloane (approved):

PR #5269, opened 2021-04-23
Swap auth header to circle token. 

teesloane:
See [pertinent conversation](https://circleci.slack.com/archives/C0KBNJGRH/p1602187242068600).

[Preview link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/swap-auth-header-preview/api/v2)

### DIFF
--- a/scripts/build_api_docs.sh
+++ b/scripts/build_api_docs.sh
@@ -31,7 +31,7 @@ build_api_v2() {
     echo "Merging in JSON patches to correct and augment the OpenAPI spec."
     jq -s '.[0] * .[1]' openapi-with-examples.json openapi-patch.json > openapi-final.json
     echo "Swapping code samples to use the circle token header."
-    sed -i '.bak' 's/authorization/Circle-Token/g' openapi-final.json; sed -i '.bak' 's/Basic REPLACE_BASIC_AUTH/:CIRCLECI_TOKEN/g' openapi-final.json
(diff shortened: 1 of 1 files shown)


teesloane:
Just to follow this up; the original plan was to `sed` replace the api json output before it's converted into a redoc site. Unfortunately, this is a bit brittle as it simply replaces the use of the word "authoriziation" to "Circle-Token" as a header across the file. 

Instead, we should try and find a way to do this based off the content in the spec, or more likely, to see if we can customize `snippet-enricher-cli` to output different authentication headers. 


```
          {
            "lang": "Node + Request",
            "source": "const request = require('request');\n\nconst options = {\n  method: 'POST',\n  url: 'https://circleci.com/api/v2/context',\n  headers: {'content-type': 'application/json', Circle-Token: ':CIRCLECI_TOKEN'},\n  body: {name: 'string', owner: {id: 'string', type: 'organization'}},\n  json: true\n};\n\nrequest(options, function (error, response, body) {\n  if (error) throw new Error(error);\n\n  console.log(body);\n});\n"
          },
          {
            "lang": "Python + Python3",
            "source": "import http.client\n\nconn = http.client.HTTPSConnection(\"circleci.com\")\n\npayload = \"{\\\"name\\\":\\\"string\\\",\\\"owner\\\":{\\\"id\\\":\\\"string\\\",\\\"type\\\":\\\"organization\\\"}}\"\n\nheaders = {\n    'content-type': \"application/json\",\n    'Circle-Token': \":CIRCLECI_TOKEN\"\n    }\n\nconn.request(\"POST\", \"/api/v2/context\", payload, headers)\n\nres = conn.getresponse()\ndata = res.read()\n\nprint(data.decode(\"utf-8\"))"
          },

PR #5271, opened 2021-04-26
Add: note on accept headers.

### LINKED ISSUE #5159: Clarify usage of Accept header for API requests
# Missing Information
We received a report about an issue when  interacting with our API in [this forum post](https://discuss.circleci.com/t/api-v2-malformed-json-bodies-returned-on-400s/39393). The heart of the issue is that when users make API calls the response content-type header is set to text, but all the APIs return JSON. This can cause some tools to break when the content type doesn't match the content.

After discussion with engineering it looks like that behavior is expected, and users should add the `Accept: application/json` header to their requests for the content-type header to be set correctly on the response.

It looks like [this information was included in an overview section](https://circleci.com/docs/api/v1/#accept-header) of the v1 docs. But there's no similar overview section [for our v2 docs](https://circleci.com/docs/api/v2/), which is where I originally looked when trying to learn more about this. I didn't notice it in the v1 docs until I did some searching on the repo here. I see that we do include the Accept header usage within the examples [in our current developer API guide](https://circleci.com/docs/2.0/api-developers-guide/), but there's no similar explanation to the significant of it as there is in the v1 docs.

I'm just raising this as an issue for now as i'm not sure the best way to approach this. But my initial thoughts are that it would be good if there was an "overview" section above the "authentication" section in the v2 docs, to be more inline with our v1 docs with providing this information. As well as including it in the current developer API guide. Since both of these are more current documentation points users are likely to encounter.

Happy to submit an initial PR with those changes if it would be helpful.

# Already Looked
https://circleci.com/docs/api/v2/

# Potential Locations
https://circleci.com/docs/api/v2/
https://circleci.com/docs/2.0/api-developers-guide/

teesloane:
Fixes #5159

### DIFF
--- a/jekyll/_cci2/api-developers-guide.md
+++ b/jekyll/_cci2/api-developers-guide.md
@@ -67,7 +67,7 @@ To add an API token, perform the steps listed below.
 {: #accept-header }
 
 It is recommended to specify an accept header in your API requests. The majority
-of API end points will return JSON by default, but some requests (primarily API
+of API endpoints will return JSON by default, but some endpoints (primarily API
 v1) return EDN if no accept header is specified.
 
 To return formatted JSON, include a `text/plain` header like so:
--- a/jekyll/_cci2/api-developers-guide.md
+++ b/jekyll/_cci2/api-developers-guide.md
@@ -60,7 +60,31 @@ To add an API token, perform the steps listed below.
     }
     ```
 
+
 **Note:** All API calls are made in the same way, by making standard HTTP calls, using JSON, a content-type, and your API token. Please note that the JSON examples shown in this document are not comprehensive and may contain additional JSON response fields not shown in the example, based on user input and fields.
+
+### Accept Header
+{: #accept-header }
+
+It is recommended to specify an accept header in your API requests. The majority
+of API end points will return JSON by default, but some requests (primarily API
+v1) return EDN if no accept header is specified.
+
+To return formatted JSON, include a `text/plain` header like so:
+
+```sh
+curl --header "Circle-Token: $CIRCLECI_TOKEN" \
+  --header 'Accept: text/plain'    \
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/api-developers-guide.md
```diff
@@ -60,8 +60,32 @@ To add an API token, perform the steps listed below.
...
 **Note:** All API calls are made in the same way, by making standard HTTP calls, using JSON, a content-type, and your API token. Please note that the JSON examples shown in this document are not comprehensive and may contain additional JSON response fields not shown in the example, based on user input and fields.
 
+### Accept Header
+{: #accept-header }
+
+It is recommended to specify an accept header in your API requests. The majority
+of API endpoints will return JSON by default, but some endpoints (primarily API
+v1) return EDN if no accept header is specified.
+
+To return formatted JSON, include a `text/plain` header like so:
```
thomasr8, line 73:
```suggestion
To return formatted JSON, include a `text/plain` header like the example shown below:
```

jekyll/_cci2/api-developers-guide.md
```diff
@@ -60,8 +60,32 @@ To add an API token, perform the steps listed below.
     }
     ```
 
+
 **Note:** All API calls are made in the same way, by making standard HTTP calls, using JSON, a content-type, and your API token. Please note that the JSON examples shown in this document are not comprehensive and may contain additional JSON response fields not shown in the example, based on user input and fields.
 
+### Accept Header
+{: #accept-header }
+
+It is recommended to specify an accept header in your API requests. The majority
```
thomasr8, line 69:
```suggestion
It is recommended that you specify an accept header in your API requests. The majority
```

thomasr8 (approved):
Just a few minor comments

PR #5277, opened 2021-04-27
CIRCLE-32157 Add documentation for the matches logic element

sulami:
This is a new config logic element which performs regular expression
matching and can be used e.g. to check a branch name.

Convenient link: https://circleci.atlassian.net/browse/CIRCLE-32157


thomasr8 (approved):

sulami:
_Keeping this until we actually roll this out._

PR #5294, opened 2021-04-29
Add Linux 2021 Q2 images

felicianotech:
Adding the new 2021 Q2 Linux images. Updated the note about 16.04 as this is the last release.


teesloane (approved):
Thanks @felicianotech

PR #5297, opened 2021-04-29
Update docs

miketwo:
Distinguish between Trigger Filters and Job Filters.

# Description
Adds descriptive wording and changes the anchor link

# Reasons
It is confusing to have two sections on a page with identical headers that apply to different parts of the configuration.


teesloane (approved):
Thanks for catching this!

PR #5318, opened 2021-05-06
[TODO JIRA LINK] Remove missing fields from audit log schema docs

zalimeni:
Some of the fields currently documented are not actually included today in our audit logs. Their addition is currently under review w/ SecOps, but for the time being, we should remove these fields from our public docs.

**Note: https://github.com/circleci/circleci-docs/blob/ae71de285dc72df4dd184fbb47e7a665e6904648/jekyll/_cci2_ja/security.md will also require an update, but it wasn't easy for me to figure out which part of the text to delete to make a clean edit.**

# Reasons
_TODO: Add Jira ticket link_


bear (approved):

zalimeni:
Squashed commits and reverted "JSON object" to "JSON blob" to match preceding bullets.

PR #5330, opened 2021-05-10
New Document - Protecting Supply Chain

teesloane:


taxonomic-blackfish:
Looks good. I'd suggest we link to it from the main sec docs too somehow. Maybe also have it appear in the nav sidebar? Not sure if it dos already

taxonomic-blackfish (approved):

teesloane:
Sounds good.  We should perhaps setup linking it in a separate PR as we are waiting on some landing page and sidebar changes and then at least this will be live. In the meantime, I will queue this up to be translated.

PR #5331, opened 2021-05-11
Update IRC Orb name in notifications.md

rfc2119:
A tiny modification in the Orb name used by IRC notifications


teesloane (approved):
Thanks for the fix!

PR #5354, opened 2021-05-21
Add additional documentation for LOS

robinsturmeit:
# Description
Adds additional documentation on how to use the Local Observability
Stack and its dashboards.

# Reasons
SERVER-412


robinsturmeit:
> @robinsturmeit I imagine @annabaker might be better suited to review the content of this review, but can you remove the `package-lock.json` changes (if deps need to change it should happen in another PR). Thanks!

It was just the lock file, but sure, removed it from the PR

nathanfishcircleci:
So Grafana includes data you would find in telegraf as well as other interesting data like CPU, etc? Telegraf is just statsd data from the services? Anyway we can make that more explicit in the documentation?

robinsturmeit:
> So Grafana includes data you would find in telegraf as well as other interesting data like CPU, etc? Telegraf is just statsd data from the services? Anyway we can make that more explicit in the documentation?

Better now?

nathanfishcircleci (approved):
Looks good.

nathanfishcircleci:
@annabaker i think once you approve this can be published.

annabaker:
Closing -- moved this work over to #5362 (@robinsturmeit thank you so much! If you append '-preview' to the branch name, it will generate the preview site which helps me ensure there are no formatting errors prior to merging, etc.)

PR #5374, opened 2021-05-26
SERVER 1077 | S3 compatible backup and restore documentation

nsmith5:
# Description

Adds documentation on how to set up backup and restore on S3 compatible storage providers.

# Reasons

Jira ticket: https://circleci.atlassian.net/browse/SERVER-1077

We've recently tested and delivered backup and restore for S3 compatible storage. This documents how to use it for customers.

### DIFF
--- a/jekyll/_cci2/server-3-operator-backup-and-restore.adoc
+++ b/jekyll/_cci2/server-3-operator-backup-and-restore.adoc
@@ -48,6 +48,11 @@ For more information, see Velero's https://velero.io/docs/v1.6/supported-provide
 
 Below, you will find instructions for creating a server 3.x backup on AWS and GCP.
 
+==== S3 Compatible Storage Prerequisites
+
+- https://docs.min.io/docs/minio-client-quickstart-guide.html[minio CLI] is
+  installed and configured for you your storage provider.
+
 ////
 
 * AWS SETUP *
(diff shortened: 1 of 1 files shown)


Williamlliam (approved):

zawzawaung-prog:
---
layout: enterprise
section: enterprise
title: "Backing up CircleCI Data"
category: [advanced-config]
order: 8
description: "How to regularly back up CircleCI Enterprise"
sitemap: false
---

This document describes how to back up your CircleCI Enterprise application so that you can recover from accidental or unexpected loss of CircleCI data attached to the Services machine:

* TOC 
{:toc}

**Note:** If you are running CircleCI in an HA configuration, you must use standard backup mechanisms for the external datastores. See the [High Availability]({{site.baseurl}}/enterprise/high-availability/) document for more information.

## Backing up the Database

If you have **not** configured CircleCI for HA, the best practice for backing up your CircleCI data is to use VM snapshots of the virtual disk acting as the root volume for the Services machine. Backups may be performed without downtime as long the underlying virtual disk supports such an operation as is true with AWS EBS. There is a small risk, that varies by filesystem and distribution, that snapshots taken without a reboot may have some data corruption, but this is rare in practice. If zero downtime backups and robustness in the face of data corruption issues are required, then an [HA configuration]({{site.baseurl}}/enterprise/high-availability/) may be the best solution.

## Backing up Object Storage

Build artifacts, output, and caches are generally stored in object storage services like AWS S3. These services are considered highly redundant and are unlikely to require separate backup. An exception is if your instance is setup to store large objects locally on the Services machine, either directly on-disk or on an NFS volume. In this case, you must separately back these files up and ensure they are mounted back to the same location on restore.

## Snapshotting on AWS EBS

There are a few features of AWS EBS snapshots that make the backup process quite easy:

1. To take a manual backup, choose the instance in the EC2 console and select Actions > Image > Create Image. 

2. Select the No reboot option if you want to avoid downtime. 
An AMI that can be readily launched as a new EC2 instance for restore purposes is created. 

It is also possible to automate this process with the AWS API.  Subsequent AMIs/snapshots are only as large as the difference (changed blocks) since the last snapshot, such that storage costs are not necessarily larger for more frequent snapshots, see [Amazon's EBS snapshot billing](https://aws.amazon.com/premiumsupport/knowledge-center/ebs-snapshot-billing/) document for details.

## Restoring From Backup

When restoring testing backups or performing a restore in production, you may need to make a couple of changes on the newly launched instance if its public or private IP addresses have changed:

1. Launch a new VM using a previous root-volume backup.
2. Ensure that the hostname configured in the settings page at port 8800 reflects the correct address. If this hostname has changed, you will also need to change it in the corresponding GitHub OAuth application settings or create a new OAuth app to test the recovery and log in to the application.
3. Update any references to the backed-up instance's public and private IP addresses in `/etc/default/replicated` and `/etc/default/replicated-operator` on Debian/Ubuntu or `/etc/sysconfig/*` in RHEL/CentOS to the new IP addresses.

## Cleaning up Build Records

While filesystem-level data integrity issues are rare and preventable, there will likely be some data anomalies in a point-in-time backup taken while builds are running on the system. For example, a build that is only half-way finished at backup time may result in missing the latter half of its command output, and it may permanently show that it is in "Running" state in the application.

If you want to clean up any abnormal build records in your database after a recovery, you can delete them by running the following commands on the Services machine replacing the example build URL with an actual URL from your CircleCI application:

```
$ circleci dev-console
# Wait for console to load
user=> (admin/delete-build "https://my-circleci-hostname.com/gh/your-org/my-project/1234") 
```

jekyll/_cci2/server-3-operator-backup-and-restore.adoc
```diff
@@ -366,6 +371,121 @@ velero-68788b675c-dm2s7   1/1     Running   0          2m
...
+
+These instructions were sourced from the Velero documentation https://velero.io/docs/v1.6/contributions/minio/[here].
+
+=== Step 1 - Configure `mc` client
+
+To start, configure `mc` to connect to your storage provider:
+
+[source,bash]
+----
+mc alias set my-provider <YOUR_ENDPOINT> <YOUR_ACCESS_KEY_ID> <YOUR_SECRET_ACCESS_KEY>
```
robinsturmeit, line 393: nit: It might be useful to assign `my-provider` to a variable and use this instead for the rest of the examples

robinsturmeit (approved):

jekyll/_cci2/server-3-operator-backup-and-restore.adoc
nsmith5: Alright, tried to express that with a variable. Limme know what you think!

robinsturmeit: Yup, that's better. You could have still called it `$PROVIDER` or something and also used vars in other places where parameters repeat themselves (e.g. `$BUCKET`), but we can definitely leave it like this :)

annabaker:
Thank you @nsmith5! I have opened #5394 which is a polish of this (with some branch renaming conventions in place so we can generate the pdf/preview site). Please see my comment there, otherwise everything looks good, thanks again.

PR #5390, opened 2021-05-31
[SERVER-1137] - remove references to the frontend loadbalancer.

soulchips:
# Description
Server 3.1 no longer uses the frontend load balancer. Instead we use traefik to route api requests to frontend. The benefit is one less loadbalancer but our docs need to reflect this change

# Reasons
[SERVER-1137](https://circleci.atlassian.net/browse/SERVER-1137)


soulchips:
Before this can be called complete, the image `server-3-architecture-diagram.png` needs to be updated.

robinsturmeit (approved):
Thank you. Some things that caught my attention (not your fault but pre-existing issues): 
* Spelling of 'frontend' is inconsistent
* The old "Server" vs "server" problem
You think you could fix that while you're at it?

robinsturmeit:
Ahem, sorry, I should have been more clear: "server" is actually the correct way 🙈

soulchips:
> Ahem, sorry, I should have been more clear: "server" is actually the correct way 🙈

Is that correct? it seems to be `Server` commonly throughout the other docs. Maybe we should park this for another ticket while we clear it up.

annabaker (approved):
Looks good, thank you, adding some polishes and adding this to the 3.1 branch.

annabaker:
> > Ahem, sorry, I should have been more clear: "server" is actually the correct way see_no_evil
> 
> Is that correct? it seems to be `Server` commonly throughout the other docs. Maybe we should park this for another ticket while we clear it up.

Yes, unfortunately a lot of this was written when I'd just started and I didn't become aware it was 's' until later :p I took care of it and it's in the 3.1 branch now

PR #5391, opened 2021-05-31
SERVER-1081 | Edit object storage documentation for server 3

nsmith5:
### Description

S3 object storage is now slightly more general as we support S3
compatible storage like Minio. Also edited the GCS object storage docs
to make them similar to the S3 docs

*Open Questions*:
- This documentation change is accurate for server 3.1 but not for 3.0. Do we have versioned documentation to gate this? I'm not sure if I need to document both versions here or if we'll simply release this documentation update once 3.1 lands?
- Do we need a little tutorial or something on how to setup a basic Minio installation? We don't really do that for how to set up an S3 bucket, but we do for setting up a Kubernetes cluster. This is about between the two in terms of complexity.. 

### Reasons
SERVER-1081


nsmith5:
Closing in favour of #5397

PR #5397, opened 2021-06-01
SERVER-1081 | Edit object storage documentation for server 3

nsmith5:
# Description
S3 object storage is now slighly more general as we support S3
compatible storage like Minio. Also edited the GCS object storage docs
to make them similar to the S3 docs

# Reasons
SERVER-1081


nsmith5:
Closing in favour of #5399

PR #5405, opened 2021-06-03
raising min terraform version for server 3.x to v0.15.4

soulchips:
# Description
As of server 3.1, the minimum required version for terraform to deploy nomad will be v0.15.4

# Reasons
https://circleci.atlassian.net/browse/SERVER-1164


robinsturmeit:
You will need to update `terraformversion` in `/jekyll/_config.yml`, too

soulchips:
> You will need to update `terraformversion` in `/jekyll/_config.yml`, too

ah thanks. I saw that but it was already older than what we had before so I wasn't sure if we should change it. What does this doc do?

robinsturmeit:
> > You will need to update `terraformversion` in `/jekyll/_config.yml`, too
> 
> ah thanks. I saw that but it was already older than what we had before so I wasn't sure if we should change it. What does this doc do?

It holds some global variables as well as defines some rendering stuff, I believe

robinsturmeit (approved):

annabaker (approved):

PR #5416, opened 2021-06-04
Update caching instructions for yarn 2

brandon-leapyear:
# Description
Updated the instructions to mention Zero Installs, and fixed the example for non-Zero Installs.

# Reasons
It seems like the caching instructions for yarn 2 are outdated. The caching instructions are also completely unneeded if you're using Zero Installs.

### DIFF
--- a/jekyll/_cci2/yarn.md
+++ b/jekyll/_cci2/yarn.md
@@ -33,7 +33,9 @@ curl -o- -L https://yarnpkg.com/install.sh | bash
 
 Yarn packages can be cached to improve CI build times.
 
-An example for Yarn 2:
+Yarn 2.x added the ability to do [Zero Installs](https://yarnpkg.com/features/zero-installs); if you're using Zero Installs, you shouldn't need to do any special caching.
+
+If you're using Yarn 2.x without Zero Installs, you can do something like this:
 
 {% raw %}
 ```yaml
@@ -49,7 +51,8 @@ An example for Yarn 2:
           name: Save Yarn Package Cache
           key: yarn-packages-{{ checksum "yarn.lock" }}
           paths:
-            - ~/.cache/yarn
+            - .yarn/cache
+            - .yarn/unplugged
 #...
 ```
 {% endraw %}


teesloane:
Hi @brandon-leapyear  - thanks for the PR. This is cool - I didn't know about Yarn Zero installs. If I understand this correctly:

> Yes, very much. To give you an idea, a node_modules folder of 135k uncompressed files (for a total of 1.2GB) gives a Yarn cache of 2k binary archives (for a total of 139MB). Git simply cannot support the former, while the latter is perfectly fine.

Yarn is creating a compressed "binary archive" and that gets checked in? If so, the first change makes sense.

The second [change proposed](https://github.com/circleci/circleci-docs/pull/5416/files#diff-d0b151db5c9fa421f98c1ea05a9fdefb64da10eead38a8e8cdac3f8313798e33R54-R55) indicates that you are setting the cache to within the project folder? If this is the case, I think the operation of cloning the repo is not really taking advantage of restoring a cache from the CircleCI cache which would live outside of your repo, and would be invalidated by a checksum, such as one based on a yarn lockfile. Let me know if this sounds incorrect - I haven't used yarn for a little while.  Thanks!

brandon-leapyear:
I'm not sure I understand your concern about the second change. If you're not using Zero Installs, you should be gitignoring the `.yarn/cache` and `.yarn/unplugged` directories. Those directories are the equivalent of a `node_modules` directory living within the project folder

teesloane:
I was referring of the switch from using a cache folder under the global user dir (`~/.cache`) to using a  single folder in the repo root.   I was thinking that if there were multiple yarn lock files (ie a project with multiple sub folders that all needed to run yarn install, that the user might want to leverage cached deps in a single place). Again, I'm not using Yarn these days, but if this is off, please let me know (and any supplementary docs would be super helpful).

brandon-leapyear:
I see. I can't find many docs, but my current understanding is this:

* Yarn 1 used to put some caches into `~/.cache`
* Yarn 1 would actually install the third party libraries to local `node_modules` directories -- typically, one directory per `package.json`

So actually, the Yarn 1.x docs aren't completely comprehensive. It should also be caching the `node_modules` folders so that latter CI runs don't have to redownload node deps.

* Yarn 2 doesn't use `~/.cache` anymore; since I don't know what Yarn 1 stored in `~/.cache`, I'm not sure where it is now, but I suspect the `.yarn/build-state.yml` and `.yarn/install-state.gz` files were the files that used to be stored in `~/.cache/yarn`
* Instead of using `node_modules` directories, Yarn 2 now stores all dependencies in `.yarn/cache/` and `.yarn/unplugged/`.

Yarn has always had exactly one `yarn.lock` file. Typically, if you have multiple subprojects, you'd have one top-level workspace for all the projects.

https://yarnpkg.com/configuration/yarnrc/#enableGlobalCache
https://yarnpkg.com/configuration/yarnrc/#cacheFolder

teesloane:
Excellent - thanks for following up and adding some documentation for reference. I will merge this in. Thanks for your contribution!

teesloane (approved):

PR #5436, opened 2021-06-15
Add clarification around prefix matching for conditional logic match

nbialostosky:
# Description
This PR will add information around the `match` conditional logic. We don't automatically prefix match, so a full pattern needs to be provided.

# Reasons
We had a customer open a ticket on this situation and the docs could be a bit more clear on the subject.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -1830,7 +1830,7 @@ argument specifications, and to a maximum depth of 100 levels.
 `matches` uses [Java regular
 expressions](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)
 for its `pattern`. It is recommended to enclose a pattern in `^` and
-`$` to avoid accidental partial matches.
+`$` to avoid accidental partial matches. However, by default we will not prefix match, a full pattern needs to be provided.
 
 **Note:**
 When using logic statements at the workflow level, do not include the `condition:` key (the `condition` key is only needed for `job` level logic statements).


teesloane (approved):

jekyll/_cci2/configuration-reference.md
```diff
@@ -1830,7 +1830,7 @@ argument specifications, and to a maximum depth of 100 levels.
 `matches` uses [Java regular
 expressions](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)
 for its `pattern`. It is recommended to enclose a pattern in `^` and
-`$` to avoid accidental partial matches.
+`$` to avoid accidental partial matches. However, by default we will not prefix match, a full pattern needs to be provided.
```
randomir, line 1833: You will not prefix match _by default_? That implies prefix matching **is possible** by turning on some hidden option?

nbialostosky: Thanks for the feedback here, prefix matching is not an option. Adding this line was intended to help clarify that, but can see why it could be interpreted otherwise.

As this seems to be directly affecting you, do you have a suggestion on language to add here that would be explicit that prefixing matching doesn't occur?

Would adjusting to something like this work?

>A full match pattern must be provided, prefix matching is not an option.

randomir: Yes, that's much clearer, thanks!

I would also remove the preceding sentence about recommended anchoring, as the pattern will be anchored regardless.

FWIW, I didn't submit PR for this as I wasn't sure what's the scope and extent of this bug, and whether you're happy handling it with docs only, or you'd prefer to address the underlying broken behavior.

nbialostosky: Thanks for the reply! I wanted to keep the note about anchoring just to ensure all customer use cases are covered. However, I added the prefix note ahead of it, to ensure that the information is as clear as possible, I'll get the PR reviewed and merged in shortly: https://github.com/circleci/circleci-docs/pull/5437

PR #5437, opened 2021-06-15
Update prefix information for matches condition

nbialostosky:
# Description
This PR will adjust the language around prefix matching to be more explicit.

# Reasons
Stemmed from a conversation with a customer. Updating to ensure the docs are as clear as possible.

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -1829,8 +1829,8 @@ argument specifications, and to a maximum depth of 100 levels.
 
 `matches` uses [Java regular
 expressions](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)
-for its `pattern`. It is recommended to enclose a pattern in `^` and
-`$` to avoid accidental partial matches. However, by default we will not prefix match, a full pattern needs to be provided.
+for its `pattern`. A full match pattern must be provided, prefix matching is not an option. Though, it is recommended to enclose a pattern in `^` and
+`$` to avoid accidental partial matches.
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -1829,8 +1829,8 @@ argument specifications, and to a maximum depth of 100 levels.
 
 `matches` uses [Java regular
 expressions](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)
-for its `pattern`. It is recommended to enclose a pattern in `^` and
-`$` to avoid accidental partial matches. However, by default we will not prefix match, a full pattern needs to be provided.
+for its `pattern`. A full match pattern must be provided, prefix matching is not an option. Though, it is recommended to enclose a pattern in `^` and
+`$` to avoid accidental partial matches.
```
randomir, line 1833: Are accidental partial matches even possible in your current implementation?

nbialostosky: No, at the moment partial matching is not possible. However, it isn't completely out of the question that we may implement that in the future. With that in mind, we are going to keep this note in the documentation in case those changes do happen. Thanks!

randomir: That's backwards, IMO -- making docs inaccurate just because you _might_ add that feature in the future.

I would at least say that **currently** full matching is assumed regardless of anchors specified, but to avoid accidental partial matches **in the future**, it's recommended to anchor the pattern. Or something along those lines.

teesloane (approved):

PR #5439, opened 2021-06-16
Add: notes on using new app store connect api vs 2fa.

### LINKED ISSUE #5380: Outdated deploying-ios example
# Locations
[Deploying-ios/#setting-up](https://circleci.com/docs/2.0/deploying-ios/#setting-up)

# Details
The current instructions suggest setting up an apple account specific to CI and configuring that apple_id and `FASTLANE_PASSWORD` in the system.  

Recently Apple made 2FA required for accounts, which means the current example requires the account owner to manually respond to the 2FA request during a build.  While the example still technically works, this is less than desirable for an automated system.  

Fastlane has [recommendations on how CI systems can authenticate](https://docs.fastlane.tools/best-practices/continuous-integration/#authenticating-with-apple-services), as well as extensive documentation on the recommended practice of generating an [appstore connect API key](https://docs.fastlane.tools/app-store-connect-api/).

teesloane:
Fixes #5380 

It looks like we may need to do a bit of a one over on the Fastlane content in our docs as things have changed from an authentication perspective. For the time being, I'm adding a section calling out some of the changes, as raised in the linked issue.

### DIFF
--- a/jekyll/_cci2/deploying-ios.md
+++ b/jekyll/_cci2/deploying-ios.md
@@ -70,6 +70,19 @@ apple_id "ci@yourcompany.com"
 app_identifier "com.example.HelloWorld"
 ```
 
+#### Two-Factor Authentication
+{: #two-factor-authentication}
+{:.no_toc}
+
+Please note that when using previous authentication methods with Fastlane (which
+was a cookie-based web session), depending on the actions in your configuration,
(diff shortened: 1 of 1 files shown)


BytesGuy (approved):
@teesloane This looks fine as a temporary placeholder until we get fresh docs drafted!

PR #5446, opened 2021-06-21
Fixed broken anchors in links from Optimizations page

OrBin:
# Description
Fixed broken anchors in links from Optimizations page.

# Reasons
The links work, but do not redirect to the specific location in the page, which a correct anchor should.


rosieyohannan:
Thank you so much for fixing these @OrBin! 🎉

rosieyohannan (approved):

PR #5447, opened 2021-06-22
Improve VM service documentation

ryanwohara:
# Description
* k8s nodes don't need to communicate
* The SSH port is 54782
* CLI commands to create the security group

# Reasons
https://circleci.atlassian.net/browse/SERVER-1229

### DIFF
--- a/jekyll/_cci2/server-3-install-hardening-your-cluster.adoc
+++ b/jekyll/_cci2/server-3-install-hardening-your-cluster.adoc
@@ -203,7 +203,15 @@ Similar to Nomad clients, there is no need for external VMs to communicate with 
 | Purpose
 
 | 22
+| Kubernetes nodes
+| Internal communication
+
+| 22
 | Nomad clients
+| Internal communication
+
+| 2376
+| Kubernetes nodes
 | Internal communication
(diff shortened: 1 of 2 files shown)


robinsturmeit (changes requested):
I commented on the Jira ticket. Opening communication on ports 22 and 2376 between cluster and external VMs is necessary.

ryanwohara:
The Jira comment doesn't seem accurate based on design docs: https://github.com/circleci/vm-service/blob/master/docs/design.md#task

robinsturmeit:
> The Jira comment doesn't seem accurate based on design docs: https://github.com/circleci/vm-service/blob/master/docs/design.md#task

Let's keep the discussion within the ticket until resolved :)

ryanwohara:
Robin is right - converting this to a draft

robinsturmeit (changes requested):
Seems like you would still need to add the k8s cluster CIDR block to the rules in the CLI examples

ryanwohara:
Super good callout ty

robinsturmeit (approved):
Thank you, this is great :)

PR #5457, opened 2021-06-25
[SERVER-1136] updating 3.1 architecture diagram

soulchips:
# Description
Update to the sever 3.x architecture diagram to reflect the new layout.

# Reasons
https://circleci.atlassian.net/browse/SERVER-1137


ryanwohara:
We should define distributor within the cluster

robinsturmeit:
Two nits:
- The communication between external VMs and cluster is missing
- The external SSH ports are given as `64535/65535`, which implies it's either of those. However, it's actually a port range, so rather `64535-65535`. For external VMs it's always `54782`.

rosieyohannan:
The note on the line coming from VCS to the services cluster, could we have a `/` between 'webhooks' and 'build' to make it clear they are three separate items?

ryanwohara:
I'm tempted to just approve it at this point but under Inspect Builds (at the bottom left) I notice it's a bit unclear now. If this is desirable phrasing then we should at least capitalize external to match the other lines.

rosieyohannan:
The only other thing I was wondering about is representing how some services can now be externalised, an asterisk maybe with a note? Happy to not do this now though if it's going to make this drag on too much. It's clear in the docs anyway.

soulchips:
> I'm tempted to just approve it at this point but under Inspect Builds (at the bottom left) I notice it's a bit unclear now. If this is desirable phrasing then we should at least capitalize external to match the other lines.

I see what you're saying. I think we can improve this.

soulchips:
> The only other thing I was wondering about is representing how some services can now be externalised, an asterisk maybe with a note? Happy to not do this now though if it's going to make this drag on too much. It's clear in the docs anyway.

I see what you're saying. I don't have a strong opinion since its already in the docs. let's see if we can get this in

robinsturmeit (approved):

rosieyohannan:
Looks good! I'm going to merge this in to the 3.1 branch now 🎉

PR #5479, opened 2021-07-01
Add app store connect api key info to ios deployment guide

BytesGuy:
Description
Update the document to reflect the new method for authenticating with app store connect for deplyoment

Reasons
MacOS document improvements initiative


teesloane (approved):
Lovely! Thanks @BytesGuy !

PR #5480, opened 2021-07-02
Removing CLI instructions for user management

robinsturmeit:
This doesn't work for server 3.x

# Description
Removing the CLI instructions for user management

# Reasons
Just spotted this, this doesn't apply for server 3


rosieyohannan:
Good spot, thanks @robinsturmeit 🎉

rosieyohannan (approved):

PR #5482, opened 2021-07-03
Clarify cache section for OSS

lesteve:
# Description

This updates the section about caching for OSS builds and tries to clarify it.

# Reasons

The symptom "the build is slow on the first PR commit because it does not seem to find the cache" is hard to google and it took me a while to find this part of the documentation.

In an ideal world there would be a from the main [cache doc page](https://circleci.com/docs/2.0/caching/#restoring-cache) to this caveat.


teesloane (approved):
Makes sense -= thanks @lesteve !

lesteve:
Thanks!

PR #5485, opened 2021-07-05
recommend a backup before starting migration

lloydChris:
# Description
We need to recommend the user take a backup of their existing instance.  This is particularly important when external datastores are, used because 3.1 will overwrite them.  Rolling back to 2.19 will require the user restore the postgres DB


jekyll/_cci2/server-3-install-migration.adoc
```diff
@@ -23,7 +23,8 @@ installation to the data stores you were using for your 2.19 instance.
 toc::[]
 
 ## Prerequisites
-. Your current CircleCI Server installation is 2.19
+. Your current CircleCI Server installation is 2.19.
+. You have taken a backup of the 2.19 instance.  If you are using external datastores, they will need to be backup up separately.
```
rosieyohannan, line 27:
```suggestion
. You have taken a backup of the 2.19 instance.  If you are using external datastores, they will need to be backed up separately.
```

rosieyohannan (approved):

soulchips (approved):

ryanwohara (approved):

PR #5498, opened 2021-07-07
We fixed SSHing into the machine executor in 3.1

ryanwohara:
We fixed this issue just before we released so we can remove it from the known issues.


robinsturmeit (approved):

rosieyohannan:
Good spot, thanks! I'll update the PDFs for this too so there's no confusion

PR #5530, opened 2021-07-14
Fix to path-filtering cookbook example

stig:
This example uses `path-filtering/filter` as a reusable command, but it's only exposed as a job in the [path-filtering orb](https://circleci.com/developer/orbs/orb/circleci/path-filtering).

Also added line breaks to the example's comments to improve readability.


teesloane (approved):
Thanks @stig !

PR #5534, opened 2021-07-14
Fixes and improvements to the dynamic config w/setup workflows example

stig:
# Description
> A brief description of the changes.

Refactor the setup workflows example to fix usage of the maven orb, and avoid the setup-job running twice. An alternative would be to add a `when` condition and not attempt to create the setup-workflow if we're no longer in setup phase.

# Reasons
> A link to a GitHub and/or JIRA issue (if applicable).
> Otherwise, a brief sentence about why you made these changes.

The maven orb exposes jobs, not reusable commands, so the previous example didn't work. cf https://circleci.com/developer/orbs/orb/circleci/maven

Using a separate continuation.yml config makes it easier to separate the setup-workflow from regular workflows.

### DIFF
--- a/jekyll/_cci2/configuration-cookbook.md
+++ b/jekyll/_cci2/configuration-cookbook.md
@@ -679,29 +679,6 @@ parameters:
     type: boolean
     default: false
 
-# our defined jobs
-jobs:
-  # the build-service-1 job uses the maven orb to build and install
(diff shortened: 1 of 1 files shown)


rosieyohannan:
closing - work duplicated in https://github.com/circleci/circleci-docs/pull/5546

PR #5538, opened 2021-07-16
Update server-3-install.adoc

ryanpedersen42:
# Description
cleaning up stray period and space in instructions.

# Reasons
looks like a typo there


rosieyohannan (approved):
Good spot, thank you!

PR #5546, opened 2021-07-22
Fix a sample config of the dynamic configuration

ganezasan:
# Description
Fix a sample config of the dynamic configuration, the current configuration is completely broken, so split the config to `config.yml` and `continue_config.yml`, and fix some small issues.

Also, add a note for the dynamic configuration. Currently, CircleCI doesn't allow to call the dynamic configuration feature more than once, and that's why I split the config.yml to `config.yml` and `continue_config.yml`.

This is an example of the failed build by using the dynamic configuration twice.
> https://app.circleci.com/pipelines/github/cheekit/circleci-setup-workflow/85/workflows/b724ba84-96e5-4852-b6fe-964cf23defde/jobs/78

# Reasons

CircleCI user pointed out about this issue.

### DIFF
--- a/jekyll/_cci2/configuration-cookbook.md
+++ b/jekyll/_cci2/configuration-cookbook.md
@@ -625,7 +625,7 @@ In the above configuration, we:
     - Continues running the pipeline based on what configuration is provided to the required `configuration_path`.
 - Lastly, we call the `setup` job defined above as a part of our `workflow`
 
-Note: You can only use one workflow per `config.yml` when using CircleCI's dynamic configuration feature
+**Note:** You can only use one workflow per `config.yml` when using CircleCI's dynamic configuration feature
 You can only run a single workflow as part of the pipeline's setup stage. This setup-workflow has access to a one-time-use token to create more workflows. The setup process does not cascade, so subsequent workflows in the pipeline cannot launch their own continuations.
 
 For a more in-depth explanation of what the `continuation` orb does, review the orb's source code in the
@@ -659,6 +659,7 @@ For example, consider a monorepo structure like the example shown below:
 An example implementation of CircleCI's dynamic configuration for the above use case can be found in the following `config.yml` and `continue_config.yml`:
 
 #### config.yml
+{: #configyml }
 
 ```yaml
 version: 2.1
@@ -694,6 +695,7 @@ workflows:
 ```
 
 #### continue_config.yml
+{: #continueconfigyml }
 
 ```yaml
 version: 2.1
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/configuration-cookbook.md
```diff
@@ -626,6 +626,7 @@ In the above configuration, we:
 - Lastly, we call the `setup` job defined above as a part of our `workflow`
 
 Note: You can only use one workflow per `config.yml` when using CircleCI's dynamic configuration feature
+Note: You can't call CircleCI's dynamic configuration feature several times in the same pipeline. It means you can't make the another workflow with `setup: true` by the CircleCI's dynamic configuration feature.
```
thomasr8, line 629:
```suggestion
Note: You are unable to call CircleCI's dynamic configuration feature several times in the same pipeline. This means that you cannot make the another workflow with `setup: true` by the CircleCI's dynamic configuration feature.
```

thomasr8 (approved):
Looks good. Just one minor comment.

jekyll/_cci2/configuration-cookbook.md
rosieyohannan: > This means that you cannot make the another workflow with `setup: true` by the CircleCI's dynamic configuration feature.

Is still unclear to me.
maybe ...

This means that you cannot make a second workflow with `setup: true` within the same `.circleci/config.yml` file.

rosieyohannan:
@ganezasan there's this PR too from the pipelines team - not sure what to do to fix this. Let me know your thoughts, and @stig https://github.com/circleci/circleci-docs/pull/5534

jekyll/_cci2/configuration-cookbook.md
stig, line 629: Throwing my own suggestion in:

```suggestion
You can only run a single workflow as part of the pipeline's setup stage. This setup-workflow has access to a one-time-use token to create more workflows. The setup process does not cascade, so subsequent workflows in the pipeline cannot launch their own continuations.
```

stig:
> @ganezasan there's this PR too from the pipelines team - not sure what to do to fix this. Let me know your thoughts, and @stig #5534

Our PRs look almost identical from a code perspective, except Taka's follows the naming in the path-filtering orb better and also omits the parameters from config.yml. I'd merge this version, to get the code example fixed, and then worry about improving the surrounding text in a subsequent PR.

rosieyohannan:
Thank you @stig!

PR #5549, opened 2021-07-23
Fix regexp to list test packages in parallel tests tutorial.

saroh:
# Description
Current command could remove unwanted characters

# Reasons
`/.py/` is not precise enough (files that contain a letter followed by `py`). It might match parts of a dir name + `.` is a catch all etc. Changed to `/\.py$/` (files that end with `.py`)

# Link
https://circleci.com/docs/2.0/parallelism-faster-jobs/#using-test-splitting-with-python-django-tests


rosieyohannan:
Thank you @saroh! 🎉

rosieyohannan (approved):

PR #5550, opened 2021-07-24
Fix broken markdown introduced in #5499

tnir:
Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

# Description
Fixes a broken markdown, introduced in #5499, in [キャッシュの復元](https://circleci.com/docs/ja/2.0/caching/#restoring-cache).

# Reasons
circleci com_docs_ja_2 0_caching_

PR #5499, opened 2021-07-07
Updated translations in _cci2_ja

heug:
# Description
Bringing master up to date with the latest Japanese documentation translations from Crowdin

PR #5551, opened 2021-07-24
Fix broken markdown and translated strings in /ja/caching

tnir:

# Description
Fixes another broken markdown, introduced in #5499, in [キャッシュの復元](https://circleci.com/docs/ja/2.0/caching/#restoring-cache). cf. #5550

# Reasons
Similar to #5549


tnir:
@ganezasan Could you review this?

cc @heug

ganezasan:
@tnir Thank you for sending a PR.
We uses `ja-localization` branch in our translation flow, and the same change is already applied. 
Also, we've released that change to the docs site. So, now the markdown issue has been solved.
> https://circleci.com/docs/ja/2.0/caching/#using-caching-in-monorepos

I'll close this PR, but thank you so much for your contribution 👍

PR #5554, opened 2021-07-25
Update circleci/node images to latest

tnir:

# Description
Updates the versions of `circleci/node` image on Docker Hub, keeping the minor/revision specifications.

# Reasons
Node.js 4.x, 7.x, 8.x, 9.x, and 10.x got EOL'ed, so we need to use 14.x LTS or 16 stable. Also `jessie` reached EOL, so these should be updated to the latest `buster`. `stretch` is still stable, but these can be updated now.

### DIFF
--- a/jekyll/_cci2/adv-config.md
+++ b/jekyll/_cci2/adv-config.md
@@ -60,7 +60,7 @@ version: 2
 jobs:
   build:
     docker:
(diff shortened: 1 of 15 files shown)


teesloane (approved):
Awesome, thanks!

PR #5562, opened 2021-07-27
[SERVER-1305] removing references to http proxy vars in 3.1

soulchips:
# Description
http proxy settings do not work currently in 3.x
We are removing this feature for now and will return to this feature later on the road map

# Reasons
https://circleci.atlassian.net/browse/SERVER-1305


ryanwohara (changes requested):
These docs are for 2.x and should not be modified. We do support proxies in 2.19.x.

soulchips:
> These docs are for 2.x and should not be modified. We do support proxies in 2.19.x.

yes you are correct. thanks for the callout

PR #5614, opened 2021-08-18
Update Node.js orb and use next-gen image

felicianotech:
This PR is mainly to swap the Node.js image. The legacy image, which is currently used, is deprecated. This switches over to the next-gen Noide.js image.

The orb is updated too because 1) why not and 2) we needed a newer version of the orb in order to use the next-gen image.

One thing to pay attention to here, this updates the Node.js version by a lot. From v8 all the way to v14. The pipeline passed successfully so that's good but we'll want to make sure there isn't any other unintended changes.


felicianotech:
Related: https://github.com/circleci/circleci-docs/pull/5615

thomasr8 (approved):

PR #5615, opened 2021-08-18
Update Ruby orb and use next-gen image

felicianotech:
This PR is mainly to swap the Ruby image. The legacy image, which is currently used, is deprecated. This switches over to the next-gen Ruby image.

The orb is updated too because 1) why not and 2) we needed a newer version of the orb in order to use the next-gen image.

---

This will also fix an occasional issue with the `build` job which is failing [due to Debian 10](https://www.feliciano.tech/blog/debian-10-apt-fix/).


felicianotech:
Related: https://github.com/circleci/circleci-docs/pull/5614

thomasr8 (approved):

PR #5642, opened 2021-08-25
Clarify local validation of files with private orb

caljess599:
# Description
Expand the instructions for validating a config file that uses a private orb

# Reasons
I had to dig around for a definition of `org-slug` and it wasn't intuitive to me that the path was just the local config file (because it never needed to be specified before).


rosieyohannan:
Hey @caljess599, thanks so much for your PR. I can see why this is confusing, the private orbs section appears before the definition of orb-slug which is here: https://circleci.com/docs/2.0/orb-intro/#identifying-orbs. What I think we should do is in fact rearrange the content slightly so we can link to this definition as well as giving the same example. I'll open a new PR and post here when ready :-)

PR #5643, opened 2021-08-25
[WEB-390] Allow recursive event tracking when `a` tag includes another element

bardan-circleci:
Ticket: [WEB-390](https://circleci.atlassian.net/browse/WEB-390)

# Description
Currently, if you click a link with child elements inside (like svg/span/img/etc), the child element triggers the click and become the target element causing our analytics tracking code to skip it over. 

# Changes
- Used CSS to make the child elements inside an `a` tag have no pointer events so they stop registering as the target element in event listeners.


j2em6h (approved):
LGTM

PR #5647, opened 2021-08-27
[CIRCLE-36533] Add copy code button docs

ccgeller:
**Ticket:** [CIRCLE-36533](https://circleci.atlassian.net/browse/CIRCLE-36533)

Changes
=======
Paired with @tshen815 
- Add in `highlightJS-badge` npm package
- Add in optimizely logic to change between treatment vs control on Copy code button

Rationale
=========

Based on the [spike](https://circleci.atlassian.net/wiki/spaces/GD/pages/6227427500/CIRCLE-36042+Copy+code+button+for+Docs+DevHub) Nav Ramabadran and the work Chantelle Gellert started with [CIRCLE-36634: Experiment - Add copy code button to devhubDEV](https://circleci.atlassian.net/browse/CIRCLE-36634) we want to make an experiment that brings the copy code button everywhere on the Doc site. For example here is a page where it’s missing: [Hello World - CircleCI ](https://circleci.com/docs/2.0/hello-world/)

Screenshots
============
<h4>Before</h4>
Screen Shot 2021-08-27 at 4 15 16 PM

<h4>After</h4>
Screen Shot 2021-08-27 at 4 10 13 PM

### DIFF
--- a/jekyll/assets/js/highlightJSBadge.js
+++ b/jekyll/assets/js/highlightJSBadge.js
@@ -1,0 +1,40 @@
+window.onload = () => {
+   // https://app.optimizely.com/v2/projects/16812830475/experiments/20580130094/variations
+
+   window.OptimizelyClient.getVariationName({
+      experimentKey: 'dd_add_copy_code_button_to_docs',
+      groupExperimentName: 'q3_fy22_docs_disco_experiment_group_test'
+    }).then(function(variation) {
+
+      if(variation === 'treatment') {
+         var options = {
+            // the selector for the badge template
+            templateSelector: "#CodeBadgeTemplate",
+      
+            // base content CSS selector that is searched for snippets
+            contentSelector: "body",
+      
+            // Delay in ms used for `setTimeout` before badging is applied
+            // Use if you need to time highlighting and badge application
+            // since the badges need to be applied afterwards.
+            // 0 - direct execution (ie. you handle timing
+            loadDelay: 0,
+      
+            // CSS class(es) used to render the copy icon.
+            copyIconClass: "fa fa-copy",     
+            // optional content for icons class (<i class="fa fa-copy"></i> or <i class="material-icons">file_copy</i>)
+            copyIconContent: "",
+      
+            // CSS class(es) used to render the done icon.
+            checkIconClass: "fa fa-check text-success",
+            checkIconContent: "",
+      
+            // function called before code is placed on clipboard that allows you inspect and modify
+            // the text that goes onto the clipboard. Passes text and code root element (hljs).
+            // Example:  function(text, codeElement) { return text + " $$$"; }
+            onBeforeCodeCopied: null        
+         };
+         window.highlightJsBadge(options);
+      }
+    });
+};
--- a/package.json
+++ b/package.json
@@ -14,21 +14,22 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@optimizely/optimizely-sdk": "^4.6.2",
-    "js-cookie": "^2.2.0",
-    "lodash.debounce": "^4.0.8",
-    "lodash.get": "^4.4.2"
+    "highlightjs-badge": "0.1.9",
+    "@optimizely/optimizely-sdk": "4.6.2",
+    "js-cookie": "2.2.0",
+    "lodash.debounce": "4.0.8",
+    "lodash.get": "4.4.2"
   },
   "devDependencies": {
-    "@babel/core": "^7.1.0",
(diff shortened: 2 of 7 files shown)


rsylvian (changes requested):
Not sure if you saw that already but it looks like the generation didn't work:

```
Generating... 
  Liquid Exception: Stamp source file does not exist: /home/circleci/circleci-docs/jekyll/assets/js/highlightJsBadge.js in /home/circleci/circleci-docs/jekyll/_layouts/classic-docs.html
  ```

File: jekyll/assets/js/highlightJSBadge.js [hunk 1]
```diff
@@ -0,0 +1,41 @@
+window.onload = () => {
+   // https://app.optimizely.com/v2/projects/16812830475/experiments/20580130094/variations
+
+   window.OptimizelyClient.getVariationName({
+      featureName: 'add_copy_code_button_to_docs',
+      groupExperimentName: 'q3_fy22_docs_disco_experiment_group_test'
+    }).then(function(variation) {
+      console.log("variation is", variation);
```
rsylvian, line 8:
```suggestion
```

✂️

src-js/app.js
```diff
@@ -1,6 +1,7 @@
 import * as Cookie from 'js-cookie';
 import * as search from './instantsearch.js';
 import * as lang from './lang.js'
+import * as HighlightJSBadge from '../node_modules/highlightjs-badge/highlightjs-badge.min.js'
```
rsylvian, line 4: Is this the only we can reference that package? Not a big fan of having the relative path to the package directly here

rsylvian (changes requested):
Any reasons why you have 6000+ changes in the `package-lock.json`? I don't think adding just `highlightjs-badge` would make that many changes?

jekyll/assets/js/highlightJSBadge.js
```diff
@@ -0,0 +1,41 @@
+window.onload = () => {
+   // https://app.optimizely.com/v2/projects/16812830475/experiments/20580130094/variations
+
+   window.OptimizelyClient.getVariationName({
+      featureName: 'add_copy_code_button_to_docs',
```
rsylvian, line 5: Please prefix you experiment name with `dd_` 🙏

src-js/app.js
rsylvian: Also, side effect: `HighlightJSBadge` is going to be marked as `unused` because it is imported here but not really used. Now I understand why you did that since you need it to be part of the `app.bundle.js` but we should at the very least add a comment of what we are doing here.

jekyll/assets/js/highlightJSBadge.js
tshen815: quick note on this: make sure to update the experiment key name in optimizely as well after updating 🙏

src-js/app.js
ccgeller: import * as highlightJSBadge from 'highlightjs-badge';
works

jekyll/assets/js/highlightJSBadge.js
```diff
@@ -0,0 +1,40 @@
+window.onload = () => {
+   // https://app.optimizely.com/v2/projects/16812830475/experiments/20580130094/variations
+
+   window.OptimizelyClient.getVariationName({
+      featureName: 'dd_add_copy_code_button_to_docs',
```
rsylvian, line 5:
```suggestion
      experimentKey: 'dd_add_copy_code_button_to_docs',
```

jekyll/assets/js/user.js
```diff
@@ -41,7 +41,7 @@ $(function() {
 });
 
 /*
-* The following functions are coming directly from what used to be site.min.js
+* The following functions are coming directly from the site.min.js
```
nramabad, line 44: ?

package.json
```diff
@@ -14,6 +14,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "highlightjs-badge": "^0.1.9",
```
nramabad, line 17: can we fix these in place. i know a lot of these are prior to your work but they should not be ^

rsylvian (changes requested):
The language inference has a lot of false positives, any chance we turn that off?

Screen Shot 2021-09-03 at 9 04 42 AM
Screen Shot 2021-09-03 at 9 05 07 AM

rsylvian (approved):
Works well and looks better too! ✅  🚢  it!
Screen Shot 2021-09-03 at 11 38 32 AM

PR #5659, opened 2021-09-02
Fix OptimizelyClient.getVariationName() so it doesn't always have to wait on userData

rsylvian:
# Description

- Transform the way to get `userId` so it doesn't have to wait on the `userDataReady` if the data is already there

# Reasons
While pairing with @ccgeller we realized that `getVariationName()` was acting strange and will not always resolve. The reason why was based on pure timing between the time the API is called and the `userDataReady` event was registered. This PR fixes that by first checking if `userData` is already present, if not it will wait on the event to resolve.

### DIFF
--- a/src-js/optimizelyClient.js
+++ b/src-js/optimizelyClient.js
@@ -81,7 +81,7 @@ class OptimizelyClient {
           resolve(null);
         });
       });
-    })
+    });
   }
 }
 
--- a/src-js/optimizelyClient.js
+++ b/src-js/optimizelyClient.js
@@ -7,6 +7,17 @@ class OptimizelyClient {
       datafile: window.optimizelyDatafile,
     });
   }
+  getUserId() {
+    return new Promise((resolve) => {
+      if (window.userData) {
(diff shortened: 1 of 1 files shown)


src-js/optimizelyClient.js
```diff
@@ -7,6 +7,17 @@ class OptimizelyClient {
       datafile: window.optimizelyDatafile,
     });
   }
+  getUserId() {
+    return new Promise((resolve) => {
+      if (window.userData) {
+        resolve((window.userData.analytics_id) ? window.userData.analytics_id : null);
+      } else {
+        window.addEventListener("userDataReady", () => {
+          resolve((window.userData.analytics_id) ? window.userData.analytics_id : null);
```
tshen815, line 16: could we include a quick comment for why we are making this check? I think what you listed in **Reason** would be perfect!

tshen815 (approved):

src-js/optimizelyClient.js
rsylvian: Good catch! Done ✅

PR #5660, opened 2021-09-03
[CIRCLE-36338] Fix navbar active items 

rsylvian:
# Description

- Remove the use of the URL parameter `?section=` and instead used the `data-section` attribute
- Remove the use of localstorage and use `data-section` attached to the `active` item to know which section to open
- Apply `slugify` to both `current_url` and `item_link` when comparing them to prevent mismatch when trailing slash is not present

# Reasons
[CIRCLE-36338](
https://circleci.atlassian.net/browse/CIRCLE-36338)

This PR improves the way both sidebars work by removing some logic around local storage and by doing a slug comparison (see Description) to prevent a mismatch of URLs. Currently, if you open this [URL](https://circleci.com/docs/2.0/server-3-install-migration/) in a private tab with 0 local storage (to simulate a brand new visitor) the section will not automatically open. 

This is because some browsers automatically add a trailing slash at the end but the URL is defined without a trailing slash in the sidenav.yml:

https://github.com/circleci/circleci-docs/blob/6e76e00a56e1a5563be00c01cf43d64c30358377/jekyll/_data/sidenav.yml#L336-L337

When we compare here, it fails because `2.0/server-3-install-migration` !== `2.0/server-3-install-migration/`:

https://github.com/circleci/circleci-docs/blob/6e76e00a56e1a5563be00c01cf43d64c30358377/jekyll/_includes/sidebar.html#L67

This PR fixes that by applying a `slugify` liquid filter which will make them match by transforming both into `2-0-server-3-install-migration` 🎉 

# Testing

Because this PR touches the way the entire sidebars (desktop and mobile) work, please make sure to try locally in all sorts of ways and let me know if you find anything suspicious! 😁

### DIFF
--- a/jekyll/assets/js/sidebar.js
+++ b/jekyll/assets/js/sidebar.js
@@ -4,39 +4,37 @@
   window.addEventListener('load', function () {
     var footer = document.querySelector('.footer');
     var sidebar = document.querySelector('.sidebar');
-    var defaultSectionName = 'getting-started';
+    const defaultSectionName = 'getting-started'
     var mobileSidebar = document.querySelector('.sidebar-mobile-wrapper');
     var mobileSidebarDefault = mobileSidebar.querySelector('[data-id="' + defaultSectionName + '"]');
-    var mobileSidebarDisplay = mobileSidebar.querySelector('.mobile-sidebar');
-    var urlParams = new URLSearchParams(window.location.search)
-    var currentSection = urlParams.get("section");
-
-    if (currentSection) {
-      localStorage.sidenavActive = currentSection;
-    }
+    //var mobileSidebarDisplay = mobileSidebar.querySelector('.mobile-sidebar');
+    //var urlParams = new URLSearchParams(window.location.search)
+    //var currentSection = urlParams.get("section");
 
     // activate default section, if nothing else is selected
-    localStorage.sidenavActive = localStorage.sidenavActive || defaultSectionName;
+    let activeSection = defaultSectionName;
+    const activePage = sidebar.querySelector('.active');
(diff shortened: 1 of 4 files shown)


jekyll/assets/js/sidebar.js
```diff
@@ -2,41 +2,36 @@
...
-    var currentSection = urlParams.get("section");
-
-    if (currentSection) {
-      localStorage.sidenavActive = currentSection;
-    }
+    const footer = document.querySelector('.footer');
+    const sidebar = document.querySelector('.sidebar');
+    const defaultSectionName = 'getting-started';
+    const mobileSidebar = document.querySelector('.sidebar-mobile-wrapper');
+    const mobileSidebarDefault = mobileSidebar.querySelector('[data-id="' + defaultSectionName + '"]');
```
teesloane, line 9: I think we should still be using es5 until we move this file into src-js and have it properly transpiled?

teesloane (changes requested):
Wow, this is awesome. I love that it removes the query param in the url when navigating by the sidebar. 

My only nit is that I think we have to keep using es5 (es6+ is used in a few places, `const`, `let`, etc) until we get this moved into the src-js directory.

ccgeller (commented):
Looks good to me! 🧚😊❤️

jekyll/assets/js/sidebar.js
rsylvian: Good point!

nramabad (approved):
LGTM!

ccgeller (approved):
Looks good to me! 👌🎶🌺👍👑

teesloane (approved):

PR #5662, opened 2021-09-03
Add AnalyticsClient to have a thin wrapper around Segment/Amplitude Analytics

rsylvian:
# Description

- Add `AnalyticsClient` with 4 static methods: `getSessionId()`, `trackAction()`, `trackUser()` & `trackPage()`
- Use `trackAction()` in the copy code [experiment](https://github.com/circleci/circleci-docs/blob/412840e56208079474debaaa2ee48082bc5f8ca2/jekyll/assets/js/highlightJSBadge.js#L36)
- Rename `optimizelyClient.js` to `optimizely.js`

# Reasons
This PR introduces an `AnalyticsClient` to help with sending events to segment/amplitude. It is heavily inspired by our internal analytics react package inside `frontplane`.  This PR also adds a `trackAction()` in the copy code experiment to track when users click the copy code button.


jekyll/assets/js/highlightJSBadge.js
```diff
@@ -10,29 +10,34 @@ window.onload = () => {
...
             checkIconContent: "",
-      
+
             // function called before code is placed on clipboard that allows you inspect and modify
             // the text that goes onto the clipboard. Passes text and code root element (hljs).
             // Example:  function(text, codeElement) { return text + " $$$"; }
-            onBeforeCodeCopied: null        
+            onBeforeCodeCopied: function(text) {
+              window.AnalyticsClient.trackAction('copy-code-clicked', {
+                page: location.pathname
```
teesloane, line 37: Should this be prefixed with `dd` ? or `docs-copy-code-clicked` ?

src-js/analytics.js
```diff
@@ -0,0 +1,56 @@
...
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN
+    const isNaN = Number.isNaN || ((value) => value !== value);
+    return !isNaN(existingSessionId) ? existingSessionId : Date.now();
+  }
+
+  /**
+   * Thin wrapper function around Segment Analytics track method
+   * @param name  Name of the action the current user performed.
+   * @param properties  A bag of metadata to send to Segment/Amplitude.
+   */
```
teesloane, line 15: Thank you for doing docstrings 🎉 !

jekyll/assets/js/highlightJSBadge.js
rsylvian: We don't prefix our events with `dd` (only experiments key) but I like the suggestion of `docs-copy-code-clicked` 🙌

nramabad: I think track actions, we typically don't prefix with dd. but it would be nice to add something like { isExperiment: true } in the options for experiments that touch exists trackActions.

since this one is new and does not occur in control, it would be superfluous.

src-js/analytics.js
```diff
@@ -0,0 +1,56 @@
+class AnalyticsClient {
```
nramabad, line 1: pls correct me if I'm wrong but you implemented all methods as static so we call the method on the class itself right? sort of like a singleton?

is there a reason for implement it as a class at all though? it doesn't look like you're using `this` or any other class construct. could this just be a number of analytics utility functions?

src-js/app.js
```diff
@@ -1,11 +1,15 @@
...
 import * as highlightJSBadge from 'highlightjs-badge';
 
 window.Cookie = Cookie;
 search.init();
 lang.init();
 
-window.OptimizelyClient = new OptimizelyClient();
\ No newline at end of file
+window.OptimizelyClient = new OptimizelyClient();
+window.AnalyticsClient = AnalyticsClient;
```
nramabad, line 14: if we want to keep this pseudo-singleton pattern, perhaps we should add a comment here explaining why we're not creating a new instance?

src-js/analytics.js
nramabad: also if we want to ensure this is used as a singleton, maybe we want to make sure multiple instances cannot be created? something like this?

`class AnalyticsClient {
  constructor() {
    throw "This is bad stuff. don't create instances mon"
}`

src-js/analytics.js
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Thin wrapper function around Segment Analytics track method
+   * @param name  Name of the action the current user performed.
+   * @param properties  A bag of metadata to send to Segment/Amplitude.
+   */
+  static trackAction(name, properties) {
+    if (isDataDogSynthetics()) return;
+
+    properties = properties ?? null;
```
ccgeller, line 19: nice! I didn't know about ??

ccgeller (approved):
Looks good to me!

src-js/analytics.js
rsylvian: This is not a Singleton but the pattern here is called Utility Class/Helper class. Because it contains only static methods, it is stateless and should not be instantiated. As you said, it could also be a bunch of functions that are populated in `window` but I prefer this pattern as it makes more sense semantically IMO

src-js/app.js
rsylvian: Yeah because it only has static methods, `AnalyticsClient` should not be instantiated. The only reason it is here is so webpack catches it and includes it in the `app.build.js`. I'll add a comment but in the near future (when we move everything to webpack) we should make it work so webpack automatically catches everything that is in `src-js/`

src-js/analytics.js
rsylvian: @ccgeller https://www.javascripttutorial.net/es-next/javascript-nullish-coalescing-operator/ 😁

src-js/analytics.js
rsylvian: All credits goes to: https://github.com/circleci/frontplane/blob/main/packages/analytics/src/analytics.ts 😂

PR #5673, opened 2021-09-07
Added clarification around the definition of certified orbs

hud-code:
# Description
Added some extra clarification around what certified orbs are. 

# Reasons
Received a customer question about what kind of security testing (if any) uncertified orbs receive, which then prompted the question around what is considered an uncertified/certified orb.


rosieyohannan (approved):
LGTM!

PR #5674, opened 2021-09-08
Add ja footers

heug:
# Description
Create Japanese versions of the footers included on all docs articles


teesloane:
Looking at this locally, I do not see translations:

image

This might be more involved than just adding the ja `includes`; it might be best to make it into a ticket (not sure which team would handle it though).

teesloane:
Pulled and tested locally. Looks good:

## Before
image

## After

image

teesloane (approved):

PR #5681, opened 2021-09-10
Update OS matrix example

### LINKED ISSUE #5611: multi-os matrix example broken and orbs outdated
# Description
The example uses two orbs node and win. The win orb has a default `shell: powershell.exe` the node orb runs a bash script to use nvm during the install phase. Resulting in errors on windows, like so:

```
#!powershell.exe -ExecutionPolicy Bypass
# Only install nvm if it's not already installed
if command -v nvm &> /dev/null; then
    echo "nvm is already installed. Skipping nvm install.";
else
    curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.37.2/install.sh | bash;
    echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV;
    echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" --no-use' >> $BASH_ENV;
    source $BASH_ENV;
fi

if [ "$NODE_PARAM_LTS" = "1" ]; then
    nvm install --lts
    nvm alias default lts/*
else
    nvm install "$NODE_PARAM_VERSION"
    nvm alias default "$NODE_PARAM_VERSION"
fi

echo 'nvm use default &>/dev/null' >> $BASH_ENV

At line:2 char:3
+ if command -v nvm &> /dev/null; then
+   ~
Missing '(' after 'if' in if statement.
At line:2 char:19
+ if command -v nvm &> /dev/null; then
+                   ~
The ampersand (&) character is not allowed. The & operator is reserved for future use; wrap an ampersand in double 
quotation marks ("&") to pass it as part of a string.
At line:11 char:3
+ if [ "$NODE_PARAM_LTS" = "1" ]; then
+   ~
Missing '(' after 'if' in if statement.
At line:11 char:5
+ if [ "$NODE_PARAM_LTS" = "1" ]; then
+     ~
Missing type name after '['.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : MissingOpenParenthesisInIfStatement
 

Exited with code exit status 1

CircleCI received exit code 1

```

Both orb version numbers are out of date. 

To fix the first problem, we would either need a documented way for how to switch the default shell of the windows orb to `shell: bash.exe`  that is included in the code-listing (and on the windows orb page). Or change the node orbs shell script to be non bash specific. I think the former is the wtg. 

To fix the second, would it be possible to at least automatically scan the documentation's code listing and update orb references whenever circle releases new orbs? Even better run use the updated code listings for some documentation driven development/testing. 

# Location
https://circleci.com/docs/2.0/configuration-cookbook/#use-matrix-jobs-to-run-multiple-os-tests 

# Steps to Reproduce
https://app.circleci.com/pipelines/github/evolvedbinary/fusion-studio/33/workflows/275dbd1d-6b07-415b-9fd8-f106b316a8b2/jobs/92

felicianotech:
Closes #5611 

I've updated the example to one that actually works. The Node.js orb doesn't support Windows so it shouldn't have been in the example.

I've verified that this example config runs: https://app.circleci.com/pipelines/github/felicianotech/cci-testing-pub/544/workflows/6836fa45-e9f5-4795-8ca6-94886906e174

### DIFF
--- a/jekyll/_cci2/configuration-cookbook.md
+++ b/jekyll/_cci2/configuration-cookbook.md
@@ -775,26 +775,27 @@ information on available elements and required parameters.
 
 Using matrix jobs is a good way to run a job multiple times with different arguments, using parameters. There are many uses for this, including testing on multiple operating systems and against different language/library versions.
 
-In the following example the `test` job is run across Linux, Windows and macOS environments, using two different versions of node. On each run of the `test` job different parameters are passed to set both the OS and the node version:
+In the following example the `test` job is run across a Linux container, Linux VM, and macOS environments, using two different versions of Node.js. On each run of the `test` job different parameters are passed to set both the OS and the Node.js version:
 
 ```yaml
 version: 2.1
 
 orbs:
-  node: circleci/node@4.0.0
-  win: circleci/windows@2.2.0
+  node: circleci/node@4.7
 
 executors:
-  linux: # linux executor using the node base image
+  docker: # Docker using the Base Convenience Image
     docker:
-      - image: cimg/node
+      - image: cimg/base:stable
         auth:
           username: mydockerhub-user
           password: $DOCKERHUB_PASSWORD  # context / project UI env-var reference
-  windows: win/default # windows executor - uses the default executor from the windows orb
-  macos: # macos executor using xcode 11.6
+  linux: # a Linux VM running Ubuntu 20.04
+    machine:
+      image: ubuntu-2004:202107-02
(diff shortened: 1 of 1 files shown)


rosieyohannan:
Thanks so much for fixing this @felicianotech!

rosieyohannan (approved):

PR #5683, opened 2021-09-10
Sentence correction

pawanbahuguna:
Correcting the sentence. Replaced `and` with `can`

# Description
Sentence seems wrong

# Reasons
A link to a GitHub and/or JIRA issue (if applicable): N/A

In the doc, the sentence was `Author and add`, which doesn't make sense as per the whole paragraph. So replace `and` with `can`.

### DIFF
--- a/jekyll/_cci2/orb-author.md
+++ b/jekyll/_cci2/orb-author.md
@@ -164,7 +164,7 @@ display:
 {: #commands }
 {:.no_toc}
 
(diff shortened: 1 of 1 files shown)


rosieyohannan:
Hey @pawanbahuguna thank you for this feedback! I think what is there is what was intended. We are saying you can author these sections and then add them to your orb. I will think of some better phrasing though and get these updated.

PR #5686, opened 2021-09-11
Default Fedora ssh client configuration fails to work

adelton:
# Description
Default Fedora ssh client configuration fails to work

# Reasons
It is necessary to downgrade the selected signature algorithms to log in to the CircleCI job via ssh from Fedora. Using `-o 'PubkeyAcceptedKeyTypes ssh-rsa'` worked for me. Maybe it will save time to others who'd be struggling with `Permission denied (publickey).` even if using the right key.

### DIFF
--- a/jekyll/_cci2/ssh-access-jobs.md
+++ b/jekyll/_cci2/ssh-access-jobs.md
@@ -130,7 +130,7 @@ debug1: Authentication succeeded (publickey).
 This sequence indicates that the key /Users/me/.ssh/id_rsa_github is the one which
 GitHub accepted.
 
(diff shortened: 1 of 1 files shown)


adelton:
I don't really have a problem with Ed25519 ... but did not also ensure that RSA with SHA 1 is now disabled on your servers? Because what the pull request showed was that SHA 1 is still accepted (while SHA 256 or SHA 512 were not).

devinmarieb:
@adelton I'm still waiting on a response to this from the people who are actively working on these things. Will follow up when I have one :)

devinmarieb:
@adelton the team working on this said that sha256 & 512 have already been added. I hope that helps.

PR #5701, opened 2021-09-17
[CIRCLE-37846] Send ExperimentViewed analytic event when a variation is returned by optimizely

rsylvian:
Ticket: [CIRCLE-37846](https://circleci.atlassian.net/browse/CIRCLE-37846)
# Description
- Add `trackExperimentViewed` once the `getVariation` resolves with a variation name
- Add a localstorage "cache' that prevents sending a bunch of duplicated `trackExperimentViewed` every time an experiment is viewed.

# Reasons
Every time a variation (`control`, `treatment` or `null`) is returned, we want to send an `ExperimentViewed` event to Amplitude so DAs can better analyze our experiments.

We also added a localstorage "cache" that prevents sending a bunch of duplicated `ExperimentViewed`. This is consistent to what we are already doing in the `@circleci/react-experiment` package.


src-js/app.js
```diff
@@ -12,5 +12,5 @@ lang.init();
 
 // adding "Clients" to the window object so they can be accessed by other js inside Jekyll
 window.Cookie = Cookie;
-window.OptimizelyClient = new OptimizelyClient();
-window.AnalyticsClient = AnalyticsClient; // because it only has static methods, AnalyticsClient is not instantiated
\ No newline at end of file
+window.AnalyticsClient = AnalyticsClient; // because it only has static methods, AnalyticsClient is not instantiated
+window.OptimizelyClient = new OptimizelyClient();
```
rsylvian, line 16: Reversed the order since OptimizelyClient need AnalyticsClient now

src-js/optimizely.js
```diff
@@ -88,4 +100,76 @@ class OptimizelyClient {
...
+
+  // assign the current experiment and its values
+  experiments[orgId][experimentKey] = { variationName, createdAt: new Date().getTime() };
+
+  try {
+    // setItem will sometimes fail with a "Quota Exceeded" exception if users have custom configurations
+    localStorage.setItem(STORAGE_KEY, JSON.stringify(experiments));
+  } catch {
+    // We're deliberately ignoring it so that it doesn't break the app. It'll mean a few extra
+    // events are emitted, but I think that's the lesser issue.
```
ccgeller, line 171: Would it be a good idea to log to error?

src-js/optimizely.js
```diff
@@ -88,4 +100,76 @@ class OptimizelyClient {
...
+  }
+}
+
 export default OptimizelyClient;
```
ccgeller, line 175: May have to set preferences per repo? As per Tony: VSCode to add it: Code > Preferences > Settings

ccgeller (approved):
Amazing I like it 🌺🥳❤️

src-js/optimizely.js
rsylvian: Technically Rollbar should log it for us already 🤩

PR #5711, opened 2021-09-20
Refactor: Move experiments to their own directory

rsylvian:
# Description
- Move `addOrbSection.js` into an `experiments/` directory
- Use babel `wildcard` plugin to be able to do `import * as Experiments from './experiments';`

# Reasons
As the Docs Disco Team, we are starting to work on a lot of experiments and it makes more sense to put them in their own directory so it is separate from the rest of the js code.


tshen815 (approved):
Looks good!

PR #5723, opened 2021-09-23
[CIRCLE-37828] Webpackify JS 🕸

teesloane:
Ticket: [CIRCLE-37828](https://circleci.atlassian.net/browse/CIRCLE-37828)

This PR moves the majority of our jekyll side javascript into our webpack bundling process. 

The following changes are present:

- remove outdated vendor code:
  - jquery cookie library (no longer used)
  - a bulk of jquery plugins from bootstrap

- moved files from jekyll/assets/js:
  - site.min.js + vendor.min.js (these have been combined)
  -  highlightJSBadge.js, main.js, nav.js, sidebar.js, user.js -> all moved to `src-js/*`

- updated webpack
  - handle a vendor bundle (which should load before our app bundle)
  - created a global substitution for jquery so that it doesn't need to be imported where it's used, and to avoid bundler mishaps between modules.


src-js/main.js
```diff
@@ -258,106 +263,11 @@ $( document ).ready(function() {
...
-
-        clickEvents.forEach(event => {
-          icon.parentElement.addEventListener(event, icon.copy);
-        });
-      });
-    }
-  });
-
-  $("article h1, article h2, article h3, article h4, article h5, article h6").filter("[id]").hover(function () {
+  $("article h2, article h3, article h4, article h5, article h6").filter("[id]").hover(function () {
```
rsylvian, line 270:
```suggestion
  $("article h1, article h2, article h3, article h4, article h5, article h6").filter("[id]").hover(function () {
```

src-js/main.js
```diff
@@ -258,106 +263,11 @@ $( document ).ready(function() {
...
-  $("article h1, article h2, article h3, article h4, article h5, article h6").filter("[id]").each(function () {
-    var isMainTitle = $(this).prop('nodeName') === 'H1';
-    $(this).append((isMainTitle
-      ? ' <a href="#'
-      : '<a href="#' + $(this).attr("id"))
-      + '"><i class="fa fa-link"></i></a>');
-    if (isMainTitle) {
-      $(this).find("i").toggle();
-    }
+  $("article h2, article h3, article h4, article h5, article h6").filter("[id]").each(function () {
```
rsylvian, line 267:
```suggestion
  $("article h1, article h2, article h3, article h4, article h5, article h6").filter("[id]").each(function () {
```

src-js/main.js
```diff
@@ -258,106 +263,11 @@ $( document ).ready(function() {
...
-        hideEvents.forEach(event => {
-          icon.parentElement.addEventListener(event, icon.hide);
-        });
-
-        clickEvents.forEach(event => {
-          icon.parentElement.addEventListener(event, icon.copy);
-        });
-      });
-    }
-  });
```
rsylvian, line 358: Where did this go?

jekyll/_includes/head.html
```diff
@@ -21,17 +21,14 @@
 	<link rel="stylesheet" href="{{ "/assets/css/footer.css" | prepend: site.baseurl }}" type="text/css" media="screen" />
 	<link rel="stylesheet" href="https://cdn.jsdelivr.net/docsearch.js/2/docsearch.min.css" />
 	<link rel="stylesheet" href="{{ "/assets/css/asciidoc.css" | prepend: site.baseurl }}">
-	<!--link rel="stylesheet" href="{{ "/assets/css/asciidoc-pygments.css" | prepend: site.baseurl }}"-->
-	<!--link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.11.0/styles/default.min.css">
 	<script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.11.0/highlight.min.js"></script>
-	<script>hljs.initHighlightingOnLoad();</script-->
+	<script>hljs.initHighlightingOnLoad();</script>
```
rsylvian, line 25: Could we have this somewhere else? like somewhere in `main.js` maybe?

src-js/app.js
```diff
@@ -1,19 +1,29 @@
...
+import * as search from './instantsearch.js';
+import * as lang from './lang.js'
+import './user.js';
+import './sidebar.js';
+import './nav.js';
+import './main.js';
+
+// imported but not used just so webpack picks it up and add it to the `app.bundle.js`
+import * as hljsBadge from 'highlightjs-badge';
+import './highlightJSBadge.js'
```
rsylvian, line 17: Can we move this to the `experiments/` directory now?

rsylvian (changes requested):
So exciting! Left couple of comments but haven't tried locally yet! Will do once some of this stuff is addressed 🙌

src-js/main.js
```diff
@@ -87,14 +89,14 @@ function detectScrollbar( element ){
 }
 
 function renderVersionBlockPopover() {
-  var badges = document.querySelectorAll(".server-version-badge");
-
```
nramabad, line 91: can you rebase w/ master? this shouldn't happen.

nramabad: i think your branch is overwriting some of the changes i merged into master like the share button stuff.

src-js/main.js
teesloane: Fixed! Made some git errors.

src-js/main.js
teesloane: Fixed! Pulled in latest version of file.

src-js/app.js
teesloane: Done.

.gitignore
```diff
@@ -28,3 +28,4 @@ src-api/build/
 vendor
 jekyll/_config_override.yml
 markdownlint-results.json
+.tool-verisons
```
rsylvian, line 31:
```suggestion
.tool-versions
```

package.json
```diff
@@ -15,16 +15,19 @@
...
     "lodash.get": "4.4.2",
     "uuid": "^8.3.2"
   },
   "devDependencies": {
-    "@babel/core": "7.1.0",
+    "@babel/core": "7.15.5",
     "@babel/plugin-proposal-nullish-coalescing-operator": "7.14.5",
-    "@babel/preset-env": "7.1.0",
+    "@babel/plugin-proposal-optional-chaining": "^7.14.5",
+    "@babel/preset-env": "7.15.6",
```
rsylvian, line 30: 🙌

.gitignore
teesloane: So that's why it wasn't picking it up... 😂

jekyll/_includes/head.html
teesloane: Done!

nramabad:
Looking good so far! I tested on my local env to verify what I could.

A few things stood out. Mainly that the app.bundle.js could be uglified, minified & inlined a little bit better by playing around with the Webpack configurations.

I noticed that UglifyJSPlugin is being required but not used as an argument in webpack.config.js. I think UglifyJS is the older way of doing uglification. Nowadays, I think it's better to use Terser(?). It may even be bundled with the latest webpack, Webpack 7.0 as I think Romain showed a few weeks ago.

Webpack currently is on 4.19.1. It would be nice to bump that up as I know keeping up with changes in Webpack can be a pain the longer we leave that out of date. If we can bump that up, we can maybe leverage Terser to get some better uglification/minification.

Lastly, if we can add a .browserslistrc that would be soooo good. I think this more of a cherry on the top but also should be p easy to add: https://github.com/browserslist/browserslist


Other than that, looks awesome! Thank you so much for removing those script tags and bringing everything into the package.json! Should be a lot more maintainable going forward!

teesloane:
Thanks for testing @nramabad ! All points duly noted - I agree with the above (although I don't think webpack has reached 7 yet -- possible typo for `5`?).

@rsylvian any thoughts on @nramabad 's points? I think we should do all of the above, but, since this touches quite a few files, I also want to get this merged in before fiddling with minification / dependencies in case it takes longer than anticipated and won't have to chase the main branch - we could create a new ticket for those steps perhaps.

If the Jekyll mini bundler @rsylvian implemented also already minifies, maybe we can ship this and circle back?

rsylvian:
I agree with most points from @nramabad. Our current webpack config already uglifies and minify but doesn't appear to inlines when going to prod. You can see that when checking the prod file in `https://circleci.com/docs/assets/app.bundle-XXXXXXXXXXX.js`. 

I'll create two following tickets to take care of the rest:

1. Move to webpack 5 and make sure inlines is handled properly in prod
2. Add `.browserslistrc`

@teesloane: Unfortunately, Jekyll mini bundler doesn't minify by itself but it supports any minifier that can read/write from stdin. I don't think it will be the right move here given the fact that everything is already processed once by webpack anyway.

package.json
```diff
@@ -15,16 +15,19 @@
   "license": "MIT",
   "dependencies": {
     "@optimizely/optimizely-sdk": "4.6.2",
+    "@popperjs/core": "^2.10.1",
     "highlightjs-badge": "0.1.9",
+    "jquery": "^3.6.0",
```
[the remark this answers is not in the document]
teesloane: Done!

src-js/app.js
```diff
@@ -1,19 +1,24 @@
...
+import * as highlightjsBadge from 'highlightjs-badge';
+// site files
 import OptimizelyClient from './optimizely.js';
 import AnalyticsClient from "./analytics.js";
+import * as search from './instantsearch.js';
+import * as lang from './lang.js'
+import './user.js';
+import './sidebar.js';
+import './nav.js';
+import './main.js';
```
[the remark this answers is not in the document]
teesloane: I was thinking `features` or `components` although the latter will likely be confused with react component folders for those coming from other MFE's. I'll go with`site` for now, and we can always change it later.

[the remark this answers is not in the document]
teesloane: Addressed. Also put optimizely, analytics and lang into a `services/` folder.

PR #5732, opened 2021-09-27
[SERVER] update nomad diagram

rosieyohannan:
Update Nomad cluster diagram to version from Design team.


thomasr8 (approved):

rsylvian:
@rosieyohannan Should this also be updated for the `ja` page? If yes, can we then remove the older unused image?

rosieyohannan:
> @rosieyohannan Should this also be updated for the `ja` page? If yes, can we then remove the older unused image?

Yes and Yes! I will do this now. Thank you for the reminder @rsylvian 👍 🎉

PR #5738, opened 2021-09-29
Add note about places where masking occurs

nbialostosky:
# Description
This update will add some clarification around the scope of secret masking.

# Reasons
This came about from an internal discussion and ensuring the documentation is up to date with the current functionality of the product.

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -45,7 +45,7 @@ The value of the environment variable will not be masked in the build output if:
 * the value of the environment variable is less than 4 characaters
 * the value of the environment variable is equal to one of `true`, `True`, `false` or `False`
 
-**Note:** Secrets Masking will only prevent the value of the environment variable from appearing in your build output. The value of the environment variable is still accessible to users [debugging builds with SSH]({{ site.baseurl }}/2.0/ssh-access-jobs).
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/env-vars.md
```diff
@@ -45,7 +45,7 @@ The value of the environment variable will not be masked in the build output if:
 * the value of the environment variable is less than 4 characaters
 * the value of the environment variable is equal to one of `true`, `True`, `false` or `False`
 
-**Note:** Secrets Masking will only prevent the value of the environment variable from appearing in your build output. The value of the environment variable is still accessible to users [debugging builds with SSH]({{ site.baseurl }}/2.0/ssh-access-jobs).
+**Note:** Secrets Masking will only prevent the value of the environment variable from appearing in your build output. If your secrets appear elsewhere, such as test results or artifacts, they won't be masked. In addition, the value of the environment variable is still accessible to users [debugging builds with SSH]({{ site.baseurl }}/2.0/ssh-access-jobs).
```
thomasr8, line 48:
```suggestion
**Note:** Secrets Masking will only prevent the value of the environment variable from appearing in your build output. If your secrets appear elsewhere, such as test results or artifacts, they will not be masked. In addition, the value of the environment variable is still accessible to users [debugging builds with SSH]({{ site.baseurl }}/2.0/ssh-access-jobs).
```

thomasr8: In general, we try to avoid using contractions whenever possible for localization/translation.

thomasr8 (approved):
Looks good.

PR #5739, opened 2021-09-30
fix flow of heading levels

rosieyohannan:
A couple of runner docs had skipped heading levels which is causing warnings when building the docs. This fixes the flow.


devinmarieb (approved):
@rosieyohannan am I allowed to do this yet? :) Looks good to me!

rosieyohannan:
> @rosieyohannan am I allowed to do this yet? :) Looks good to me!

of course yes and thank you!

PR #5742, opened 2021-09-30
Add URL query key to enable experiments

rsylvian:
# Description
- Add a `force-all` query key to enable all experiments by default

# Reasons
We have a preview system but when sharing these links with stakeholders the experiments are not able to resolve. This URL query key helps short-circuiting the optimizely client


nramabad:
do we want to remove the key from local storage if not present?

nramabad (approved):
Left a question. Otherwise LGTM.

PR #5790, opened 2021-10-14
[CIRCLE-38483]: Side Nav in Configuring CircleCI Does Not Scroll

tshen815:
**Ticket:** [CIRCLE-38483](https://circleci.atlassian.net/browse/CIRCLE-38483)

# Changes

- update class name to `full-height`
- add `fixed` position and `max-height` to allow sidenav to scroll with page 

# Rationale
The side nav that we have on the page in production is currently sticking to the top which would make the feature redundant as once you scroll past it on the page you would no longer be able to use it. As the Configuring CircleCI page is rather long, having the side nav scroll with the page as users move down would allow them to click to certain sections thus improving the overall experience. 

# Screen Shots

**Before:**

https://user-images.githubusercontent.com/57234605/137387277-a82da1f7-9da6-44f8-a4b7-84ba48297c7e.mp4

**After:**

https://user-images.githubusercontent.com/57234605/137385782-cb4241b1-22fd-42c7-8d99-c9d53638275e.mp4

### DIFF
--- a/jekyll/_layouts/classic-docs.html
+++ b/jekyll/_layouts/classic-docs.html
@@ -53,13 +53,13 @@
 
         {% if page.toc != false %}
         <aside class="full-height">
-          <h6 class="toc-heading">On this page</h6>
+          <h5 class="toc-heading">On this page</h5>
           <div class="toc-indent">
             {{ content | toc_only }}
           </div>
           {% if page.suggested != false %}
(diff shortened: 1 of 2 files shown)


rosieyohannan:
Would it be possible to add up to H6 in this change too now it's scrolling? Maybe that's a separate job though :-)

PR #5836, opened 2021-10-22
[CIRCLE-38569] Migrate css styles to sass, imported with webpack

teesloane:
See [ticket](https://circleci.atlassian.net/browse/CIRCLE-38569).

This PR:

- removes the process of scss by jekyll.
- makes webpack process scss in development and production
- aggregate separate css requests from head into scss.

Preview PR visible here: http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/tees-CIRCLE-38569-webpack-styles-preview/?force-all

*Stats*

| Production (old) | Preview (this PR) |
| -----------------|----------------------|
| Requests: 96 | Requests: 50 |
| Transferred: 1.8mb |    Transferred: 1.4mb |
| Time: 3:47  seconds | Time:  1:38 seconds


src-js/styles/_instantsearch.scss
```diff
@@ -85,8 +87,8 @@
 }
 
 .result-item-wrap {
-  $result-vertical-spacing: $spacing-medium-vertical / 4;
-  $half-gutter-width: $grid-gutter-width / 2;
+  $result-vertical-spacing: math.div($spacing-medium-vertical, 4);
```
ccgeller, line 90: never seen this before neat!

ccgeller (approved):
Nice enjoyed this 🥳🤩🧚

webpack.config.js
```diff
@@ -31,6 +36,17 @@ module.exports = {
...
+      {
+        test: /\.s[ac]ss$/i,
+        use: [
+          // fallback to style-loader in development
+          process.env.NODE_ENV !== 'production'
+            ? 'style-loader'
+            : MiniCssExtractPlugin.loader,
+          'css-loader',
+          'sass-loader',
+        ],
```
rsylvian, line 48: Doesn't seems like the CSS is minified? http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/tees-CIRCLE-38569-webpack-styles-preview/assets/app-2b4a22891844d95d2762af302515cbd0.css

teesloane: Added webpack minification.

jekyll/assets/css/footer.css
```diff
@@ -1,115 +0,0 @@
-/**
```
teesloane, line 1: I migrated this code into the other existing footer.scss file.

rsylvian (changes requested):
Tested locally and I have this weird flash where the site is loaded without CSS for couple frames:

Kapture 2021-11-02 at 17 54 48

Same thing when I navigate from page to page.

.circleci/config.yml
```diff
@@ -154,11 +154,11 @@ jobs:
           name: "Prepare JS assets" # Compile our final, production-ready JavaScript.
           command: yarn webpack-prod
       - run:
-          name: "Persist JS assets" # Move our JS into our workspace.
+          name: "Persist JS assets" # Move our JS+CSS into our workspace.
           command: |
             set -exu
             mkdir -p /tmp/workspace/js
-            mv jekyll/assets/js/*.bundle.js /tmp/workspace/js/
+            mv jekyll/assets/js/* /tmp/workspace/js/
```
rsylvian, line 161: Does this mean we are putting the `app.css` file inside the `js/` directory?

webpack.config.js
```diff
@@ -43,4 +61,7 @@ module.exports = {
       message: /export 'default'/,
     },
   ],
+  optimization: {
+    minimizer: [new CssMinimizerPlugin()],
+  },
```
rsylvian, line 66: Thanks 👏

.circleci/config.yml
teesloane: Yep, I have two questions about this:

- how do you feel about `styles` being under `src-js`? I could potentially put it outside src-js but that seems like it's not idiomatic to import from outside the src-js folder? (or perhaps we rename src-js to `src`- and then have a top level `js` and `styles` folder?) 
- The above is why webpack exports the final css file adjacent to the bundled js, I could move it out of the js folder at pipeline build time _or_ see if webpack can configure this a bit better.

teesloane:
> Tested locally and I have this weird flash where the site is loaded without CSS for couple frames:
> 
> Kapture 2021-11-02 at 17 54 48
> 
> Same thing when I navigate from page to page.

The [flashing on unstyled content](https://stackoverflow.com/a/36455916) is a result of using the style loader, which enables hot reloading with styles. This will not happen in production thanks to [these lines](https://github.com/circleci/circleci-docs/pull/5836/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fR45-R47). 

I can remove this ternary and default to just using the `MiniCssExtractPlugin.loader` but this will result in recompiling an output file in development which will trigger a full re-render of the head template (causing a full jekyll reload of 140~seconds).

.circleci/config.yml
rsylvian: Yeah, option 1 seems to make more sense to me! I'm also okay if we want to output files in a directory called `bundles/` (or a better name) and have inside the `app.js` & `app.css`.

rsylvian (approved):
Awesome work 👍 
Thanks for addressing all the feedback 🤩

Issue #5856, opened 2021-10-27
Name key not documented in Jobs in Workflow Section of the Configuration Reference 


cadecairos:
# Locations
[Jobs in Workflow section](https://circleci.com/docs/2.0/configuration-reference/#jobs-in-workflow)

# Details
There are two problems I noticed in this section:

1. [The section claims the valid keys for the maps in the workflow jobs list are: `requires`, `context`, `type`, and `filters`, and does not mention the also valid key `name`](https://github.com/circleci/circleci-docs/blob/73286a0211720095b7e07ea0717a04afb56a90cc/jekyll/_cci2/configuration-reference.md?plain=1#L1508)
2. [In the following section where you would expect to see `name` described, it appears to have been accidentally merged with the description of `requires` ](https://github.com/circleci/circleci-docs/blob/73286a0211720095b7e07ea0717a04afb56a90cc/jekyll/_cci2/configuration-reference.md?plain=1#L1527)

In summary, the jobs section should identify the key `name` as a valid key, and the documentation below needs some reformatting to display the name key properly.

devinmarieb:
@cadecairos You can check out the [PR here](https://github.com/circleci/circleci-docs/pull/5867) if you'd like to give any feedback on this before it's merged next week :) Thanks!

PR #5867, opened 2021-10-29
[WIP] Separate out name key from require key

devinmarieb:
# Description
Taking the `name` key out of the same table as the `require` key because it is its own valid key and needs its own space.

# Reasons
The `name` key was likely accidentally merged with the `require` key, but it needs its own definition and table.


thomasr8 (approved):

devinmarieb:
closes #5856

PR #5876, opened 2021-11-01
[CIRCLE-38683]: Add Property to Amplitude event to see what sections are being shared

tshen815:
**Ticket:** [CIRCLE-38683](https://circleci.atlassian.net/browse/CIRCLE-38683)

# Changes

- add new `section` property to `trackAction` for  tracking analytics to know explicitly 

# Reasons

- Adding new metadata property to our tracking analytics in amplitude for the share section button to explicitly know which sections users are sharing so we can prioritize those to update and kept up to date.  

# Screenshot

Screen Shot 2021-11-01 at 2 04 12 PM

### DIFF
--- a/src-js/site/main.js
+++ b/src-js/site/main.js
@@ -293,6 +293,7 @@ $(document).ready(function () {
         });
       },
       copy(event) {
+        let url = event.target.href;
         event.preventDefault();
         navigator?.clipboard
           .writeText(event.target.href)
@@ -316,7 +317,7 @@ $(document).ready(function () {
             window.AnalyticsClient.trackAction('docs-share-button-click', {
               page: location.pathname,
               success: true,
-              section: event.target.href,
+              section: url.substring(url.indexOf('#')),
             });
           })
           .catch((error) =>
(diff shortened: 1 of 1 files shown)


src-js/site/main.js
```diff
@@ -316,6 +317,7 @@ $(document).ready(function () {
             window.AnalyticsClient.trackAction('docs-share-button-click', {
               page: location.pathname,
               success: true,
+              section: url.substring(url.indexOf('#')),
```
rsylvian, line 320: Looks good but I don't think that will work when clicking on the main document title since there is no section? For example in https://circleci.com/docs/2.0/migrating-from-buildkite/ :

Screen Shot 2021-11-01 at 4 32 16 PM

tshen815: Thats a great point. I can try to find a way to account for that (or use the full url path) but as we also know the `page`, do you think this could work since we know `#` will denote the title link for the page that we are on when we review the data?

src-js/site/main.js
```diff
@@ -316,6 +322,7 @@ $(document).ready(function () {
             window.AnalyticsClient.trackAction('docs-share-button-click', {
               page: location.pathname,
               success: true,
+              section: section,
```
rsylvian, line 325:
```suggestion
              section,
```

rsylvian (approved):

PR #5879, opened 2021-11-01
[CIRCLE-38681]: Add property to Amplitude event to see which snippets are copied (Docs)

tshen815:
**Ticket:** [CIRCLE-38681](https://circleci.atlassian.net/browse/CIRCLE-38681)

# Changes

- add `codeSnippetPosition` property as `trackAction` metadata 

# Reasons

- we want to add more metadata to our tracking analytics, in this case to let use know explicitly what code snippets are being copied so that we may better prioritize them for updates and upkeep

### DIFF
--- a/src-js/experiments/highlightJSBadge.js
+++ b/src-js/experiments/highlightJSBadge.js
@@ -34,6 +34,7 @@ window.onload = () => {
         onBeforeCodeCopied: function (text) {
           window.AnalyticsClient.trackAction('docs-copy-code-clicked', {
             page: location.pathname,
+            code: text,
           });
           return text;
         },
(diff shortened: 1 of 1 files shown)


rsylvian (changes requested):
Sadly, I don't think adding the entire snippet is going to give us many insights. Can we track the position instead? 

For example, if I click that copy code button in https://circleci.com/docs/2.0/migrating-from-buildkite/:

Screen Shot 2021-11-01 at 4 29 28 PM

It should track that I clicked the second snippet of the page. That way we can truly understand which snippets are the most popular for a given page.

rsylvian (approved):

PR #5881, opened 2021-11-01
[CIRCLE-38539] Add DataDog Synthetics Tests as CI Job

nramabad:
# Description

We currently have two DD Synthetics tests:

1 for the landing page: https://app.datadoghq.com/synthetics/details/sfs-zqk-cfq

and 1 for the config reference page: https://app.datadoghq.com/synthetics/details/747-gcy-3tk

We need to make them part of our CI/CD so we get alerted in a PR if we are breaking something:  

# Impact

Running Synthetic tests in our CI pipeline will help us with earlier issue detection 
[CIRCLE-38539](https://circleci.atlassian.net/browse/CIRCLE-38539)


.circleci/config.yml
```diff
@@ -116,6 +116,12 @@ workflows:
           filters:
             branches:
               only: master
+      - datadog-synthetics:
+          context:
+            - web-ui-datadog
+          requires:
+            - deploy
```
rsylvian, line 123: I am not sure this is the right way to go. At this point, we are only running the tests after it deploys on `master`. If we broke something it's already too late.

I would really like it if we could test during the PR stage so we can have a signal that our proposed changes are about to break something. I was under the impression that we could tell the CLI to overwrite the test's URL: https://docs.datadoghq.com/synthetics/cicd_testing/?tab=npm#further-configuration

Anything we can do there?

rsylvian (changes requested):

tests.synthetics.json
```diff
@@ -0,0 +1,10 @@
+{
+    "tests": [
+        {
+            "id": "sfs-zqk-cfq"
```
ccgeller, line 4: what is this for? Are the id's specific for anything?

nramabad: these come from the DD synthetic test. i would include a comment w/ a URL but this is JSON instead of YAML.

.circleci/config.yml
```diff
@@ -314,27 +383,58 @@ jobs:
           name: Update Algolia Index
           command: |
             ALGOLIA_API_KEY=$ALGOLIA_PRIVATE_KEY bundle exec jekyll algolia --source jekyll --config jekyll/_config.yml
-  deploy:
+
+  deploy-preview:
```
nramabad, line 387: I moved the jobs around so that they match the order in which they are specified in `build_deploy` worflow

rsylvian (approved):
Nice job 👍  Thanks for addressing my feedback! Left a question but besides this it looks all good!

.datadog/development.synthetics.json
```diff
@@ -0,0 +1,10 @@
+{
+    "tests": [
+        {
+            "id": "as2-nqg-wcu"
+        },
+        {
+            "id": "3tu-7d6-qk7"
+        }
+    ]
```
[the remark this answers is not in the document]
nramabad: I just noticed this on the Tunneling's doc page: 

`Note: When using the testing tunnel, your tests' locations are overriden by a location that depends on your Datadog account region.`
https://docs.datadoghq.com/synthetics/testing_tunnel

ccgeller (approved):
Well done! 🥳🌺🤩

PR #5906, opened 2021-11-07
Improve ja env var page

bufferings:
# Description

Improved Japanese translation and added small fixes for English version I found.

En

* Fixed typo
* Fixed links

Ja

* Fixed header
* Fixed links
* Added translation to the untranslated sentences
* Fixed wrong translations which seems to translate different parts

# Reasons

I read the Japanese page and found there were several small things which can be improved.


thomasr8 (approved):

devinmarieb:
@bufferings Sorry for the long delay, but our _ja pages are set up a little differently and unfortunately we can not accept PRs on our Japanese content at this time. I will create an issue/ticket out of this to see if we can handle it internally. Thank you!

PR #5915, opened 2021-11-09
Fix: getSessionId not defined - now defined.

teesloane:
I needed to prepend the class the static method is called from.


ccgeller (approved):
🤩🥳🌺Great find!

PR #5921, opened 2021-11-10
Fix: Add a `experimentContainer` to accurately track `Experiment Viewed` events

rsylvian:
# Description
- Change `getVariationName()` definition to require a `experimentContainer` attribute
- Return early when tracking `Experiment Viewed` if the `experimentContainer` is not in the DOM
- Update all experiments to provide an `experimentContainer` attribute

# Reasons
We realized that we track an experiment as viewed as soon as we get a resolution from optimizely (`control` or `treatment`). This works well for experiments that are globally seen across the entire site (like the `Doc/DevHub merge` exp) but produces bogus results for experiments that are only focusing on a subset of pages (like the `Recomended Resources` exp).

By providing an `experimentContainer` we can make sure that the experiment is in the DOM before tracking a view.


teesloane (approved):

ccgeller (approved):
🥰🙏🦾Looks good to me

PR #5931, opened 2021-11-11
Chore: Remove `.idea/` folder

rsylvian:
# Description
- Remove `.idea/*`

# Reasons
This directory is created by a JetBrains IDE and holds user preferences that don't need to be in git


teesloane (approved):

ccgeller (approved):
Great ! 🥰🦾🙏

PR #5932, opened 2021-11-12
Fix: Remove jekyll-minibundle

rsylvian:
# Description
- Remove dependency to `jekyll-minibundle` in Gemfile & jekyll `config.yml`.
- Update includes to not use `jekyll-minibundle` (called `ministamp`) anymore.

# Reasons
@teesloane and I are unsure at this point why/if `jekyll-minibundle` is useful. It is causing a bunch of problems in dev because it is [incompatible](jekyll-minibundle) with `--incremental`. Some of the examples includes: broken icons in the sidebar and JS files not found.

Currently, it is only used to md5 fingerprint some assets (not all) in prod and it doesn't seem like other non-fingerprinted assets are giving us any trouble? I'm suggesting that we try this for a while and monitor datadog/rollbar to look for any issues.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,4 @@ COPY Gemfile Gemfile.lock ./
 RUN gem install bundler && bundle install
 RUN bundle exec jekyll clean
 
-ENV JEKYLL_MINIBUNDLE_MODE=development
-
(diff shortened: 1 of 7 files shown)


ccgeller (approved):
Everything still works for me! 😁😊🙏

teesloane (approved):

PR #5935, opened 2021-11-15
redefine on which branches server PDFs are built

rosieyohannan:
Previously, the server PDF build script: /scripts/build_pdfs_asciidoc.sh was run for any branch starting with `server/*`. This needs to change to `*/server` so that we can start branch names with the jira ticket reference. 

I am not sure I've done it right though! 😬


ccgeller (approved):
Great! 🧚👑🌺

PR #5937, opened 2021-11-15
Fix: add missing experimentContainer.

teesloane:
Fixes ["(unknown): Script error."](https://rollbar.com/circle/circleci-docs/items/9/)

### DIFF
--- a/src/js/site/main.js
+++ b/src/js/site/main.js
@@ -346,7 +346,7 @@ $(document).ready(function () {
   window.OptimizelyClient.getVariationName({
     experimentKey: 'dd_share_section_icon_test',
     groupExperimentName: 'q3_fy22_docs_disco_experiment_group_test',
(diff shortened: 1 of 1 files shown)


src/js/site/main.js
```diff
@@ -346,6 +346,7 @@ $(document).ready(function () {
   window.OptimizelyClient.getVariationName({
     experimentKey: 'dd_share_section_icon_test',
     groupExperimentName: 'q3_fy22_docs_disco_experiment_group_test',
+    experimentContainer: '.article-container',
```
rsylvian, line 349:
```suggestion
    experimentContainer: '.external-link-tag-wrapper',
```

rsylvian (approved):
Thanks for taking care of it. Made a suggestion about using a container that is more specific to the experiment.

PR #5939, opened 2021-11-15
Add new section for scheduled pipelines feature

kbatuigas:
# Description
Scheduled pipelines is a new feature that allows pipelines to be triggered based on a schedule. This feature is a new method of schedule management. It can currently be configured via the v2 API (UI coming soon).

We'll base this draft on the doc currently available here: https://github.com/CircleCI-Public/api-preview-docs/blob/master/docs/scheduled-pipelines.md 

# Reasons
This PR will remain in draft status until around GA, so things may still be moved around in terms of content organization.


h42zhu (approved):
LGTM from a technical standpoint. 

The scheduled pipelines UI should be available now for open preview.

radhigulati (commented):
Looks good! I'm sure this will get changed but just need to update the part about the UI now that it is available "At this time, a UI for scheduled pipelines is not yet available."

kbatuigas:
> Two minor things:
> 
> * @h42zhu do schedules actually auto-cancel pipelines?
> * One of the other benefits we should include is consolidation of schedules. Maybe something like "Consolidated management of common schedules"
> 
> If we can take care of the above, I'll approve.

Sounds good, is there a preferred section where we should add consolidation of schedules? @salilsu

h42zhu:
> Two minor things:
> 
> * @h42zhu do schedules actually auto-cancel pipelines?
> * One of the other benefits we should include is consolidation of schedules. Maybe something like "Consolidated management of common schedules"
> 
> If we can take care of the above, I'll approve.

@salilsu The pipelines triggered by schedules will enable auto-cancel in `builds-service` by default (if the project feature flag is set and the branch is not a default branch). This makes me think that we perhaps should expose this option to allow users to opt-in or out of autocanceling when creating schedules.

salilsu:
> > Two minor things:
> > 
> > * @h42zhu do schedules actually auto-cancel pipelines?
> > * One of the other benefits we should include is consolidation of schedules. Maybe something like "Consolidated management of common schedules"
> > 
> > If we can take care of the above, I'll approve.
> 
> Sounds good, is there a preferred section where we should add consolidation of schedules? @salilsu

@kbatuigas I think in the "Overview" section, with the other bullet points

salilsu:
> > Two minor things:
> > 
> > * @h42zhu do schedules actually auto-cancel pipelines?
> > * One of the other benefits we should include is consolidation of schedules. Maybe something like "Consolidated management of common schedules"
> > 
> > If we can take care of the above, I'll approve.
> 
> @salilsu The pipelines triggered by schedules will enable auto-cancel in `builds-service` by default (if the project feature flag is set and the branch is not a default branch). This makes me think that we perhaps should expose this option to allow users to opt-in or out of autocanceling when creating schedules.

@h42zhu Let's see what the customer feedback is, we can re-evaluate in the future. Given some of the discussions from standup, maybe we want to reconsider how we manage auto-cancellation generally.

jekyll/_cci2/scheduled-pipelines.md
```diff
@@ -0,0 +1,189 @@
...
+}'
+```
+
+For additional information, refer to the **Schedule** section under the [open-api docs](https://circleci.com/docs/api/v2/).
+
+#### Use project settings
+{: #project-settings }
+{:.no_toc}
+
+1. In the CircleCI application, select the project, and navigate to **Project Settings**.
```
rosieyohannan, line 82: People sometimes struggle to find project settings - I've tried to make it clear in the first step here, maybe use the same, or similar description? https://circleci.com/docs/2.0/managing-api-tokens/#creating-a-project-api-token

jekyll/_cci2/scheduled-pipelines.md
```diff
@@ -0,0 +1,189 @@
...
+2. Navigate to **Triggers**.
+3. To create a new schedule, click **Add Scheduled Trigger**.
+4. Define the new schedule's name, timetable, pipeline parameters, and attribution actor, then save the trigger.
```
rosieyohannan, line 85: When I look at the feature I have the choice of the actor being me, or a Scheduled Actor - would be good to define what that is somewhere maybe? @salilsu?

jekyll/_cci2/scheduled-pipelines.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Migrate scheduled workflows to scheduled pipelines
+{: #migrate-scheduled-workflows }
+
+Currently, using scheduled workflows has some limitations:
```
rosieyohannan, line 90: Maybe we can add a bit more =of an intro to why people should Migrate. "The current method for scheduling work on your projects is to us the scheduled workflows feature. This feature has some limitations so you should consider migrating your scheduled workflows to the scheduled pipelines feature. The limitations include:" or something?

jekyll/_cci2/scheduled-pipelines.md
```diff
@@ -0,0 +1,189 @@
...
+        - build
+    ```
+
+#### Add workflows filtering
+{: #workflows-filtering }
+{:.no_toc}
+
+As a scheduled pipeline is essentially a triggered pipeline, it will run every workflow in the config.
+
+One way to implement workflows filtering is by using the pipeline values. For example:
```
rosieyohannan, line 134: Can we link to the pipeline values doc here?

jekyll/_data/sidenav.yml
```diff
@@ -48,6 +48,8 @@ en:
   children:
     - name: Features
       children:
+        - name: Scheduled Pipelines
```
rosieyohannan, line 51: I think maybe adding this a bit further down below the big features, maybe after env vars? @thomasr8

rosieyohannan (commented):
Thanks for this @kbatuigas!! just left some comments and suggestions, let me know if you want me to explain any of them.

salilsu (approved):
Looks good to me.

jekyll/_cci2/scheduled-pipelines.md
devinmarieb: @rosieyohannan and @salilsu Since I'm unsure of the context, I will leave this for now and we can add in at a later date.

jekyll/_data/sidenav.yml
devinmarieb: @rosieyohannan I will also leave this for now. I'm unsure of what is best here. Since we're suggesting people migrate away from workflows, it makes sense to me to have this above workflows in the navigation.

devinmarieb (approved):
good to go

PR #5940, opened 2021-11-16
Help users find examples by linking to them from where they are referenced

blag:
# Description
This PR simply adds a few links to examples and detailed documentation where they are referenced, so users can get a complete picture of how to use pipeline variables. I also de-duplicated a bit of the truthy/falsy explanation since it's more thoroughly covered in the linked-to section.

# Reasons
It's difficult to find the reference documentation for `when`/`unless` clauses and logic statements/expressions when you're digging through the documentation. If you are immensely familiar with the documentation, you know to look in the configuration reference page for this information, but users don't have that knowledge.


rosieyohannan:
Thank you so much @blag! This is really helpful! 🎉

rosieyohannan (approved):

PR #5941, opened 2021-11-16
Fix: argument order for trackExperimentViewed

rsylvian:
# Description
- Bring back parameters in order of how they are supposed to be called

# Reasons
Parameters bacame out of order after a quick refactor and so the `experimentContainer` was using the `experimentId` and the `experimentId` was using the `experimentContainer` 😬


ccgeller (approved):
Woo! Experiment container no longer a number thank you!

PR #5945, opened 2021-11-17
update runner image extension docs to use sudo

jrahme-cci:
# Description
A change to the runner docker image requires sudo to be used for some commands that require admin privileges. This change updates our extension docs to note that. 

# Reasons
Keep docs up to date with the current state of the runner docker image.


devinmarieb (approved):
thanks!

PR #5958, opened 2021-11-19

sebastian-lerner:
recommend larger resource class and/or parallelism


rosieyohannan:
@sebastian-lerner good catch! Yes I guess we were recommending the wrong thing previously - workflows are no help if a single job is timing out 👍 🎉

rosieyohannan (approved):

PR #5966, opened 2021-11-23
Fix: remove globing to prevent tests from spilling into the app.bundle.js

rsylvian:
Preview: [link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/rs/fix/remove-globing-to-avoid-tests-in-bundle-preview/?force-all)

# Description
- Use an `import/export` pattern in each directory to precisely bring in the build what is needed 

# Reasons
While working on [DD-181](https://circleci.atlassian.net/browse/DD-181), @teesloane and I realized that with the current setup, tests were spilling into the bundle and were creating problems in production. This PR ensures that only the files we actually need make it to the bundle.

# Considerations
Please thoroughly test the [preview build](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/rs/fix/remove-globing-to-avoid-tests-in-bundle-preview/?force-all) as we changed the way everything is imported. Thank you 🙏

### DIFF
--- a/src/js/services/analytics.test.js
+++ b/src/js/services/analytics.test.js
@@ -1,7 +1,6 @@
 import AnalyticsClient from './analytics.js';
 import { default as CookieOrginal } from 'js-cookie';
 import glob from '../../../jest/global';
-import { describe } from 'jest-circus';
 
 jest.mock('js-cookie');
 const Cookie = CookieOrginal;
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -2,23 +2,19 @@
 
 import * as Cookie from 'js-cookie';
 import * as highlightjsBadge from 'highlightjs-badge';
-// site files
-import OptimizelyClient from './services/optimizely.js';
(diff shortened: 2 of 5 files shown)


src/js/services/analytics.test.js
```diff
@@ -1,7 +1,6 @@
 import AnalyticsClient from './analytics.js';
 import { default as CookieOrginal } from 'js-cookie';
 import glob from '../../../jest/global';
-import { describe } from 'jest-circus';
```
rsylvian, line 4: this is probably automatically imported by my VS Code 😬

tshen815 (approved):
All the side nav sections are working well in preview link and locally, thanks for the speedy resolution!

teesloane (approved):

src/js/services/analytics.test.js
inderashah007: that no original web from CircleCI became i'm develop that

PR #5970, opened 2021-11-23
Provide separate installation guides for AWS and GCP platforms

rosieyohannan:
We have had feedback that the server 3 installation guide is difficult to follow in part because steps for both AWS and GCP installations are presented together on the same pages. There are two ways to tackle this:

* use asciidoc conditionals to present AWS or GCP steps depending on which attribute is set: https://docs.asciidoctor.org/asciidoc/latest/directives/ifdef-ifndef/#ifdef. I've added `ifdef` statements around AWS and GCP content using the attributes: `env-aws` and `env-gcp` just to one page for initial testing - I think all that's needed is a switcher on the docs page so the reader can switch to their choice of either AWS or GCP and the content will change accordingly - might be really complicated but I hope not!

* have two versions of each installation doc - 4 docs in total currently so there would be 8 - this is excessive because large parts of the text apply to both platforms.

THIS PR IS NOW JUST TO UPDATE THE PDF VERSION OF THE INSTALLATION GUIDE TO PROVIDE AN AWS AND GCP VERSION


rosieyohannan:
I'm going to change this to use inverse logic `ifndef` otherwise we will need to set both GCP and AWS attributes to true to get the content to show in the web-based docs.

catherinepope (approved):
Looks really good to me!

jekyll/_cci2/server-3-install.adoc
```diff
@@ -150,11 +150,11 @@ Or select IAM role and provide:
 * *Role ARN* - 
 https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html[Role ARN for Service Accounts] (Amazon Resource Name) for S3 bucket access.
 
-endif::env-gcp[]
+// Stop hiding from GCP PDF:
```
rosieyohannan, line 153: @catherinepope not sure what happened here, is replacing rather than adding the comment a mistake maybe?

catherinepope: Hmm, possibly a keyboard spasm! Sorry about that.

PR #5974, opened 2021-11-23
[DD-166] Add link + highlight for each insight metrics glossary

ccgeller:
Jira link: [DD-166](https://circleci.atlassian.net/jira/software/c/projects/DD/boards/477?modal=detail&selectedIssue=DD-166)
Preview [link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-166-add-link-highlighting-preview/?force-all)  

Note: You need to do ?highlight before you do #id.
Examples:
[/?highlight#overallSuccesRate-definition](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-166-add-link-highlighting-preview/2.0/insights-glossary/?highlight#overallSuccesRate-definition)

[/?highlight#workflowRuns-definition](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-166-add-link-highlighting-preview/2.0/insights-glossary/?highlight#workflowRuns-definition)

Alternative
[/?highlight=#overallSuccesRate-definition](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-166-add-link-highlighting-preview/2.0/insights-glossary/?highlight=#overallSuccesRate-definition)

[/?highlight=#workflowRuns-definition](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-166-add-link-highlighting-preview/2.0/insights-glossary/?highlight=#workflowRuns-definition)

# Description
Insight team wants to be able to link to each of the 13 metric terms that can be found on this page: Insights metrics glossary - CircleCI 
While we originally looked into a more dynamic approach, we decided to go with the easy solution first and just handle this page.

# What changed
- For each term, add a unique anchor (#runs, #totalDuration,…) so we can directly link to it via the URL Link Within a Page Using an Anchor 
- Add support for a new URL parameter called highlight which will highlight the term that we are linking to (by using a styled css)

# Before 
Screen Shot 2021-11-23 at 5 50 40 PM

# After
Screen Shot 2021-11-25 at 10 54 05 AM

# Todo:
- [x] Review colors with @danbeere 
color: #1A649A;
background-color: rgba(25, 99, 154, 0.05);

### DIFF
--- a/jekyll/_cci2/insights-glossary.md
+++ b/jekyll/_cci2/insights-glossary.md
@@ -25,15 +25,16 @@ General metrics appear across the Insights experience and can refer to different
 entities, depending on the context. For example, the `Runs` metric may refer to a
 count of workflow executions or jobs, depending on the context.
 
-| Term                  | Definition                                                                                                                                                                                                                                                                                                             |
-|-----------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| Runs                  | The count of executions in a selected time window.                                                                                                                                                                                                                                                                     |
-| Total Duration        | The sum of execution times for all workflows in a project.                                                                                                                                                                                                                                                             |
-| P95 Duration          | The 95th percentile of execution times for an entity in a selected time window (i.e. 95% of runs completed in the same or less amount of this time). <br/> _The 95th percentile is a standard measure used to interpret performance data. It provides a measure of max value when outlier or transient values are excluded._ |
-| P50 (median) Duration | The median execution time. <br/> _Medians are a better measure of central tendency than arithmetic means because they are more resilient to_ skewness _in datasets_.                                                                                                                                                         |
-| Total Credits         | The sum of credits consumed during execution.                                                                                                                                                                                                                                                                          |
-| Success Rate          | The percentage of runs that completed successfully, calculated by `100 x (Successful Runs / All Runs)`                                                                                                                                                                                                                   |
+| Term                                                 | Definition                                                                                                                                                                                                                                                                                                             |
+|------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| {::nomarkdown}<div id="insights-runs">Runs</div>{:/} | The count of executions in a selected time window.                                                                                                                                                                                                                                                                     |
+| {::nomarkdown}<div id="insights-totalDuration">Total Duration</div>{:/}         | The sum of execution times for all workflows in a project.                                                                                                                                                                                                                                                             |
+| {::nomarkdown}<div id="insights-p95Duration">P95 Duration</div>{:/}            | The 95th percentile of execution times for an entity in a selected time window (i.e. 95% of runs completed in the same or less amount of this time). <br/> _The 95th percentile is a standard measure used to interpret performance data. It provides a measure of max value when outlier or transient values are excluded._ |
+| {::nomarkdown}<div id="insights-p50Duration">P50 (median) Duration</div>{:/} | The median execution time. <br/> _Medians are a better measure of central tendency than arithmetic means because they are more resilient to_ skewness _in datasets_.                                                                                                                                                         |
+| {::nomarkdown}<div id="insights-totalCredits">Total Credits</div>{:/}         | The sum of credits consumed during execution.                                                                                                                                                                                                                                                                          |
+| {::nomarkdown}<div id="insights-successRate">Success Rate</div>{:/}         | The percentage of runs that completed successfully, calculated by `100 x (Successful Runs / All Runs)`                                                                                                                                                                                                                   |
 {: class="table table-striped"}
+{: #insights-table}
 
 ## Organization-level metrics
 {: #organization-level-metrics }
@@ -44,11 +45,12 @@ Organization-level metrics allow you to analyze your organization’s performanc
 
 | Term                    | Definition     |
 |-------------------------|----------------|
-| Workflow Runs | The count of all workflows executions for an organization for the selected projects in the selected time frame. |
-| Total Workflow Duration | The total execution time of all workflows runs for an organization within the selected projects and time frame. |
-| Total Credits | The total credits consumed across all selected projects in an organization. |
-| Overall Success Rate | The percentage of runs that completed successfully across all runs in the selected projects and workflows. (calculated by `100 x (All Successful Runs/ All Runs)`) |
+| {::nomarkdown}<div id="insights-workflowRuns">Workflow Runs</div>{:/} | The count of all workflows executions for an organization for the selected projects in the selected time frame. |
+| {::nomarkdown}<div id="insights-totalWorkflowDuration">Total Workflow Duration</div>{:/} | The total execution time of all workflows runs for an organization within the selected projects and time frame. |
+| {::nomarkdown}<div id="insights-totalCredits">Total Credits</div>{:/} | The total credits consumed across all selected projects in an organization. |
+| {::nomarkdown}<div id="insights-overallSuccesRate">Overall Success Rate</div>{:/} | The percentage of runs that completed successfully across all runs in the selected projects and workflows. (calculated by `100 x (All Successful Runs/ All Runs)`) |
 {: class="table table-striped"}
+{: #insights-table}
 
 ## Workflow Metrics
 {: #workflow-metrics }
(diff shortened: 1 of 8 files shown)


jekyll/_cci2/insights-glossary.md
```diff
@@ -25,15 +25,16 @@ General metrics appear across the Insights experience and can refer to different
...
+| Term                                                 | Definition                                                                                                                                                                                                                                                                                                             |
+|------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| {::nomarkdown}<div id="insights-runs">Runs</div>{:/} | The count of executions in a selected time window.                                                                                                                                                                                                                                                                     |
```
ccgeller, line 30: I tried to just add {: #insights-runs} but it did not work. If I don't add in `{::nomarkdown}` then the `div` is interpreted just as plain text

src/js/site/main.js
```diff
@@ -395,3 +395,17 @@ $(document).ready(function () {
...
+$(document).ready(function () {
+  var urlParts = window.location.href.split('#');
+  if (urlParts && urlParts[1]) {
+    var highlight = urlParts[1].split('?');
+    if (highlight[1] == 'highlight') {
+      $('#' + highlight[0]).each(function () {
+        $(this).attr('class', 'highlight');
+      });
+    }
+  }
```
ccgeller, line 410: I decided to use jquery and css to do the highlight instead of using a styled <mark></mark>

rsylvian (commented):
@ccgeller and I talked this afternoon about some of the gotchas we discovered in the preview build. Chantelle is going to do more work on it 🙌

thomasr8 (approved):

src/js/site/main.js
```diff
@@ -395,3 +395,17 @@ $(document).ready(function () {
       }
     });
 });
+
+// insights page column highlights
+// highlight will only work if inside of #inights-table
+$(document).ready(function () {
+  var urlParts = window.location.href.split('#');
+  if (urlParts && urlParts[1]) {
+    var highlight = urlParts[1].split('?');
```
tshen815, line 404: lets us either `let` or `const` since we have es6 now 😄

ccgeller: Good catch!

src/js/site/main.js
rsylvian: Let's add some tests for this please 🙏

src/js/site/main.js
```diff
@@ -395,3 +395,26 @@ $(document).ready(function () {
...
+  const isHighlighted = searchParams.has('highlight');
+  let hash = window.location.hash;
+  if (isHighlighted && hash) {
+    $(hash).attr('class', 'highlight');
+
+    window.addEventListener(
+      'hashchange',
+      () => {
+        let hashNew = window.location.hash;
+        $(hash).attr('class', '');
```
rsylvian, line 413: This could be unreliable depending on function timings.

What we usually do is a reset first (remove the `highlight` class on everything that has the `highlight` class) and then continue with your logic. This prevents having to know the previous hash. 

```suggestion
        $('.highlight').removeClass('highlight');
```

src/js/site/main.js
```diff
@@ -395,3 +395,26 @@ $(document).ready(function () {
...
+
+// This function will highlight a dom element with the ID found in the URL fragment
+// Added to draw attention to specific entries in only the insights table
+// Highlight can be generalized to rest of documentation later see main.scss
+$(document).ready(function () {
+  const searchParams = new URLSearchParams(window.location.search);
+  const isHighlighted = searchParams.has('highlight');
+  let hash = window.location.hash;
+  if (isHighlighted && hash) {
+    $(hash).attr('class', 'highlight');
```
rsylvian, line 407:
```suggestion
    $(hash).addClass('highlight');
```

src/js/site/main.js
```diff
@@ -395,3 +395,26 @@ $(document).ready(function () {
...
+
+    window.addEventListener(
+      'hashchange',
+      () => {
+        let hashNew = window.location.hash;
```
rsylvian, line 412:
```suggestion
        const hashNew = window.location.hash;
```

rsylvian (changes requested):
Works great on my end! 🎉  - Made a couple of suggestions on how to improve the main function 😊

src/js/site/insightsTableScroll.js
```diff
@@ -0,0 +1,23 @@
+// This function will highlight a dom element with the ID found in the URL fragment
+// Added to draw attention to specific entries in only the insights table
+// Highlight can be generalized to rest of documentation later see main.scss
+export function insightsTableScroll() {
```
teesloane, line 4: This function could be used to run on any article / page. Can we rename this to something more general?

src/js/site/insightsTableScroll.test.js
```diff
@@ -0,0 +1,75 @@
...
+  describe('insightsTableScroll', () => {
+    beforeEach(() => {
+      jest.spyOn(window, 'addEventListener').mockImplementationOnce(() => {
+        const hashNew = window.location.hash;
+        $('.highlight').removeClass('highlight');
+        $(hashNew).attr('class', 'highlight');
+      });
+    });
+
+    it('ishighlighted false add class', () => {
```
teesloane, line 24: Please change the text of the `it` block to be more descriptive / fix spelling.

src/js/site/insightsTableScroll.test.js
```diff
@@ -0,0 +1,75 @@
...
+      Object.defineProperty(window, 'location', {
+        writable: true,
+        value: location,
+      });
+
+      insightsTableScroll();
+      expect($(window.location.hash)[0].classList[0]).not.toEqual('highlight');
+    });
+
+    it('ishighlighted true no added class', () => {
```
teesloane, line 39: Same as above, this does not read very easily.

src/js/site/insightsTableScroll.test.js
```diff
@@ -0,0 +1,75 @@
+import { insightsTableScroll } from './insightsTableScroll';
+
+window.$ = require('src-api/source/javascripts/lib/_jquery.js');
```
teesloane, line 3: We shouldn't require jquery from src-api...
I susepct you have written some tests that use `$` and we don't have jquery available in our test runs. So one of two possible fixes are:

1. Make jquery available by installing it as a package and making it importable to jest tests.
2. Use vanilla js dom selectors.

I think `2` makes the most sense here and we can pair on setting this up if you want.

src/js/site/insightsTableScroll.test.js
```diff
@@ -0,0 +1,75 @@
...
+
+document.body.innerHTML = '<div id="overallSuccesRate-definition">' + '</div>';
```
teesloane, line 5: Is there a reason you are using string concatenation here? It seems that there is no content in this html that you are concat'ing between the two divs? IN which case you could do this?

`'<div id="overallSuccesRate-definition"></div>';`  
?

src/js/site/insightsTableScroll.test.js
```diff
@@ -0,0 +1,75 @@
...
+        search: '?highlight',
+        hash: '#overallSuccesRate-definition',
+      };
+      Object.defineProperty(window, 'location', {
+        writable: true,
+        value: location,
+      });
+      let oldHash = window.location.hash;
+      insightsTableScroll();
+      expect($(oldHash)[0].classList[0]).toEqual('highlight');
```
teesloane, line 51: An example here to use js instead of jquery:

```
var oldHashEl = document.querySelector(oldHash)
expect(oldHashEl).... // etc

```

teesloane (changes requested):
Nice work! I requested a couple small changes. Let me know if you have any questions.

src/js/site/insightsTableScroll.test.js
ccgeller: At first I had some items in the middle. Simply just didn't clean up thanks :)

src/js/site/insightsTableScroll.test.js
ccgeller: That be great!

src/js/site/insightsTableScroll.js
ccgeller: Renamed to highlightURLHash, since it actually doesn't do any scrolling only highlighting

teesloane (approved):

src/js/site/highlightURLHash.test.js
```diff
@@ -0,0 +1,74 @@
+import { highlightURLHash, addHighlightClassIfInUrl } from './highlightURLHash';
+
+document.body.innerHTML = '<div id="overallSuccesRate-definition"></div>';
```
rsylvian, line 3: This should not be here as it will be in every test run after that. Suggesting that you put it in the `beforeEach` and perform a removal in the `afterEach`

rsylvian (approved):

PR #5975, opened 2021-11-24
[DD-35] Promote HighlightJS copy code snippet button experiment & Improve design

nramabad:
[DD-35](https://circleci.atlassian.net/browse/DD-35)
[Preview link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/docs-disco/promote-highlightjs-copy-code-btn-preview/2.0/writing-yaml/?force-all)

# Description

Now that we know that the Copy code button experiment is successful, we want to slightly tweak the design so it matches more closely our guidelines. 

First, we don’t want to always see the button, and only show it when hovering on the code snippet (this is consistent with what other big players are doing in the industry. For example Github):
image

On hover, button appears:
image

When pressing button, it gets slightly darker
image

Clicking on it copies the code in the pasteboard and display a visual cue that everything was successful: 
image

# Designs:

[Clickthrough prototype](https://www.figma.com/proto/HrgpUU2M3ED7OT0p0f99bH/Docs-Disco-Experiments?page-id=3818%3A7850&node-id=3897%3A7916&viewport=312%2C48%2C0.1&scaling=scale-down&starting-point-node-id=3897%3A7916)

Note: Please disregard everything that has to do with the Was this helpful? feedback mechanism. This will most likely be a future experiment we are going to work on.

Copy code button: [Default](https://www.figma.com/file/HrgpUU2M3ED7OT0p0f99bH/Docs-Disco-Experiments?node-id=3897%3A7976)
Copy code button: [Hover](https://www.figma.com/file/HrgpUU2M3ED7OT0p0f99bH/Docs-Disco-Experiments?node-id=3897%3A7995)
Copy code button: [Press](https://www.figma.com/file/HrgpUU2M3ED7OT0p0f99bH/Docs-Disco-Experiments?node-id=3897%3A8008)
Copy code button: [Copied](https://www.figma.com/file/HrgpUU2M3ED7OT0p0f99bH/Docs-Disco-Experiments?node-id=3897%3A7921)

**Design note:** This button style and it’s states is based on the secondary button component in Compass Design System

# Impact:

More consistent (and cleaner) design across our online properties

# Before:

**Unclicked & Unhovered**
Screen Shot 2021-11-24 at 12 25 23 PM

**Clicked**
Screen Shot 2021-11-24 at 12 26 27 PM

# After:

**Unclicked & Unhovered**
Screen Shot 2021-11-24 at 12 29 24 PM
**Hovered**
Screen Shot 2021-11-24 at 12 29 30 PM
**Full Video**
https://user-images.githubusercontent.com/42252054/143309468-3095fce9-b715-4e8c-b2cf-3e7566813392.mov


src/styles/main.scss
```diff
@@ -507,12 +507,37 @@ h6{
 }
 
 
-/*
-*  Experiment Key: dd_add_copy_code_button_to_docs
-*  https://app.optimizely.com/v2/projects/16812830475/experiments/20580130094/variations
-*/
+.code-badge {
+	opacity: 0!important;
```
nramabad, line 511: use of `!important` is how highlight js extension recommends to add styling tweaks much to the chagrin CSS enthusiasts everywhere.

src/js/services/highlightJSBadge.js
```diff
@@ -0,0 +1,71 @@
...
+    // CSS class(es) used to render the done icon.
+    checkIconClass: 'fa fa-check text-success',
+    checkIconContent: '',
+
+    // function called before code is placed on clipboard that allows you inspect and modify
+    // the text that goes onto the clipboard. Passes text and code root element (hljs).
+    // Example:  function(text, codeElement) { return text + " $$$"; }
+    onBeforeCodeCopied: function (text, codeElement) {
+      if (!codeSnippetButtons) {
+        codeSnippetButtons = document.getElementsByClassName('code-badge');
```
nramabad, line 31: no longer have to find this on every copy code click on a page. do it once per page and cache it.

src/js/services/highlightJSBadge.js
```diff
@@ -0,0 +1,71 @@
...
+      return text;
+    },
+  };
+  window.highlightJsBadge(options);
+
+  if (!codeSnippetButtons) {
+    codeSnippetButtons = document.getElementsByClassName('code-badge');
+  }
+
+  const eventType = `${window.PointerEvent ? 'pointer' : 'mouse'}down`;
```
nramabad, line 54: chrome switched to `pointerdown` rather than `mousedown` in Chrome 55.

src/js/services/highlightJSBadge.js
nramabad: we use `code-badge` instead of `hljs` because it is useful to the `mousedown` darker background transitory state logic.

src/js/services/highlightJSBadge.js
```diff
@@ -0,0 +1,71 @@
...
+    // Example:  function(text, codeElement) { return text + " $$$"; }
+    onBeforeCodeCopied: function (text, codeElement) {
+      if (!codeSnippetButtons) {
+        codeSnippetButtons = document.getElementsByClassName('code-badge');
+      }
+
+      const codeSnippetPosition =
+        Array.prototype.indexOf.call(
+          codeSnippetButtons,
+          codeElement.previousSibling,
```
nramabad, line 37: we use `previousSibling` to get the sibling `code-badge` node to the `codeElement`'s `hljs` `div`.

src/js/services/highlightJSBadge.js
```diff
@@ -0,0 +1,71 @@
...
+    // function called before code is placed on clipboard that allows you inspect and modify
+    // the text that goes onto the clipboard. Passes text and code root element (hljs).
+    // Example:  function(text, codeElement) { return text + " $$$"; }
+    onBeforeCodeCopied: function (text, codeElement) {
+      if (!codeSnippetButtons) {
+        codeSnippetButtons = document.getElementsByClassName('code-badge');
+      }
+
+      const codeSnippetPosition =
+        Array.prototype.indexOf.call(
```
nramabad, line 35: instead of invoking spread operator and then indexOf, we can just use indexOf directly.

ccgeller (approved):
All worked for me! Thanks for the extra comments in code! 👌🤩👌

tshen815 (approved):
pulled and tested locally, good stuff!

src/js/services/highlightJSBadge.js
```diff
@@ -0,0 +1,71 @@
...
+    },
+  };
+  window.highlightJsBadge(options);
+
+  if (!codeSnippetButtons) {
+    codeSnippetButtons = document.getElementsByClassName('code-badge');
+  }
+
+  const eventType = `${window.PointerEvent ? 'pointer' : 'mouse'}down`;
+  for (let i = 0; i < codeSnippetButtons.length; i++) {
```
teesloane, line 55: I think we should use a forEach here. More readable, less prone to errors.

teesloane (changes requested):

rsylvian (changes requested):
I still need to spend more time reviewing but one thing I noticed is that the check icon doesn't match Dan's design:

yours:
Screen Shot 2021-11-29 at 9 03 52 AM

design:
image

Can we fix this, please?

src/js/services/highlightJSBadge.js
nramabad: I can't say I've heard of the errors thing. do you have something I can read ab0ut that?

nramabad: plain for loops are faster than forEach hence why. i thought it made sense here given there could be a lot of codeSnippets on a page.

i can change it for readability reasons though.

teesloane (approved):

nramabad:
> I still need to spend more time reviewing but one thing I noticed is that the check icon doesn't match Dan's design:
> 
> yours: Screen Shot 2021-11-29 at 9 03 52 AM
> 
> design: image
> 
> Can we fix this, please?

Addressed

rsylvian:
@nramabad Sorry but the icon still doesn't match on my end:

Screen Shot 2021-11-29 at 5 14 14 PM

The checkmark appears more full (is it the right svg?) and the icon is too small compared to the requested design. 

Something else I am also noticing is that the button doesn't show the `cursor: pointer` unless you are hovering on the icon and the background of the button is not getting darker:

----

Kapture 2021-11-29 at 17 13 38

nramabad:
@rsylvian 

is there a reason we want to use an SVG instead of the default font awesome icons? We use font awesome icon for also the copy icon.

I can fix the cursor pointer thing.

The background only gets darker on press/click/mousedown as specified in the both the ticket and the figma. Please let me know if that was typo.

nramabad:
@rsylvian

so it seems like when i made styling tweaks, it broke the `mousedown` button color change. It is fixed now.

I've also gone through all the buttons / icons to make sure the pointer functionality and pixel dimensions are to specs of the ticket/figma mockup.

src/js/services/highlightJSBadge.js
rsylvian: I think that's a great comment, could you please put it in the code? Might help others in the future 🙌

rsylvian (approved):
Beautiful! LGTM ✅

Issue #5961, opened 2021-11-22
Conditional parameters in run command


andormarkus:
# Missing Information
Hi CircleCI Team,

In the <https://circleci.com/docs/2.0/reusing-config/#parameter-types>  the following syntax comes up.
`<<# parameters.all >> -a <</ parameters.all >><<^ parameters.short >> -l <</ parameters.short >>`

The is don't know how this called syntax/template is not well documented. Please can you add documentation or reference where I can read more about this syntax.

Thank you,
Andor

devinmarieb:
Hi @andormarkus, that syntax is for pipeline parameters, which you can read more about here:
https://circleci.com/docs/2.0/pipeline-variables/#pipeline-parameters-in-configuration
If this is not what you were looking for or not what you expected, let us know. 

I will go over the code snippets with an engineer to make sure they are up to date and are accurate examples. Thanks!

andormarkus:
Hi @devinmarieb, I know these are pipeline parameters. 

These are some special kind of logical statement for pipeline parameters.

devinmarieb:
@andormarkus Thanks for the clarification of what you were looking for. I will see if I can get this clarified.

devinmarieb:
@andormarkus Sorry for the late response, but with the holiday in the US last week it's was slower moving than usual. The syntax in the code example is mustache, however, after discussing with the engineers, they've determined it is out of date and we will be updating it to a newer example. While this is not guaranteed to be the final version of what we will replace the code snippet with, this is what has been suggested thus far as a possible replacement for a boolean example:
```
version: 2.1
commands:
  npm-install:
    parameters:
      clean:
        description: Perform a clean install
        type: boolean
        default: false
    steps:
      - when:
          condition: << parameters.clean >>
          steps:
            - run: npm clean-install
      - when:
          condition:
            not: << parameters.clean >>
          steps:
            - run: npm install
```

andormarkus:
@devinmarieb, we really like mustache because it allows us to dynamically change parameters which are otherwise immutable. I know several orb uses mustache and in our side if mustache support would go it would require a massive refactoring.

Job orb example:
```yaml
executor: <<^ parameters.self-hosted-executor >>executor-base-docker<</ parameters.self-hosted-executor >><<# parameters.self-hosted-executor >>executor-base-selfhosted<</ parameters.self-hosted-executor >>

parameters:
  self-hosted-executor:
    default: false
    description: "Self hosted executor to be used."
    type: boolean
[...]

PR #5984, opened 2021-11-29
Add workflow example to attach_workspace section

m477r1x:
# Description
Added note and example of required workflow step needed to persist assets from one job to the next.

# Reasons
I spent several days trying to figure out why assets were being successfully persisted from my first job, but the `attach_workspace` step in my second job was constantly erroring and downloading 0B of assets. Hopefully this note in the documentaion will help others know that this is required in order for `attach_workspace` to work correctly.


MrC-netizen:
#5961

PR #5989, opened 2021-11-30
Add runner autoscaling documentation

Stuart133:
# Description

A user guide to autoscaling runner instances

# Reasons

We have added endpoints to support an autoscaling solution - This documentation will help customers use these endpoints to set up their autoscalers


devinmarieb:
@Stuart133 I am going to look over this today. As a general observation after chatting with Jacob last week, we determined that using the word "autoscaling" is not very accurate and can be misleading for customers. If we change all instances of "autoscaling" to just "scaling" would that be okay?

devinmarieb:
@Stuart133 if you're fine with the changes that were made let me know and I can merge.

Stuart133:
> @Stuart133 if you're fine with the changes that were made let me know and I can merge.

I'm happy but I think @stalmy wanted to look at this as well

stalmy:
I have a couple things to discuss offline with y'all

devinmarieb (approved):

PR #5991, opened 2021-11-30
[WEB-610] Update long cookie expirations

tfe:
**Ticket:** [WEB-610](https://circleci.atlassian.net/browse/WEB-610)

NOTE: I was unable to test this locally, due to a problem installing the `liquid-c` gem I was unable to resolve in reasonable time. Someone should give this a try in a dev or staging environment before merging.

## Description

The goal of this ticket is to reduce our long (10-year) expirations on various cookies to 2 years. This commit changes the expiry that we write when creating cookies.

In addition, since cookie setting is conditional, we can't count on those being re-created with the current expiry date on each visit. So we add and call a function to update the expiration date on an existing cookie (but won't set the cookie if it doesn't already exist).

See also: [WEB-611]

## How to test

There are two paths to test:

#### Pre-existing cookies

A pre-existing cookie has its expiry shortened by loading a page with this new code. So load a preview page, e.g. http://static-preview.circleci.com/master/ and see that your `cci-customer` cookie has a far-future expiration date (10 years, 2031).

Then load a page from this branch, e.g. http://static-preview.circleci.com/cookie-expiry-update/, and verify your `cci-customer` cookie has the new shorter expiry date (2 years, 2023).

I did notice that in Firefox, it seemed to cache the value for this on the "Storage" tab of the dev tools, and only showed the new one when I closed and reopened the tab.

#### New cookies

This one is easier. Just delete your `cci-customer` cookie for the preview domain (or clear all cookies), load the new page, and check for the desired 2-year date.

[WEB-611]: https://circleci.atlassian.net/browse/WEB-611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/js/site/user.js
+++ b/src/js/site/user.js
@@ -1,4 +1,5 @@
 import Cookies from 'js-cookie';
+import { updateCookieExpiration } from '../utils';
 
 function setUserData(userData) {
   window.userData = userData;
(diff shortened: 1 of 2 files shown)


rsylvian (approved):
Thanks!

PR #5994, opened 2021-12-01
Feat: snippet feedback form experiement

teesloane:
[Ticket: DD-41](https://circleci.atlassian.net/browse/DD-41)

**Status:** In draft, blocked by https://github.com/circleci/circleci-docs/pull/5975

[Preview link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/tees-DD-41-preview/?force-all).

## Overview

This change adds a requested feature, as an experiment, to enable user's to provide feedback on the quality of a snippet whenever they click the `copy code` button in the top right corner. 

I will add screenshots and before/afters once out of draft mode.

### DIFF
--- a/jest/global.js
+++ b/jest/global.js
@@ -6,6 +6,8 @@ global.analytics = {
 
 global._DATADOG_SYNTHETICS_BROWSER = false;
 
+global.$ = jest.fn();
+
 class IntersectionObserver {
   constructor() {}
 
--- a/src/js/experiments/index.js
+++ b/src/js/experiments/index.js
@@ -3,6 +3,7 @@ import highlightjsBadge from './highlightJSBadge';
 import relatedResources from './relatedResources';
 import surfaceFullExampleConfig from './surfaceFullExampleConfig';
 import videoTutorials from './videoTutorials';
+import snippetFeedback from './snippetFeedback';
 
 export default {
   forceAll,
@@ -10,4 +11,5 @@ export default {
   relatedResources,
   surfaceFullExampleConfig,
   videoTutorials,
+  snippetFeedback,
 };
--- a/src/js/experiments/snippetFeedback.js
+++ b/src/js/experiments/snippetFeedback.js
@@ -62,6 +62,7 @@ class SnippetFeedback {
       onClick: (el) => {
         if (this.currentState === 'hasCopiedCode') {
           this.currentState = 'clickedYes';
+          this.trackYesOrNoButton('yes');
           this.constructFeedbackForm();
         }
       },
@@ -78,6 +79,7 @@ class SnippetFeedback {
       onClick: (el) => {
         if (this.currentState === 'hasCopiedCode') {
           this.currentState = 'clickedNo';
+          this.trackYesOrNoButton('no');
           this.constructFeedbackForm();
         }
       },
@@ -85,11 +87,32 @@ class SnippetFeedback {
     return noBtn;
   }
 
+  trackYesOrNoButton(yesOrNoString) {
+    // actions we track are: "docs-snippet-helpful-no"  and "docs-snippet-helpful-yes"
+    window.AnalyticsClient.trackAction(
+      `docs-snippet-helpful-${yesOrNoString}`,
+      {
+        originatingSnippet: this.snippetElement.textContent,
+        wasHelpful: yesOrNoString,
+        timeOfButtonClick: Date.now(),
+      },
+    );
+  }
+
+  trackFormSubmission(formContent) {
+    window.AnalyticsClient.trackAction(`docs-snippet-helpful-form-submission`, {
+      originatingSnippet: this.snippetElement.textContent,
+      feedback: formContent,
+      // for diffing this against the time that the trackYesOrNoButton was clicked
+      // as well as possible the time the copy code button was clicked (TODO add that)
+      timeOfFormSubmission: Date.now(),
+    });
+  }
+
   // We create somedom elements that constitut a feedbackform each element here
   // needs to have a uuid so that we can pull the value out of the dom once the
   // user has submitted the feedback.
   constructFeedbackForm() {
-    console.log('this current state is ', this.currentState);
     if (
       this.currentState === 'clickedNo' ||
       this.currentState === 'clickedYes'
@@ -108,9 +131,10 @@ class SnippetFeedback {
         kind: 'button',
         className: 'text-btn',
         text: 'Send',
-        onClick: (e) => {
-          console.log(textForm.value);
-          this.wasThisHelpfulContainer.innerHTML = "<div>Thank you for your feedback!</div>"
+        onClick: () => {
+          this.trackFormSubmission(textForm.value);
+          this.wasThisHelpfulContainer.innerHTML =
+            '<div>Thank you for your feedback!</div>';
         },
       });
 
@@ -122,34 +146,34 @@ class SnippetFeedback {
       this.currentState = 'hasOpenedFeedbackForm';
     }
   }
-
-  /**
-   * submitFeedbackForm pulls user feedback out of the dom and sends the content
-   * to amplitude
-   *
-   * */
-  submitFeedbackForm() {}
 }
 
-// flow
-//
-// 1. every code snippet copy button gets a click listener
-// 1a. on click, we greate a new SnippetFeedback
-// 1b. we store the associated snippt in the class,
-
+/**
+ * init queries the dom for all snipppets
+ * TODO: when that snippet's COPY CODE button is clicked...
+ *
+ * */
 function init() {
   // add event listeners to all code snippets on page
 
-  const snippets = document.querySelectorAll('code.hljs');
-  [...snippets].forEach((snippetBlock) => {
-    snippetBlock.addEventListener(
-      'click',
-      (pointerEvent) => {
-        let snippetElement = pointerEvent.target;
-        const snippetFeedback = new SnippetFeedback(snippetElement);
-      },
-      { once: true },
-    );
+  window.OptimizelyClient.getVariationName({
+    experimentKey: 'dd_docs_snippet_feedback_experiment_test',
+    groupExperimentName: 'q4_fy22_docs_disco_experiment_group_test',
+    experimentContainer: '.hljs',
+  }).then((variation) => {
+    if (variation === 'treatment') {
+      const snippets = document.querySelectorAll('code.hljs');
+      [...snippets].forEach((snippetBlock) => {
+        snippetBlock.addEventListener(
+          'click',
+          (pointerEvent) => {
+            let snippetElement = pointerEvent.target;
+            new SnippetFeedback(snippetElement);
+          },
+          { once: true },
+        );
+      });
+    }
   });
 }
 
--- a/src/js/experiments/snippetFeedback.js
+++ b/src/js/experiments/snippetFeedback.js
@@ -9,68 +9,61 @@
  * */
 export class SnippetFeedback {
   constructor(snippetElement) {
+    // Dom elements
     this.snippetElement = snippetElement;
     this.codeSnippetContainer = null;
     this.wasThisHelpfulContainer = null;
-    // mini state machine indicator - this is used to prevent more than one form being spawned per code block
-    // or to prevent a user from clicking "yes" and already clicking "yes" (or no).
-    this.currentState = 'empty'; // empty | hasCopiedCode | clickedYesOrNo | hasOpenedFeedbackForm | submittedFeedback
-    this.showWasThisHelpfulPrompt();
+    this.feedbackForm = null;
+    // For tracking user's response to be included in the final feedback form.
+    this.wasThisHelpful = null;
+    this.renderWasThisHelpfulPrompt();
   }
+
+  // -- Render functions
+  // These are responsible for constructing and attaching new dom elements
+  // based on the user's flow in the feedback process.
+  //
 
   /**
    * Construct a "Was this helpful" line, and inject it below the code snippet.
-   * Add on click handlers to the yes/no that will update class state, and send
-   * off amplitude data as well. */
-  showWasThisHelpfulPrompt() {
+   */
+  renderWasThisHelpfulPrompt() {
     this.codeSnippetContainer = this.snippetElement.closest('div.highlight');
-    if (this.currentState === 'empty') {
-      // Create some dom elements:
-      const container = this.makeElement({
-        kind: 'div',
-        className: 'was-this-helpful',
-      });
-      const textPrompt = this.makeElement({
-        kind: 'span',
-        text: 'Was this helpful?',
-      });
-      const slash = this.makeElement({ kind: 'span', text: ' / ' });
-      const yesButton = this.constructYesNoButton({ text: 'yes' });
-      const noButton = this.constructYesNoButton({ text: 'no' });
-      // append dom nodes.
-      container.appendChild(textPrompt);
-      container.appendChild(yesButton);
-      container.appendChild(slash);
-      container.appendChild(noButton);
-      this.wasThisHelpfulContainer = container;
-      this.codeSnippetContainer.appendChild(container);
-      // after creating dom elements, move the currentState forward
-      this.currentState = 'hasCopiedCode';
-    }
+    // Create some dom elements:
+    const container = this.makeElement({
+      kind: 'div',
+      className: 'was-this-helpful',
+    });
+    const textPrompt = this.makeElement({
+      kind: 'span',
+      text: 'Was this helpful?',
+    });
+    const slash = this.makeElement({ kind: 'span', text: ' / ' });
+    const yesButton = this.renderYesNoButton({ text: 'yes' });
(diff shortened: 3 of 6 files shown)


tshen815:
Could we also have the snippetFeedback appear when users click on the copy code icon? Right now it only pops up when users click on the actual code snippet block. We may get more opportunities to get feedback if the snippet appears when the code is copied as well. 

Outside of that I pulled and tested locally, everything looks and works great!

src/js/experiments/index.js
```diff
@@ -3,11 +3,13 @@ import highlightjsBadge from './highlightJSBadge';
...
 import videoTutorials from './videoTutorials';
+import snippetFeedback from './snippetFeedback';
 
 export default {
   forceAll,
   highlightjsBadge,
   relatedResources,
   surfaceFullExampleConfig,
   videoTutorials,
+  snippetFeedback,
```
tshen815, line 14: I think we may have to import this as `SnippetFeedback` ?  Was getting a console warning on it

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,188 @@
...
+ * init queries the dom for all snipppets
+ * TODO(blocked): when that snippet's COPY CODE button is clicked... (waiting for graduation of copy code buttons)
+ *
+ * */
+function init() {
+  // add event listeners to all code snippets on page
+  window.OptimizelyClient.getVariationName({
+    experimentKey: 'dd_docs_snippet_feedback_experiment_test',
+    groupExperimentName: 'q4_fy22_docs_disco_experiment_group_test',
+    experimentContainer: '.hljs',
```
tshen815, line 168: can we add a link to the experiment to optimizely here?

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * Renders a feedback form to the user, provided they have clicked
+   * on a "yes" or "no" button in the 'was this helpful' prompt.
+   * */
+  renderFeedbackForm() {
+    if (!this.feedbackForm) {
+      const container = this.makeElement({ kind: 'div', className: 'form' });
+      const prompt = this.makeElement({
+        kind: 'div',
+        text: 'Any way we can improve?',
```
tshen815, line 83: If its not too much effort, could we include a different message between `yes` and `no`? I feel like having the same message for both could be a bit confusing. Maybe something like `'How can we improve'` if users click `no`.

src/js/experiments/index.js
teesloane: Can you post the console warning you are getting?

src/js/experiments/snippetFeedback.js
teesloane: I'm going to defer to design on that one, but I think this is a good point.

tshen815: Screen Shot 2021-12-01 at 2 01 43 PM

jest/global.js
```diff
@@ -6,6 +6,8 @@ global.analytics = {
 
 global._DATADOG_SYNTHETICS_BROWSER = false;
 
+global.$ = jest.fn();
```
nramabad, line 9: this may or may not conflict w/ my unit test PR.

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,250 @@
...
+ *
+ * */
+export class SnippetFeedback {
+  constructor(snippetElement) {
+    // Dom elements
+    this.snippetElement = snippetElement;
+    this.codeSnippetContainer = null;
+    this.wasThisHelpfulContainer = null;
+    this.feedbackForm = null;
+    this.maxCharCount = 240;
```
nramabad, line 17: properties like this could easily be part of an optional properties/options argument with default fallback values if nothing is specified.

Also it's preferable for constants to be all uppercase.

https://softwareengineering.stackexchange.com/questions/319688/what-is-the-history-for-naming-constants-in-all-uppercase

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,250 @@
...
+      timeOfFormSubmission: Date.now(),
+    });
+  }
+
+  // -- Helpers --
+
+  /**
+   * Construct a dom element, setting classes, text content, class, onclick etc.
+   * */
+  makeElement({ kind, text, className, onClick }) {
```
nramabad, line 210: maybe some of these class methods like this should be made private?

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,250 @@
...
+   * trackYesOrNoButton sends a trackAction to aplitude.
+   * actions are: "docs-snippet-helpful-no"  and "docs-snippet-helpful-yes"
+   * */
+  trackYesOrNoButton(yesOrNoString) {
+    window.AnalyticsClient.trackAction(
+      `docs-snippet-helpful-${yesOrNoString}`,
+      {
+        originatingSnippet: this.snippetElement.textContent,
+        location: window.location.pathname,
+        timeOfButtonClick: Date.now(),
```
nramabad, line 184: not sure how this is done elsewhere but it's probably better for chart-making purposes and readability to convert this to an ISO string.

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,250 @@
...
+   * */
+  trackFormSubmission(formContent) {
+    window.AnalyticsClient.trackAction(`docs-snippet-helpful-form-submission`, {
+      originatingSnippet: this.snippetElement.textContent,
+      feedback: formContent,
+      location: window.location.pathname,
+      wasThisHelpful: this.wasThisHelpful,
+      // for diffing this against the time that the trackYesOrNoButton was clicked
+      // as well as possible the time the copy code button was clicked (TODO add that)
+      timeOfFormSubmission: Date.now(),
```
nramabad, line 201: same as above

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,250 @@
...
+ * */
+function init() {
+  // add event listeners to all code snippets on page
+  window.OptimizelyClient.getVariationName({
+    experimentKey: 'dd_docs_snippet_feedback_experiment_test',
+    groupExperimentName: 'q4_fy22_docs_disco_experiment_group_test',
+    experimentContainer: '.hljs',
+  }).then((variation) => {
+    if (variation === 'treatment') {
+      const snippets = document.querySelectorAll('code.hljs');
```
nramabad, line 233: seems similar to what we're doing in highlight js. do we want to consolidate this into one?

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,250 @@
...
+      [...snippets].forEach((snippetBlock) => {
```
nramabad, line 234: for (let snippetBlock of snippets)

nramabad: or a conventional for loop since index is useful for trackActions

nramabad: or if you want to use the forEach, you could call it directly off the prototype.

main point is the spread feels superfluous.

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,250 @@
...
+        snippetBlock.addEventListener(
+          'click',
+          (pointerEvent) => {
+            let snippetElement = pointerEvent.target;
+            new SnippetFeedback(snippetElement);
```
nramabad, line 239: we should probably be passing the index of the snippetElement so we can track back the index of the code snippet.

src/js/experiments/snippetFeedback.js
nramabad: we're dong a lot pretty eagerly when there's no guarantee the user will even copy a snippet to begin with.

I think it would behoove us to consolidate this part code with the highlight js service and consider how we want to do this in a more lazy fashion - whether that's by only initializing this feedback instance when a cursor hovers over a code snippet block OR by using intersection observer API.

src/js/experiments/snippetFeedback.js
teesloane: We do not have access to public/private keywords in the version of JavaScript we are using.

src/js/experiments/snippetFeedback.js
teesloane: Done.

src/js/experiments/snippetFeedback.js
teesloane: Done

src/js/experiments/snippetFeedback.js
nramabad: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#browser_compatibility am I missing something?

if we're not able to, can we follow convention of using _ before the method name? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_With_Private_Class_Features

src/js/experiments/snippetFeedback.js
teesloane: > I think it would behoove us to consolidate this part code with the highlight js service and consider how we want to do this in a more lazy fashion - whether that's by only initializing this feedback instance when a cursor hovers over a code snippet block OR by using intersection observer API.

So in this case, we'd only initiate the `getVariationName` call when we when we see a code block scroll into view? I feel like the additional code complexity (and having to run a callback on scrolling) could be more complex than an async call that won't block anything anyway?

src/js/experiments/snippetFeedback.js
tshen815: I think this may be what Nav is referring to: https://github.com/circleci/circleci-docs/blob/9022f6e1eae6196f6f714da6b97f9b569616ee91/src/js/services/highlightJSBadge.js#L31

Previously it was `const codeSnippets = [...document.getElementsByClassName('hljs')];` in `onBeforeCodeCopied`

teesloane: Hmm, I'm not sure, what do you think @nramabad ? Should we create some kind of singleton that holds some global state or whatever for common elements that come up like this?

src/js/experiments/snippetFeedback.js
teesloane: Done!

src/js/experiments/snippetFeedback.js
teesloane: Done

src/js/experiments/snippetFeedback.js
nramabad: do you think we should change the logic in highlight js to be the same? these should ideally track in the same way.

no payload restriction but is there a cost associated with storing the entire code snippet repeatedly? I'm not familiarly with the pricing of our Amplitutude & Segment. @rsylvian can you chime in?

I'm not sure how I feel about CTRL+F to find a location/page of a snippet in our code base. Content changes are a valid concern but I think we can see when changes were made in git history to understand our charts.

src/js/experiments/snippetFeedback.js
nramabad: the getVariationName should ideally be called as early as possible since that's a latency issue (and also only needs to run once per page anyways).

What can be lazier is just the scripts to add the input for a given snippet. We can discuss this in a huddle if this is still unclear.

src/js/experiments/snippetFeedback.js
rsylvian: We should not store the entire code snippet. While all the above concerns are valid, having an entire code snippet in the metric is not going to produce data/charts that are usable according to DA. Please store location instead 🙏

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,261 @@
...
+      }, 0);
+    }
+  }
+
+  /**
+   * Render the char count and set it's result on the respective element.
+   * */
+  _renderCharCount() {
+    let charCount = this.currCharCount;
+    let charCountLimited =
```
nramabad, line 170:
```suggestion
    let charCountLimited = Math.min(SnippetFeedback.MAX_CHAR_COUNT, charCount)
```

src/js/experiments/snippetFeedback.js
nramabad: sounds good to me. just only thing is to ensure you're able to write mocks/tests for it.

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,261 @@
...
+        </div>`;
+    this.currCharCountElement.innerHTML = el;
+  }
+
+  // -- Trackers --
+  // These are responsible for sending data based on user input
+  // to amplitude.
+
+  /**
+   * _trackYesOrNoButton sends a trackAction to aplitude.
```
rsylvian, line 186:
```suggestion
   * _trackYesOrNoButton sends a trackAction to amplitude.
```

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,261 @@
...
+   * actions are: "docs-snippet-helpful-no"  and "docs-snippet-helpful-yes"
+   * */
+  _trackYesOrNoButton(yesOrNoString) {
+    const date = new Date();
+    window.AnalyticsClient.trackAction(
+      `docs-snippet-helpful-${yesOrNoString}`,
```
rsylvian, line 192: I think we should keep everything into one event and have the yes/no as a meta:

`docs-snippet-feedback, {helpful: true}`
`docs-snippet-feedback, {helpful: false}`

It would be easier to track how much feedback we got and not hard to separate the data between helpful/not helpful either. Thoughts?

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,261 @@
...
+  _trackFormSubmission(formContent) {
+    const date = new Date();
+    window.AnalyticsClient.trackAction(`docs-snippet-helpful-form-submission`, {
+      originatingSnippet: this.snippetElement.textContent,
+      feedback: formContent,
+      location: window.location.pathname,
+      wasThisHelpful: this.wasThisHelpful,
+      // for diffing this against the time that the _trackYesOrNoButton was clicked
+      // as well as possible the time the copy code button was clicked (TODO add that)
+      timeOfFormSubmission: date.toISOString(),
```
rsylvian, line 214: Amplitude events already have a time associated with it so I don't think that is necessary:

Screen Shot 2021-12-07 at 10 49 00 PM

src/js/experiments/snippetFeedback.js
rsylvian: Agree with @nramabad. Usually, for this sort of work if you really need the element in the DOM you will only add it once and just show it / move it to the right position once needed.

src/js/experiments/snippetFeedback.js
teesloane: I'm not sure. I imagine that many users will decline to send written feedback. If we track yes/no, at least we know how many clicked it ? If there is an associated cost that we're paying per event tracked I understand why we might not want to do it, but if not why not track it and then just not use it if it's not useful?

src/js/experiments/snippetFeedback.js
teesloane: I'm not sure I'm clear on this, so let's pair on this tomorrow after I clean up other issues and fix UI suggestions.

src/js/experiments/snippetFeedback.js
rsylvian: @teesloane What I am suggesting is just for the Click on `Yes` | `No`. Basically, group them under the same event name.

teesloane:
> I don't know how updated the preview is but it doesn't seem like the design matches the figma prototype. Not sure if you are running into issues to reproduce the design but just wanted to let you know that I am happy to pair with you if you want to work on it together 🙌
> 
> Screen Shot 2021-12-08 at 4 58 28 PM Screen Shot 2021-12-08 at 4 59 17 PM

Dan sent me a Loom and I've made the changes - the preview is up to date now!

src/js/experiments/snippetFeedback.js
teesloane: Oh! Sorry, I misunderstood. I'll fix that.

src/js/experiments/snippetFeedback.js
```diff
@@ -0,0 +1,271 @@
...
+      const prompt = this._makeElement({
+        kind: 'div',
+        className: 'can-we-improve',
+        text: 'Any way we can improve?',
+      });
+
+      const textForm = this._makeElement({
+        kind: 'textarea',
+      });
+      textForm.placeholder = 'Your feedback...';
```
rsylvian, line 103: Can we set a `maxlength` of `240` too, please?

ccgeller (approved):
Working for me!

tshen815 (approved):
This looks great, thanks for making all the changes!

rsylvian (approved):
LGTM! Let's deploy and do a production testing before officially releasing the experiment to the users 🙌

PR #6007, opened 2021-12-03
Reflect changes in sign up process - sign up with email

kbatuigas:
# Description
Describe process for signing up with email, which allows a new user to explore the app first and connect their code later.

# Reasons
New users can now sign up for a CircleCI account using only an email address (in addition the GitHub and Bitbucket logins).


thomasr8 (approved):
Looks good Kat!

PR #6008, opened 2021-12-03
Remove broken third party service from IP ranges docs

SemanticallyNull:
# Description
We use A records to share IP ranges with our customers, but they were shared using a service which converts DNS to JSON. This was down for several hours last night promting customer complaints that were outside of our control.

We should remove the third party service and share the A records directly with the customer, making sure they know that they can use whatever tool they want to retrieve them, but including an example.

# Reasons
* [Slack thread](https://circleci.slack.com/archives/C028JJL4AMN/p1638471095021200)

### DIFF
--- a/jekyll/_cci2/ip-ranges.md
+++ b/jekyll/_cci2/ip-ranges.md
@@ -118,15 +118,21 @@ IP address ranges for core services (used to trigger jobs, exchange information 
 
 The machine-consumable lists have also been updated to reflect the new IP address ranges.
 
-**Machine-consumable lists can be found below:**
+**Machine-consumable lists can be found by querying the DNS A records below:**
 
-- IP address ranges *for jobs*: [DNS A record](https://dnsjson.com/jobs.knownips.circleci.com/A.json).
+- IP address ranges *for jobs*: `jobs.knownips.circleci.com`.
 
-- IP address ranges *for core services*: [DNS A record](https://dnsjson.com/core.knownips.circleci.com/A.json).
+- IP address ranges *for core services*: `core.knownips.circleci.com`.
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/ip-ranges.md
```diff
@@ -118,16 +118,22 @@ IP address ranges for core services (used to trigger jobs, exchange information
 
 The machine-consumable lists have also been updated to reflect the new IP address ranges.
 
-**Machine-consumable lists can be found below:**
+**Machine-consumable lists can found by querying the DNS A records below:**
```
kbatuigas, line 121:
```suggestion
**Machine-consumable lists can be found by querying the DNS A records below:**
```

jekyll/_cci2/ip-ranges.md
```diff
@@ -118,16 +118,22 @@ IP address ranges for core services (used to trigger jobs, exchange information
...
 
-- IP address ranges *for core services*: [DNS A record](https://dnsjson.com/core.knownips.circleci.com/A.json).
+- IP address ranges *for core services*: `core.knownips.circleci.com`.
 
-- *All IP address ranges*:  [DNS A record](https://dnsjson.com/all.knownips.circleci.com/A.json).
+- *All IP address ranges*:  `all.knownips.circleci.com`.
 
 During the preview phase, this list may change. You should check regularly for updates, at least once a week.
 
+To query these, you can use any DNS resolver. Here's an example using dig with the default resolver:
```
kbatuigas, line 131:
```suggestion
To query these, you can use any DNS resolver. Here's an example using `dig` with the default resolver:
```

kbatuigas (changes requested):
Thank you! Just a couple of minor suggestions but looks good to me otherwise.

kbatuigas (approved):
Thank you, looks good!

PR #6020, opened 2021-12-06

ogii:
Adding language to reflect that the large resource class is now available to users on the Free Plan

# Description
I have added language to state that the `large` resource class is available for the linux builds on the free plan.

# Reasons
The resources available for the Free Plan were recently upgraded, and our documentation should reflect that.


rosieyohannan (approved):
Thanks for this!

PR #6021, opened 2021-12-06

sebastian-lerner:
point to the actual source of truth for resource classes and availability


rosieyohannan (approved):
Thanks @sebastian-lerner !

PR #6022, opened 2021-12-06
Add schedule related pipeline values

h42zhu:
# Description
Add the new pipeline values related scheduled pipelines to the public docs

# Reasons
As the scheduled pipelines feature went into GA, the customers should know that there're new pipeline values that are available to use in the config when it's a scheduled pipeline


devinmarieb (approved):
Made a small adjustment to remove contractions for readability purposes. Otherwise looks great, thanks!

Fernando-Abreu (approved):

PR #6028, opened 2021-12-07
[DD-72] Add unit test optimizely client

ccgeller:
Jira link: [DD-72](https://circleci.atlassian.net/browse/DD-72 )
Preview link: [link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-72-add-unit-test-OptimizelyClient-preview/?force-all)

This story is for adding unit tests for the OptimizelyClient as it is a critical part of our experiment resolving process.

# Description
Work in progress, add unit tests to optimizely client

# What changed
- Function inside of optimiely had to be exported so that tests could be ran on them
- Keys were exported such that tests are dependant on latest key

# Reasons
Better test coverage

# Impact
Unit tests help make sure we are not breaking anything critical, especially when it comes to how we resolve experiments (in that case the OptimizelyClient)

# Definition of Done:
Critical aspects of the OptimizelyClient file are thoroughly tested. This includes:

getUserId
getVariationName
trackExperimentViewed
isExperimentAlreadyViewed
storeExperimentParticipation

### DIFF
--- a/src/js/services/optimizely.js
+++ b/src/js/services/optimizely.js
@@ -2,8 +2,10 @@ import * as optimizelySDK from '@optimizely/optimizely-sdk';
 import { v4 as uuidv4 } from 'uuid';
 import { isProduction } from '../utils';
 import Cookies from 'js-cookie';
+// testing not working unless add this
+import { forceAll } from '../experiments/forceAll';
 
-const COOKIE_KEY = 'cci-org-analytics-id';
+export const COOKIE_KEY = 'cci-org-analytics-id';
 export const STORAGE_KEY = 'growth-experiments-participated';
 const optimizelyLogLevel = isProduction() ? 'error' : 'info';
 optimizelySDK.setLogLevel(optimizelyLogLevel);
@@ -64,6 +66,7 @@ class OptimizelyClient {
       // it means the current user is not ready to see an experiment and so
       // getVariationName() will resolve to "null"
       const orgId = Cookies.get(COOKIE_KEY) ?? null;
+
       if (!orgId) {
         return resolve(null);
       }
--- a/src/js/services/optimizely.js
+++ b/src/js/services/optimizely.js
@@ -148,7 +148,7 @@ class OptimizelyClient {
 // trackExperimentViewed checks if we alredy have sent the Experiment Viewed
 // event to segment/amplitude by looking into the localstorage.
 // If not, it builds the properties needed and call trackAction with it
-export function trackExperimentViewed(
+export const trackExperimentViewed = (
   orgId,
   experimentKey,
   experimentContainer,
@@ -156,7 +156,7 @@ export function trackExperimentViewed(
   variationName,
   variationId,
   userId,
-) {
+) => {
   // don't track user if the experiment is not present in the current page
   if (!$(experimentContainer).length) {
     return;
@@ -182,11 +182,11 @@ export function trackExperimentViewed(
     // store experiment participation in localstorage
     storeExperimentParticipation(orgId, experimentKey, variationName);
   }
-}
+};
 
 // isExperimentAlreadyViewed checks in the localstorage if we already
 // marked the experiment as viewed
-export function isExperimentAlreadyViewed(orgId, experimentKey) {
+export const isExperimentAlreadyViewed = (orgId, experimentKey) => {
   try {
     const experiments = JSON.parse(localStorage.getItem(STORAGE_KEY));
     return (
@@ -199,15 +199,15 @@ export function isExperimentAlreadyViewed(orgId, experimentKey) {
     // Uglify /w browserlist force us to do catch (_)
     return false;
   }
-}
+};
 
 // storeExperimentParticipation stores the experiment variationName in the
 // localstorage
-export function storeExperimentParticipation(
+export const storeExperimentParticipation = (
   orgId,
   experimentKey,
   variationName,
-) {
+) => {
   if (!orgId || !experimentKey || !variationName) {
     return;
   }
@@ -243,6 +243,6 @@ export function storeExperimentParticipation(
     // We're deliberately ignoring it so that it doesn't break the app. It'll mean a few extra
     // events are emitted, but I think that's the lesser issue.
   }
-}
+};
 
 export default OptimizelyClient;
--- a/src/js/services/optimizely.js
+++ b/src/js/services/optimizely.js
@@ -147,7 +147,7 @@ class OptimizelyClient {
 // trackExperimentViewed checks if we alredy have sent the Experiment Viewed
 // event to segment/amplitude by looking into the localstorage.
 // If not, it builds the properties needed and call trackAction with it
-export function trackExperimentViewed (
+export function trackExperimentViewed(
   orgId,
   experimentKey,
   experimentContainer,
@@ -181,14 +181,13 @@ export function trackExperimentViewed (
     // store experiment participation in localstorage
     storeExperimentParticipation(orgId, experimentKey, variationName);
   }
-};
+}
 
 // isExperimentAlreadyViewed checks in the localstorage if we already
 // marked the experiment as viewed
 export function isExperimentAlreadyViewed(orgId, experimentKey) {
   try {
     const experiments = JSON.parse(localStorage.getItem(STORAGE_KEY));
-
     return (
       experiments &&
       Object.prototype.hasOwnProperty.call(experiments, orgId) &&
@@ -199,11 +198,15 @@ export function isExperimentAlreadyViewed(orgId, experimentKey) {
     // Uglify /w browserlist force us to do catch (_)
     return false;
   }
-};
+}
 
 // storeExperimentParticipation stores the experiment variationName in the
 // localstorage
-export function storeExperimentParticipation(orgId, experimentKey, variationName) {
+export function storeExperimentParticipation(
+  orgId,
+  experimentKey,
+  variationName,
+) {
   if (!orgId || !experimentKey || !variationName) {
     return;
   }
@@ -239,6 +242,6 @@ export function storeExperimentParticipation(orgId, experimentKey, variationName
     // We're deliberately ignoring it so that it doesn't break the app. It'll mean a few extra
     // events are emitted, but I think that's the lesser issue.
   }
-};
+}
 
 export default OptimizelyClient;
--- a/src/js/services/optimizely.js
+++ b/src/js/services/optimizely.js
@@ -4,7 +4,7 @@ import { isProduction } from '../utils';
 import Cookies from 'js-cookie';
 
 const COOKIE_KEY = 'cci-org-analytics-id';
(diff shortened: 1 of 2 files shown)


src/js/services/optimizely.js
```diff
@@ -2,9 +2,11 @@ import * as optimizelySDK from '@optimizely/optimizely-sdk';
 import { v4 as uuidv4 } from 'uuid';
 import { isProduction } from '../utils';
 import Cookies from 'js-cookie';
+// testing not working unless add this
+import { forceAll } from '../experiments/forceAll';
```
nramabad, line 6: ?

rsylvian: +1, I am not sure I follow

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,265 @@
...
+  });
+
+  // Function used in TrackExperimentViewed
+  describe('isExperimentAlreadyViewed()', () => {
+    beforeEach(() => {
+      // ensure local storage clear
+      window.localStorage.clear();
+    });
+    it('experiment has not been viewed before', () => {
+      storeExperimentParticipation(orgId, experimentKey + '456', variationName);
```
rsylvian, line 100:
```suggestion
```

✂️

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,265 @@
...
+      trackExperimentViewed(
+        orgId,
+        experimentKey,
+        ['container item'],
+        '5',
+        variationName,
+        '5',
+        '5',
+      );
+      expect(spy).not.toHaveBeenCalled();
```
rsylvian, line 136: suggestion: Here we should test that the `trackAction` has been called with the correct parameters

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,265 @@
...
+        orgId,
+        experimentKey,
+        ['container item'],
+        '5',
+        variationName,
+        '5',
+        '5',
+      );
+      expect(spy).toHaveBeenCalled();
+    });
```
rsylvian, line 151: How is this different than the previous one?

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,265 @@
...
+      window.localStorage.clear();
+      jest.clearAllMocks();
+    });
+
+    afterEach(() => {
+      delete global.AnalyticsClient;
+      jest.resetAllMocks();
+    });
+
+    it('expect trackExperimentViewed to have bene viewed', () => {
```
rsylvian, line 122:
```suggestion
    it('expect trackExperimentViewed to call trackAction', () => {
```

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,265 @@
...
+  describe('getVariationName()', () => {
+    afterEach(() => {
+      jest.clearAllMocks();
+      jest.resetAllMocks();
+    });
+
+    it('test forceAll is true return treatment', async () => {
+      forceAll.mockImplementation(() => true);
+      await expect(client.getVariationName(options)).resolves.toBe('treatment');
+      expect(forceAll.mock.calls.length).toBe(1);
```
rsylvian, line 163: We don't need to test for that since this is a mock. line 162 is enough 👍

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,265 @@
...
+      const errorObject = { error: 'Missing required options' };
+      forceAll.mockImplementation(() => false);
+      await expect(client.getVariationName(null)).rejects.toEqual(errorObject);
+    });
+
+    it('test forceAll is false valid options no cookie', async () => {
+      forceAll.mockImplementation(() => false);
+      await expect(client.getVariationName(options)).resolves.toBe(null);
+      expect(forceAll.mock.calls.length).toBe(1);
+    });
```
rsylvian, line 176: I'm not sure I understand what this is testing?

rsylvian (changes requested):
Great progress so far! Left couple suggestions/comments. There are a lot of tests so I need to spend more time reviewing tomorrow 🙌

src/js/services/optimizely.js
ccgeller: I can examplain why I did it but I do not know a better way to fix the issue. I get `ReferenceError: forceAll is not defined`. I think because its a unit test not an E2E test it doesn't know what forceAll is, so if I import forceAll then it knows what it is.

src/js/services/optimizely.test.js
ccgeller: Specifically I was testing if there is no orgId because there is no cookie saved
      if (!orgId) {
        return resolve(null);
      }
 It was a branch test, I can make the test message more clear.

src/js/services/optimizely.test.js
ccgeller: there is a not in the previous one. The previous one sets up data with
storeExperimentParticipation(orgId, experimentKey, variationName);
so the first one
if(!isExperimentAlreadyViewed) will be false and the second one it will be true

src/js/services/optimizely.test.js
ccgeller: It would actually be the second one because the first one skips over trackAction but its a good point!

src/js/services/optimizely.test.js
ccgeller: I just realized my test name is confusing I updated it

src/js/services/optimizely.js
nramabad: why can't you use globs to solve that?

ccgeller: glob worked thank you!

src/js/services/optimizely.js
```diff
@@ -181,14 +182,13 @@ const trackExperimentViewed = (
     // store experiment participation in localstorage
     storeExperimentParticipation(orgId, experimentKey, variationName);
   }
-};
+}
 
 // isExperimentAlreadyViewed checks in the localstorage if we already
 // marked the experiment as viewed
-const isExperimentAlreadyViewed = (orgId, experimentKey) => {
+export function isExperimentAlreadyViewed(orgId, experimentKey) {
```
rsylvian, line 189: why remove the `const` and the `=>` function? 

This should work:
`export const isExperimentAlreadyViewed = (orgId, experimentKey) => {`

src/js/services/optimizely.js
```diff
@@ -147,15 +148,15 @@ class OptimizelyClient {
...
+export function trackExperimentViewed(
   orgId,
   experimentKey,
   experimentContainer,
   experimentId,
   variationName,
   variationId,
   userId,
-) => {
+) {
```
tshen815, line 159: could we also do a `export const trackExperimentViewed = (parameters.....) => {}` here?

src/js/services/optimizely.js
```diff
@@ -199,11 +199,15 @@ const isExperimentAlreadyViewed = (orgId, experimentKey) => {
     // Uglify /w browserlist force us to do catch (_)
     return false;
   }
-};
+}
 
 // storeExperimentParticipation stores the experiment variationName in the
 // localstorage
-const storeExperimentParticipation = (orgId, experimentKey, variationName) => {
+export function storeExperimentParticipation(
```
tshen815, line 206: same here as the previous two comments, unless if there is a reason for switching back to function :)

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,263 @@
+import OptimizelyClient, {
+  storeExperimentParticipation,
+  isExperimentAlreadyViewed,
+  trackExperimentViewed,
+  STORAGE_KEY,
+  COOKIE_KEY,
+} from './optimizely';
+import AnalyticsClient from '../services/analytics.js';
+import glob from '../../../jest/global';
```
tshen815, line 9: 👏

src/js/services/optimizely.js
ccgeller: Both work the same it was more keeping it consistent. It seems like some files are written one way and some are written another. I was matching how the most recent tests were done with highlightURLHash, Users and sidebar. I will add it to blue tape item but put it how you suggested.

src/js/services/optimizely.js
ccgeller: Same comment as above.

src/js/services/optimizely.js
ccgeller: Same comment as above. I was matching how the most recent tests were done with highlightURLHash, Users and sidebar.

ccgeller: I think think you are saying that the standard should be using the arrow function and we should move things towards using that and not move things towards being back to function.

nramabad (approved):
LGTM!

src/js/services/optimizely.js
```diff
@@ -64,6 +64,7 @@ class OptimizelyClient {
       // it means the current user is not ready to see an experiment and so
       // getVariationName() will resolve to "null"
       const orgId = Cookies.get(COOKIE_KEY) ?? null;
+
```
rsylvian, line 67:
```suggestion
```

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,241 @@
...
+
+  afterEach(() => {
+    glob.userData = null;
+    glob.forceAll = null;
+    jest.clearAllMocks();
+    jest.resetAllMocks();
+  });
+
+  describe('getUserId()', () => {
+    it('test if userData exists but analytics_id undefined', () => {
```
rsylvian, line 44: usually, test names should describe what they are expecting rather than testing. For example for this one that would be: `returns null when userData is empty`

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,241 @@
...
+          .spyOn(client.client, 'getVariation')
+          .mockImplementationOnce(() => 'treatment');
+        jest
+          .spyOn(client.client, 'getVariation')
+          .mockImplementationOnce(() => 'treatment');
+        return client.getVariationName(options).then((data) => {
+          expect(data).toBe('treatment');
+        });
+      });
+    });
```
rsylvian, line 213: I'm struggling to understand what this is testing?

src/js/services/optimizely.test.js
```diff
@@ -0,0 +1,241 @@
...
+        variationName,
+        '5',
+        '5',
+      );
+      expect(spy).toHaveBeenCalledWith('Experiment Viewed', expect.any(Object));
+    });
+  });
+
+  describe('getVariationName()', () => {
+    it('test forceAll is true return treatment', async () => {
```
rsylvian, line 146: Same comment as above for test names. Here I will do:

```suggestion
    it('returns treatment when forceAll is true', async () => {
```

rsylvian (changes requested):

rsylvian (approved):
Paired with @ccgeller on this. LGTM ✅

PR #6029, opened 2021-12-07
Update to newer Xcode image in macOS example config

alexazl:
# Description
Updated the example macOS config to point to a newer Xcode image.

# Reasons
Since many customers look at our docs when getting started (and often copy the examples directly), we want to encourage them to use the latest images.


devinmarieb (approved):
Thanks!

PR #6031, opened 2021-12-08
[DD-151]: Minimize images for Prod build

tshen815:
**Ticket:** [DD-151](https://circleci.atlassian.net/browse/DD-151)
**Preview Link:** http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/minimize-images-preview/?force-all

# Changes
- add `process_images.sh` script to recursively compress images with imagemin while keeping the original directory structure for final output
- add `imagemin-pngquant` abd `imagemin-cli` as devDependencies 
- add `Minimize-Images` job to `build-deploy` workflow 

# Description
Currently the images we use for the docs site takes up around 49.2 MB. We want to reduce the overall size of the images to improve performance when loading a page that uses images and to have a smaller bundle size when uploading to S3.

# Before
Screen Shot 2021-12-07 at 6 20 47 PM

# After
Screen Shot 2021-12-07 at 6 21 15 PM


.circleci/config.yml
```diff
@@ -199,6 +205,7 @@ jobs:
             mv scripts/datadog-synthetics-preview.js /tmp/workspace/scripts/datadog-synthetics-preview.js
             mv scripts/datadog-synthetics-ci.sh /tmp/workspace/scripts/datadog-synthetics-ci.sh
             mv scripts/postinstall.sh /tmp/workspace/scripts/postinstall.sh
+            mv scripts/process_images.sh /tmp/workspace/scripts/process_images.sh
```
tshen815, line 208: moving script for compressing images to `/tmp/workspace/` for ssh

.circleci/config.yml
```diff
@@ -98,6 +98,12 @@ workflows:
             - js_build
             - build_server_pdfs
             - build_api_docs
+      - optimize-images:
+          requires:
+            - build              
+          # filters: 
+          #   branches:
+          #     only: master
```
tshen815, line 106: currently testing out the new job, will comment out afterwards to have `optimize-images` run only when merging with master

.circleci/config.yml
nramabad: nit: can we keep rename it to process-image.sh or rename the other files to follow this naming convention?

teesloane:
Possible scope creep question:  are we considering resizing of images to be "optimization"?  For example, if someone adds a screenshot to a document that is 3500x2200 pixels, we could resize that to be half the width and it would (arguably) still be sufficient for docs).

rsylvian:
> Possible scope creep question: are we considering resizing of images to be "optimization"? For example, if someone adds a screenshot to a document that is 3500x2200 pixels, we could resize that to be half the width and it would (arguably) still be sufficient for docs).

I agree with you but this is definitely out of scope and should not be attempted as part of that ticket. Plus, `imagemin` doesn't handle resizing so we would have to use something like `imagemagick` instead

.circleci/config.yml
tshen815: uncommented filter for `minimize-images` job to have it run only when we merge into master due to successful testing of `minimize-images` in preview branch

tshen815:
> Possible scope creep question: are we considering resizing of images to be "optimization"? For example, if someone adds a screenshot to a document that is 3500x2200 pixels, we could resize that to be half the width and it would (arguably) still be sufficient for docs).

You make a great point, I've updated the job name to `Minimize Images` to better reflect what its actually accomplishing.

package.json
```diff
@@ -22,7 +22,8 @@
     "webpack-prod": "NODE_ENV=production webpack --mode=production",
     "webpack-watch": "webpack --watch --mode=development",
     "test": "jest --runInBand",
-    "test:ci": "yarn test --coverage"
+    "test:ci": "yarn test --coverage",
+    "minimize-img": "./scripts/process-images.sh _site/docs/assets/img"
```
tshen815, line 26: directory passed as argument to `process-images.sh` needs to match file path in `/tmp/workspace/`

teesloane:
Final point - can we make sure that we get both docs and design eyes on this before we merge to master? It'd be great if we can get multiple thumbs-up on how the actual compressed image quality looks before we commit to compressing on every push.

tshen815:
> Final point - can we make sure that we get both docs-team and design-team eyes on this before we merge to master? It'd be great if we can get multiple thumbs-up on how the actual compressed image quality looks before we commit to compressing on every push.

@danbeere would you be able to review this on the behalf of the design team? Happy to walk you through it or take a look with you.

ccgeller (approved):
Great work on this!

.circleci/config.yml
```diff
@@ -98,6 +98,12 @@ workflows:
             - js_build
             - build_server_pdfs
             - build_api_docs
+      - minimize-images:
+          requires:
+            - build              
+          filters: 
+            branches:
+              only: master
```
nramabad, line 106: we should probably run this for preview branches as well to help catch if an image is breaking.

tshen815: Good call on this, will be addressed

package.json
nramabad: I would ask @rsylvian how we feel about hardcoding this w/ `docs` in the path. I see this can work if we're only running it only on master branch but it creates 2 issues.

1. If we ever make changes to our S3 bucket structure, this is a hardcoded path not reliant on `JEKYLL_BASENAME`.
2. no true testability of images w/ preview branches

tshen815: Ill see if romain is able to sync on that today

.circleci/config.yml
rsylvian: I'm not against this! Good call

package.json
rsylvian: Good callout @nramabad. Would be great to avoid hardcoding the path if that's an easy lift.

nramabad (approved):
LGTM! Thanks for going extra distance to rm the hardcoded paths!

.circleci/config.yml
```diff
@@ -73,7 +73,7 @@ commands:
       - run:
           name: Populate JEKYLL_BASENAME env var
           command: |
-            if [ ${CIRCLE_BRANCH} = master -o ${CIRCLE_BRANCH} = main ]; then
+            if [ ${CIRCLE_BRANCH} = master -o ${CIRCLE_BRANCH} = main -o ${CIRCLE_BRANCH} = /.*-preview/ ]; then
```
rsylvian, line 76: 👏 👏 👏 👏

rsylvian (approved):
Awesome! Everything looks good on my end ✅

PR #6042, opened 2021-12-09
Add banner for scheduled workflows, fix link

devinmarieb:
# Description
Adding a banner to the scheduled workflows section of our docs to indicate that scheduled workflows are sunsetting in the spring. I also randomly found a link in an unrelated page that was an incorrect format, so I fixed that.

# Reasons
There are still a lot of people who haven't migrated scheduled workflows to scheduled pipelines after communication has gone out about it. Hopefully this banner will help prevent new users from setting them up as well as remind users that the feature is being phased out.
s_workflow_banner


thomasr8 (approved):
Looks good!

salilsu (changes requested):
I think we should be more specific about when it is going to be removed. We've announced a sunset date of June 3rd, 2022.
cc @radhigulati

PR #6043, opened 2021-12-10
Orbs view UI

kbatuigas:
# Description
The CircleCI app now includes an Orbs page, allowing members of an organization to view a list as well as details of orbs authored by their organization.

# Reasons
Previously, a list of an organization's orbs could only be retrieved via the CLI, and the CLI command does not list private orbs by default. Additionally, while public orbs' source and documentation can easily be viewed on the Dev Hub, this information for private orbs is not easily accessible. The orbs (list) page and individual details page make this information more visible and accessible to users.


thomasr8 (approved):
Looks Great!

PR #6049, opened 2021-12-13
Fix/machine image listing

### LINKED ISSUE #5907: Searching for machine image info
# Locations
https://circleci.com/docs/2.0/executor-types/#using-machine

# Details
The text at the top:

Legacy images with the prefix "circleci/" will be deprecated on December 31, 2021. For faster builds, upgrade your projects with next-generation convenience images.

Links to here: 
https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/

But these images aren't compatible with the machine executor...

Where is the list of available machine images?

I found this, which seemed useful, but, is an announcement (not in docs) and is only for a specific image.
https://discuss.circleci.com/t/linux-machine-executor-images-october-q4-update/37847

There's this, but, no discussion of what's on each image:
https://circleci.com/docs/2.0/configuration-reference/#machine

Googling a specific image I can find the announcement for it and thus get its requirements, so here's the latest one:
https://discuss.circleci.com/t/linux-machine-executor-images-2021-july-q3-update/40772

Am I missing something here? Is there a doc with all this info in one place?

rosieyohannan:
Fixes #5907 link to dev hub and discuss from config reference to get people to the right place to find full machine image listings.

future work - we should have a wider docs effort to ensure we're linking to the dev hub as the source of truth when appropriate

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -368,7 +368,7 @@ jobs:
 ##### Available `machine` images
 {: #available-machine-images }
 
-**Specifying an image in your config file is strongly recommended.** CircleCI supports multiple machine images that can be specified in the `image` field:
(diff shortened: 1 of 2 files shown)


devinmarieb (approved):
👍

PR #6053, opened 2021-12-13
Put GHE & CircleCI on same domain

ryanwohara:
SERVER-1497

Warn users that failing to host CircleCI & GHE on the same
domain will lead to images failing to load. (This is a GH bug)


thomasr8 (approved):
Looks good 👍

PR #6058, opened 2021-12-13
Move submitting a PR

devinmarieb:
# Description
Move the PR instructions up, as well as making a note in the first section that this process is different.

# Reasons
Internally we have some non-developers who are asked to make changes to their docs for their teams. This makes the instructions a little more clear for non-developers.


thomasr8 (approved):
Looks great! 👍

thomasr8 (approved):
Looks good! This should be really helpful to all of us, as well as other users who want to develop and build locally.

PR #6059, opened 2021-12-13
Scheduled Pipelines are not yet available in 3.x

vjpandian:
# Description
Scheduled pipelines are not yet available in Server 3.x and will not be available until 3.4. This needs to be accurately reflected in our Docs page.

# Reasons
Received a question from a customer about Scheduled Pipelines not being available in their installation. They noticed that the feature is being [listed as available in Server 3](https://circleci.com/docs/2.0/scheduled-pipelines/#overview) in our docs page.


devinmarieb (approved):
thanks!

PR #6061, opened 2021-12-14
DD-265 surface last updated date on the docs page

ccgeller:
Jira link: [DD-265](https://circleci.atlassian.net/browse/DD-265)
Preview link: [link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-265-Surface-last-updated-date-on-the-docs-page-preview/2.0/about-circleci/?force-all)

# Description
Show last updated date under the title of the page so user know how outdated docs are. This plugin can potentially be used: GitHub - gjtorikian/jekyll-last-modified-at: A Jekyll plugin to show the last_modified_at time of a post. 

# Reasons
Users are aware when a doc is outdated

# Before
Screen Shot 2021-12-14 at 5 02 56 PM

# After
/docs/2.0/dynamic-config/
Screen Shot 2021-12-21 at 3 25 33 PM

/docs/2.0/about-circleci/
Screen Shot 2021-12-21 at 3 23 07 PM

/docs/2.0/glossary/
Screen Shot 2021-12-20 at 9 45 55 PM

## Tooltip 
/docs/2.0/about-circleci/
Screen Shot 2021-12-21 at 3 23 21 PM


thomasr8 (approved):
Looks good. Have a question though. Is there a way for us to only expose the last updated date internally? I am a little unsure about whether we want to expose this data point externally. I will check with the rest of the docs team to see how they feel.

tshen815:
Was just curious, for the longer dates did we still want it in the month format? Or did we want to do something like **>6+ MONTHS AGO** for the pages that were updated past a certain time? 

Otherwise everything looks great, awesome job on this Chantelle!! 🍾 🍾 

Screen Shot 2021-12-15 at 5 24 45 PM

heug (approved):

rosieyohannan:
Hi! Docs team have some concerns about this change. This information can be relatively meaningless. Docs will appear as recently updated even if a small change/typo fix is made to a page. Just wondering if we can look at the reasons behind this feature work to check it will be helpful to readers?

devinmarieb:
I also mirror the concerns about what a user is actually getting out of this if we can't attribute meaning to the date. I worry that most updates will be meaningless and potentially cause confusion for a user spending time looking for a change to a code snippet or something when in fact the change was just a typo. My thought was that either we keep this information internal as a way to know which pages need looking at or updating, or possibly only put the dates user facing on the most popular pages, like the configuration reference.

devinmarieb:
> Was just curious, for the longer dates did we still want it in the month format? Or did we want to do something like **>6+ MONTHS AGO** for the pages that were updated past a certain time?
> 
> Otherwise everything looks great, awesome job on this Chantelle!! 🍾 🍾
> 
> Screen Shot 2021-12-15 at 5 24 45 PM

If we do end up actually doing this, I think that this is exactly what we should do for older pages that haven't been updated in a while.

thomasr8:
Another option we have talked about was the idea of having a Docs Changelog visible to our users. This could be where we surface changes made to our docs so users could go there and see what has changed since the last time they visited. I'm not sure if there is a way to automate this or not, but it might be something worth exploring.

This is also probably "pie in the sky" but I seem to remember an idea being talked about where we might be able to determine the last time a user visited our docs site, and then show them what has happened since their last visit. Not sure if this is possible or not because of security and privacy concerns though.

src/js/site/main.js
```diff
@@ -402,3 +402,72 @@ $(document).ready(function () {
...
+    if (interval >= 1) {
+      return render(interval, 'hour');
+    }
+    interval = Math.floor(seconds / 60);
+    if (interval >= 1) {
+      return render(interval, 'minute');
+    }
+    interval = Math.floor(seconds);
+    return render(interval, 'second');
+  }
```
teesloane, line 462: This seems like something we might want to move to our [utils](https://github.com/circleci/circleci-docs/blob/407b3e51565e647f55e707eaea2b2d1f05d85f8e/src/js/utils.js) file.

src/js/site/main.js
```diff
@@ -402,3 +402,72 @@ $(document).ready(function () {
...
+
+  if (
+    document.getElementById('time-posted-on') &&
+    document.getElementById('time-posted-ago')
+  ) {
+    var date = Date.parse(
```
teesloane, line 468: const?

src/styles/main.scss
```diff
@@ -708,3 +681,73 @@ Fixing styling for the code example tables in the "Migrating From..." pages
   top: calc(50% - 11px);
   left: 18px;
 }
+
+// CSS styling for the date modifiex X ago
```
teesloane, line 685:
```suggestion
// CSS styling for the date modified X ago
```

src/styles/main.scss
```diff
@@ -708,3 +681,73 @@ Fixing styling for the code example tables in the "Migrating From..." pages
...
+  font-family: Roboto;
+  font-style: normal;
+  font-weight: normal;
+  font-size: 15px;
+  color: #343434;
+}
+
+.quick-info-list img {
+  margin: 0px 3px 3px 0px;
+}
```
teesloane, line 706: Since this is scss you can nest the `img` selector underneath the original `.quick-info-list` declaration on line 686.

src/styles/main.scss
```diff
@@ -708,3 +681,73 @@ Fixing styling for the code example tables in the "Migrating From..." pages
...
+  border: none;
+  font-family: Roboto;
+  font-style: normal;
+  font-weight: normal;
+  font-size: 15px;
+  line-height: 18px;
+  padding: 0px;
+  cursor: pointer;
+
+  &:not(:nth-child(1)):not(:nth-child(2)):before {
```
teesloane, line 720: This can be a bit of a confusing read - can you add a comment above it to explain why we have it?

src/styles/main.scss
```diff
@@ -708,3 +681,73 @@ Fixing styling for the code example tables in the "Migrating From..." pages
...
+  }
+}
+
+.server-version-wrapper {
+  padding: 10px;
+  flex-grow: 1;
+  border-left: 1px solid #DDDDDD;
+}
+
+#tooltiptime {
```
teesloane, line 737: I think we usually use kebab-case for our css.
```suggestion
#tooltip-time {
```

teesloane (changes requested):
Nice work so far. Added a few comments. I also noticed the home page has a bar on it:

image

Gemfile
```diff
@@ -19,6 +19,8 @@ group :jekyll_plugins do
   gem 'jekyll-target-blank'
   gem 'jekyll-toc'
   gem 'jekyll-asciidoc'
+  gem 'jekyll-minibundle'
```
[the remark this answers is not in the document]
ccgeller: Hmm i got it from a rebase this is a mistake

kbatuigas (approved):
Thanks!

teesloane (approved):

jekyll/_layouts/classic-docs.html
```diff
@@ -27,20 +27,33 @@
...
+              <div id='arrow' data-popper-arrow></div>
             </div>
-            {% if page.version %}
-            <div class="server-version-wrapper">
-              {% for version in page.version %}
-              <span class="badge badge-primary">
-                <button class="server-version-badge" aria-describedby="tooltip">{{ version }}</button>
+
+            <div class="quick-info-list">
+              <time aria-describedby="tooltip-time" id="time-posted-on" datetime="{{  page.last_modified_at | date_to_xmlschema }}">
```
nramabad, line 40: I like how semantic this is. this is a good opportunity to add sn itemprop SEO schema tag as well:
https://schema.org/datePublished

Let's reach out to SEO to see what they think.

more on how to use schema: https://schema.org/docs/gs.html#microdata_itemscope_itemtype

src/js/site/main.js
```diff
@@ -402,3 +403,40 @@ $(document).ready(function () {
 
 // Currently this function is only used for the insights table
 $(document).ready(highlightURLHash());
+
+// update date shown to be X ago
+$(document).ready(function () {
```
nramabad, line 408: `$()` is shorthand for `$(document).ready()`

src/js/site/main.js
```diff
@@ -402,3 +403,40 @@ $(document).ready(function () {
 
 // Currently this function is only used for the insights table
 $(document).ready(highlightURLHash());
```
nramabad, line 405:
```suggestion
$(highlightURLHash);
```
nit but can you try this without the function invocation? i believe it should work and not give a vscode warning.

src/js/utils.js
```diff
@@ -54,3 +54,35 @@ export function updateCookieExpiration(cookieName, newExpiration) {
   // Re-set cookie with same values and new expiration date
   Cookies.set(cookieName, existingValue, { expires: newExpiration });
 }
+
+// change date into how long ago that date is relative to when view page
+export function dateHowLongAgo(date) {
```
nramabad, line 59: sorta feeling like this file is long enough to be broken into several files under a util folder.

it's RIGHT about at that cusp which I'd start considering. up to you but def should break it out if add any more than this.

src/js/utils.js
```diff
@@ -54,3 +54,35 @@ export function updateCookieExpiration(cookieName, newExpiration) {
...
+  function render(n, unit) {
+    return n + ' ' + unit + (n === 1 ? '' : 's') + ' ago';
+  }
+
+  let seconds = Math.floor((new Date() - date) / 1000);
+
+  let interval = Math.floor(seconds / (60 * 60 * 24 * 365));
+  if (interval >= 1) {
```
nramabad, line 67: nit: can we do this as a switch-case?

src/js/utils.test.js
```diff
@@ -79,4 +79,102 @@ describe('Utils', () => {
       expect(result).toBe(false);
     });
   });
+
+  describe('dateHowLongAgo', () => {
```
nramabad, line 83: 👏

src/js/utils.test.js
```diff
@@ -79,4 +79,102 @@ describe('Utils', () => {
...
+    let date;
+    beforeEach(() => {
+      date = new Date();
+    });
+    afterEach(() => {
```
nramabad, line 88: this is an unnecessary afterEach, right?

jekyll/_layouts/classic-docs.html
ccgeller: Will added to DD-266

src/js/dateHowLongAgo.js
```diff
@@ -0,0 +1,23 @@
+// change date into how long ago that date is relative to when view page
+export function dateHowLongAgo(date) {
+  function render(n, unit) {
```
nramabad, line 3: nit: can we rename render to something more semantic? like `formatDate` or `dateFormatHelper`?

src/styles/main.scss
```diff
@@ -708,3 +682,76 @@ Fixing styling for the code example tables in the "Migrating From..." pages
...
+  font-size: 15px;
+  line-height: 18px;
+  padding: 0px;
+  color: #343434;
+
+  // We want to put a dot between every tag and there is undetermined number of tags
+  // First child is the tag icon
+  // Second child is the first tag name
+  &:not(:nth-child(1)):not(:nth-child(2)):before {
+    content:"\A";
```
nramabad, line 725:
```suggestion
    content: "\A";
```

src/js/dateHowLongAgo.js
ccgeller: Can do!

src/styles/main.scss
ccgeller: Good catch!

src/js/site/main.js
```diff
@@ -401,4 +402,41 @@ $(document).ready(function () {
...
 
 // Currently this function is only used for the insights table
-$(document).ready(highlightURLHash());
+$(highlightURLHash);
+
+// update date shown to be X ago
+$(function () {
+  // tooltip code for posted on time
+  const tooltiptime = document.getElementById('tooltip-time');
+  let timeposted = document.getElementById('time-posted-on');
```
nramabad, line 411:
```suggestion
  const timeposted = document.getElementById('time-posted-on');
```

src/js/site/main.js
```diff
@@ -401,4 +402,41 @@ $(document).ready(function () {
...
+  let popperInstance = null;
+  let showEvents = ['mouseenter', 'focus'];
```
nramabad, line 413:
```suggestion
  const showEvents = ['mouseenter', 'focus'];
```

src/js/site/main.js
```diff
@@ -401,4 +402,41 @@ $(document).ready(function () {
...
+  let hideEvents = ['mouseleave', 'blur'];
```
nramabad, line 414:
```suggestion
  const hideEvents = ['mouseleave', 'blur'];
```

nramabad: can you also move the `showEvents` and `hideEvents` on line 274 & 275 respectively to the upper scope and use them instead of re-declaring?

tshen815 (approved):
Awesome job on this Chantelle!

nramabad (approved):

PR #6065, opened 2021-12-14
Add macOS resource class options table

alexazl:
# Description
We're adding a new resource class option to the macOS executor.

# Reasons
Since the new `macos.x86.medium.gen2` resource option has the same basic specs as our standard `medium`, I've added the processor speeds to the table so customers will more easily understand what the differences are between the various options.


thomasr8 (approved):
Looks good! I say we release tomorrow when it goes live.

alexazl:
@thomasr8 Any harm in merging it today? We silently released today, assuming that early birds and EMEA folks will start testing it out first thing tomorrow.

thomasr8:
@alexazl : No harm in releasing today. If we have already silently launched this, let's have it go live today.

PR #6070, opened 2021-12-15
Remove reference to -ja folder, fix link

devinmarieb:
# Description
We can't accept PRs to our translated content since it will be overwritten by Crowdin. I've removed the reference to the -ja folder for local development. I also updated a url.

# Reasons
We've received a PR for translated content from last year that has been sitting. We unfortunately can't do anything with this as it will just get overwritten by Crowdin. Removing this line and only referencing the english folder will imply that is the location to make any changes allowed.


thomasr8 (approved):
Looks good! Thank you for catching this.

PR #6074, opened 2021-12-15
update for newer elixir image on dockerhub

jerdog:
The convenience image we link them to has been updated to 1.11.4, but this example config is stuck on 1.6.5. The updates bring the Dockerfile over to match.

# Description
Bring over the dockerfile from dockerhub so that our doc matches what the convenience image config is

# Reasons
Our docs for Docker Layer Caching were out of sync with what we're actually linking to, with the potential to cause confusion.


jekyll/_cci2/docker-layer-caching.md
```diff
@@ -117,56 +117,95 @@ Let's use the following Dockerfile to illustrate how Docker Layer Caching works.
...
+FROM elixir:1.11.4
 
-# Make apt non-interactive
+# make Apt non-interactive
 RUN echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90circleci \
   && echo 'DPkg::Options "--force-confnew";' >> /etc/apt/apt.conf.d/90circleci
 
 ENV DEBIAN_FRONTEND=noninteractive
 
+# Debian Jessie is EOL'd and original repos don't work.
```
thomasr8, line 128:
```suggestion
# Debian Jessie is EOL'd and original repos don not work.
```

thomasr8 (approved):
Looks good! Just one minor comment.

jekyll/_cci2/docker-layer-caching.md
jerdog: I'll edit it and remove the extra `n` @thomasr8

jerdog:
@thomasr8 I updated the word usage - can merge it now

PR #6089, opened 2021-12-19
Update pipeline-variables.md

kelvintaywl:
# Description

Moved the note on `pipeline.git.base_revision` from outside of the table listing to within the pipeline parameter's description itself.
Upon discussion with a customer, it was mentioned that we should placed it "closer" to the source, for better call to attention to this known limitation.

I've also made the `Note:` stanzas bold, based on the styling of most other documentation pages i see here on our Docs.

# Reasons

Updates based on [a customer feedback](https://circleci.zendesk.com/agent/tickets/101081).


devinmarieb (approved):
Thanks!

PR #6093, opened 2021-12-21
[TD-124] Document test metadata size limits

circlecai:
# Description
Document the strict limits we recently added on test metadata sizes the previously would have had undefined poor behavior.

# Reasons
https://circleci.atlassian.net/browse/TD-124


circlecai:
since it only affects the API, this will be better documented there

PR #6102, opened 2021-12-26
fix a wrong spacing in the 2nd config.yml

toshiaizawa:
Using `.circleci/config.yml` as indicated in Using the workflows functionality gives an error as follows:

Unable to parse YAMLwhile parsing a block mapping in 'string', line 22, column 9:          - run: echo "A more familiar hi" # ...             ^

The proposed spacing (killing two spaces) will eliminate the error.

# Description
This fix will eliminate the error caused by extra spacing that makes the list parsed incorrectly.

# Reasons
It's such a simple change that I went ahead with this PR instead of submitting an issue.


kurumai (approved):
@toshiaizawa  LGTM! Thank you for your contribution!!

PR #6105, opened 2021-12-27
updated `AWS_DEFAULT_REGION` to `AWS_REGION`

ShaleenAg:
I followed the docs to deploy to ECS, but it was throwing an error during deployment in the pipeline. Following the logs, it looks like aws expects `$AWS_REGION` variable for region. Adding an environment variable with name `AWS_REGION` fixed that

# Description
updated `AWS_DEFAULT_REGION` environment variable to `AWS_REGION`

# Reasons
I followed the docs to deploy to ECS, but it was throwing an error during deployment in the pipeline. Following the logs, it looks like aws expects `$AWS_REGION` variable for region. Adding an environment variable with name `AWS_REGION` fixed that

### DIFF
--- a/jekyll/_cci2/ecs-ecr.md
+++ b/jekyll/_cci2/ecs-ecr.md
@@ -61,7 +61,7 @@ Variable                 | Description
 -------------------------|------------
 AWS_ACCESS_KEY_ID        | Security credentials for AWS.
 AWS_SECRET_ACCESS_KEY    | Security credentials for AWS.
-AWS_DEFAULT_REGION       | Used by the AWS CLI.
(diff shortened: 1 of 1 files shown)


devinmarieb:
Hi @ShaleenAg we've done some digging on this issue and we were wondering if you by any chance have a build link for the failure you were observing? We ran some tests and `AWS_DEFAULT_REGION` is still working in every instance, though `AWS_REGION` will override that if present.

devinmarieb:
@ShaleenAg I'm going to close this for now. If you can provide further information around your build failure we can revisit this. We were unable to replicate the issue.

PR #6108, opened 2021-12-30
docs: Fix anchor typos

MatheusdeMelo:
# Description
Fix anchor to follow page links

# Reasons
Page links call the anchor with a underscore but the anchor doesn't have it


thomasr8 (approved):

devinmarieb:
@thomasr8 I'm closing this because it's stuck in the pipeline, but in addition the way this page is set up is a pattern for keys and if we were to change it, we would use dashes (-) in our URLs and anchors, not underscores.

PR #6131, opened 2022-01-05
Add workflow banner, fix broken link

devinmarieb:
# Description
Add banner notifying users of sunsetting scheduled workflows.
Fix link in a server doc.
Correct instances of rollback as a noun vs roll back as a verb.

# Reasons
There is still a significant number of users who are using scheduled workflows who have not migrated yet. Hopefully this will be an additional reminder to migrate to scheduled pipelines.


kbatuigas (approved):
Looks good!

PR #6133, opened 2022-01-05
Update links/header text

devinmarieb:
# Description
Taking care of some English suggestions to links/headers that were a part of an open source PR that couldn't be addressed because it was majority Japanese content updates which can't be merged.

# Reasons
Some of the internal docs links were entered as external links + some of those links were going to confusing URLs based on headers which also needed updating.


thomasr8 (approved):
Looks good 👍

PR #6135, opened 2022-01-05
Fix yark typo, include quotes

ryanwohara:
We typoed yarn.lock and were missing quotes.
Also fixed some whitespacing in the translated content


MikeyYeahYeah (approved):
Not sure if we need an additional member of docs team to approve as well

devinmarieb (approved):
Thanks!

Mayuko-Sa (approved):

PR #6136, opened 2022-01-05
Update: don't auto open sidebar on mobile.

teesloane:
[Ticket](https://circleci.atlassian.net/browse/DD-306)

### Description

On mobile the sign up cta was being pushed below the fold.

### Screenshots

Screen Shot 2022-01-05 at 4 03 29 PM


tshen815 (approved):
A beautiful and elegant solution!

PR #6137, opened 2022-01-05
Translate legacy conv ref to next-gen ones

felicianotech:
Completing the items I have from the tracker.

### DIFF
--- a/jekyll/_cci2/databases.md
+++ b/jekyll/_cci2/databases.md
@@ -104,7 +104,7 @@ This section describes additional optional configuration for further customizing
 {:.no_toc}
 
 PostGIS is available and can be used like this:
-`circleci/postgres:12.0-postgis`
+`cimg/postgres:12.0-postgis`
 
 ### Using binaries
 {: #using-binaries }
--- a/jekyll/_cci2/databases.md
+++ b/jekyll/_cci2/databases.md
@@ -31,7 +31,7 @@ In the primary image the config defines an environment variable with the `enviro
 Set the POSTGRES_USER environment variable in your CircleCI config to `postgres` to add the role to the image as follows:
 
 ```yml
-      - image: circleci/postgres:14.0
+      - image: cimg/postgres:14.0
         auth:
           username: mydockerhub-user
           password: $DOCKERHUB_PASSWORD  # context / project UI env-var reference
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/databases.md
```diff
@@ -31,27 +31,25 @@ In the primary image the config defines an environment variable with the `enviro
...
 ```
 
-This Postgres image in the example is slightly modified already with `-ram` at the end. It runs in-memory so it does not  hit the disk and that will significantly improve the testing performance on this PostgreSQL database by using this image.
-
 {% raw %}
 
 ```yml
-version: 2
+version: 2.1
 jobs:
```
kbatuigas, line 46: Do we need to include the `orbs` stanza as well where we invoke `circleci/browser-tools`?

kbatuigas (commented):
@felicianotech I saw that the postgres and python images are still in `circleci` and not the `cimg` namespace so I want to double check if that's intended.

jekyll/_cci2/databases.md
felicianotech: No, I updated it to remove browsers completely.

felicianotech:
@kbatuigas I don't know what I was smoking when I made this PR but clearly I made a lot of mistakes. I've corrected the things you point out, thank you.

jekyll/_cci2/databases.md
```diff
@@ -101,17 +99,12 @@ When the database service spins up, it automatically creates the database `circl
...
 
-The default `circleci/postgres` Docker image uses regular persistent storage on disk.
-Using `tmpfs` may make tests run faster and may use fewer resources. To use a variant
-leveraging `tmpfs` storage, just append `-ram` to the `circleci/postgres` tag (i.e.,
-`circleci/postgres:9.6-alpine-ram`).
-
-PostGIS is also available and can be combined with the previous example:
-`circleci/postgres:9.6-alpine-postgis-ram`
+PostGIS is available and can be used like this:
+`circleci/postgres:12.0-postgis`
```
kbatuigas, line 107:
```suggestion
`cimg/postgres:12.0-postgis`
```

kbatuigas (changes requested):

kbatuigas (approved):

PR #6141, opened 2022-01-06
Brian's updates

brivu:


devinmarieb:
@felicianotech are these PRs by any chance duplicate work to what we are doing on docs? A lot of our PRs for image updates have been stuck in a review stage. I would hate for people to start doing work that's already being done. cc @brivu

devinmarieb:
@brivu I know some of the new images won't work on version 2 with the new -browsers variant because it requires the browser orb, which in turn requires version 2.1. I'm not entirely sure which images this effects, but has that been taken into consideration here? I just want to double check as I see a lot of -browser updates. If we need to keep both versions for reference, we might need to create code tabs for server 2, server 3 and cloud.

jekyll/_cci2/language-php.md
```diff
@@ -56,7 +56,7 @@ version: 2
 jobs: # a collection of steps
   build: # runs not using Workflows must have a `build` job as entry point
     docker: # run the steps with Docker
-      - image: circleci/php:7.1-node-browsers # ...with this image as the primary container; this is where all `steps` will run
+      - image: cimg/php:8.0.14-node # ...with this image as the primary container; this is where all `steps` will run
```
felicianotech, line 59: This line below is no longer valid in next-gen images:

```
- run: sudo docker-php-ext-install zip
```

1. The zip extension is already installed. If we just wanted to show an example of how to install something else, we would use `pecl`.

jekyll/_cci2/language-php.md
```diff
@@ -56,7 +56,7 @@ version: 2
 jobs: # a collection of steps
   build: # runs not using Workflows must have a `build` job as entry point
     docker: # run the steps with Docker
-      - image: circleci/php:7.1-node-browsers # ...with this image as the primary container; this is where all `steps` will run
```
felicianotech, line 59: Looks like we dropped browsers. If this isn't a real project to run, then I think that's fine. If it is, then the browser tools orb and the browser variant are needed.

jekyll/_cci2/language-php.md
```diff
@@ -123,7 +123,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/php:7.1-node-browsers
+      - image: cimg/php:8.0.14-node
```
felicianotech, line 126: Same comments below as I provided for above.

jekyll/_cci2/sample-config.md
```diff
@@ -1000,11 +1000,11 @@ jobs:
 
   rake_test:
     docker:
-      - image: circleci/ruby:2.4-node-jessie
+      - image: circleci/ruby:3.1.0-node
         auth:
           username: mydockerhub-user
           password: $DOCKERHUB_PASSWORD  # context / project UI env-var reference
-      - image: circleci/postgres:9.4.12-alpine
+      - image: cimg/postgres:14.0.0
```
felicianotech, line 1007: The tag `14.0.0` doesn't exist. It would be `14.0`.

jekyll/_cci2/sample-config.md
```diff
@@ -1022,11 +1022,11 @@ jobs:
 
   precompile_assets:
     docker:
-      - image: circleci/ruby:2.4-node-jessie
+      - image: circleci/ruby:3.1.0-node
         auth:
           username: mydockerhub-user
           password: $DOCKERHUB_PASSWORD  # context / project UI env-var reference
-      - image: circleci/postgres:9.4.12-alpine
+      - image: cimg/postgres:14.0.0
```
felicianotech, line 1029: Same as above.

felicianotech (changes requested):

Williamlliam:
@ThailandDistributionAgentoffice™



ในวันที่ ศ. 7 ม.ค. 2022 03:54 น. Brian Vu ***@***.***>
เขียนว่า:

jekyll/_cci2/language-php.md
brivu: > If it is, then the browser tools orb and the browser variant are needed.

Do you mean I to drop the node variant and add the browsers variant with browser tool orbs?
Or did you mean I need to add the browsers image + orb in addition to the node variant?

jekyll/_cci2/language-php.md
brivu: To clarify, do you mean `sudo docker-php-ext-install pecl` ?

PR #6143, opened 2022-01-06
Reorder steps for `launchctl` on mac

afranco07:
# Description
This PR reorders the steps for enabling the `launchctl` service on mac to make it less error prone for users who are installing runner for the first time

# Reasons
Jira: [RT-289](https://circleci.atlassian.net/browse/RT-289)

Screenshot:

Screen Shot 2022-01-06 at 17 17 53

### DIFF
--- a/jekyll/_cci2/runner-installation-mac.adoc
+++ b/jekyll/_cci2/runner-installation-mac.adoc
@@ -104,16 +104,15 @@ Copy the following to `/Library/LaunchDaemons/com.circleci.runner.plist`, owned 
 
 == Enable the `launchd` service
 
-If you are following these instructions for a second time, you should unload the following existing service:
-
-```bash
-sudo launchctl unload '/Library/LaunchDaemons/com.circleci.runner.plist'
-```
-
 Now you can load the service:
 
 ```bash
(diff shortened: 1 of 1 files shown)


stalmy:
Does this unload step make more sense if we say something like "if the service is already running, unload the previous running version first:"   or similar?

Does indenting the unload step here make it more obviously not for the mainstream first-install case?  Do we have conventions on formatting for these peripheral steps, Devin?

Stuart133 (approved):

devinmarieb (approved):
Thanks!

devinmarieb:
@stalmy Maybe I'm missing context, but the NOTE in the steps seems to address the "if the service is already running" issue in a way that makes sense (to me at least). If you feel it doesn't, we can adjust.

And for formatting, this is an asciidoc file, so this would be the best way to denote this. We are discussing to see if there is a possibility to change the icons for asciidoc notes/warnings as none of us really like them all that much.

@afranco07 If you think this is good to go regarding Steve's note, just ping me and I can merge if you don't have the ability to merge.

stalmy:
@devinmarieb , it's fine to merge.  We can revisit additional tweaks in a future revision.  thanks

PR #6150, opened 2022-01-07
Update scheduled-pipelines.md

ekrivchenko:
fix typo


devinmarieb:
@ekrivchenko Thank you! This typo is already being fixed in a larger PR that is in review, but thanks for looking out!

PR #6097, opened 2021-12-23
[DD-267]: Request New Language Form 

tshen815:
**Ticket:[ DD-267](https://circleci.atlassian.net/browse/DD-267)
Preview Link:** http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/request-new-language-preview/?force-all

# Changes
- add globe svg
- add new language requester form
- add appropriate style and functionality for language request form
- refactor code to changes languages to allow for dynamic URL for preview builds
- update href links with `{{ site.baseurl }}` for Japanese hompage 

# Description
Including new feature in the language selector to allow users to request languages that they would like to see as translation options for the future.

# Considerations
Please see associated PRs:

- #6084 
- #6096 

# Amplitude
Screen Shot 2021-12-23 at 9 29 40 AM

# Screenshots:

**Before:**
Screen Shot 2021-12-23 at 12 17 34 PM

**After**
Screen Shot 2021-12-23 at 12 16 08 PM

# Demo:

https://user-images.githubusercontent.com/57234605/148326854-cb70ce2b-165f-44fb-b022-35c709774b42.mp4

PR #6152, opened 2022-01-07
[DD-314]: Refactor Language Submission Process

tshen815:
**Ticket:** [DD-314](https://circleci.atlassian.net/browse/DD-314)
**Preview:** http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/request-new-language-preview/?force-all

# Changes
- use `<button>` in place of `<a>` with correct styles when disabled
- refactor how language form input is processed and submitted to amplitude 

# Description
Addressing  a bug where the language input is sent to amplitude multiple times upon submission on click due to how it is being handled and processed in lang.js . We are also refactoring the Submit button to be more SEO friendly. 

# Considerations
Refactor associated with: #6097

### DIFF
--- a/src/js/services/lang.js
+++ b/src/js/services/lang.js
@@ -68,7 +68,6 @@ const handleChangeLanguageNav = () => {
 const languageRequest = () => {
   const langForm = $('.lang-form');
   const submitBtn = $('#submit-btn');
-  const input = $('#lang-req')[0];
 
   // Add styles for input form when active
   langForm.on('click', () => {
@@ -82,17 +81,18 @@ const languageRequest = () => {
   });
 
   // Toggle submit button UI and button state based on user input
-  langForm.on('keyup', () => {
-    submitBtn.removeAttr('disabled');
-    if (input.value == '') {
+  langForm.on('keyup', (e) => {
+    if (e.currentTarget.value == '') {
       submitBtn.attr('disabled', 'disabled');
+    } else {
+      submitBtn.removeAttr('disabled');
     }
   });
 
   submitBtn.on('click', () => {
-    langForm.css('pointer-events', 'none');
+    langForm.addClass('lang-submitted');
     window.AnalyticsClient.trackAction('New Language Request', {
-      requestedLanguage: input.value,
+      requestedLanguage: langForm[0].value,
       browserNativeLang: window.navigator.language,
       app: 'Docs',
       location: window.location.href,
@@ -100,7 +100,7 @@ const languageRequest = () => {
     });
     // Swap out button with submit message after submission
     submitBtn.replaceWith(
-      '<span id=lang-submitted>' + 'Thank you for your help' + '</span>',
+      `<span id=lang-submitted>Thank you for your help</span>`,
     );
   });
 };
--- a/src/js/services/lang.js
+++ b/src/js/services/lang.js
@@ -65,9 +65,11 @@ const handleChangeLanguageNav = () => {
 /*
   Handle functionality and UI changes for language request input form and new language submission when an input is provided
 */
-const langForm = $('.lang-form');
-
 const languageRequest = () => {
+  const langForm = $('.lang-form');
+  const submitBtn = $('#submit-btn');
+  const input = $('#lang-req')[0];
+
   // Add styles for input form when active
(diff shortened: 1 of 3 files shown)


src/js/services/lang.js
```diff
@@ -65,9 +65,11 @@ const handleChangeLanguageNav = () => {
 /*
   Handle functionality and UI changes for language request input form and new language submission when an input is provided
 */
-const langForm = $('.lang-form');
-
 const languageRequest = () => {
+  const langForm = $('.lang-form');
+  const submitBtn = $('#submit-btn');
+  const input = $('#lang-req')[0];
```
rsylvian, line 71: `.langForm` and `#lang-req` are effectively the same element so that seems redundant

src/js/services/lang.js
```diff
@@ -78,40 +80,30 @@ const languageRequest = () => {
...
-  if (input.value.length > 0) {
-    // If user input present, adjust UI and add event listener to submit data on click to amplitude
-    submitBtn.css({ opacity: '100%', cursor: 'pointer' });
-    submitBtn.on('click', () => {
-      clearInterval(checkLangInput);
-      langForm.css('pointer-events', 'none');
+  // Toggle submit button UI and button state based on user input
+  langForm.on('keyup', () => {
+    submitBtn.removeAttr('disabled');
+    if (input.value == '') {
```
rsylvian, line 87: seems like you can use `this.value` instead and no need for keeping an input variable in that case 😊

rsylvian (approved):
LGTM but left two suggestions

src/js/services/lang.js
tshen815: addressed!

src/js/services/lang.js
```diff
@@ -78,40 +80,30 @@ const languageRequest = () => {
...
     });
-  }
-  // If no user input present, adjust UI and remove event from submit btn
-  if (!input.value.length) {
-    submitBtn.css({ opacity: '50%', cursor: 'not-allowed' });
-    submitBtn.off('click');
-  }
+    // Swap out button with submit message after submission
+    submitBtn.replaceWith(
+      '<span id=lang-submitted>' + 'Thank you for your help' + '</span>',
```
nramabad, line 103: why the concat?

src/js/services/lang.js
```diff
@@ -78,40 +80,30 @@ const languageRequest = () => {
...
 
-  if (input.value.length > 0) {
-    // If user input present, adjust UI and add event listener to submit data on click to amplitude
-    submitBtn.css({ opacity: '100%', cursor: 'pointer' });
-    submitBtn.on('click', () => {
-      clearInterval(checkLangInput);
-      langForm.css('pointer-events', 'none');
+  // Toggle submit button UI and button state based on user input
+  langForm.on('keyup', () => {
+    submitBtn.removeAttr('disabled');
```
nramabad, line 86: i think this logic could be simpler. `if (input.value == '') submitBtn.attr('disabled', 'disabled') else ...removeAttr(... els`

src/js/services/lang.js
```diff
@@ -78,40 +80,30 @@ const languageRequest = () => {
...
-        app: 'Docs',
-        location: window.location.href,
-        path: window.location.pathname,
-      });
-      // Swap out button with submit message after submission
-      submitBtn.replaceWith(
-        '<span id=lang-submitted>' + 'Thank you for your help' + '</span>',
-      );
+  submitBtn.on('click', () => {
+    langForm.css('pointer-events', 'none');
```
nramabad, line 93: can this be in css instead?

src/js/services/lang.js
tshen815: `this.value` is coming back as undefined, but we could use `langForm.val()`?

src/js/services/lang.js
rsylvian: ^ Good point

src/js/services/lang.js
tshen815: missed that, done!

src/js/services/lang.js
tshen815: good call

PR #6155, opened 2022-01-08
[DD-313] Reduce image dimensions on Artifacts article

rsylvian:
# Description
- Resize the image to 1920px wide (max layout size) instead of the original 6250px
- Optimize image from 241kb to 105kb

# Reasons
Image is unnecessarily too big and is causing LCP issues


tshen815 (approved):
Looks great locally!

PR #6176, opened 2022-01-11
Fix: Check for IntersectionObserver existence before using it

rsylvian:
# Description
- Check for typeof `IntersectionObserver` before using it.

# Reasons
While the `IntersectionObserver` is [fairly well supported everywhere](https://caniuse.com/intersectionobserver) we have a few errors every day on [DD](https://app.datadoghq.com/rum/error-tracking?issueId=f63406d4-6899-11ec-b77f-da7ad0900002&start=1641921449048&end=1641925049048&paused=false) that suggests that it might not be available for everyone completely. This PR prevents calling it without checking first.


tshen815 (approved):
Thanks for addressing this!

PR #6188, opened 2022-01-13
try to proactively answer questions about difference between network transfer & IP ranges

sebastian-lerner:
@devinmarieb what do you think? too wordy?

# Description

# Reasons

### DIFF
--- a/jekyll/_cci2/ip-ranges.md
+++ b/jekyll/_cci2/ip-ranges.md
@@ -143,7 +143,7 @@ Notifications of a change to this list will be sent out by email to all customer
 ## Pricing
 {: #pricing }
 
-Pricing will be calculated based on data usage of jobs opted into the IP ranges feature, however, only the traffic of the opted-in jobs will be counted. It is possible to mix jobs with and without the IP ranges feature within the same workflow or pipeline.  Data used to pull in the Docker image to the container before the job starts executing will _not incur usage costs_ for jobs with IP ranges enabled.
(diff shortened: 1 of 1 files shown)


devinmarieb:
@sebastian-lerner I don't think it's too wordy. I did update the linking format on one of the links in there and also made a few small changes if you want to take a look in my commit. I think we are trying to move away from capitalizing things like "Network Transfer" unless we are specifying a particular button/tab in the UI.

sebastian-lerner:
hmmmm i am trying to indicate "Network Transfer & Egress" that you see in the Plan Usage page.  I didn't realize that we show customers in the Projects tab a total of per-project Network Transfer & Egress so we may want to adjust the wording to reflect that instead of directing them to the "Network tab".  Want me to take a stab at re-wording?

sebastian-lerner:
Screen Shot 2022-01-13 at 1 20 09 PM

thomasr8:
@sebastian-lerner I just made a few very minor edits and changes that I noticed. These were very small (a typo and a few spacing fixes), so nothing major.

devinmarieb:
@sebastian-lerner Yeah since there's definitely a difference between overall network transfer (which a user can still see in the app) and the billable network transfer, I think the clarification needs to be made. I didn't read this as speaking about billable usage specifically I guess. I still wouldn't capitalize "network transfer" in this instance, but I would consider adding the word "egress", and I think a distinction should be made that we are only talking about billable network transfer and egress here.

PR #6190, opened 2022-01-13
[DD-274] Add cypress e2e testing framework

chantelle11:
[Ticket link ](https://circleci.atlassian.net/jira/software/c/projects/DD/boards/477?modal=detail&selectedIssue=DD-353)

# Description
This was submitted on the Blue tape list: https://circleci.atlassian.net/wiki/spaces/GD/pages/6325928038/Blue+Tape+List

As a first pass we would want that story to purely focus on bringing the cypress testing framework and as a proof of concept test one of our features.

# Impact
Adding cypress tests will allow us to be more confident we are not breaking anything when releasing

# Changes
Add cypress package and write tests for sidebar

# How to test
Run the application with `yarn dev`
Run cypress with `yarn cypress:open` to get gui
Or
Run cypress with `yarn test:e2e:run` to run without gui


yarn.lock
```diff
@@ -2,22 +2,17 @@
 # yarn lockfile v1
```
chantelle11: I updated a lot of versions.

.circleci/config.yml
```diff
@@ -82,7 +82,6 @@ commands:
 # Workflows orchestrate a set of jobs to be run;
 # the jobs for this pipeline are # configured below
 workflows:
-
```
rsylvian, line 85: Let's revert that change, please 🙏

yarn.lock
rsylvian: There are 3,738 changes for the lock file which sounds like a lot for installing one package and updating another one. Are you sure it is accurate?

chantelle11: It seems like a lot was just out of date. delete yarn lock and then did `yarn install`

cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+  "baseUrl": "https://ui.circleci.com",
+  "defaultCommandTimeout": 10000
```
nramabad, line 3: seems like a long timeout

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
+/// <reference types="cypress" />
```
nramabad, line 1: what is this comment?

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+
+describe('sidebar', () => {
+  // Based on sidebar.js
+  const defaultSectionName = 'getting-started';
+  const basepath = '/docs/2.0/'
+  beforeEach(() => {
+    // Takes to starting url which is baseURL
+    cy.visit(`${basepath}`)
```
nramabad, line 9: why interpolate?

cypress/integration/sidebar.spec.js/sidebar.spec.js
chantelle11: this is standard in cypress files to allow for types. when you do cy. it will auto complete

cypress.json
chantelle11: default is 4000 i can put it back to default

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      // Check defaultSectionName open then close defaultSectionName
+      cy.get('nav.sidebar').children('.sidebar-item-group').children('ul').children(`[data-section="${defaultSectionName}"]`).should('not.have.class', 'closed').children('.list-wrap').click()
+      // Check defaultSectionName closed then open
+      cy.get('nav.sidebar').children('.sidebar-item-group').children('ul').children(`[data-section="${defaultSectionName}"]`).should('have.class', 'closed').click().should('not.have.class', 'closed')
+    })
+
+    it('should bring you to right url', () =>{
+      // Click second element in getting started which is migration-intro
+      cy.get('nav.sidebar').children('.sidebar-item-group').children('ul').children(`[data-section="${defaultSectionName}"]`).children().next().click()
+      cy.url().should('include', 'migration-intro')
```
nramabad, line 36: do we want to hardcode stuff like this? is there anyway this can be done in a less hardcoded fashion?

cypress/integration/sidebar.spec.js/sidebar.spec.js
chantelle11: good catch! remnants of changing my mind :)

cypress/integration/sidebar.spec.js/sidebar.spec.js
nramabad: can we derive all the URLs in this file from https://github.com/circleci/circleci-docs/blob/ea8c1f7c6409952ec35b7e2b397a89b3691a0f07/jekyll/_data/sidenav.yml ?

chantelle11: https://docs.cypress.io/api/commands/children#Selector 
There is a lot of chaining that takes place unless we place id's on things. Then we can just get the id we want no matter where it is in the DOM

nramabad: i don't have a problem with the ids/classes. its the URLs.

but in the future if a technical writer wants to modify some links, I'm not sure if they're comfortable fishing around in here to modify this as well. Any simple link change would become a 2 step process.

nramabad: actually i just reviewed the logic here @chantelle11 

https://github.com/circleci/circleci-docs/blob/5f29a7f3aaab04db1ffafdc03b950ad67565947e/jekyll/_includes/sidebar.html

all the data-sections are coming from `sidenav.yaml`. I think the correct way to do this is rely on this Single Source of Truth.

In fact, since you have a SSOT, I think you can basically do a for loop over `sidenav.yaml` and create a bunch of cypress tests for everything in YAML file.

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+describe('sidebar', () => {
+  // Based on sidebar.js
+  const defaultSectionName = 'getting-started';
+  const basepath = '/docs/2.0/'
+  beforeEach(() => {
+    // Takes to starting url which is baseURL
+    cy.visit(basepath)
+  })
+
+  describe('Full screen', () => {
```
rsylvian, line 12:
```suggestion
  describe('Desktop', () => {
```

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+    beforeEach(() => {
+      cy.viewport(1280, 720)
+    })
+
+    it('should load sidebar in full screen', () => {
```
rsylvian, line 17:
```suggestion
    it('should load sidebar', () => {
```

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      cy.get(`[data-section="${defaultSectionName}"]`).should('exist')
+      cy.get('nav.sidebar').children('.sidebar-item-group').children('ul').should('have.length.greaterThan', 0)
+    })
+
+    it('should close getting started and open pipelines', () => {
+      // Check defaultSectionName open then close defaultSectionName
+      cy.get('nav.sidebar').children('.sidebar-item-group').children('ul').children(`[data-section="${defaultSectionName}"]`).should('not.have.class', 'closed').children('.list-wrap').click()
+      // Check defaultSectionName closed then open
+      cy.get('nav.sidebar').children('.sidebar-item-group').children('ul').children(`[data-section="${defaultSectionName}"]`).should('have.class', 'closed').click().should('not.have.class', 'closed')
+    })
```
rsylvian, line 31: Let's split that test in two, please 🙏

cypress/integration/sidebar.spec.js/sidebar.spec.js
rsylvian: +1 to have the `sidenav.yml` being the source of truth. We have been waiting on cypress for a while now so if it is something that requires extensive work, let's create a ticket for the next sprint so we can ship this ASAP.

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+
+    it('should bring you to right url', () =>{
+      // Click second element in getting started which is migration-intro
+      cy.get('nav.sidebar').children('.sidebar-item-group').children('ul').children(`[data-section="${defaultSectionName}"]`).children().next().click()
+      cy.url().should('include', 'migration-intro')
+      cy.get(`[href="${basepath}migration-intro/"]`).should('have.class', 'active')
+    })
+
+    it('should load a page open the right section in the sidebar', () =>{
```
rsylvian, line 40:
```suggestion
    it('should open the right section in the sidebar on page load', () =>{
```

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      cy.get(`[href="${path}"]`).should('have.class', 'active')
+    })
+  })
+
+  describe('Mobile', () => {
+    beforeEach(() => {
+      cy.viewport(320, 568)
+    })
+
+    it('should load sidebar closed in mobile', () => {
```
rsylvian, line 53:
```suggestion
    it('should have sidebar closed by default', () => {
```

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      // mobile sidebar
+      cy.get('nav.mobile-sidebar').should('exist')
+      cy.get('nav.mobile-sidebar').children('ul').should('have.length.greaterThan', 0)
+
+      // sidebar open and close button
+      cy.get('[data-target="#global-nav"]').should('exist')
+      cy.get('[data-target="#global-nav"]').should('have.class', 'collapsed')
+    })
+
+    it('should open sidebar in mobile', () => {
```
rsylvian, line 66:
```suggestion
    it('should open sidebar when clicking hamburger menu, () => {
```

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      })
+
+      it('should close getting started and open pipelines', () => {
+        // Check defaultSectionName closed then Open defaultSectionName
+        cy.get('nav.mobile-sidebar').children('ul').children(`[data-section="${defaultSectionName}"]`).should('have.class', 'closed').children('.list-wrap').click()
+        // Check defaultSectionName open then close defaultSectionName
+        cy.get('nav.mobile-sidebar').children('ul').children(`[data-section="${defaultSectionName}"]`).should('not.have.class', 'closed').children('.list-wrap').click()
+        // Check defaultSectionName closed
+        cy.get('nav.mobile-sidebar').children('ul').children(`[data-section="${defaultSectionName}"]`).should('have.class', 'closed')
+      })
```
rsylvian, line 84: Same comment as above, this can be refactored in smaller tests and they don't have to be that specific. All we need to know is that when opening one section, the others are closing.

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,95 @@
...
+
+      it('should bring you to right url', () =>{
+        // Open getting started
+        cy.get('nav.mobile-sidebar').children('ul').children(`[data-section="${defaultSectionName}"]`).should('have.class', 'closed').children('.list-wrap').click()
+        // Click second element in getting started which is migration-intro
+        cy.get('nav.mobile-sidebar').children('ul').children(`[data-section="${defaultSectionName}"]`).children().next().click()
+        cy.url().should('include', 'migration-intro')
+      })
+    })
```
rsylvian, line 93: This is basically the same thing as the test in the `Desktop section` and chance we can share it and only have to write the test once?

rsylvian (changes requested):

cypress/integration/sidebar.spec.js/sidebar.spec.js
chantelle11: the nav.sidebar structure is different than the nav.mobile-sidebar structure. maybe with the refactor that nav suggested

cypress/integration/sidebar.spec.js/sidebar.spec.js
chantelle11: When you open 1 section others do not close automatically

cypress/integration/sidebar.spec.js/sidebar.spec.js
chantelle11: I think what I need to do is take the yml make it into a json object, parse the json object and then use the object as the source of truth. im just trying to figure out how to parse the yml still

jekyll/_data/sidenav.yml
```diff
@@ -116,10 +116,10 @@ en:
     - name: Machine
       children:
         - name: Overview
-          link: 2.0/executor-types/
+          link: 2.0/executor-types/#using-machine
           hash: using-machine
         - name: Pre-installed software
-          link: 2.0/docker-to-machine
+          link: 2.0/docker-to-machine#pre-installed-software
```
rsylvian, line 122: I don't think we should change the URLs in that file. Or at least not as part of that PR. Any reason you tweaked that?

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,145 @@
...
+      // start where configuration is closed
+      cy.get('nav.sidebar div.sidebar-item-group').children('ul').children(`[data-section="configuration"]`).should('have.class', 'closed').click().should('not.have.class', 'closed')
+      cy.get('nav.sidebar div.sidebar-item-group').children('ul').children(`[data-section="configuration"]`).children('ul.subnav').children('ul').children('li.subnav-item').children(`[href="${basepath}configuration-reference/"]`).click()
+      // Navigate to configuration-reference now `configuration` is open and configuration-reference is active 
+      cy.url().should('include', 'configuration-reference')
+      cy.get('nav.sidebar div.sidebar-item-group').children('ul').children(`[data-section="configuration"]`).should('not.have.class', 'closed')
+      cy.get('nav.sidebar div.sidebar-item-group').children('ul').children(`[data-section="configuration"]`).children('ul.subnav').children('ul').children('li.subnav-item').children(`[href="${basepath}configuration-reference/"]`).should('have.class', 'active')
+    })
+
+    it('should bring you to right url for hompage', () =>{
```
rsylvian, line 81:
```suggestion
    it('should bring you to right url for homepage', () =>{
```

cypress/integration/sidebar.spec.js/sidebar.spec.js
```diff
@@ -0,0 +1,145 @@
+/// <reference types="cypress" />
```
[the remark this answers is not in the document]
chantelle11: My mistake

src/js/services/lang.js
```diff
@@ -9,15 +9,15 @@ const languages = {
   en: {
     name: 'English',
     url: 'docs/',
-    id: 'globalNavLangEng',
-    domEl: getElById('globalNavLangEng'),
+    id: 'globalNavLang-en',
+    domEl: getElById('globalNavLang-en'),
```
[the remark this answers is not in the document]
chantelle11: Taking a look at test 'should go through sidebar yml to see if all links exist in EN then JA' explains it well. Basically the yml file is split into 2 the JA version and the EN version. To avoid parsing the yml file twice and reusuing the code i changed the button name that you need to press in order to change the language. 

In hinde site I could have done each time the yml file goes into a new language then go to the new url but I figured that the button would have the url so if the URL ever changes I wouldn't have to update it in 2 places but I assumed there will always be a button to select the language just a matter of different location. 

I am up for changing but that was the logic behind the choice.

jekyll/_data/sidenav.yml
chantelle11: the yml file link did not match the link in the <a> tags so it seems like an inconsistency issue. The 'should go through sidebar yml to see if all links exist in EN then JA' test just goes through and makes sure all links exist. I had a hard time trying to get cypress to look for a href that only contains the link even if its a partial link. I could only get it to work if it was exact match. I can look into it further but that is the reason for the change

src/js/services/lang.js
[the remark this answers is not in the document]
chantelle11: The url for JA ends in ja but it doesn't end in /en for english. It was easier to just change the button name as it seemed to have a consistent pattern for testing in the future. You need to have the right language selected / correct language url to show the side bar for the specific language. Let me know if have any other questions but I think the test kinda explains the logic as well :)

cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+  "baseUrl": "https://ui.circleci.com",
```
teesloane, line 2: Will this work when we run cypress in CI?

jekyll/_data/sidenav.yml
teesloane: Hmm, I think it's okay to change this, but if we are changing it we are effectively doubling up on hashes with the use of the hash key on line `120`. This was originally used to construct urls to get around our issues with using query-params + local-storage to indicate where the user is in the docs when they reload the page and the sidebar jumps to the document they are on. It seems that we no longer use the query params (I think @rsylvian might have refactored that a while ago). If that's the case, we could make this change, but we should remove the use of `hash` and then get rid of it where it appears in sidebar.html:

```
                  <a class="{% if url_slug == link_slug %} active {% endif %}"
                    href="{{ site.baseurl }}/{{ item.link }}{% if item.hash %}#{{item.hash}}{% endif %}" data-section="{{ section_slug }}">
                    {{display_name}}
                  </a>
```

src/js/services/lang.js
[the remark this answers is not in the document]
teesloane: if I understand correctly, you are using the dynamic values of the sidebar to select the values from the javascript in this line:

```
cy.get(`#globalNavLang-${key}`).click()
```

I think this is fine to do - we should standardize our use of languages to be `ISO 639-1` (two letter-codes) when we can anyway.

teesloane (commented):
Added a few comments - but this is looking great. Lots of awesome work. I'm still a little new to cypress, but from my understanding this will be super helpful to have. 

One other remaining question I have is - do we want to integrate this into our CI pipeline as part of this PR?

cypress.json
chantelle11: When you run in CI you can configure an override to the default baseURL. Alternatively I can remove the baseURL and add in both script commands the override manually so there is no confusing. In addition when looking int Nav code for the spun up enviroment it was configured for ui.circleci.com so I did not find an issue with making it the default as the enviroment in CI was the same

jekyll/_data/sidenav.yml
chantelle11: Good catch! I did not see double # in the html or the test would fail but I will look into this code if can now be removed

src/js/services/lang.js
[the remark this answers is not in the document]
chantelle11: i believe that the languages is 2 letter codes 'en' and 'ja' but good call out for future languages and yes you are correct on the process

jekyll/_data/sidenav.yml
chantelle11: I think since we want to do a source of truth it be good to not make these changes after the fact from sidenav.yml

chantelle11: I didn't realize sidenav had `hash: using-docker` so I don't have to alter sidenav.yml itself my mistake

teesloane (approved):

src/js/services/lang.js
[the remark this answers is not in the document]
tshen815: this is a very cool way to do this, good stuff chantelle!

PR #6191, opened 2022-01-13
Update faq.md

sebastian-lerner:
@devinmarieb

### DIFF
--- a/jekyll/_cci2/faq.md
+++ b/jekyll/_cci2/faq.md
@@ -305,7 +305,7 @@ You can choose to pay for premium features per active user, compute, and optiona
 {: #how-do-I-calculate-my-monthly-storage-and-network-costs }
 {:.no_toc}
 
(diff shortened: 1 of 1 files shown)


thomasr8 (approved):
Looks good 👍

devinmarieb:
@sebastian-lerner I added IP ranges to the summary of the pricing and changed the style of how we notify about what tab to look for (bold). Since we say what the credit costs are for N&S I also think it's fine to add the credit costs here as well. Right now this is the only place in the docs where we will speak specifically about credits. I'm not going to merge this until we know whether or not the discrepancy in the numbers is a bug or not.

thomasr8:
@devinmarieb I agree about being able to add credit costs, as long as we don't add too much specificity here, since we have seen that credit costs can change pretty quickly, meaning we may need to update this page on a regular basis.

PR #6196, opened 2022-01-14
Clarify use of CIRCLE_USERNAME variable

catherinepope:
# Description
Explained that CIRCLE_USERNAME variable available only if user has CCI account.

# Reasons

https://circleci.atlassian.net/browse/DOCSTEAM-53


rosieyohannan (approved):
Thank you!!

PR #6210, opened 2022-01-18
Fix orb page numbers and indentation

devinmarieb:
# Description
After some trial and error I found a markdown format that cleans up the numbered lists and indentations. 

# Reasons
The orbs page was confusing as the instructional list was breaking format.
Markdown numbered lists break when things like code blocks and images are added to the lists. With proper formatting and some HTML, the lists can be cleaned up.

**Before:**
before
**After:**
after


kbatuigas (approved):
lgtm!

PR #6215, opened 2022-01-18
Add: batch resizes images in docs.

teesloane:
[Ticket](https://circleci.atlassian.net/browse/DD-335)

🖼 - add resizing of images to python find_big_images.py script.
👾 - squishes their bits and bytes down to size. (anything beyond 1920x in width or height gets limited to 1920px.)
🤙 - uses sweet thumbnail feature to avoid skewed images.

🔗  - [Preview](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-335-resize-big-images-preview/?force-all)


scripts/find_big_images.py
```diff
@@ -71,13 +72,27 @@ def print_report():
...
         dict_writer.writeheader()
         dict_writer.writerows(Images)
 
+
 get_img_data()
-print_report()
-write_csv()
+# print_report()
+# write_csv()
+resize_image()
```
chantelle11, line 98: Now that the code is run should resize_image() be default commented like write and print?

chantelle11 (approved):
Looks good to me! 🧚✌️💕

nramabad (changes requested):
so the thing I'm seeing here is that for some of these images, shrinking the image dimensions is actually increasing the image file size? I'm guessing this is down to how png does things like anti-aliasing(?)

I'm sorta skeptical if we're going to see benefits, given [according to Google](https://web.dev/lcp/#how-is-an-element's-size-determined) LCP of an image is typically determined by the size of an image as displayed in the viewport. it would be good run Lighthouse on `master-preview` vs `<your branch>-preview` on a page URL that has an image touched in this PR. (preferrably 2 - one with the decrease file size and one with the increase file size)

I ran it ooc and these were my results:
After: Screen Shot 2022-01-19 at 7 02 46 PM
Before: Screen Shot 2022-01-19 at 7 02 37 PM

Seems like in this example the image size increases and the LCP score does too (although I admit Lighthouse scores are always a bit suspect and more ballpark compared to something like PageSpeed)

If you're able to demonstrate the improvement and add it to the PR description, i think that would be much better than flying blind on 90 file changes.

nramabad:
Here's the apples to apples Lighthouse on Desktop since it's a bit more relevant.

Before: image
After: image

The contrast in LCP is more stark here since Desktop unlike Mobile, the image is above the fold content.

teesloane:
Wow, interesting insights @nramabad . Thanks for quantifying these changes. I'll do some more testing today. Interestingly, I've realized that pngs in all ascidocs seem to not be working in the preview.

teesloane:
Found [this thread](https://stackoverflow.com/questions/33892585/why-does-resizing-a-jpeg-image-using-pil-increases-file-size) that points at some possible reasons file sizes are increasing after resize.

teesloane:
I manually compressed the resized images and they are all now smaller. Our light house score on the caching page (picked to match what @nramabad posted above) has not improved; it has gotten worse:

image

However, the callout to "properly size images" was not on my report, so at least that seems improved.

rsylvian (approved):
LGTM ✅  Glad TinyPNG worked!

🚢  🇮🇹

nramabad (approved):
LGTM!

PR #6221, opened 2022-01-19
[DD-334] Track if Users Enable Dark Mode

tshen815:
[Ticket](https://circleci.atlassian.net/browse/DD-334)
[Preview](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/track-darkmode-preview/?force-all)

# Changes

- add `trackAction` for user dark mode preference
- set storage key is user preference already recorded

# Description
We are gauging the interest of users to switch to dark mode since developers tend to work in dark mode and it’s considered as table-stakes

# Amplitude
Screen Shot 2022-01-19 at 4 36 30 PM

### DIFF
--- a/src/js/site/main.js
+++ b/src/js/site/main.js
@@ -426,8 +426,8 @@ $(function () {
   Check if users have dark mode enabled
   Set key if user response has not already been tracked to ensure we dont get multiple events from the same user
  */
-$(document).ready(() => {
-  const storageKey = 'provided-dark-mode-reponse';
+const trackDarkModePreference = () => {
+  const storageKey = 'provided-dark-mode-response';
   if (localStorage.getItem(storageKey)) {
     return;
   } else {
@@ -438,4 +438,5 @@ $(document).ready(() => {
         window.matchMedia('(prefers-color-scheme: dark)').matches,
     });
   }
-});
+};
+trackDarkModePreference();
--- a/src/js/site/main.js
+++ b/src/js/site/main.js
@@ -422,10 +422,20 @@ $(function () {
   });
 });
 
+/*
+  Check if users have dark mode enabled
+  Set key if user response has not already been tracked to ensure we dont get multiple events from the same user
+ */
 $(document).ready(() => {
-  window.AnalyticsClient.trackAction('User Dark Mode Preference', {
-    darkModeEnabled:
-      window.matchMedia &&
-      window.matchMedia('(prefers-color-scheme: dark)').matches,
-  });
+  const storageKey = 'provided-dark-mode-reponse';
+  if (localStorage.getItem(storageKey)) {
+    return;
+  } else {
+    localStorage.setItem(storageKey, true);
+    window.AnalyticsClient.trackAction('User Dark Mode Preference', {
+      darkModeEnabled:
+        window.matchMedia &&
+        window.matchMedia('(prefers-color-scheme: dark)').matches,
+    });
+  }
 });
--- a/src/js/site/main.js
+++ b/src/js/site/main.js
@@ -421,3 +421,11 @@ $(function () {
     });
   });
 });
+
+$(document).ready(() => {
+  window.AnalyticsClient.trackAction('User Dark Mode Preference', {
+    darkModeEnabled:
+      window.matchMedia &&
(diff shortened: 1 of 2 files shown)


src/js/site/main.js
```diff
@@ -421,3 +421,21 @@ $(function () {
     });
   });
 });
+
+/*
+  Check if users have dark mode enabled
+  Set key if user response has not already been tracked to ensure we dont get multiple events from the same user
+ */
+$(document).ready(() => {
+  const storageKey = 'provided-dark-mode-reponse';
```
nramabad, line 430:
```suggestion
  const storageKey = 'provided-dark-mode-response';
```

src/js/site/main.js
```diff
@@ -421,3 +421,21 @@ $(function () {
...
+  if (localStorage.getItem(storageKey)) {
+    return;
+  } else {
+    localStorage.setItem(storageKey, true);
+    window.AnalyticsClient.trackAction('User Dark Mode Preference', {
+      darkModeEnabled:
+        window.matchMedia &&
+        window.matchMedia('(prefers-color-scheme: dark)').matches,
```
nramabad, line 438: I just want to verify that the trackUser allows us to match the analytics_id to a trackAction. in cases where localStorage cannot cover such as when we have users using multiple browsers/machines.

teesloane (changes requested):
This looks great - I like the localstorage idea.

Can we extract the document ready callback into a function? I believe we are going to be refactoring those out pretty soon anyway, so we might as well start at it now.

src/js/site/main.js
tshen815: analytics_id is User ID in Amplitude, so we can group the data using User ID for an event thats associated with a trackAction to see all their results.

tshen815:
> This looks great - I like the localstorage idea.
> 
> Can we extract the document ready callback into a function? I believe we are going to be refactoring those out pretty soon anyway, so we might as well start at it now.

Sure thing! Just to clarify, are you suggesting to use a IIFE for this new extracted function?

teesloane:
> > This looks great - I like the localstorage idea.
> > Can we extract the document ready callback into a function? I believe we are going to be refactoring those out pretty soon anyway, so we might as well start at it now.
> 
> Sure thing! Just to clarify, are you suggesting to use a IIFE for this new extracted function?

Erp, sorry that wasn't clear! I mean something like this:

```js
$(document).ready(() => {
  checkIfUserIsUsingDarkMode()
  // other()
  // functions()
  // we()
  // will()
  // call()
})

```

It's a bit of a stylistic thing but it will clean up our code in the long run, imo.

src/js/site/main.js
```diff
@@ -421,3 +421,21 @@ $(function () {
     });
   });
 });
+
+/*
+  Check if users have dark mode enabled
+  Set key if user response has not already been tracked to ensure we dont get multiple events from the same user
+ */
+$(document).ready(() => {
```
nramabad, line 429: dumb q but have you tried this without the document.ready? I'm unsure if we need it.

also to echo @teesloane, I think it would be preferrable to extract this into a unit testable function. and then call said function in app.js.

tshen815: You are correct sir! no `document.ready` needed. Ive imported the function to `app.js` where its being invoked

tshen815:
> > > This looks great - I like the localstorage idea.
> > > Can we extract the document ready callback into a function? I believe we are going to be refactoring those out pretty soon anyway, so we might as well start at it now.
> > 
> > 
> > Sure thing! Just to clarify, are you suggesting to use a IIFE for this new extracted function?
> 
> Erp, sorry that wasn't clear! I mean something like this:
> 
> ```js
> $(document).ready(() => {
>   checkIfUserIsUsingDarkMode()
>   // other()
>   // functions()
>   // we()
>   // will()
>   // call()
> })
> ```
> 
> It's a bit of a stylistic thing but it will clean up our code in the long run, imo.

Ah gotcha! No worries, Ive currently exported the function from `main.js` and imported it to `app.js`. Should we create a ticket and seperate PR to have everything called in once place in `main.js` like you suggested?

rsylvian (approved):
LGTM ✅

teesloane (approved):

PR #6228, opened 2022-01-19
[DD-293] Fix header Nav Height when Expanded

tshen815:
[Ticket](https://circleci.atlassian.net/browse/DD-293)
[Preview](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/header-preview/?force-all)

# Changes

- set header height with media query for views larger than 768px

# Description
Addressing bug where header becomes broken and expands in height if users shrink down the window, opens and closes the drop down menu, and then expands the window view again. Users have to reload the page to fix issue.

### DIFF
--- a/src/styles/_global-nav.scss
+++ b/src/styles/_global-nav.scss
@@ -13,7 +13,6 @@ $btn-selected-green: #00701B;
 $btn-outer-blue: #0078CA;
 $btn-hover-blue: #0062B0;
 $btn-selected-blue: #004A95;
-$desktop-header-nav-height: 68.5px;
 
 // Copy of outer blog-nav, renmaed to global-nav
 ///////////////////////////////////////////////////////////////////////////////
@@ -147,7 +146,7 @@ $desktop-header-nav-height: 68.5px;
     display: flex;
     flex-direction: row;
     align-items: center;
-    min-height: 48px;
+    max-height: 48px;
   }
 }
 
@@ -461,13 +460,6 @@ a.nav-skip-button {
 
 .no-display {
   display: none;
-}
-
-// Ensures header does not expand after opening the dropdown menu and expanding the window view size
-@media (min-width: $screen-sm) {
-  .global-nav--component .global-nav--row-desktop {
-    height: $desktop-header-nav-height;
-  }
 }
 
 // -- Docs Disco Experiments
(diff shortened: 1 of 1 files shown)


src/styles/_global-nav.scss
```diff
@@ -462,6 +462,13 @@ a.nav-skip-button {
   display: none;
 }
 
+// Ensures header does not expand after opening the dropdown menu and expanding the window view size
+@media (min-width: $screen-sm) {
+  .global-nav--component .global-nav--row-desktop {
+    height: 68.5px;
```
rsylvian, line 468: Where is that number coming from? It seems rather random 😄?

chantelle11: It might be a good idea to set a variable for the height and use it across different screen widths. The use case seems to be fixed for me now though.

tshen815: @rsylvian The height is from the height of the `global-nav--component` when the window size is > 768px. Basically just making sure the header stays at that height when the window expands back to > 768px.

tshen815: @chantelle11 unfortunately that another bug that we talked about :( I made a ticket for it in our backlog [here](https://circleci.atlassian.net/browse/DD-355)

rsylvian:
@tshen815 Something is wrong with your PR. Please rebase with `master` 🙂

rsylvian (changes requested):
@tshen815 and I paired on this and we found a better way to handle this

src/styles/_global-nav.scss
```diff
@@ -146,7 +146,7 @@ $btn-selected-blue: #004A95;
     display: flex;
     flex-direction: row;
     align-items: center;
-    min-height: 48px;
```
rsylvian, line 149: We should probably keep both now that I think about it. We also want to make sure that the header is never under 48px.

rsylvian (approved):
✅  LGTM! One quick comment to address and this is ready to be shipped 🚢

src/styles/_global-nav.scss
tshen815: good call, done!

src/styles/_global-nav.scss
chantelle11: Sounds good!

PR #6235, opened 2022-01-20
Fix: Bring back DD synthetics tests for CI

rsylvian:
# Description
- Use new Tests IDs for tests running on CI

# Reasons
The DD synthetics tests got removed by mistake and so we had to create new ones and adjust the IDs in the `.datadog` directory. 

# Results
Screen Shot 2022-01-19 at 6 20 30 PM

### DIFF
--- a/.datadog/development.synthetics.json
+++ b/.datadog/development.synthetics.json
@@ -1,10 +1,10 @@
 {
-    "tests": [
-        {
-            "id": "as2-nqg-wcu"
-        },
-        {
-            "id": "3tu-7d6-qk7"
-        }
-    ]
+  "tests": [
+    {
+      "id": "wvz-xga-k4d"
+    },
+    {
+      "id": "9gh-kp8-7pp"
+    }
+  ]
 }
--- a/.datadog/production.synthetics.json
+++ b/.datadog/production.synthetics.json
@@ -1,10 +1,10 @@
 {
-    "tests": [
-        {
-            "id": "sfs-zqk-cfq"
-        },
-        {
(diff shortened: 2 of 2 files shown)


nramabad:
I think you need the port number :4000 in the host URL for it to work.

Also if you're able to add them like {{ HOST_URL }}/{{ JEKYLL_BASENAME }}/ & {{ HOST_URL }}/{{ JEKYLL_BASENAME }}/2.0/configuration-reference/ and then specify the environment variable in the test, that would be better.

The preview synthetics use these same test but pass arguments for the 2 environment variables.

rsylvian:
> I think you need the port number :4000 in the host URL for it to work.

That fixed it, thanks!

> Also if you're able to add them like {{ HOST_URL }}/{{ JEKYLL_BASENAME }}/ & {{ HOST_URL }}/{{ JEKYLL_BASENAME }}/2.0/configuration-reference/ and then specify the environment variable in the test, that would be better.
> 
> example: https://app.circleci.com/pipelines/github/circleci/circleci-docs/35331/workflows/33b0abad-9d26-41c9-9170-c4f53ba229ed/jobs/104378
> 
> The preview synthetics use these same test but pass arguments for the 2 environment variables.

Not a bad idea, I add this to the Blue Tape list. Trying to keep this PR focused on just bringing them back

chantelle11 (approved):
Looks good to me! The issue I had when I found this is that the CI would pass but there was actually errors if you opened the tab. If possible if the tests are not found the CI should actually fail. 🥳🧚🌺

PR #6236, opened 2022-01-20
Moved kotsadm-kots error to correct column

catherinepope:
# Description
Moved kotsadm-kots error to correct column

### DIFF
--- a/jekyll/_cci2/server-3-overview.adoc
+++ b/jekyll/_cci2/server-3-overview.adoc
@@ -238,8 +238,9 @@ but can also be configured to be exported to other places (i.e. Datadog or Prome
 
 | kotsadm-kots
 | Licensing
-| The main KOTS application. Runs the KOTS Admin Console where upgrades and configuration of server take place	No admin console available.
+| The main KOTS application. Runs the KOTS Admin Console where upgrades and configuration of server take place.
 | No upgrades or configuration possible for server.
+No admin console available.
 |
 
 | kotsadm-migrations


jekyll/_cci2/server-3-overview.adoc
```diff
@@ -238,8 +238,9 @@ but can also be configured to be exported to other places (i.e. Datadog or Prome
 
 | kotsadm-kots
 | Licensing
-| The main KOTS application. Runs the KOTS Admin Console where upgrades and configuration of server take place	No admin console available.
+| The main KOTS application. Runs the KOTS Admin Console where upgrades and configuration of server take place.
 | No upgrades or configuration possible for server.
+No admin console available.
```
rosieyohannan, line 243:
```suggestion
| No upgrades or configuration possible for server. No admin console available.
```

rosieyohannan: I think this should be in the "what happens if it fails" column?

catherinepope: Doesn't that shift it into the Notes column? This is what my version looks like for me in Docker:
kotsadm

rosieyohannan: oh ok cool, the pic looks right so let's go with that, sorry!

rosieyohannan: merge away!

catherinepope: Those tables are really eye-boggling! 👀

PR #6170, opened 2022-01-11
Feat: Add a `yarn dev` cmd for local dev without Docker

rsylvian:
# Description
- Add a `yarn dev` command that runs everything needed locally
- Add documentation to `local-development.md`

# Reasons
It seems that Jekyll & Docker don't mix well together and so this PR creates a new command for those that wish to run everything locally. The first results are pretty exciting because caching is done locally and build tools can use the full power of the machine (instead of only half allocated CPUs in Docker). The very first generation is taking the usual duration (120-140s) but the following ones are blazing fast. It seems like content changes are `under <5s`, CSS changes `under <1s` and js changes around `2-3s`.

Special thanks to @devinmarieb, @chantelle11, @teesloane & @tshen815 for beta-testing the changes and helping me uncovering some issues ahead of time 🙌

PR #6239, opened 2022-01-20
Feat: Add a `yarn dev` cmd for local dev without Docker

rsylvian:
This is a re-run of https://github.com/circleci/circleci-docs/pull/6170

# Description
- Add a `yarn dev` command that runs everything needed locally
- Add documentation to `local-development.md`

# Reasons
It seems that Jekyll & Docker don't mix well together and so this PR creates a new command for those that wish to run everything locally. The first results are pretty exciting because caching is done locally and build tools can use the full power of the machine (instead of only half allocated CPUs in Docker). The very first generation is taking the usual duration (120-140s) but the following ones are blazing fast. It seems like content changes are `under <5s`, CSS changes `under <1s` and js changes around `2-3s`.

Special thanks to @devinmarieb, @chantelle11, @teesloane & @tshen815 for beta-testing the changes and helping me uncovering some issues ahead of time 🙌


jekyll/_config_production.yml
```diff
@@ -4,3 +4,40 @@ rollbar_env: production
...
+  - scope:
+      path: ""
+    values:
+      toc: true
+
+  # Disable helpful resources section by default
+  - scope:
+      path: ""
+    values:
+      suggested: false
```
rsylvian, line 37: This was the reason for the incident. Originally it was set as: 

```
defaults:
  - scope:
        path: ""
      values:
        sitemap: true
```

But that has the side effect of replacing the entire `default` object instead of merging that field with what is coming from `_config.yml`

chantelle11 (approved):
Works for me! 🤩🥳🌺

devinmarieb (approved):
🙌

PR #6247, opened 2022-01-21
nomad-autoscaler is an anchor not a page

ryanwohara:
`nomad-autoscaler` is an anchor not a page, so it 404's in its current state.
Also fixed some whitespace issues.


rosieyohannan:
ooo good spot thanks! I'll check the changelog too

rosieyohannan (approved):

PR #6253, opened 2022-01-22
Create It isn’t a need to have to

Mesjeudis10:
No allowing a person to be where they aren’t supposed to live 

giving personal permission to an individual to go to where they need to be

if not to other is not


Mesjeudis10:
> No allowing a person to be where they aren’t supposed to live
> 
> giving personal permission to an individual to go to where they need to be
> 
> if not to other is not

``

PR #6261, opened 2022-01-24
Update supported k8s chart for server

ryanwohara:
SERVER-1537
SERVER-1551

It's worth noting here that we technically support k8s 1.22 on server,
but KOTS is the limiting factor as it only supports up to 1.21.
https://kots.io/kotsadm/installing/system-requirements/#kubernetes-version-compatibility


rosieyohannan (approved):
Thank you!

PR #6268, opened 2022-01-25
Fix broken numbers and indentations on page

devinmarieb:
# Description
Fixing the markdown numbered lists and indentations on this page as they were broken.

**Before**
Screen Shot 2022-01-25 at 10 03 18 AM

**After**
Screen Shot 2022-01-25 at 10 03 38 AM

# Reasons
The lists kept resetting at 1 because of the code blocks, which breaks the markdown logic.


kbatuigas (approved):
Looks great!

PR #6290, opened 2022-01-28
Removed outdated note about Scheduled Workflows and random line breaks

catherinepope:
# Description
- Removed outdated note about Scheduled Workflows
- Removed random line breaks

# Reasons
Spotted while working on DOCSTEAM-59!


devinmarieb (approved):
thanks!

PR #6302, opened 2022-01-31
[FIX] Initialize lang.js when DOM is Loaded

tshen815:
# Changes

- initializing `lang.js` when the DOM is fully loaded in `app.js`

# Description
We are getting errors in rollbar due to `langValue.domEl` being null as we are trying to add an event to the element before the DOM has been fully loaded and the element has been defined

# Reasons
Addressing the errors from rollbar [here](https://rollbar.com/circle/circleci-docs/items/2901/?item_page=0&)

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -16,7 +16,6 @@ window.OptimizelyClient = new services.OptimizelyClient();
 
 import './site';
 
-services.lang.init();
 services.rum.init();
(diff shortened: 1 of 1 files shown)


chantelle11 (approved):
Perfect! lets go!

rsylvian (approved):

PR #6305, opened 2022-02-01
[DD-396] Update Design for Language Selector

tshen815:
[Ticket](https://circleci.atlassian.net/browse/DD-396)
[Preview Final Design](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/lang-requestor-preview/?force-all)
**NOTE**: Will need to review changes to design with @danbeere 

# Changes

- add `greenCheck.svg` for selected language
- update design for selected language

# Description
Updating the design of our language selector to match the current design that is present in outer.

# Reasons
The design changes were most likely to account for people that have trouble seeing certain colors. We should make the current language selector adhere to the same design for both consistency and accessibility.

# Screenshots
**Before**:
Screen Shot 2022-02-01 at 10 57 47 AM
**After**:
Screen Shot 2022-02-02 at 5 49 27 PM


src/styles/_langselector.scss
```diff
@@ -69,24 +69,31 @@ $border-col: rgba(229,231,235, 1);
...
     margin: 4px 0px;
     margin-top: 0px;
-    padding: 4px 16px;
+    padding: 4px 20px;
     font-size: 15px;
     font-family: 'Roboto', 'Helvetica Neue', Helvetica, Arial, sans-serif;
     line-height: 18px;
     &.selected {
-      background: #F3F3F3;
+      // background: #F3F3F3;
```
chantelle11, line 77: Nit: but can we just remove this?

src/styles/_langselector.scss
```diff
@@ -69,24 +69,31 @@ $border-col: rgba(229,231,235, 1);
...
       @media (max-width: $screen-xs-max) {
         font-weight: bold;
       }
     }
 
     &:hover {
-      color: $btn-hover-blue;
+      background: #F3F3F3;
```
chantelle11, line 84: Should we possibily share this background color?
```$nav-bg: #F3F3F3 !default;```

File: src/js/services/lang.js [hunk 3]
```diff
@@ -45,7 +47,7 @@ const setLanguageSelectorOnLoad = () => {
     languages[window.currentLang] ||
     languages[window.navigator.language] ||
     languages['en'];
-  $(currentLang.domEl).addClass('selected');
+  $(currentLang.activeLang).toggleClass('no-display', false);
```
chantelle11, line 50: I think this only works because we are actually changing the url to change the language. I think it works but just something to keep in mind.

chantelle11 (commented):
Great looking forward to seeing more!! 🌺👍💕

src/styles/_langselector.scss
tshen815: good call!

src/js/services/lang.js
```diff
@@ -11,13 +11,15 @@ const languages = {
     url: 'docs/',
     id: 'globalNavLang-en',
     domEl: getElById('globalNavLang-en'),
+    activeLang: getElById('select-lang-en'),
```
teesloane, line 14: Do we need to add this `activeLang` property? We could use the existing`domEl` query and target the image tag within it? In pseudocode:

$(currentLang.domEl > "classname_of_checkmark_image").toggleClass etc.

This way we don't have to add a unique id to every img tag for each language we have.

teesloane: I acknowledge that that the above is a nit.

nramabad (changes requested):
I don't think you want `.no-display`. `display: none` does not "reserve" the space in the DOM for your element. You want an empty space when a language is not selected instead of the entire line to shift to the left.

Try using `visibility: visible` & `visibility: hidden` instead: https://www.w3schools.com/cssref/pr_class_visibility.asp

src/js/services/lang.js
tshen815: very cool! I wasnt aware that we could do that. Ill make the updates 👍

tshen815:
> I don't think you want `.no-display`. `display: none` does not "reserve" the space in the DOM for your element. You want an empty space when a language is not selected instead of the entire line to shift to the left.
> 
> Try using `visibility: visible` & `visibility: hidden` instead: https://www.w3schools.com/cssref/pr_class_visibility.asp

This is actually something that I need to check in with @danbeere about during our design review tomorrow as Im not sure what alignment he would prefer between the language options, and the selected check in relation to the language input box.

nramabad:
> > I don't think you want `.no-display`. `display: none` does not "reserve" the space in the DOM for your element. You want an empty space when a language is not selected instead of the entire line to shift to the left.
> > Try using `visibility: visible` & `visibility: hidden` instead: https://www.w3schools.com/cssref/pr_class_visibility.asp
> 
> This is actually something that I need to check in with @danbeere about during our design review tomorrow as Im not sure what alignment he would prefer between the language options, and the selected check in relation to the language input box.

definitely consult w/ Dan. I'm just going off what's in outer, it looks like they keep the space when the line isn't selected which to me looks like visibility hidden.

src/js/services/lang.js
tshen815: I was having some trouble with the syntax, tried a few other ways including using a template literal but it didnt seem to be able to target the child element. I'm using `$(currentLang.domEl).children('.select-lang-wrap').toggleClass('no-display', false);` and its working nicely 👌

tshen815:
> > > I don't think you want `.no-display`. `display: none` does not "reserve" the space in the DOM for your element. You want an empty space when a language is not selected instead of the entire line to shift to the left.
> > > Try using `visibility: visible` & `visibility: hidden` instead: https://www.w3schools.com/cssref/pr_class_visibility.asp
> > 
> > 
> > This is actually something that I need to check in with @danbeere about during our design review tomorrow as Im not sure what alignment he would prefer between the language options, and the selected check in relation to the language input box.
> 
> definitely consult w/ Dan. I'm just going off what's in outer, it looks like they keep the space when the line isn't selected which to me looks like visibility hidden.

I mocked up a few designs to go over with dan tomorrow, for this one I'm using visibility http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/lang-select-design-2-preview/?force-all

chantelle11 (commented):
On outer it seems like it is bolded the selected text plus the check mark. Was the bold checked with @danbeere ? Other than that I think it works great for me :)
Screen Shot 2022-02-03 at 2 00 06 PM

nramabad:
> > > > I don't think you want `.no-display`. `display: none` does not "reserve" the space in the DOM for your element. You want an empty space when a language is not selected instead of the entire line to shift to the left.
> > > > Try using `visibility: visible` & `visibility: hidden` instead: https://www.w3schools.com/cssref/pr_class_visibility.asp
> > > 
> > > 
> > > This is actually something that I need to check in with @danbeere about during our design review tomorrow as Im not sure what alignment he would prefer between the language options, and the selected check in relation to the language input box.
> > 
> > 
> > definitely consult w/ Dan. I'm just going off what's in outer, it looks like they keep the space when the line isn't selected which to me looks like visibility hidden.
> 
> I mocked up a few designs to go over with dan tomorrow, for this one I'm using `visibility` http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/lang-select-design-2-preview/?force-all

Looks better to my eye. One question for @danbeere is it looks like we're diverging from outer for the padding-left/right? Outer's padding looks more. If this is the intent, then no worries. Just wondering.

The one thing @tshen815 if you can fix is related to when a user makes a selection. There is some delay before the new page loads. Can we leave the selected option bolded to indicate the user successfully clicked and it's just a delay page loading. I think this would be really nice given that Japan has a lot of response time delay.

Above and beyond would be adding a spinning loading icon or something but that's Dan's call.

nramabad (approved):
Left some questions/feedback. Looks good otherwise. I'm going to approve but if you can look into that feedback, that would be great!

tshen815:
> > > > > I don't think you want `.no-display`. `display: none` does not "reserve" the space in the DOM for your element. You want an empty space when a language is not selected instead of the entire line to shift to the left.
> > > > > Try using `visibility: visible` & `visibility: hidden` instead: https://www.w3schools.com/cssref/pr_class_visibility.asp
> > > > 
> > > > 
> > > > This is actually something that I need to check in with @danbeere about during our design review tomorrow as Im not sure what alignment he would prefer between the language options, and the selected check in relation to the language input box.
> > > 
> > > 
> > > definitely consult w/ Dan. I'm just going off what's in outer, it looks like they keep the space when the line isn't selected which to me looks like visibility hidden.
> > 
> > 
> > I mocked up a few designs to go over with dan tomorrow, for this one I'm using `visibility` http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/lang-select-design-2-preview/?force-all
> 
> Looks better to my eye. One question for @danbeere is it looks like we're diverging from outer for the padding-left/right? Outer's padding looks more. If this is the intent, then no worries. Just wondering.
> 
> The one thing @tshen815 if you can fix is related to when a user makes a selection. There is some delay before the new page loads. Can we leave the selected option bolded to indicate the user successfully clicked and it's just a delay page loading. I think this would be really nice given that Japan has a lot of response time delay.
> 
> Above and beyond would be adding a spinning loading icon or something but that's Dan's call.

Updated the preview link with the newest styles as per requested by dan this morning after our design review. I made the current selected language bold as well (`font-weight: 700`) in addition the the check mark and current hover behavior. Ill run the loading icon by dan and see if thats something we can add.

tshen815:
> On outer it seems like it is bolded the selected text plus the check mark. Was the bold checked with @danbeere ? Other than that I think it works great for me :) Screen Shot 2022-02-03 at 2 00 06 PM

Applied the changes to bold!

PR #6308, opened 2022-02-01
Update image upgrade example

devinmarieb:
# Description
A duplicate of a stuck image PR that updates the example of how to change the namespace for the new images.

# Reasons
People need to be aware they need to change the namespace of the image, not just the version number.


kbatuigas (approved):
Thank you!

PR #6315, opened 2022-02-02
Add CTA to the top of the code pane in API v2

teesloane:
# Description
It was requested that we add a clarifying call to action to the top code samples pane in our API v2 documentation.

Ticket: [DD-323](https://circleci.atlassian.net/browse/DD-323)

Todo: Waiting on suggestions or changes to copy.

# Reasons
Users were expression confusion at the blank space at the top of the page.

# Screenshots 

## Before
image

## After

image

### DIFF
--- a/src-api/v2/template.hbs
+++ b/src-api/v2/template.hbs
@@ -13,6 +13,30 @@
       padding: 0;
       margin: 0;
     }
+
+    .side-section-cta {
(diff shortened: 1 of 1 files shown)


rsylvian (approved):
Approving so we don't have to review again once we have the final copy! Glad we were able to do something about that 🙌

PR #6328, opened 2022-02-04
add note about dangers of very specific cache keys

rosieyohannan:
https://circleci.atlassian.net/browse/DOCSTEAM-141?atlOrigin=eyJpIjoiYWRmYjBmMGI2YzZiNGZlOWIwYzBjOTA3ZjYzNjQ5ZDYiLCJwIjoiaiJ9

### DIFF
--- a/jekyll/_cci2/caching.md
+++ b/jekyll/_cci2/caching.md
@@ -29,7 +29,7 @@ Caching is particularly useful with **package dependency managers** such as Yarn
 
 Automatic dependency caching is not available in CircleCI, so it is important to plan and implement your caching strategy to get the best performance. Manual configuration enables advanced strategies and fine-grained control. See the [Caching Strategies]({{site.baseurl}}/2.0/caching-strategy/) and [Persisting Data]({{site.baseurl}}/2.0/persist-data/) guides for tips on caching strategies and management.
 
-This document describes the manual caching options available, the costs and benefits of a chosen strategy, and tips for avoiding problems with caching. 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/caching.md
```diff
@@ -277,8 +277,11 @@ Template | Description
 
 The following example demonstrates how to use `restore_cache` and `save_cache`, together with templates and keys in your `.circleci/config.yml` file.
 
+This example uses a _very_ specific cache key. Making your caching key more specific gives you greater control over which branch or commit dependencies are saved to a cache, it is important be aware that this can also have a significant increase on storage usage. For tips on optimizing your caching strategy, see the [Caching Strategies]({{site.baseurl}}/2.0/caching-strategy) guide.
```
rosieyohannan, line 280:
```suggestion
This example uses a _very_ specific cache key. Making your caching key more specific gives you greater control over which branch or commit dependencies are saved to a cache. However, it is important be aware that this can **significantly increase** your storage usage. For tips on optimizing your caching strategy, see the [Caching Strategies]({{site.baseurl}}/2.0/caching-strategy) guide.
```

devinmarieb (approved):
@rosieyohannan looking at this makes me think we should add a section to the style guide about how we go about implementing banners (unless I missed it -- I look at the SG yesterday). Sometimes we use HTML and sometimes we just use the class. Would be good to figure out what's best (especially from an accessibility stand point - my guess is HTML is probably better).

PR #6326, opened 2022-02-04
[DD-408] Use PrismJS for code highlighting 

rsylvian:
Ticket: [DD-408](https://circleci.atlassian.net/browse/DD-408)
Preview: [link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/rs/feat/add-prismjs-preview/?force-all
)

# Description
- Replace HighlightJS with PrismJS
- Update all `.adoc` & `.md` to use `shell` instead of `sh` (more [info](https://github.com/circleci/circleci-docs/pull/6326#discussion_r799119956))
- Add babel prismjs plugin to be able to only bring JS/CSS we need (rather than all PrismJS+plugins)
- Update [snippetFeedback experiment](https://github.com/circleci/circleci-docs/pull/6326/files#diff-fb79bc521a0dfe6d7459e382d8772b5809578ed1e0358d1cc42efb737901a7e0) to be compatible with PrismJS
-  Update highlight theme to closely match our in-ap config editor

# Reasons
Recently, we have been experiencing a few issues with highlightJS. We experimented with PrismJS and it is great for our use cases. 

# Screenshots

before:
screely-1644029702129

after:
screely-1643942689300

PR #6324, opened 2022-02-03
[BUG FIX] Node references in Python

nramabad:
# Description
Bug fix for Node.js references in Python tutorial

# Reasons
Typos can be confusing.

PR #6330, opened 2022-02-04
Use /tmp/ instead of ~/ in Mocha example for Collecting Test Metadata

noahlitvin:
# Description
This suggests using `/tmp/junit` instead of ` ~/junit` (per the sample code under _Metadata collection in custom test steps_) in the Mocha example for Collecting Test Metadata.

# Reasons
I couldn't get this working without making this change.

### DIFF
--- a/jekyll/_cci2/collect-test-data.md
+++ b/jekyll/_cci2/collect-test-data.md
@@ -143,16 +143,16 @@ A working `.circleci/config.yml` section for testing might look like this:
     steps:
       - checkout
       - run: npm install
-      - run: mkdir ~/junit
+      - run: mkdir /tmp/junit
       - run:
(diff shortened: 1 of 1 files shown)


thomasr8 (approved):

ESA-Security-Group (approved):
#6324 #6326 #6332

chadchabot (changes requested):
I don't think it should matter whether the path is  `~/junit` or `/tmp/junit` as the directory is created as part of the job steps.

Are you able to share an example of a build (or logs from your machine) that failed when using the previous junit path? Maybe there was a permission issue when creating the directory or something along those lines.

Looking at the other paths to various testing tool output locations, they all point to `~/TOOL_NAME`. I don't expect junit to be special in this regard, needing a specific and different path value.

PR #6332, opened 2022-02-04
Update macOS docs for freemium launch

alexazl:
# Description
Remove warnings in docs that using macOS executor requires a paid plan and isn't supported on server.

# Reasons
With today's launch of freemium support for macOS, the paid plan prerequisite is no longer accurate. Additionally, Server has supported macOS since v3.1


kbatuigas (approved):
Thank you, @alexazl !

PR #6343, opened 2022-02-08
Remove duplicate content

rosieyohannan:
Steps to view network and storage usage are shown in the Persisting Data page. This PR removes these steps from the Workspaces doc and points to the Persisting Data page instead.

### DIFF
--- a/jekyll/_cci2/workspaces.md
+++ b/jekyll/_cci2/workspaces.md
@@ -93,17 +93,7 @@ Workspaces are stored for up to 15 days. Workspaces are not shared between pipel
 
 When using self-hosted runners there is a network and storage usage limit included in your plan. Once your usage exceeds your limit charges will apply. These charges are based on your accrued overages.
 
-### Viewing network egress usage
-{: #viewing-network-egress-usage }
-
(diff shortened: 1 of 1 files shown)


devinmarieb (commented):
This is fine for now, but we might just want to try and remember that we did this here. Again, the information in the Persisting Data page about billing was only mean to be temporary, as a landing page where users could see this information now that the feature was opened up to them (which has been delayed by 3 months now). In the future I think the FAQ page might be better. If you think the information in the FAQ also works, it might be better to point there for credit/billing info.
https://circleci.com/docs/2.0/faq/#how-do-I-calculate-my-monthly-storage-and-network-costs

rosieyohannan:
@devinmarieb I'm going to merge this as is and then make a ticket to look at this more closely. Looks to me like the overview in the Persisting Data page is more conceptual about Network and Storage usage - valid dev docs thing? Whereas the FAQ going deeper into pricing calculations which I totally agree is more of a FAQ and support area. Totally happy to change this though, can discuss in the ticket if that's OK.

devinmarieb (approved):
@rosieyohannan That's fine! I will likely be deleting the billing section in the Persisting Data page at the end of March. Instead it will link out to the FAQ page.

PR #6345, opened 2022-02-08
[DD-291] Minify HTML output

teesloane:
# Description
Ticket - [DD-291](https://circleci.atlassian.net/browse/DD-291)

This Pr introduces HTML compression [via liquid](http://jch.penibelst.de/). 
[Preview Link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-291-preview/?force-all)

|Document | Before | After |
|---- | -------| ------|
| Configuration Reference | 387K | 295 kb |
| workflows                         |  241K  | 161 kb |

Benchmarks on size were handled by building the site locally and comparing the index.html file sizes for each document (rather than testing the difference between prod and preview sites as the former uses gzip encoding while the latter does not). 

Edit: This PR will compress many html files and while it shouldn't affect our JS, we still need to do diligent manual testing: please test the sidebar, footer, lang picker, mobile, experiment appearance and functionality if possible!


jekyll/_layouts/compress.html
```diff
@@ -0,0 +1,10 @@
...
...l dt dd figure figcaption main div table caption colgroup col tbody thead tfoot tr td th" | split: " " %}{% else %}{% assign _clippings = site.compress_html.clippings %}{% endif %}{% for _element in _clippings %}{% assign _edges = " <e;<e; </e>;</e>;</e> ;</e>" | replace: "e", _element | split: ";" %}{% assign _content = _content | replace: _edges[0], _edges[1] | replace: _edges[2], _edges[3] | replace: _edges[4], _edges[5] %}{% endfor %}{% if _profile and _clippings %}{% assign _profile_clippings = _content | size | plus: 1 %}{% endif %}{{ _content }}{% if _profile %} <table id="compress_html_profile_{{ site.time | date: "%Y%m%d" }}" class="compress_html_profile"> <thead> <tr> <td>Step <td>Bytes <tbody> <tr> <td>raw <td>{{ content | size }}{% if _profile_endings %} <tr> <td>endings <td>{{ _profile_endings }}{% endif %}{% if _profile_startings %} <tr> <td>startings <td>{{ _profile_startings }}{% endif %}{% if _profile_comments %} <tr> <td>comments <td>{{ _profile_comments }}{% endif %}{% if _profile_collapse %} <tr> <td>collapse <td>{{ _profile_collapse }}{% endif %}{% if _profile_clippings %} <tr> <td>clippings <td>{{ _profile_clippings }}{% endif %} </table>{% endif %}{% endif %}
```
chantelle11, line 10: This seems like a very long line of code. Is it possible to make this more readable? Or is this the result of compression?

rsylvian (changes requested):
I think that method might not be the best because it is also "compressing" (or rather removing `\n`) for the code snippet inside:`<pre><code>`

http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-291-preview/2.0/migrating-from-github/

Are you able to skip some sections?

jekyll/_layouts/compress.html
teesloane: Good question. So... this `compress.html` file layout is the actual "library" that does the compression, and we use it as the parent template for classic docs so that everything inside it gets compressed. This format is actually the way the library is released, so I just copied and pasted it in (there is no un-minified version, but that would be nice!)

chantelle11 (approved):
Nice work on this! ✌️🌺👑

rsylvian (changes requested):
@tshen815 and I were testing and we realized that this is behaving a little strange:

If you check the source of this [link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/html-minify-preview/2.0/migrating-from-github/?force-all), there are 62 empty lines in the source:

Screen Shot 2022-02-15 at 6 33 46 PM

teesloane:
> @tshen815 and I were testing and we realized that this is behaving a little strange:
> 
> If you check the source of this [link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/html-minify-preview/2.0/migrating-from-github/?force-all), there are 62 empty lines in the source:
> 
> Screen Shot 2022-02-15 at 6 33 46 PM

Good catch.

This was because I was using an unminified version of `_layouts/compress.html` to address [this comment](https://github.com/circleci/circleci-docs/pull/6345#discussion_r802222413) but I've reverted - there should no longer be 62 lines of empty text at the beginning of each document.

tshen815 (approved):
Thanks for addressing the feedback!

rsylvian (approved):

PR #6348, opened 2022-02-09
DD-319 click to enlarge photo

chantelle11:
Ticket: [DD-319](https://circleci.atlassian.net/browse/DD-319)
Preview: [link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-319-click-to-enlarge-image-preview/?force-all)

# Note
The preview build does not load pictures in asciidoc. This has been verified in other preview builds outside of this one. If you build the application locally you will see the images in asciidoc.

# Description
Ability for users to click on the image which results in the enlarged image so they can read the text in the image clearly

# Impact
Enable users to enlarge the image if the original image is to dense and content in it is not legible

# Things to consider
Close caption is not part of original design but see Pending questions
When the user clicks on an image it enlarges with a legible content
Clicking anywhere when open or X-icon closes the image (X-icon color changes when hover)
Thoughts on max-width: 1024px
No max height to not squish images

# Design 
[Image opened](https://www.figma.com/file/HrgpUU2M3ED7OT0p0f99bH/Docs-Disco-Experiments?node-id=5098%3A9978) 

[Image opened hover](https://www.figma.com/file/HrgpUU2M3ED7OT0p0f99bH/Docs-Disco-Experiments?node-id=5098%3A9980)

[Clickable prototype](https://www.figma.com/proto/HrgpUU2M3ED7OT0p0f99bH/Docs-Disco-Experiments?page-id=5089%3A10001&node-id=5098%3A9981&viewport=312%2C48%2C0.27&scaling=scale-down&starting-point-node-id=5098%3A9981)

# Answers
1. Yes add close captions for images that have alt
2. Don't have to cover banner for enlarge image
3. X changes colours when you hover it
4. Add cursor pointer

# Before
Static picture

Screen Shot 2022-02-08 at 10 13 50 PM

# After
**Note** the white in the background is because the image has a white background and is not transparent making this effect
smalldemoOfClickOpenBig


src/js/site/expandImage.js
```diff
@@ -0,0 +1,37 @@
+/*
+ * Add enlarge image capability for easier readability
+ */
+export function expandImageOnClick() {
+  // create references to the modal...
+  var modal = document.getElementById('expandModal');
```
tshen815, line 6: can we use `const` here?

tshen815: also, could we use `document.getElementByClass('modal')` instead? There doesnt seem to be another element with the class `modal` and the id `expandModal` has no css applied to it as well. 

you could still do `document.getElementById('expandModal')` but then I dont think we need the class, and we can change the css for modal to `#expandModal {}`?

src/js/site/expandImage.js
```diff
@@ -0,0 +1,37 @@
...
+
+  // get all images without class icon or logo, inside or article but also have src `.png`
+  // it is most likely that a icon or logo is src `.svg`
+  let images = document.querySelectorAll(
+    "article img:not(a):not(.icon):not(.logo)[src*='.png']",
```
tshen815, line 11: cool stuff here, didnt know you could do all that!

src/js/site/expandImage.js
```diff
@@ -0,0 +1,37 @@
...
+  );
+
+  // create reference to add caption
+  let captionText = document.getElementById('caption');
+
+  for (let i = 0; i < images.length; i++) {
+    let img = images[i];
+    // and attach our click listener for this image.
+    img.onclick = function () {
+      modal.style.display = 'block';
```
tshen815, line 21: can we add a class here with `display: block` instead of setting the css? Similar to what you're doing for `img.classList.add('enlarge');`

tshen815:
This looks great! 🤩 Just left a few comments on it. Also I noticed that some pictures dont have a caption as its missing an `alt` attribute. For SEO and best practices maybe we should do an audit and add them in?

src/js/site/expandImage.js
chantelle11: Great catch!

chantelle11:
> This looks great! 🤩 Just left a few comments on it. Also I noticed that some pictures dont have a caption as its missing an `alt` attribute. For SEO and best practices maybe we should do an audit and add them in?

I agree! At the very least we should put it part of our best practices and try to add alts.

src/js/site/expandImage.js
chantelle11: it can't be const because i need to change the model with the close caption but it can be `let`

tshen815: gotcha, could we change the css for `.modal` to `#expandModal`? The modal class doesnt seem to be used anywhere else.

tshen815: Im not sure I follow for the close caption? I was able to test it out and `const` seems to be working well for modal 👍

tshen815:
Something else came into mind when I was testing your branch! Could we add a `cursor: pointer` to the images on hover? I feel like that will make it more apparent that the images are now clickable.

src/js/site/expandImage.js
chantelle11: Can do the css changes! good point!

tshen815 (approved):
Looks great to me! Thanks for making the edits! 🎉

src/js/site/main.js
```diff
@@ -287,6 +288,9 @@ $(document).ready(function () {
 // Currently this function is only used for the insights table
 $(highlightURLHash);
 
+// This function is used to be able to expand images when you click them
+$(expandImageOnClick);
```
teesloane, line 292: What does wrapping a function in `$(<my func>)` do? 

I'm guessing it replaces the need to use jquery's document.onready call. If this is the case, can we add this function call into one of the existing onready's, perhaps [this one](DD-319-click-to-enlarge-image)?

teesloane (commented):
Nice work — this is such a good user experience improvement.

I added a minor question about document ready calls out of curiousity, but I'm still approving as this looks good to go to me.

teesloane (approved):

src/js/site/main.js
chantelle11: I have added it here
```
// Used to call functions on document on ready
$(function() {
  // Currently this function is only used for the insights table
  highlightURLHash();
  // This function is used to be able to expand images when you click them
  expandImageOnClick();
});
```

jekyll/_layouts/classic-docs.html
```diff
@@ -47,6 +47,12 @@
             {% if page.collection != 'cci1' and page.page-type != "index" and "classic-docs-experimental|classic-docs|enterprise" contains page.layout %}
             {% include_cached doc-footer.html page=page site=site %}
             {% endif %}
+
```
rsylvian, line 50:
```suggestion
```

✂️

src/styles/main.scss
```diff
@@ -678,3 +678,82 @@ Fixing styling for the code example tables in the "Migrating From..." pages
...
+  left: 0;
+  top: 0;
+  width: 100%;
+  /* Full width */
+  height: 100%;
+  /* Full height */
+  overflow: auto;
+  /* Enable scroll if needed */
+  background-color: rgba(0, 0, 0, 0.8);
+  /* Black w/ opacity */
```
rsylvian, line 705:
```suggestion
  position: fixed;
  z-index: 1;
  padding-top: 120px;
  left: 0;
  top: 0;
  width: 100%;
  height: 100%;
  overflow: auto;
  background-color: rgba(0, 0, 0, 0.8);
```

rsylvian: Personal preference, but I don't think these comments are necessary

src/js/site/expandImage.js
```diff
@@ -0,0 +1,42 @@
...
+ * Add enlarge image capability for easier readability
+ */
+export function expandImageOnClick() {
+  // create references to the modal...
+  const modal = document.getElementById('expandModal');
+
+  // get all images without class icon or logo, inside or article but also have src `.png`
+  // it is most likely that a icon or logo is src `.svg`
+  const images = document.querySelectorAll(
+    "article img:not(a):not(.icon):not(.logo)[src*='.png']",
```
rsylvian, line 11: since you are avoiding the `icon` class, are all set on the content side to make sure that all the icons have an icon class already?

rsylvian (changes requested):

src/js/site/expandImage.js
chantelle11: So the main deciding factor is that its .png for none icon and .svg for icon. The icon class and logo class is because I have found an icon that was a png but it is rare. So it gives the extra check.

rsylvian (approved):

devinmarieb (approved):
Awesome! From the styles you've chosen, feel free to let us know what kind of images work best with the modal (with a shadow border, cropped tight, etc). We should make our style guide a little more rigid for images now that we have this feature 🥳

PR #6357, opened 2022-02-10
Fix network transfer guidance

aaronstillwell:
Only applicable to self-hosted runners

# Description
Update network transfer guidance to reflect the fact that it only applies to restoring caches/workspaces to self-hosted runners.

# Reasons
Current guidance is misleading.

### DIFF
--- a/jekyll/_cci2/faq.md
+++ b/jekyll/_cci2/faq.md
@@ -313,8 +313,7 @@ To calculate monthly storage costs from your daily usage, click on the **Storage
 {: #network }
 {:.no_toc}
 
-Billing for network usage is only applicable to traffic from CircleCI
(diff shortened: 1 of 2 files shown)


devinmarieb:
@stalmy and @aaronstillwell This is fine, but just as a heads up, we are currently overhauling an entire section, which includes the Persisting Data page. That page is going to become an overview page for a new section, which means the content is slightly changing. There's too much repetitive (and out of date) content in multiple places in the Caching, Persisting Data, and Optimization pages. The billing information will be taken out of the Persisting Data page soon as it won't make sense on this page anymore in the new structure. Instead we will link out to the Doc's FAQ page where the billing information will live.

devinmarieb (approved):

PR #6366, opened 2022-02-14
Update ssh-access-jobs.md

ogii:
# Description
Previously rerunning jobs with ssh would not create a visible job on the UI for the user to click on. 
This was recently changed to create the jobs inside of the same pipeline, and they are now shown on the pipelines page for a project.

# Reasons
The documentation is outdated, and as such I am changing the language to better reflect how the feature currently works.


rosieyohannan (approved):
Thanks so much for the update, much appreciated 🙏

PR #6368, opened 2022-02-14
We don't yet support k8s 1.22

ryanwohara:
SERVER-1632

We found a bug in our k8s 1.22 support. Until we update prometheus this will be the case.


rosieyohannan (approved):
Thank you @ryanwohara!

PR #6384, opened 2022-02-17
Update executor-types.md

devinmarieb:
Banner update - changed wording to indicate the deprecation has already happened.


rosieyohannan (approved):
Thank you!

PR #6389, opened 2022-02-17
Add dedicated host image manifest column

alexazl:
# Description
Replace UI testing column with bare metal image manifest column.

# Reasons
VM and bare metal (for dedicated hosts) images may have slight differences, so it's important to provide the exact configuration to our customers. UI testing is only unsupported by our older images, which will be deprecated within the next few months. Replacing this column with the bare metal manifest prevents overcrowding in the table.


kbatuigas (approved):
Thank you!

PR #6396, opened 2022-02-19
[DD-417] Update Signup CTA Copy

tshen815:
[Ticket](https://circleci.atlassian.net/browse/DD-417)
[Preview Link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-417/update-signup-CTA-experiment-preview/?force-all)
**NOTE:** Tested with the work @rsylvian did on #6427 guest user experiments. Was able to get treatment variation for as an unsigned in user.

To test out please log out of CCI and view in preview link. If you see the control variation delete your local host and cookies and refresh until you see variation copy

# Changes

- update copy for CTA button in header

# Description
Update the sign-up CTA button verbiage for non-logged-in users to "Start Building for Free" so it's more encouraging to sign-up.

# Considerations
We originally wanted to run this as an experiment but realized that we are unable to do so due to the user not being able to be segmented into a control or treatment group with Optimizely as they are not signed in. 

# Tracking Events

- `Clicked New Signup CTA`

# Screenshots
**Before:**
Screen Shot 2022-02-18 at 4 29 16 PM

**After:**
Screen Shot 2022-02-18 at 4 29 04 PM

### DIFF
--- a/src/js/experiments/signUpCTA.js
+++ b/src/js/experiments/signUpCTA.js
@@ -3,6 +3,7 @@ window.OptimizelyClient.getVariationName({
   experimentKey: 'dd_update_signup_cta_test',
   groupExperimentName: 'q1_fy23_docs_disco_experiment_group_test',
   experimentContainer: '.global-nav--footer',
+  guestExperiment: true,
 }).then((variation) => {
   if (variation === 'treatment') {
     const ctaBtn = document.getElementById('signup-cta');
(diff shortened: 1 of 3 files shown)


src/js/experiments/signUpCTA.js
```diff
@@ -0,0 +1,15 @@
+// https://app.optimizely.com/v2/projects/16812830475/experiments/21229840508/variations
+window.OptimizelyClient.getVariationName({
+  experimentKey: 'dd_update_signup_cta_test',
+  groupExperimentName: 'q1_fy23_docs_disco_experiment_group_test',
+  experimentContainer: '.global-nav--footer',
+  guestExperiment: true,
```
tshen815, line 6: `guestExperiment` is false by default. If we are running an experiment on an guest/not signed up user we have to add in this option and set it to true.

PR #6402, opened 2022-02-22
Remove deprecated remote Docker versions

felicianotech:
Removing old Docker versions that we either don't want new users on or are technically gone (see [Discuss](https://discuss.circleci.com/t/remote-docker-updates/43026)).

We will have a new Docker version that I will add in April.


catherinepope (approved):
👍

PR #6410, opened 2022-02-23
prepare for new UI -  do not merge yet

sebastian-lerner:
context: 

we are adding some new information to the Resources tab under the CPU/RAM utilization graphs to aid customers in predicting cost of enabling the IP ranges feature.  I want to prepare the docs and ideally add a screenshot as well.  something like this:

image

### DIFF
--- a/jekyll/_cci2/ip-ranges.md
+++ b/jekyll/_cci2/ip-ranges.md
@@ -143,7 +143,7 @@ IP ranges usage is visible in the "Plan Usage" page of the CircleCI app:
 
 ![Screenshot showing the location of the IP ranges feature]({{ site.baseurl }}/assets/img/docs/ip-ranges.png)
 
-On the Resources tab within the Job Details UI page, customers can view approximations of network transfer for any Docker job, even those without the IP ranges feature enabled.  This approximation can be used to predict the cost of enabling the IP ranges feature on a job without having to turn the feature on:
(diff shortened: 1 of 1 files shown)


sebastian-lerner:
we probably want a better screenshot though.  once the UI is live, i can take a cleaner screenshot that shows more of the screen as well

sebastian-lerner:
@devinmarieb what do you think about this screenshot?

image

or should we show the whole screen?

sebastian-lerner:
i guess with this screenshot it's hard to tell that we're talking about the bottom tile, not the RAM usage one.  is it weird if i add a line around the network transfer tile in the image? or is there a different way we like to do it with docs?

devinmarieb:
@sebastian-lerner We mark up images. Usually in that lime green color. Don't ask me why it's that lime green color, but that's what we seem to use. I can find a hex code for you if you need it. Usually as much of the UI as possible is good unless it makes it hard to see what's going on. We do now have a feature in the docs that allows a user to click on an image to enlarge it, so that's super helpful for detailed images.

sebastian-lerner:
okay cool, i'm hoping to have this go live today or monday, so i'll grab a new screenshot that shows more of the screen, add lime green line, and paste it in here.  if you have the hex that'd be helpful

sebastian-lerner:
Screen Shot 2022-02-25 at 1 45 08 PM

is that the right lime green?

devinmarieb (approved):

devinmarieb:
Closing as we needed to make these changes locally.

PR #6416, opened 2022-02-24
server-1647 Added Notes for using Private Image Repository

atulsingh0:
# Description
Enabling the Nomad Client to use Private Image repo, rather than public docker repository

# Reasons
Customers want to use their own image repository to pull the image i.e. - `circleci/picard`

### DIFF
--- a/jekyll/_cci2/server-3-install-build-services.adoc
+++ b/jekyll/_cci2/server-3-install-build-services.adoc
@@ -381,12 +381,12 @@ Once you have configured the environmental variables and contexts, rerun the rea
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
 ==== Enabling Nomad Client for External Image Repository
-To enable a Nomad client to download images from an external or private repo, you can implement one of the two solutions suggested below:
+To enable the Nomad client so it can download images from an external or private repo, you can implement one of the two solutions suggested below:
 
-* Change the image registry from public to private/internal network so the Nomad client can download the images without any authentication.
-* Embed the private image repository credentials in a Nomad client image at `/root/.docker/config.json`, assuming the credentials are static.
+* Keep the image registry as `Public` within `Internal network` so the Nomad client can download the images without any authentication.
+* Or, Embed the private image repository credentials in the Nomad client image at `/root/.docker/config.json`, assuming the credentials are static.
 
-NOTE: If you are using AWS, GCP, or any other cloud image registry, most of them are using temporary image repository credentials (for example, AWS ECR,GCP Container Registry), valid to use for a limited time. As soon as those credentials expire, your Nomad client will be unable to download the image from a private image repository if the image does not exist in a Nomad client. You must refresh the credentials again on those Nomad clients to pull the new images.
+NOTE: If you are using Amazon Web Services (AWS), Google Cloud  Platform (GCP), or any other cloud image registry, most of these registries use temporary image repository credentials, valid to use for a limited time. As soon as those credentials expire, your Nomad client will be unable to download the image from the private image repository if the image does not already exist in a Nomad client. You must refresh the credentials again on those Nomad clients to pull the images.
 
 
 === VM service
--- a/jekyll/_cci2/server-3-install-build-services.adoc
+++ b/jekyll/_cci2/server-3-install-build-services.adoc
@@ -380,14 +380,9 @@ Once you have configured the environmental variables and contexts, rerun the rea
 
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
-==== Enabling Nomad Client for External Image Repository
-To enable the Nomad client so it can download images from an external or private repo, you can implement one of the two solutions suggested below:
+==== Hosting Build Agent Image in External Image Repository
 
-* Keep the image registry as `Public` within `Internal network` so the Nomad client can download the images without any authentication.
-* Or, Embed the private image repository credentials in the Nomad client image at `/root/.docker/config.json`, assuming the credentials are static.
-
-NOTE: If you are using Amazon Web Services (AWS), Google Cloud  Platform (GCP), or any other cloud image registry, most of these registries use temporary image repository credentials, valid to use for a limited time. As soon as those credentials expire, your Nomad client will be unable to download the image from the private image repository if the image does not already exist in a Nomad client. You must refresh the credentials again on those Nomad clients to pull the images.
-
+If you are hosting build agent image (`circleci/picard`), used by Nomad client, in an external or private repo, you have to keep the private image registry as Public and unauthenticated.
 
 === VM service
 
--- a/jekyll/_cci2/server-3-install-build-services.adoc
+++ b/jekyll/_cci2/server-3-install-build-services.adoc
@@ -380,6 +380,15 @@ Once you have configured the environmental variables and contexts, rerun the rea
 
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/server-3-install-build-services.adoc
```diff
@@ -380,6 +380,15 @@ Once you have configured the environmental variables and contexts, rerun the rea
 
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
+==== Enabling Nomad Client for External Image Repository
+To enable a Nomad client to download images from an external or private repo, you can implement one of the two solutions suggested below:
+
+* Change the image registry from public to private/internal network so the Nomad client can download the images without any authentication.
+* Embed the private image repository credentials in a Nomad client image at `/root/.docker/config.json`, assuming the credentials are static.
+
+NOTE: If you are using AWS, GCP, or any other cloud image registry, most of them are using temporary image repository credentials (for example, AWS ECR,GCP Container Registry), valid to use for a limited time. As soon as those credentials expire, your Nomad client will be unable to download the image from a private image repository if the image does not exist in a Nomad client. You must refresh the credentials again on those Nomad clients to pull the new images.
```
thomasr8, line 389:
```suggestion
NOTE: If you are using Amazon Web Services (AWS), Google Cloud  Platform (GCP), or any other cloud image registry, most of these registries use temporary image repository credentials (for example, AWS ECR,GCP Container Registry), valid to use for a limited time. As soon as those credentials expire, your Nomad client will be unable to download the image from a private image repository if the image does not exist in a Nomad client. You must refresh the credentials again on those Nomad clients to pull the new images.
```

thomasr8 (approved):
Just one minor comment

jekyll/_cci2/server-3-install-build-services.adoc
```diff
@@ -380,6 +380,15 @@ Once you have configured the environmental variables and contexts, rerun the rea
 
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
+==== Enabling Nomad Client for External Image Repository
+To enable the Nomad client so it can download images from an external or private repo, you can implement one of the two solutions suggested below:
+
+* Keep the image registry as `Public` within `Internal network` so the Nomad client can download the images without any authentication.
+* Or, Embed the private image repository credentials in the Nomad client image at `/root/.docker/config.json`, assuming the credentials are static.
+
+NOTE: If you are using Amazon Web Services (AWS), Google Cloud  Platform (GCP), or any other cloud image registry, most of these registries use temporary image repository credentials, valid to use for a limited time. As soon as those credentials expire, your Nomad client will be unable to download the image from the private image repository if the image does not already exist in a Nomad client. You must refresh the credentials again on those Nomad clients to pull the images.
```
lloydChris, line 389: I don't think we want to include such a general statement about all cloud image registries.  I think we can remove this note altogether.

jekyll/_cci2/server-3-install-build-services.adoc
```diff
@@ -380,6 +380,15 @@ Once you have configured the environmental variables and contexts, rerun the rea
 
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
+==== Enabling Nomad Client for External Image Repository
+To enable the Nomad client so it can download images from an external or private repo, you can implement one of the two solutions suggested below:
+
+* Keep the image registry as `Public` within `Internal network` so the Nomad client can download the images without any authentication.
+* Or, Embed the private image repository credentials in the Nomad client image at `/root/.docker/config.json`, assuming the credentials are static.
```
lloydChris, line 387: I think this point deserves more details:
- Are we telling the customer to publish their own AMI to use in the nomad cluster?
- what is this file (`/root/.docker/config.json`) supposed to look like, can we publish an example or a link to dockers documentation
- what if the credentials are not static?

jekyll/_cci2/server-3-install-build-services.adoc
```diff
@@ -380,6 +380,15 @@ Once you have configured the environmental variables and contexts, rerun the rea
 
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
+==== Enabling Nomad Client for External Image Repository
+To enable the Nomad client so it can download images from an external or private repo, you can implement one of the two solutions suggested below:
+
+* Keep the image registry as `Public` within `Internal network` so the Nomad client can download the images without any authentication.
```
lloydChris, line 386: We don't reference an `Internal network` anywhere else, is there another name we can use that is more explicit.  Is it the VPC, whatever network the cluster is in?  Or maybe we don't need to mention it at all, that the registry is public and unauthenticated is detail enough

lloydChris, line 386: nit:
```suggestion
* Keep the image registry as Public within `Internal network` so the Nomad client can download the images without any authentication.
```

jekyll/_cci2/server-3-install-build-services.adoc
```diff
@@ -380,6 +380,15 @@ Once you have configured the environmental variables and contexts, rerun the rea
 
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
+==== Enabling Nomad Client for External Image Repository
+To enable the Nomad client so it can download images from an external or private repo, you can implement one of the two solutions suggested below:
```
lloydChris, line 384: If this section gets moved to Configure and Deploy, it can probably refer to the `circleci/picard` image directly, to avoid confusion with all the other circleci images or the customers own images

lloydChris (changes requested):
The [configure and deploy](https://github.com/circleci/circleci-docs/blob/dee565b312d1c7edb065c0b4ab6eda50817504f8/jekyll/_cci2/server-3-install-build-services.adoc#configure-and-deploy) section may need to reference the new field also.  Regardless, this section should precede the Nomad Clients Validation section.

jekyll/_cci2/server-3-install-build-services.adoc
atulsingh0: Ripped off this sections, as this is required customer to alter the nomad client image.

jekyll/_cci2/server-3-install-build-services.adoc
atulsingh0: done

jekyll/_cci2/server-3-install-build-services.adoc
```diff
@@ -380,6 +380,10 @@ Once you have configured the environmental variables and contexts, rerun the rea
 
 image::realitycheck-pipeline.png[Screenshot showing the realitycheck project building in the CircleCI app]
 
+==== Hosting Build Agent Image in External Image Repository
+
+If you are hosting build agent image (`circleci/picard`), used by Nomad client, in an external or private repo, you have to keep the private image registry as Public and unauthenticated.
```
lloydChris, line 385: This could probably all go in the Configure and Deploy section.  and it could probably just say something like "consult customer support to set up"

lloydChris (changes requested):
The setting needs to be described in the Configure and Deploy section

Given that this feature is only used in very unique situations, It might be good to answer the question "why should i care/when do i use this setting?" for the customer here.  Probably suggest consulting Customer support to set up, which is a pattern we have used in the past

jekyll/_cci2/server-3-install-build-services.adoc
atulsingh0: moving to configure and deploy.

ryanwohara (approved):

lloydChris (approved):

PR #6427, opened 2022-02-26
Add `guestExperiment`option for OptimizelyClient

rsylvian:
# Description
- Add a `guestExperiment` option for `getVariationName()` function
- Add more test to validate that `getUserId()` works well with logged in user & guest users

# Reasons
We want to start to run experiments on guest users. We can call `analytics.user().anonymousId()` to get a consistent anonymous `userId` that can be used to resolve the experiment variation. More info: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/identity/#anonymous-ids

# Screesnhots
Trying the implementation on @tshen815 experiment: https://github.com/circleci/circleci-docs/pull/6396

Screen Shot 2022-02-28 at 5 19 33 PM
Screen Shot 2022-02-28 at 5 19 36 PM

"resetting" the device (by clearing local storage and cookie), we also get `control` since you get assigned a new `anonymousId`:
Screen Shot 2022-02-28 at 5 25 03 PM

### DIFF
--- a/jest/global.js
+++ b/jest/global.js
@@ -13,7 +13,7 @@ global.analytics = {
   user: jest.fn(() => userFn),
 };
 
-global.global._DATADOG_SYNTHETICS_BROWSER = false;
+global._DATADOG_SYNTHETICS_BROWSER = false;
 
 class IntersectionObserver {
   constructor() {}
--- a/jest/global.js
+++ b/jest/global.js
@@ -2,13 +2,18 @@ import $ from 'jquery';
 
 global.$ = global.jQuery = $;
 
+const userFn = {
+  anonymousId: jest.fn(),
+};
+
 global.analytics = {
   page: jest.fn(),
   track: jest.fn(),
   identify: jest.fn(),
+  user: jest.fn(() => userFn),
 };
 
-global._DATADOG_SYNTHETICS_BROWSER = false;
+global.global._DATADOG_SYNTHETICS_BROWSER = false;
 
 class IntersectionObserver {
   constructor() {}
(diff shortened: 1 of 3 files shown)


tshen815 (approved):
Tested earlier on my PR #6396 working well!

jest/global.js
```diff
@@ -2,13 +2,18 @@ import $ from 'jquery';
...
+
 global.analytics = {
   page: jest.fn(),
   track: jest.fn(),
   identify: jest.fn(),
+  user: jest.fn(() => userFn),
 };
 
-global._DATADOG_SYNTHETICS_BROWSER = false;
+global.global._DATADOG_SYNTHETICS_BROWSER = false;
```
teesloane, line 16: why do we have two: `global.global` ?

src/js/services/optimizely.js
```diff
@@ -60,16 +66,26 @@ class OptimizelyClient {
       // defines additional attributes we will want to send to optimizely to qualify/disqualify a user
       const attributes = options.attributes ?? {};
 
+      // capture if we are trying to run this experiment as a guest experiment
+      // defautl to false as most of our experiments are for logged in users
```
teesloane, line 70:
```suggestion
      // default to false as most of our experiments are for logged in users
```

jest/global.js
rsylvian: was a typo! Good catch 😊  fixed in https://github.com/circleci/circleci-docs/pull/6427/commits/476e19cf6d1f4b79bc422dfd9cef48c04ec00230

src/js/services/optimizely.js
rsylvian: Thanks!

PR #6439, opened 2022-02-28
Call out contexts & project specific env vars are not possible in sec…

jabsci:
…ondary containers

# Description
Adding in the following line to example 2 of Setting an environment variable in a container:
```
**Note**: While hard coded environment variable values will be passed on correctly to the secondary or service container, contexts or project specific environment variables won't be interpolated for non-primary containers.
```

# Reasons:
- Adding a note to highlight the inability to pass in project or context specific env vars to secondary/service containers as it's currently not possible.

### DIFF
--- a/jekyll/_cci2/env-vars.md
+++ b/jekyll/_cci2/env-vars.md
@@ -422,6 +422,8 @@ jobs:
 
 The following example shows separate environment variable settings for the primary container image (listed first) and the secondary or service container image.
 
+**Note**: While hard coded environment variable values will be passed on correctly to the secondary or service container, contexts or project specific environment variables won't be interpolated for non-primary containers.
+
 ```yaml
 version: 2.1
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/env-vars.md
```diff
@@ -422,6 +422,8 @@ jobs:
 
 The following example shows separate environment variable settings for the primary container image (listed first) and the secondary or service container image.
 
+**Note**: While hard coded environment variable values will be passed on correctly to the secondary or service container, contexts or project specific environment variables won't be interpolated for non-primary containers.
```
kbatuigas, line 425:
```suggestion
**Note**: While hard-coded environment variable values will be passed on correctly to the secondary or service container, contexts or project specific environment variables will not be interpolated for non-primary containers.
```
We don't currently have this in our own style guide, but I'll follow Microsoft's style guide here and use a hyphen.
We do avoid contractions when we can! https://circleci.com/docs/style/style-voice/#word-choices

kbatuigas (changes requested):
Thanks @jabsci ! Just a couple of tiny suggestions.

kbatuigas (approved):

PR #6442, opened 2022-02-28
Runner UI preview

devinmarieb:
# Description
- [Preview Link Web App Installation Guide](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DOCSTEAM-157_db_runnerUI-preview/2.0/runner-installation/)
- [Preview Link CLI Installation Guide](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DOCSTEAM-157_db_runnerUI-preview/2.0/runner-installation-cli/)
- [Preview Link Runner Overview](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DOCSTEAM-157_db_runnerUI-preview/2.0/runner-overview/)
- [Preview Link Runner Concepts](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DOCSTEAM-157_db_runnerUI-preview/2.0/runner-concepts/)
- [Preview Link macOS](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DOCSTEAM-157_db_runnerUI-preview/2.0/runner-installation-mac/)
- [Preview Link Linux](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DOCSTEAM-157_db_runnerUI-preview/2.0/runner-installation-linux/)
- [Preview Link Docker](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DOCSTEAM-157_db_runnerUI-preview/2.0/runner-installation-docker/)

# Reasons
UI is updating to allow for a quick creation of namespaces and resource classes which makes the web app preferable to the CLI. The docs are adjusting to this workflow by providing both options as well as breaking out some runner pages into more easily digestible smaller pages.


sebastian-lerner:
putting first part of feedback here.  first set goes top to bottom in the page.  i'm stopping halfway thru to go to some meetings and will pick this back up in a new comment.

first set of feedback:
- "authenticating and installing self-hosted runners..." --> can we just say "Installing"?  It might also be worth being specific up here that an org admin needs to accept the terms of service in the org settings
- "(installed as part of coreutils on Linux apt/yum, and on macOS via brew)" --> do you need a comma before the ", and"?  Also is it worth linking to https://www.gnu.org/software/coreutils/?
- what is the difference between the first set of utilities and the second set of utiliites? why isn't it all just one set?
- does the text next to the big information "i" apply to Server customers today? It would be good to confirm with Steve whether or not a server customer today can use the Web UI to actually go thru the install.  if yes, then it's good to leave as is. if no, we should move that no?
- "In order to install self-hosted runners, you will need to create a namespace and authentication token. Please note that to create resource classes and tokens, you need to be an organization administrator in the VCS provider." if it's my first time using CircleCI and Runner, "namespace" and "authentication token" are terms that mean nothing to me.  Can we put this in the section where we actually talk about namespaces and authentication tokens and just say "You need to be an admin in your vcs to perform certain steps below"? and call out in the steps which one require being an admin.  or does it need to be the very first thing we say in the instructions? 
- more to come.....

other things that came to mind:
- Right now, I don't think you can actually add a second Runner via the UI if you have an existing resource class with a Runner, it might be worth calling that out as a known limitation?  especially if we'll have existing Runner customers reading these docs since that will be the case for a lot of them
- Step 5, is it worth including such a large snippet of code? and if yes, should we be more precise about what exactly should be copy/pasted? I'm imagining someone going through using Runner for the first time, they probably only need copy the resource_class line if they have an existing config with existing jobs.

sebastian-lerner:
second part of feedback, picking up from where the first set leaves off:

- "Create a namespace and resource class for your organization’s self-hosted runner resources." --> can we get rid of the second use of the word "resources" and just say "runners"?
- same feedback as before wrt orbs.  if it's my first time using CCI, it is a confusing sentence
- "The default token" is there a reason we call it the "default" token instead of the "resource class token". when i can see the UI again i'll check but does it explicitly say "default token"?
- "After authenticating" --> is authentication a part of the steps in step #3?
- for #4, is it even worth including the links to the other pages? it just feels like it adds more clutter and isn't neccessary?
- "You will be able to see a list of your self-hosted runners on the CircleCI web app, on the inventory page." --> should this be a step #6 to just close the loop and say click on the link that takes you to your inventory which lets you see the runners you just installed? instead of it being a standalone thing
- same feedback as before as having a better definition for resource class in the Overview page

will there ever be a way to see the photos or do we have to wait until it's live? i can't remember what you mentioned

devinmarieb:
A few things we still need answers about from engineering:
- Deleting runners
- Mac root/permission code
- Server compatibility

sebastian-lerner:
can we adjust the language of the item in the to specify that someone designated as an administrator in their VCS need to agree to the Runner Terms in the CircleCI organization settings page?

sebastian-lerner:
"Currently, you can only create one runner per resource class through the UI. If you would like to add more runners to your resource class, you can use the CLI."

might be worth putting this at the end of the instructions instead of having be the before the instructions?

devinmarieb:
@sebastian-lerner for the first comment, I worry if we put it at the end, someone could start the process and not realize it's not available to do. If we put it at the end, there is a bold assumption they will read the whole page before they start the process (or at all). I don't mind moving it, but I'm not sure putting it at the end is really all that helpful given that most people don't fully read docs to begin with.

sebastian-lerner:
@devinmarieb fair! let's leave it

devinmarieb:
@sebastian-lerner I know there was talk somewhere in Slack about having dynamic links, so we could include links to the inventory page and such, but I imagine that's not something that's included/in scope for the initial release.

sebastian-lerner:
@devinmarieb wdym about dynamic links?

sebastian-lerner:
@devinmarieb the definition for resource class is awesome, thank you for putting that there.  i think we may want to add in one sentence elaborating why you would even want a pool of runners in the first place.  

right now, we define resource class as a label.  to make it even more clear, we may want to add in an example that says something like: "if you want to install 3 runners on 3 mac minis and distinguish those runners from the runners you have running on 3 aws ec2 instances, you may choose to have a mac-mini resource class that you associate the 3 runners to and only send jobs that run on macOS to that resource class and a separate resource class for your 3 aws instances called "aws-ec2-linux" where you send your linux jobs to.  this lets you pool resources and associate at the job level in config.yml which set of runner resources to send a job to".  something like that? i think that would make it crystal clear.  leaving the definition as just a label still doesn't feel like enough for a new customer to understand the concept.

also minor thing but can we adjust it to say "a label you choose to represent a pool of self-hosted runners" instead of "your pool".

sebastian-lerner:
"Resource class labels help you identify a pool of self-hosted runners, which allow you to set up your configuration to send jobs to specific places. For example, if you have multiple machines running macOS, and multiple machines running AWS EC2, you could create resource class labels for each of these, macOS and aws-ec2-linux, respectively. At the job level in your .circleci/config.yml, you can associate which set of self-hosted runner resources to send a job to based on the resource class." 

-->

"Runner "resource classes" help you create pools of self-hosted runners.  Pooling runners with resource classes enables you to isolate a group of self-hosted runners for a specific subset of jobs. For example, if you have one set of machines running macOS and one set of machines running Linux, you could create resource class labels for each: 'macOS' and 'linux', respectively. At the job level in your .circleci/config.yml, you can [associate](https://circleci.com/docs/2.0/runner-overview/#referencing-your-runner-on-a-job) jobs that are intended to run on macOS to the 'macOS' resource class and the jobs intended for Linux machines to the 'linux' resource class."

@devinmarieb what do you think about this new wording?

devinmarieb:
@sebastian-lerner Also just so you're aware, our style guide is against using regular words/phrases/partial sentences as links. If we are linking out to something, we need to explicitly say what it is. In this case, we won't link the word "associate" rather indicate we are suggesting a different section of that page to look at for reference.

sebastian-lerner:
https://circleci.slack.com/archives/CP4LXVD1T/p1648518006673039

We can update images now

sebastian-lerner:
where we say "navigate to **Runners**", we need to change it to "Self-hosted Runners"

sebastian-lerner:
under the section "CircleCI Web app installation", in step 1, it says "Self-Hostd Runners" instead of "Self-Hosted"

sebastian-lerner:
in the definition of resource classes: "to send jobs to specific places" sounds odd.  suggestion below which would replace everything after the first sentence in that paragraph.

Pooling runners with resource classes enables you to isolate a group of self-hosted runners for a specific subset of jobs. For example, if you have one set of machines running macOS and one set of machines running Linux, you could create resource class labels for each: 'macOS' and 'linux', respectively. At the job level in your .circleci/config.yml file, you can associate jobs that are intended to run on macOS with the 'orgname/macOS' resource class and the jobs that are intended for Linux machines with the 'orgname/linux' resource class.

jekyll/_cci2/runner-on-kubernetes.adoc
```diff
@@ -15,21 +15,21 @@ toc::[]
...
+* Go through the steps to <<runner-installation#authentication,create and authenticate a CircleCI self-hosted runner resource class>>.
 
 === CircleCI server installation
 
-When installing the Helm chart for use with a CircleCI server installation the `agentVersion` will need to be set to the pinned version specified in the  xref:runner-installation.adoc#runner-for-server-compatibility[Runner Installation] instructions. The `LAUNCH_AGENT_API_URL` will also need to be set as an environment variable either via the `--set` flag or in the `env` section of the `values.yaml` file, specifying the hostname or address of the server installation.
+When installing the Helm chart for use with a CircleCI server installation the `agentVersion` will need to be set to the pinned version specified in the  xref:runner-installation.adoc#runner-for-server-compatibility[Self-hosted Runner Installation] instructions. The `LAUNCH_AGENT_API_URL` will also need to be set as an environment variable either via the `--set` flag or in the `env` section of the `values.yaml` file, specifying the hostname or address of the server installation.
```
thomasr8, line 30:
```suggestion
When installing the Helm chart for use with a CircleCI server installation, the `agentVersion` will need to be set to the pinned version specified in the  xref:runner-installation.adoc#runner-for-server-compatibility[Self-hosted Runner Installation] instructions. The `LAUNCH_AGENT_API_URL` will also need to be set as an environment variable either via the `--set` flag or in the `env` section of the `values.yaml` file, specifying the hostname or address of the server installation.
```

thomasr8 (approved):
Hi @devinmarieb This is awesome. This is so well-written! I made a few suggestions/recommendations, so please feel free to commit whatever suggestions you would like.

sebastian-lerner:
devin, in our linux install instructions the "Ubuntu/Debian" section after the existing commands, can we add: 

"Consider adding the following command if you have an existing Cloud config file and want to make sure that existing config commands that work on Cloud work on your self-hosted Runner without errors.  Note that this enables code to execute root commands on your machine and changes that the job makes to the system may persist after the job as well.

```echo "circleci ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers```

@ryanpedersen42 what do you think about that language?

ryanpedersen42:
@sebastian-lerner place seems fine to me if just for Ubuntu. Would we also want to do it for the CentOS/RHEL as well though? 

In terms of messaging, trying to figure out the best way. I think it might make sense to focus on the orbs as the main blocker. Something like this:
"Consider adding the following command if you would like to use certified orbs that work on Cloud work on your self-hosted Runner without errors. Note that this enables code to execute root commands on your machine and changes that the job makes to the system may persist after the job as well."

But also don't know if that's calling too in the weeds?

sebastian-lerner:
no i like that a lot.  i think the focus on orbs is good and i'd rather err on the side of being explicit about the risk.  @devinmarieb sound good to you?

i think we add to RHEL as well, devin can we get that in there additionally?

devinmarieb:
@sebastian-lerner @ryanpedersen42 The wording is still a little awkward so I'll tweak it a bit, but this still doesn't really answer the question of where this command goes/where the user needs to put it.

sebastian-lerner:
I think we're saying just after the mkdir and chown command snippet in the Ubuntu/Debian and CentOS/RHEL headers.  as a new snippet

devinmarieb:
@sebastian-lerner Before you were mentioning the config file though. These aren't config commands, these are bash commands. So I take it this has nothing to do with the config, and the new command is also a bash command?

sebastian-lerner:
yup these arent config commands.  thsese commands are needed so that when a customer wants to use certain config commands with a runner job, they work fine and without error

PR #6456, opened 2022-03-02
Remove pre-installed software section

devinmarieb:
# Description
Removing the software section. Will adjust and link out to more specific resources when those links are available or when the information in included in DevHub.

Apologies for a very unclear commit message - I came back to this and had forgotten what I had done on that particular change 🙃 

# Reasons
This section kept getting out of date too quickly and there was a lack of communication around when changes happened.


catherinepope (approved):
👍  to the change and 😆  to commit message.

PR #6460, opened 2022-03-02
[DD-446] Implement option to disable read time

chantelle11:
Ticket [Link](https://circleci.atlassian.net/browse/DD-446)

# Description
In [DD-425: Move the "top reasons why my build failed..." from support to docDEV IN PROGRESS](https://circleci.atlassian.net/browse/DD-425) we want to disable the read-time because it might not completely make sense for this page (no one read the entire page). We should add an option as a front matter variable to let this be customizable.

# Changes
- A variable called readTime: false can be set in a markdown page to disable the read time for the output page
- by defaults readTime needs to be true for all pages (in _config).

# How to test
You will have to test it locally by adding in new `readTime` value into md file
Screen Shot 2022-03-02 at 3 37 06 PM

# Screenshots
## Before
Screen Shot 2022-03-02 at 3 42 30 PM

## After
Screen Shot 2022-03-02 at 3 52 47 PM

### DIFF
--- a/jekyll/_includes/quick-info-list.html
+++ b/jekyll/_includes/quick-info-list.html
@@ -5,6 +5,7 @@
   <div id='arrow' data-popper-arrow></div>
 </div>
 <div class="quick-info-list">
+  {% if page.readTime != false %}
   <img alt="Last updated" src="{{ site.baseurl }}/assets/img/docs/duration.svg">
   <time aria-describedby="tooltip-time" id="time-posted-on" datetime="{{  page.last_modified_at | date_to_xmlschema }}">
       {% assign dateStart = page.last_modified_at | date: '%s' %}
@@ -28,6 +29,7 @@
       • {{ ert }} min read
     </span>
   </time>
+  {% endif %}
   
   {% if page.version %}
   <div class="server-version-wrapper">
--- a/jekyll/_includes/quick-info-list.html
+++ b/jekyll/_includes/quick-info-list.html
@@ -5,7 +5,6 @@
   <div id='arrow' data-popper-arrow></div>
 </div>
 <div class="quick-info-list">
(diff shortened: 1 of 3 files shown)


jekyll/_includes/quick-info-list.html
```diff
@@ -5,6 +5,7 @@
   <div id='arrow' data-popper-arrow></div>
 </div>
 <div class="quick-info-list">
+  {% if page.readTime != false %}
   <img alt="Last updated" src="{{ site.baseurl }}/assets/img/docs/duration.svg">
   <time aria-describedby="tooltip-time" id="time-posted-on" datetime="{{  page.last_modified_at | date_to_xmlschema }}">
```
rsylvian, line 10: The readtime option should not hide the Last updated date

rsylvian (changes requested):

jekyll/_includes/quick-info-list.html
```diff
@@ -5,6 +5,7 @@
   <div id='arrow' data-popper-arrow></div>
 </div>
 <div class="quick-info-list">
+  {% if page.readTime != false %}
```
teesloane, line 8: Super nit, but I feel that `!= false` is a bit tricky to read. Is it possible to just use `{% if page.readTime %}` in liquid? And if not ... just use `{% if page.readTime == true %}`?

rsylvian: Yup great point!

chantelle11: I was following how we do page.toc I can update no worries! I agree makes sense

rsylvian (approved):
Looks good ✅

tshen815 (approved):
nice work!

PR #6463, opened 2022-03-02
`OptimizelyClient.getVariationName()` should return null when analytics object is not available

rsylvian:
# Description
- Check for `analytics` & `analytics.user()` existance before requesting the `anonymousId`

# Reasons

We are seeing multiple errors since we launched the new guest user experiment resolver. They are most likely happening because Ad blockers are preventing segment to load.
image

### DIFF
--- a/src/js/services/optimizely.test.js
+++ b/src/js/services/optimizely.test.js
@@ -76,7 +76,7 @@ describe('Optimizely Service logged-in users', () => {
       await expect(client.getUserId(true)).resolves.toBe(null);
     });
 
-    it("returns null when analytics doesn't exists", async () => {
+    it("returns null when analytics doesn't exist", async () => {
       glob.analytics = null;
       await expect(client.getUserId(true)).resolves.toBe(null);
     });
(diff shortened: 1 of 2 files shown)


src/js/services/optimizely.test.js
```diff
@@ -75,6 +75,11 @@ describe('Optimizely Service logged-in users', () => {
       glob.userData = {};
       await expect(client.getUserId(true)).resolves.toBe(null);
     });
+
+    it("returns null when analytics doesn't exists", async () => {
```
rsylvian, line 79:
```suggestion
    it("returns null when analytics doesn't exist", async () => {
```

tshen815 (approved):
Thanks for addressing this! 👍

src/js/services/optimizely.test.js
```diff
@@ -75,6 +75,11 @@ describe('Optimizely Service logged-in users', () => {
       glob.userData = {};
       await expect(client.getUserId(true)).resolves.toBe(null);
     });
+
+    it("returns null when analytics doesn't exist", async () => {
```
chantelle11, line 79: great idea to add test for it

PR #6469, opened 2022-03-03
[FIX]DD-446: Disable Read Time Fix

tshen815:
# Changes

- revert styles
- declare `!= false` for pages to show readtime

# Description
Fixing a bug that occurred in #6460 for disabling read time

# Considerations
If we want to disable the read time for a page, we have to add the `readtime: false` variable and value to the front matter
Screen Shot 2022-03-03 at 3 43 27 PM

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -548,8 +548,6 @@ Fixing styling for the code example tables in the "Migrating From..." pages
   font-weight: normal;
   font-size: 15px;
   color: #343434;
-  // border-right: 1px solid #DDDDDD;
-  // margin-right: 16px;
 }
 
 // badge tags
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -548,8 +548,8 @@ Fixing styling for the code example tables in the "Migrating From..." pages
   font-weight: normal;
   font-size: 15px;
   color: #343434;
-  border-right: 1px solid #DDDDDD;
-  margin-right: 16px;
+  // border-right: 1px solid #DDDDDD;
+  // margin-right: 16px;
```
chantelle11: I would remove over commenting out
```suggestion
```

chantelle11 (approved):
Nice

PR #6478, opened 2022-03-07
remove pre-installed software info

devinmarieb:
# Description
Removing pre-installed software information.

# Reasons
This information will live on DevHub in the future.


thomasr8 (approved):

devinmarieb:
Converting this to a draft because this was the wrong Docker page -- but I imagine this will need to be removed as well. Will double check.

PR #6482, opened 2022-03-08
Update security.md

HennaAbbas:
We now only support retrieving audit logs for up to 12 months.

# Description
Updated docs to communicate policy that we only support retrieval of audit logs for up to 1 year. 

# Reasons
Previously we supported retrieval of audit logs indefinitely. Due to a change in how/where we store this data, audit logs are now only supported for up to 1 year.

### DIFF
--- a/jekyll/_cci2/security.md
+++ b/jekyll/_cci2/security.md
@@ -73,6 +73,8 @@ Complete Audit logs may be downloaded from the Audit Log page within the Admin s
 
 **Note:** In some situations, the internal machinery may generate duplicate events in the audit logs. The `id` field of the downloaded logs is unique per event and can be used to identify duplicate entries.
 
+Per [our data retention policy](https://circleci.com/privacy/#information) audit logs can be retrieved for up to 12 months. 
+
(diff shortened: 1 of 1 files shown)


rosieyohannan:
Thanks so much @HennaAbbas, I'll just push up a change to also add this note to the server-specific security page if that's ok?

devinmarieb:
@rosieyohannan I made a small change just to get it to build (added a comma).

HennaAbbas:
> Thanks so much @HennaAbbas, I'll just push up a change to also add this note to the server-specific security page if that's ok?

@rosieyohannan server customers will not be affected by this policy. Let me know if you need me to specify this in our docs.

rosieyohannan:
@HennaAbbas yes if you wouldn't mind just adding "This does not affection server installations." or something to avoid any confusion that would be great :-D

devinmarieb:
@HennaAbbas bumping this to make sure you saw Rosie's comment.

rosieyohannan (approved):

PR #6493, opened 2022-03-09
`OptimizelyClient.getVariationName()` should return null when analytics object is not available

rsylvian:
# Description
- Add a `try...finally` around the  `analytics.user().anonymousId()` expression

# Reasons
3rd time's a charm! This is a follow-up of https://github.com/circleci/circleci-docs/pull/6463 & https://github.com/circleci/circleci-docs/pull/6470


tshen815 (approved):
nice! Good call with the `anonymousId` variable, thanks for working on this!

PR #6500, opened 2022-03-10
make clear that regardless of parallelism level, deploy step must be …

sebastian-lerner:
…removed

# Description

# Reasons

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -1199,7 +1199,7 @@ There is no direct replacement for the `deploy` step if you are using parallelis
 
 ###### *Example*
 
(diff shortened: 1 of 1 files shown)


devinmarieb:
@sebastian-lerner would it make more sense that this clarification is added in the main paragraph, not in one example? Otherwise someone might miss this information if that particular example does not apply to them (since there's two examples to look at).

sebastian-lerner:
yeah that sounds good, we can move it up there

devinmarieb (approved):

sebastian-lerner:
@devinmarieb "A config file that uses the deprecated deploy step and has parallelism > 1 will need to be converted to remove the deploy step. See example below (this code is deprecated, do not copy):"

Can we reword this so it sounds like this is just a generic example of a config with parallelism?  right now it might be read as only jobs with parallelism have to be fixed and that's why we're showing you that example (even though we added the thing at the top).  

sorry, that was my original intent with the original PR

PR #6502, opened 2022-03-10
[DD-465]  CircleCI-Docs: Add `Experiment Entry` event on docs onload

chantelle11:
# Description
We need to add a trackAction on the docs onload to understand incoming activity

# Reason
Request from engagement team

# Testing
Looker [link](https://circleci.looker.com/dashboards/662?Page+Name=job)

### DIFF
--- a/src/js/services/optimizely.js
+++ b/src/js/services/optimizely.js
@@ -43,6 +43,15 @@ class OptimizelyClient {
       }
     });
   }
+
+  // Experiment entered tracking
+  getTrackExperimentEntry() {
+    let orgId = Cookies.get(COOKIE_KEY) ?? null;
+    this.getUserId().then((userId) => {
+      trackExperimentEntry(userId, orgId, 'docs');
+    });
+  }
+
   // getVariationName is always guaranteed to resolve with either "null" or a variation name.
   // This is consistent to what getVariation from the optimizely-sdk does.
   //
@@ -264,6 +273,19 @@ export const storeExperimentParticipation = (
     // Uglify /w browserlist force us to do catch (_)
     // We're deliberately ignoring it so that it doesn't break the app. It'll mean a few extra
     // events are emitted, but I think that's the lesser issue.
+  }
+};
+
+export const trackExperimentEntry = (userId, orgId, projectId) => {
+  if (userId && orgId) {
+    const properties = {
+      id: uuidv4(),
+      timestamp: Date.now().toISOString(),
+      userId,
+      orgId,
+      projectId,
+    };
+    window.AnalyticsClient.trackAction('Experiment Entry', properties);
   }
 };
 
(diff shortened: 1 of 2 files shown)


src/js/services/optimizely.js
```diff
@@ -267,4 +276,17 @@ export const storeExperimentParticipation = (
   }
 };
 
+export const trackExperimentEntry = (userId, orgId, projectId) => {
```
tshen815, line 279: Is there any reason why we are exporting this function? We dont seem to be importing it anywhere and you are initializing `getTrackExperimentEntry` in `app.js` which has `trackExperimentEntry` in its function body.

tshen815:
Do we need to set up an amplitude chart for this event as well?

rsylvian (changes requested):
I don't think this work should be part of the `OptimizelyClient` since it is not used to resolve experiments. In the end, it is just some tracking about the number of users that _could_ be part of future experiments.

src/js/services/optimizely.js
chantelle11: code updated thanks!

chantelle11: I exported it to be reusable but code has been updated. You are correct it wasn't needed.

chantelle11:
> Looks good, just added a quick question for you. Also, could you add the ticket to the PR body? And do we need to set up an amplitude chart for this event as well?

Don't need amplitude chart see PR body for update and details

tshen815 (approved):

src/js/services/optimizely.js
```diff
@@ -43,6 +43,7 @@ class OptimizelyClient {
       }
     });
   }
+
```
rsylvian, line 46:
```suggestion
```

src/js/services/trackExperimentEntry.js
```diff
@@ -0,0 +1,25 @@
...
+    if (userId && orgId) {
+      const projectId = 'docs';
+
+      const properties = {
+        id: uuidv4(),
+        timestamp: new Date().toISOString(),
+        userId,
+        orgId,
+        projectId,
+      };
```
rsylvian, line 17:
```suggestion
      const properties = {
        id: uuidv4(),
        timestamp: new Date().toISOString(),
        userId,
        orgId,
        pageName: 'docs',
      };
```

rsylvian (changes requested):

src/js/services/optimizely.js
chantelle11: I actually prefer it better with the space. Usually there is a space between comments and the function above. Ill change it back since I am not changing code in this area just why I added it

src/js/services/trackExperimentEntry.js
chantelle11: Thanks!

PR #6447, opened 2022-03-01
[DD-423] Experiment: Add cta buttons to "first-steps".

teesloane:
[Ticket](https://circleci.atlassian.net/browse/DD-423)

This PR adds an experiment that renders CTA signup buttons to the docs/2.0/first-steps document. The styles of the buttons are based on our design and what is on outer.

- Amplitude tracking key: `dd_first-steps-signup-cta_test`
- Amplitude data payload: ` {button: <button.innerText>}`
- Experiment [link](https://app.optimizely.com/v2/projects/16812830475/experiments/21253750305/variations).

### Before
Screen Shot 2022-03-01 at 3 56 20 PM

### After

Screen Shot 2022-03-01 at 3 56 03 PM

Mobile:

image

PR #6508, opened 2022-03-10
Add: track event to control link.

teesloane:
https://github.com/circleci/circleci-docs/pull/6447 did not add track control events. This PR adds that.

### DIFF
--- a/src/js/experiments/signupLinksCta.js
+++ b/src/js/experiments/signupLinksCta.js
@@ -31,6 +31,15 @@ function handleClickedLink() {
   });
 }
 
+$('a[href="https://circleci.com/signup/"]').click((e) => {
+  window.AnalyticsClient.trackAction(
+    'dd_first-steps-signup-cta_test_control_link',
+    {
+      text: e.target.innerText,
+    },
+  );
+});
+
 // https://app.optimizely.com/v2/projects/16812830475/experiments/21253750305/variations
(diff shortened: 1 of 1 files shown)


tshen815:
Is this specifically to track an event when users are segmented into the control group? Would we need to gate the new trackAction you made in a conditional? 

something like:
`if (variation === 'control') {// new trackAction for control events here} `

tshen815 (commented):
Is this specifically to track an event when users are segmented into the control group? Would we need to gate the new trackAction you made in a conditional? 

something like:
`if (variation === 'control') {// new trackAction for control events here} `

teesloane:
> Is this specifically to track an event when users are segmented into the control group? Would we need to gate the new trackAction you made in a conditional?
> 
> something like: `if (variation === 'control') {// new trackAction for control events here} `

Oh, good catch, I will move that into a control block.

src/js/experiments/signupLinksCta.js
```diff
@@ -31,6 +31,7 @@ function handleClickedLink() {
   });
 }
 
+
```
rsylvian, line 34:
```suggestion
```

chantelle11 (approved):
Nice! 👍😁😊

tshen815 (approved):
looks good!

PR #6516, opened 2022-03-14
Fix broken resource class link

devinmarieb:
# Description
Changes link from resource_class to resourceclass

# Reasons
resource_class is not the appropriate URL and the link was just going to the top of the config ref


catherinepope (approved):
👍

PR #6532, opened 2022-03-15
Update banner in Persist Data

devinmarieb:
# Description
Update April 1 to May 1 and add clarification about the ability to test usage controls.

# Reasons
We are allowing customers to test out usage controls for 30 days before we start new charges.


thomasr8 (approved):
Looks good 👍

PR #6546, opened 2022-03-17
fix link tag in doc to fix link from app

rosieyohannan:
There is a link from the app project settings > ssh keys to the config reference. The link was broken. This PR fixes the link

Screenshot 2022-03-17 at 11 48 15
.


catherinepope (approved):
👍

PR #6554, opened 2022-03-18
Fix: date does not reset.

teesloane:
# Description
- The original code had the "lastSeen" date being reset every time the counter reset. This would result in the popup not  accurately representing the duration of "one day" between being shown.

### DIFF
--- a/src/js/experiments/popupCta.js
+++ b/src/js/experiments/popupCta.js
@@ -35,6 +35,21 @@ function getPopupData() {
 }
 
 /**
+ * @lastDay is string representation of a date, coming from localStorage
+ * returns `true` if the user is visiting the same day as stored in local storage
+ * */
+function isStillTheSameDay (lastDay) {
+  let stillTheSameDay = false;
+  const now = new Date();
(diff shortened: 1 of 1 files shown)


src/js/experiments/popupCta.js
```diff
@@ -43,26 +58,20 @@ function getPopupData() {
...
 function canShowPopup() {
-  let now = new Date();
-  let stillTheSameDay = false;
   const popupData = getPopupData();
-  if (popupData.lastSeen !== null) {
-    const popupLastSeen = new Date(popupData.lastSeen);
-    stillTheSameDay =
-      now.getFullYear() === popupLastSeen.getFullYear() &&
-      now.getMonth() === popupLastSeen.getMonth() &&
-      now.getDate() === popupLastSeen.getDate();
```
teesloane, line 54: This block ( was extracted into the function "isStillTheSameDay()" as it's now being called twice. I passed in the local storage value rather than having `isStillTheSameDay` be responsible for running `getPopupData` since the local storage data was available in the outer scope both places that it is invoked.

rsylvian (approved):

PR #6558, opened 2022-03-21
Fix docker image name in note

merwan:
# Description

The note references the legacy circle CI image for python and a version which does not match the code example above.

# Reasons

Let's make the doc consistent


catherinepope (approved):
👍  Thank you, Merouane!

PR #6559, opened 2022-03-21
add css to allow use of asciidoc sidebars

rosieyohannan:
* Some features of asciidoctor require additional CSS to work. 
* This PR is to add the CSS required to use sidebars: https://docs.asciidoctor.org/asciidoc/latest/blocks/sidebars/
* The admonitions look bad (as they do always!) but we can tackle that in another ticket :-D

They look like this - I just added some random content to show how they look - the whole "Interesting" box is a "sidebar":
Screenshot 2022-03-21 at 14 32 13

### DIFF
--- a/src/styles/_asciidoc.scss
+++ b/src/styles/_asciidoc.scss
@@ -53,7 +53,7 @@ span.icon>.fa {
 }
 
 .sidebarblock{
-  border-style:1px solid #dddddd;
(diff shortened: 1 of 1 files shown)


devinmarieb:
@rosieyohannan I would use margin since the goal is to have the spacing between separate elements.

rsylvian (approved):

PR #6564, opened 2022-03-21
Add: two-up include for new quick start guide

teesloane:
# Description
[Ticket](https://circleci.atlassian.net/browse/DD-451)

- add a new include for a "two-up" component where the left side is text and the right side is an image.
- this PR adds no content - just the structure for the include.
  - however, one can view an example of it implemented at [this preview link](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/DD-451-two-up-preview/2.0/workflows/)

## Screenshots &tc

Here is a screen-recording representing breakpoints. I think it should change sooner, but I just used the standard mobile breakpoint we use everywhere.

https://user-images.githubusercontent.com/12987958/159349694-fc5855ae-093c-4d7f-aa38-c002fc45cd2c.mp4

### DIFF
--- a/jekyll/_includes/two-up.html
+++ b/jekyll/_includes/two-up.html
@@ -1,0 +1,18 @@
+{% comment %}
+twoup is a jekyll include for rendering "two-pane" content: an image and text side by side.
+  * Params: @title    - section title
+  *         @content  - dynamic paragraph content
+  *         @imageURL - string pointing to image asset
+  *         @imageAlt = Alt text for the image.
+{% endcomment %}
+
+<div class="include-two-up">
+  <div class="left-side">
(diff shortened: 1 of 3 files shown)


thomasr8 (approved):

tshen815 (changes requested):
It looks great so far! Just two small things:

1. I think `color: #000000` needs to be applied to the section heading and content text. I added it to the div with the `section-heading` class and it worked well
2. Based on the figma design, I think we can remove `margin-bottom: 16px` from `section-heading` as well. I think that looks closer to the design?

Awesome stuff on this 👍

teesloane:
This is how things are looking when I remove the sidebar, and fixed the padding to show the spacing relationship between multiple two-up panes:

image

PR #6481, opened 2022-03-07
[DD 425] Migrate Troubleshooting Document from Support to Docs

teesloane:
# Description
- Migrate 30 troubleshooting articles from support into a single document.
- Each individual article is now under a `<details>` folding html element.

This is still blocked as we are pending how articles will be categorized.

PR #6565, opened 2022-03-21
Add: scroll margin top to h1's

teesloane:
I thought I added this small change and pushed this to #6481 , but I did not, and then I merged the previous branch.


tshen815 (approved):

chantelle11 (approved):
Works for me great job!

PR #6578, opened 2022-03-23
Document the OpenID Connect ID token.

aterweele:
# Description
New documentation page describing the OpenID Connect token available in CircleCI jobs that use a context.

# Reasons
Document a new feature.


devinmarieb:
Will upload the asciidoc file in a bit that is more inline with the new way we are adding docs + style changes

PR #6591, opened 2022-03-24
[DD-454] - Add "new" badge to sidebar for Quickstart Guide

teesloane:
[Ticket](https://circleci.atlassian.net/browse/DD-454)
[Preview](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/ts-dd-454-preview/?force-all)

*Note*: 
I tried to set the base branch for merge to be DD-450, but GitHub responded with:
> There are no new commits between base branch 'DD-450/getting-started-experiment-boilerplate' and head branch 'tees/DD-454/getting-started-new-badge'

This is maybe because I accidentally pushed my code to DD-450. I undid my commits but I'm not sure why I'm unable to set this Pr to merge  to `DD-450/getting-started-experiment-boilerplate`

## Before

image

## After 
image

### DIFF
--- a/src/js/experiments/gettingStartedGuide.js
+++ b/src/js/experiments/gettingStartedGuide.js
@@ -1,5 +1,21 @@
 // https://app.optimizely.com/v2/projects/16812830475/experiments/21268663100/variations
 
+/**
+ * Changes "Your First Green Build" in the docs sidebar to "Quickstart Guide"
+ * Adds a "NEW" badge to the sidebar item as well.
+ **/
+function addNewBadgeToSidebar() {
+  const NEW_SIDEBAR_HTML = `
+    <a class="" href="/docs/2.0/getting-started/"
+       data-section="getting-started" data-proofer-ignore="">
+      <span>Quickstart Guide</span>
+      <img style="margin: -2px 0px 0px 4px;" src="/docs/assets/img/compass/new.svg" alt="New">
+    </a>
+`;
+
+  $("li > a[href='/docs/2.0/getting-started/']").replaceWith(NEW_SIDEBAR_HTML);
+}
+
 // NOTE: experimentContainer can be updated once we have the other components of this experiment created and in place
 window.OptimizelyClient.getVariationName({
   experimentKey: 'dd_getting_started_docs_test',
@@ -8,6 +24,7 @@ window.OptimizelyClient.getVariationName({
   guestExperiment: false,
 }).then((variation) => {
   if (variation === 'treatment') {
+    addNewBadgeToSidebar();
     // console.log('IN TREATMENT GROUP');
   }
   if (variation === 'control') {
--- a/src/js/experiments/gettingStartedGuide.js
+++ b/src/js/experiments/gettingStartedGuide.js
@@ -6,7 +6,7 @@
  **/
 function addNewBadgeToSidebar() {
   const isGettingStartedPage =
-    window.location.pathname == '/docs/2.0/getting-started/'
+    window.location.pathname == '/docs/2.0/getting-started/';
 
   const NEW_SIDEBAR_HTML = `
     <a class="${isGettingStartedPage ? 'active' : ''}"
--- a/src/js/experiments/gettingStartedGuide.js
+++ b/src/js/experiments/gettingStartedGuide.js
@@ -5,8 +5,14 @@
  * Adds a "NEW" badge to the sidebar item as well.
  **/
 function addNewBadgeToSidebar() {
+  const isGettingStartedPage =
+        window.location.pathname == '/docs/2.0/getting-started/' ||
+        window.location.pathname == '/docs/ja/2.0/getting-started/';
+
   const NEW_SIDEBAR_HTML = `
-    <a class="" style="display: flex; align-items: center" href="/docs/2.0/getting-started/"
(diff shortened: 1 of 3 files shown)


src/js/experiments/gettingStartedGuide.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Changes "Your First Green Build" in the docs sidebar to "Quickstart Guide"
+ * Adds a "NEW" badge to the sidebar item as well.
+ **/
+function addNewBadgeToSidebar() {
+  const NEW_SIDEBAR_HTML = `
+    <a class="" href="/docs/2.0/getting-started/"
+       data-section="getting-started" data-proofer-ignore="">
+      <span>Quickstart Guide</span>
+      <img style="margin: -2px 0px 0px 4px;" src="/docs/assets/img/compass/new.svg" alt="New">
```
tshen815, line 12: should we use a dynamic path for the img src for preview?

tshen815 (approved):
Looks great! Left one small comment, wasnt sure if we wanted a preview build for this but tested it out locally 👍

chantelle11 (commented):
Looking great!! Could we get a preview link for this?

src/js/experiments/gettingStartedGuide.js
```diff
@@ -8,6 +24,7 @@ window.OptimizelyClient.getVariationName({
   guestExperiment: false,
 }).then((variation) => {
   if (variation === 'treatment') {
+    addNewBadgeToSidebar();
     // console.log('IN TREATMENT GROUP');
```
chantelle11, line 28: Nit: we can remove this?

chantelle11 (commented):
Looking good for me! just want to see preview link :)

src/js/experiments/gettingStartedGuide.js
teesloane: We are unable to access {{site.baseurl}} in js. @rsylvian - do you think we should just use base64?

src/js/experiments/gettingStartedGuide.js
teesloane: The console log? That's part of the base branch, which I assumed we should not be changing until the end in case we are merging all the respective parts of the epic at once.

src/js/experiments/gettingStartedGuide.js
rsylvian: Why do we need an image for that? can we just do it in CSS?

chantelle11 (commented):
To my understanding treatment is to be always on for preview branches? But I see it on locally but not on in preview
Screen Shot 2022-03-28 at 1 16 55 PM

Screen Shot 2022-03-28 at 1 17 06 PM

chantelle11 (commented):
Quick start guide in the side nav does not highlight when clicked
Screen Shot 2022-03-28 at 1 18 18 PM

teesloane:
> Quick start guide in the side nav does not highlight when clicked Screen Shot 2022-03-28 at 1 18 18 PM

good catch, I'll fix that.

src/js/experiments/gettingStartedGuide.js
```diff
@@ -1,5 +1,27 @@
 // https://app.optimizely.com/v2/projects/16812830475/experiments/21268663100/variations
 
+/**
+ * Changes "Your First Green Build" in the docs sidebar to "Quickstart Guide"
+ * Adds a "NEW" badge to the sidebar item as well.
+ **/
+function addNewBadgeToSidebar() {
+  const isGettingStartedPage =
+    window.location.pathname == '/docs/2.0/getting-started/' ||
+    window.location.pathname == '/docs/ja/2.0/getting-started/';
```
rsylvian, line 10: this is wrong because the new badge should show up regarding of the current page?

rsylvian (changes requested):

src/js/experiments/gettingStartedGuide.js
teesloane: No, this is just being used to determine if we are on the getting started page so that we can show that the link is active.

chantelle11: /docs will need to be changed to branch name in order for it to work in the preview branch but the issue is that the badge doesn't show up in the preview branch despite treatment should always be turned on during preview

src/js/experiments/gettingStartedGuide.js
```diff
@@ -1,5 +1,27 @@
...
+    <a class="${isGettingStartedPage ? 'active' : ''}"
+       style="display: flex; align-items: center"
+       href="/docs/2.0/getting-started/"
+       data-section="getting-started" data-proofer-ignore="">
+      <span>Quickstart Guide</span>
+      <span class="getting-started-new-badge"> NEW </span>
+    </a>
+`;
+
+  $("li > a[href='/docs/2.0/getting-started/']").replaceWith(NEW_SIDEBAR_HTML);
```
chantelle11, line 22: this needs to be not /docs to work in preview build i think same as above

src/js/experiments/gettingStartedGuide.js
rsylvian: @teesloane oh! yeah I misread the code below 🤦‍♂️

rsylvian (approved):

PR #6599, opened 2022-03-25
Add note re:ssh-key-add & env vars

CobyPear:
# Description
Adds a note regarding the `ssh-key-add` step and storing fingerprints in CircleCI environment variables.

# Reasons

There's some [info on this in the discussion board here](https://discuss.circleci.com/t/using-an-environment-variable-inside-of-add-ssh-keys-step/27302) but it doesn't explicitly state this. I think putting this in the docs will help people who are new to the CircleCI platform and are trying to adhere to perceived security best practices.


rosieyohannan (approved):

rosieyohannan:
Thank you so much @CobyPear and sorry it's taken so long to get this change in!

PR #6602, opened 2022-03-25
Ensure all images have alt text

rosieyohannan:
For accessibility, we should always include alt text when including images in our docs. This PR adds those that were missing.


catherinepope (approved):
👍

PR #6611, opened 2022-03-28
Added AsciiDoc example

catherinepope:
# Description
Added AsciiDoc example

# Reasons
https://circleci.atlassian.net/browse/DOCSTEAM-195

### DIFF
--- a/docs/Content-Reuse.md
+++ b/docs/Content-Reuse.md
@@ -20,7 +20,9 @@ Site variables are defined within `jekyll/_config.yml` as key value pairs:
 my-variable-name: my-variable-value
 ```
 
-### Markdown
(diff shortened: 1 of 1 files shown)


docs/Content-Reuse.md
```diff
@@ -20,8 +20,33 @@ Site variables are defined within `jekyll/_config.yml` as key value pairs:
 my-variable-name: my-variable-value
```
rosieyohannan, line 20: @catherinepope I think we also actually need to separate out how to add the variables to the `_config` file. SO for asciidoc they need defining as attributes under `asciidoctor:`

catherinepope: Wouldn't they only go under `asciidoctor` for PDFs? I can't see anywhere to add a non-AsciiDoc attribute - where would they go?

rosieyohannan:
Thank you @catherinepope !!

PR #6613, opened 2022-03-28
Update environment variables

trulex:
# Description
Replace outdated environment variables with new.

# Reasons
Replaced environment variables are no longer used in the script so the build fails when it's configured as described currently.


kelvintaywl (approved):
This looks good to me, as per our latest aws-ecr Orb @ 8.0.0

Screen Shot 2022-04-13 at 8 20 46

thomasr8:
Thanks @kelvintaywl That is what I thought, but I wanted to make sure.

PR #6618, opened 2022-03-30
Fix: only show "reading time" in tooltip if it exists for doc.

teesloane:
[Ticket](https://circleci.atlassian.net/browse/DD-497)

# Description
- Fix: conditionally show "reading time" popper text if reading time is enabled for that document.

# Reasons
Original bug: 
image

### DIFF
--- a/jekyll/_includes/quick-info-list.html
+++ b/jekyll/_includes/quick-info-list.html
@@ -1,7 +1,9 @@
 {% if page.page-type != "index" and page.page-type != "homepage" %}
 <div id="tooltip" role="tooltip" class="tooltip-popover"></div>
 <div id="tooltip-time" role="tooltip" class="tooltip-popover">
-  Last updated • Reading time
+  <span>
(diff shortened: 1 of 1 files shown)


thomasr8 (approved):

chantelle11 (approved):
Looks good to me thanks!

jekyll/_includes/quick-info-list.html
```diff
@@ -1,7 +1,9 @@
 {% if page.page-type != "index" and page.page-type != "homepage" %}
 <div id="tooltip" role="tooltip" class="tooltip-popover"></div>
 <div id="tooltip-time" role="tooltip" class="tooltip-popover">
-  Last updated • Reading time
+  <span>
+    <span>Last updated</span> {% if page.readtime %} <span>• Reading time</span>{% endif %}
```
rsylvian, line 5: are we sure formating is okay? is it not going to do `Last updated• Reading time` instead?

PR #6620, opened 2022-03-31
Init impl of auto assignment of reviewer for l10n

makotom:
# Description

This introduces a GitHub action workflow to assign a reviewer randomly if a new PR is created `master <- ja-localization`.

# Reasons

* We need a reviewer to merge l10n works.
* We don't want to overwhelm a single person with reviews.


makotom:
This is an example of how the job would be triggered. https://github.com/circleci/circleci-docs/runs/5769036687 (It will go through without skips if the head branch is `ja-localization` and the base branch is `master`.)

chantelle11 (approved):
LGTM

PR #6625, opened 2022-03-31
Add onlyQualifyGuests option to accurately show experiments to only guests

rsylvian:
# Description
A brief description of the changes.

# Reasons
A link to a GitHub and/or JIRA issue (if applicable).
Otherwise, a brief sentence about why you made these changes.

PR #6626, opened 2022-03-31
Revert "Add `onlyQualifyGuests` option to accurately show experiments to only guests"

chantelle11:
Reverts circleci/circleci-docs#6625


tshen815 (approved):
This was the last merged PR before we got an alert from datadog rum about a failed test. Approving to see if it solves the issue.

PR #6637, opened 2022-04-04
Storage feature

devinmarieb:
# Description
Since storage controls are not available to free users:
- Slight change in wording on the persisting data page.
- Add storage controls to list of features on paid plans.

# Reasons
Figured it would be nice to add on the plans page that this feature exists.


thomasr8 (approved):
Looks good 👍

PR #6641, opened 2022-04-05
remove orb author server page and use warnings instead

rosieyohannan:
Recently we made a separate orb authoring page for server v3.x.

This PR reverts that change and instead uses a note to indicate which parts of the main orb auth doc are not available on server.

Reason: there was too much duplication, which would lead to errors and increased workload for upkeep :-D


catherinepope (approved):
👍

PR #6644, opened 2022-04-05
[Dd 455] - homepage redesign preview

teesloane:
# Description
- [Ticket](https://circleci.atlassian.net/browse/DD-455)

- adds badges to the home page, different from the language guide experiment (as that experiment is still running)
- cleans up how mixins are imported as they were being done improperly before (they should be included in the main.scss file, not in individual files)
- created an include for the new badge, as well as a respective scss file.

### Screenshot

image


src/styles/main.scss
```diff
@@ -323,6 +330,7 @@ body.homepage{
 		display: flex;
 		flex-wrap: wrap;
 		width: 100%;
+   flex-direction: column;
```
chantelle11, line 333: Nit, I think there is a space issue?

chantelle11 (changes requested):
On prod I can go to '/docs' and be redirected to '/docs/' on this branch I can't
https://circleci.com/docs

chantelle11 (approved):
LGTM

teesloane:
Please note that currently the home page looks like this until we decide what to do with the results of the lang tutorial experiment:

image

src/styles/pages/_getting-started.scss
```diff
@@ -1,19 +1,3 @@
...
-  font-weight: 500;
-  line-height: 1rem;
-  letter-spacing: 0.01em;
-  text-transform: uppercase;
-  border-radius: 100px;
-  color: rgb(255, 255, 255);
-  background-color: rgb(0, 137, 51);
-  white-space: nowrap;
-  margin-left: 8px;
+  @include cta-badge;
```
tshen815, line 2: wow nice I didnt know we could do this

tshen815 (changes requested):
This looks fantastic, and thanks for adding it into `gettingStartedGuide` already! Just noticed a few very minor things:

1. There is some text-decoration for the Quickstart Guide and the New badge where its underlined if we open the link in a new tab
2. Small typo of "here" in the first sentence 
3. I think the border radius should be set to 4px for the badges 

Happy to approve afterwards! 🎉 

Screen Shot 2022-04-05 at 1 10 34 PM

tshen815 (approved):

PR #6647, opened 2022-04-05
add 20.10.12 to list of available docker versions

JTunis:
# Description
Docker version `20.10.12` is available to use, so this adds it to the list of available versions
Screen Shot 2022-04-05 at 1 27 21 PM

# Reasons
Keeps the list of available Docker versions up-to-date


thomasr8 (approved):

kelvintaywl (approved):
LGTM!

i confirmed 20.10.12 is available indeed:

Screen Shot 2022-04-13 at 8 26 31

PR #6660, opened 2022-04-07
Use cimg/* images for example on /adv-config/

tnir:
Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

# Description
Use `cimg/*` images for example on [/adv-config/](https://circleci.com/docs/2.0/adv-config/).

# Reasons
`circleci/node` and `circleci/python` images are legacy and has been deprecated on 2021-12-31.

> Legacy images with the prefix "circleci/" were [deprecated](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034) on December 31, 2021.

# Appendix
```
$ docker run --rm -it cimg/node:lts-browsers cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

```
$ docker run --rm -it cimg/python:3.6.15 cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.2 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

cf. Ubuntu 20.04, the base image of `cimg/python:3.6`, has `psql` command in [postgresql-client-12](https://packages.ubuntu.com/focal/postgresql-client-12) instead of [postgreql-client-9.6](https://packages.debian.org/stretch/postgresql-client-9.6).

PR #6674, opened 2022-04-11
Use valid PG image version and update credentials

tnir:

# Description
- Uses valid PG image version (9.6.22 or 9.6.24 are available in PG 9.6),
- updates credentials (username) defined in `TEST_DATABASE_URL`.

# Reasons
`POSTGRES_USER` is default to `postgres`: https://github.com/CircleCI-Public/cimg-postgres/blob/main/docker-entrypoint.sh#L213

Follows up #6660

### DIFF
--- a/jekyll/_cci2/adv-config.md
+++ b/jekyll/_cci2/adv-config.md
@@ -224,10 +224,10 @@ jobs:
           username: mydockerhub-user
           password: $DOCKERHUB_PASSWORD  # context / project UI env-var reference
         environment:
-          TEST_DATABASE_URL: postgresql://root@localhost/circle_test
+          TEST_DATABASE_URL: postgresql://postgres@localhost/circle_test
 
     # Service container image
-      - image: cimg/postgres:9.6.5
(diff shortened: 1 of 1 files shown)


tnir:
@devinmarieb What do you think?

devinmarieb:
@tnir This looks fine for now thanks! I'm just double checking with engineering since we use both 
`TEST_DATABASE_URL: postgresql://root@localhost/circle_test`
`TEST_DATABASE_URL: postgresql://postgres@localhost/circle_test`
in various places in our docs. The version you've identified to update is fine, but we will likely update to just `<version>` in the future since there are a lot to chose from.

tnir:
> I'm just double checking with engineering since we use both
> TEST_DATABASE_URL: postgresql://root@localhost/circle_test
> TEST_DATABASE_URL: postgresql://postgres@localhost/circle_test
> in various places in our docs.

@devinmarieb Not sure how you finished to double-check that, but in #6700 I clarified the problem around it, which exists for years.

PR #6683, opened 2022-04-12
Use anchor/reference to reduce duplication in cache keys

exoego:
# Description

Replaced duplicated cache keys with anchor (`&cache-key`) and reference (`*cache-key`)

# Reasons
To encourage users to reduce duplication and makes it easier to maintain confing.


rosieyohannan:
Thanks so much for this idea @exoego! 🎉

rosieyohannan (approved):

PR #6686, opened 2022-04-13
orb-publishing-process update

danielholdsworth:
# Description
updated the orb-publishing-process.png
# Reasons
The publishing process has changed since v11


devinmarieb:
@danielholdsworth It was a good effort. Would you like me to upload it instead?

PR #6706, opened 2022-04-19
Using maven, update circleci cli split sentence in the example

dariocazas:
# Description
The split sentence don't work well, copy from the Gradle example run fine https://circleci.com/docs/2.0/language-java/#sample-configuration

# Reasons
The previous circleci split sentence return this error:
```
Error autodetecting timing type, falling back to weighting by name. Autodetect no matching filename or classname.  If file names are used, double check paths for absolute vs relative.
```

### DIFF
--- a/jekyll/_cci2/language-java-maven.md
+++ b/jekyll/_cci2/language-java-maven.md
@@ -223,13 +223,17 @@ jobs:
     steps:
       - checkout
       - run: |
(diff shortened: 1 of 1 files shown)


rosieyohannan:
Thank you for this contribution, and sorry it's taken so long to respond! This page has now been archived. There will be an updated language guide soon.

PR #6709, opened 2022-04-20
Replace insights snapshot badge image with screenshot of README

kbatuigas:
# Description
Instead of a screenshot of just the badge itself, we'll use a screenshot of a project README with the badge embedded.

# Reasons
Makes it clearer that it's not an image of a Mac window.


catherinepope (approved):
👍

PR #6716, opened 2022-04-21
[DD-542] Inline API v1 resources

teesloane:
# Description
Ticket - [DD#542](https://circleci.atlassian.net/browse/DD-542)

This PR inlines javascript and stylesheets so that the api v1 does not depend on the hosting of those files (making it easier to migrate things over to new docs-platform)

- takes all the JS that is vendored in lib/app and put it into layout.erb.
  - uses a cdn when possible
- vendors the functionality of the [middleman-inline plugin](https://github.com/miloshadzic/middleman-inline) - why? The gem was not working. 
- why aren't we using the `inline_js` method in the .erb layout? Because Sprockets uses something. called `execjs` to actually validate your javascript - and it stumbles over the vendored jquery.

Let me know if you have any questions! You can view the inlined API v1 [here](http://circleci-doc-preview.s3-website-us-east-1.amazonaws.com/tees/DD-542/inline-api-v1-preview/api/v1/#trigger-a-new-job)

### DIFF
--- a/src-api/Gemfile
+++ b/src-api/Gemfile
@@ -5,11 +5,9 @@ source 'https://rubygems.org'
 gem 'middleman', '~> 4.3.7'
 gem 'middleman-syntax', '~> 3.0.0'
 gem 'middleman-autoprefixer', '~> 2.7.1'
-gem 'middleman-inline'
 gem 'middleman-sprockets', '~> 4.1.0'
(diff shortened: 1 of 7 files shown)


chantelle11 (approved):
Everything looks like its working the same as before!

PR #6722, opened 2022-04-22
docs: warn new orb repos must be empty

KyleTryon:
# Description
Add warning to make clear that new orb repos MUST be empty


catherinepope (approved):
👍

PR #6724, opened 2022-04-25
Typo fix

azwok:
# Description
Minor typo fix

# Reasons
Fix a minor typo in a reference to an example in the following block.


catherinepope (approved):
👍  Many thanks, @azwok !

PR #6737, opened 2022-04-26
style collapsible blocks and code callouts

rosieyohannan:
# Description
Collapsable blocks - https://docs.asciidoctor.org/asciidoc/latest/blocks/collapsible/
Code callouts - https://docs.asciidoctor.org/asciidoc/latest/verbatim/callouts/

Both would be useful features but some CSS work is required

I've made a test page in this PR: `test-asciidoc`

## Collapsable blocks
The text used for the block label is styled with the same class used for figure titles - centre aligned and italic. This does not look good, we need something that is obviously different and clickable. I've noticed if I remove the `display: block` setting the little expand arrow appears:

Screenshot 2022-04-26 at 14 36 25

Screenshot 2022-04-26 at 14 38 06

## Callouts

The numbers in the snippet and the corresponding text should appear inside circles, like here: https://docs.asciidoctor.org/asciidoc/latest/verbatim/callouts/#callout-icons, but they don't:

Screenshot 2022-04-26 at 14 39 36

If you look at the preview URL the callout "bubbles" are rendering in the text (although they don't look great), and you can see them for a second when refreshing the page in the snippet, but that formatting is being overridden.

Any help with these two features much appreciated!


rsylvian:
@rosieyohannan Can we meet to talk bout that? With Docs Detangling, all styling is stripped away from the content so if you need something custom we should coordinate so we (`docs-disco`) can build it for you.

PR #6747, opened 2022-04-28
Fix: search hits not clickable on chrome v101

rsylvian:
# Description
- Remove blanket `pointer-events: none;` on all link's child

# Reasons
It was reported to us that search results are not clickable on Chrome latest update. I was able to reproduce on `Chrome Version 101.0.4951.41`

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -76,12 +76,6 @@ a {
     text-decoration: none;
   }
 
-  &>* {
-    // to prevent the child elements inside a tag from registering
-    // as the target in event listeners (for analytics tracking)
-    pointer-events: none;
-  }
-
   &.active {
     color: $link-text-color !important;
   }


src/styles/main.scss
```diff
@@ -76,12 +76,6 @@ a {
     text-decoration: none;
   }
 
-  &>* {
-    // to prevent the child elements inside a tag from registering
-    // as the target in event listeners (for analytics tracking)
-    pointer-events: none;
-  }
```
rsylvian, line 83: @tshen815 I am assuming I can't just remove all of it but we should probably not target every `a` children either?

chantelle11 (approved):
Approved!

src/styles/main.scss
tshen815: agreed, Im not sure when this was set in place but targeting every `a` seems risky

tshen815 (approved):

PR #6752, opened 2022-04-28
add note about pytest 6.1

rosieyohannan:
# Description
This PR adds a note about using a legacy XML format with pytest if you want timing-based test splitting.

# Reasons
Using pytest 6.1 the default junit_family format changed to no longer include filenames, which will break timings-based test splitting. The format will need to be changes to `xunit1` for this feature to work. 

# Content Checklist
Please follow our style when contributing to CircleCI docs. Our style guide is here: https://circleci.com/docs/style/.


catherinepope (approved):
👍

PR #6762, opened 2022-04-29
Change UI instructions for org id

devinmarieb:
# Description
Updating the instructions for a user to find their org id.

# Reasons
The UI to find an organization ID is changing from Context page to Overview page in the UI.


catherinepope (approved):
👍

PR #6767, opened 2022-05-02
switch "web app" for "UI"

sebastian-lerner:
i think we'd also need to change the wording in: `snippets/runner-platform-prerequisites.adoc` but i'm not sure how to get to that file?  and the wording in the left-hand table of contents.

also at the bottom of: `runner-overview.adoc` , the "see also" would need to change as well but i couldnt figure out how to do that...

### DIFF
--- a/jekyll/_cci2/runner-installation.adoc
+++ b/jekyll/_cci2/runner-installation.adoc
@@ -2,7 +2,7 @@
 version:
 - Cloud
 ---
-= Installing Self-Hosted Runners with the Web App
+= Installing Self-Hosted Runners with the CircleCI UI
 :page-layout: classic-docs
 :page-liquid:
 :page-description: Find resources to install CircleCI self-hosted runners on Linux, macOS, Docker, and Windows platforms.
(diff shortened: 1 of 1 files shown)


sebastian-lerner:
if the docs team feels incredibly strongly that we must use the term "web app", then feel free to decline this pull request.

i just really / very strongly prefer "CircleCI UI" over "web app"...

rosieyohannan:
Hey @sebastian-lerner interesting! We had an internal discussion about this and chose "web app" as standard. But! I'm very interested to hear your thoughts, nothing is set in stone. We just want consistency :-D

sebastian-lerner:
got it, thanks for the update.  to me, "web app" sounds old and legacy.
 it doesn't exactly exude the feeling of modern software.  that's really
it, but i understand needing consistency across the docs.

devinmarieb:
@sebastian-lerner and @rosieyohannan if we want to get into technicalities, a "UI" is just a singular part of a web application. While it might sound more modern, "UI" is not technically correct when referring to a whole program/application.

sebastian-lerner:
yeah i understand that, but is the goal to be 100% technically correct or
to use language that resonates with developers and makes us sound like a
modern software company?  i would say it's the latter.  i think developers
understand UI can reference the application when we have a technical
product where you dont actually have to interact with the UI/web app to use.

devinmarieb:
@sebastian-lerner You should start a poll in engineering 😆 Either way though, we use web app currently and we won't update this page to be inconsistent.

PR #6773, opened 2022-05-04
fix redirects

rosieyohannan:
# Description
redirects were broken due to missing `/`


devinmarieb:
@rosieyohannan I'm confused by the second commit. Why is there a non-Japanese redirect on a Japanese page?

rosieyohannan:
@devinmarieb yeah I'm not 100% sure but I planned to open a separate PR for that. I just reverted the change so I wouldn't need a review from the translations team. Sorry should have made that clear!

PR #6785, opened 2022-05-06
update snippet to correct RAM values

rosieyohannan:
# Description
Update RAM column in machine executor table in config ref

# Reasons
Values were in fact disc size.


sebastian-lerner (approved):

catherinepope (approved):
👍

PR #6814, opened 2022-05-17
stop uppercase subsections linking to pages

rosieyohannan:
# Description
Changes the sidebar layout to standardise the use of subsections - uppercase subsections no longer link to pages

# Reasons
A few subsections were providing duplicate links to overview pages. This was causing confusion for docs disco when testing out new platform: https://circleci.slack.com/archives/C0313KL1X9B/p1652734542280669

### DIFF
--- a/jekyll/_data/sidenav.yml
+++ b/jekyll/_data/sidenav.yml
@@ -19,7 +19,6 @@ en:
         - name: FAQ
           link: 2.0/faq/
     - name: Migration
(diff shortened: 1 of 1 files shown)


devinmarieb (approved):
I assume this is a hold over from when we used "Overview" as a heading but also a link? I haven't been paying attention, have we been making those "Overview" headers non-linkable (I think that's the way to go, it's confusing).

PR #6863, opened 2022-05-26
Fix which agent agentVersion is for in the Runner K8s docs

christian-stephen:
# Description
The `agentVersion` in the Helm chart is actually for `launch-agent`. See [here](https://github.com/CircleCI-Public/circleci-runner-docker/blob/main/launch-agent/Dockerfile#L3).

# Reasons
See above.

# Content Checklist


devinmarieb:
@christian-stephen could you remove the changes from the Japanese file? Those files get updated automatically when changes to the English files happen. Otherwise this PR is likely to get held up in the translation process.

christian-stephen:
https://circleci.atlassian.net/browse/RT-578

PR #6996, opened 2022-06-23
Update faq.md

ogii:
Updating the Japanese text to reflect the correct cost for 840 credits ($50.0 to $0.50). The English version correctly states $0.5

# Description
Updating the correct cost of 840 credits to $0.5.

# Reasons
We should be consistent in our messaging for pricing to customers.


Mayuko-Sa (approved):
Thank you!

PR #6998, opened 2022-06-23
Fix git tag workflows example

fynsta:
# Description
Change "test" to "build" in git tag workflow example.

# Reasons
It required "build", which was not a defined job for that workflow. I assume "test" was meant?


rosieyohannan:
Good spot @fynsta thanks so much for fixing this!

rosieyohannan (approved):

PR #7000, opened 2022-06-24
Feat: Docs Detangling Tooling

rsylvian:
# Description
- Add custom Jekyll plugins to generate JSON data that can be used by `docs-platform`
- Remove `src-shared` submodule because it is handled by `docs-platform` directly now 
- Adapt CI config to make sure it still builds correctly
- Add @teesloane's work on API building
- Cleaned up various small pieces that are not needed anymore (JS experiments, JS UI)

# Reasons
This is the first step in getting some of that Docs Detangling work to `master`. Once this is merged, this will allow us to run `docs-platform` directly from `master` instead of the `rs/feat/docs-detangling` branch which will make the content publishing workflow much more streamlined.

PR #7004, opened 2022-06-24
Fix: Do not alter jekyll internal objects directly

rsylvian:
# Description
- Use `dup` to copy the jekyll objects

# Reasons
https://github.com/circleci/circleci-docs/pull/7000 introduced a minor bug that directly alters the jekyll internal objects rather than copying them beforehand. This was introducing a lot of unnecessary extra metadata that was not needed. This is explaining why Algolia refused to `index` our content because the data will be too large:

Screen Shot 2022-06-24 at 3 38 33 PM

### DIFF
--- a/jekyll/_plugins/json_data.rb
+++ b/jekyll/_plugins/json_data.rb
@@ -4,13 +4,13 @@ def data_to_json(site)
   # add sidenav
   output = output.push({
     name: 'sidenav',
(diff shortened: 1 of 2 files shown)


chantelle11 (approved):
LGTM

Nivl (approved):
Looks good to me!

PR #7116, opened 2022-07-20
[DD-734] Add `HTMLEntities` to force decoding html character

rsylvian:
Ticket: [DD-734](https://circleci.atlassian.net/browse/DD-734)

# Description
- Use `HTMLEntities.new.decode` on the document title before it goes to the JSON file

# Reasons
In Asciidoc, titles are automatically escaped with HTML encoding which is then reflected on the site. This PR forces to decode the title before writing it to the HTML. More info here: https://docs.asciidoctor.org/asciidoc/latest/subs/replacements/

# Before
Screen Shot 2022-07-19 at 6 46 20 PM

# After
Screen Shot 2022-07-19 at 6 45 40 PM

### DIFF
--- a/jekyll/_plugins/json_pages.rb
+++ b/jekyll/_plugins/json_pages.rb
@@ -1,4 +1,5 @@
 require 'htmlcompressor'
+require 'htmlentities'
 
 def doc_to_json(document, site)
   puts 'processing json for: ' + document.relative_path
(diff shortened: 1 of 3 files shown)


chantelle11 (approved):
thank you!

jekyll/_plugins/json_pages.rb
```diff
@@ -9,6 +10,10 @@ def doc_to_json(document, site)
   output['content'] = compressor.compress(document.content)
   output['file_name'] = document.relative_path
 
+  # cleanup title from all HTML encoding introduced by asciidoc
+  # see here for more info: https://docs.asciidoctor.org/asciidoc/latest/subs/replacements/
+  output['title'] = HTMLEntities.new.decode document.data['title']
```
chantelle11, line 15: yeah all would be great but this is awesome!

PR #7137, opened 2022-07-22
Update curl command/resolve issue 7133

devinmarieb:
# Description
Update curl command to use double quotes as single quotes are not compatible with Windows. (the -d should fix the one line that still uses single quotes).

# Reasons
[Issue 7133](https://github.com/circleci/circleci-docs/issues/7133)


kbatuigas (approved):
Thanks!

PR #7198, opened 2022-08-01
Specify the registry for server 4.x installs

ryanwohara:
The tricky part is specifying the org as null
and we don't want customers to be confused.

(Also some whitespace)

Question: do we want to just specify `<registry>` and it will be provided to them? I'm unclear how much information is provided them with the license.


lloydChris (approved):

lloydChris:
> Question: do we want to just specify `<registry>` and it will be provided to them? I'm unclear how much information is provided them with the license.

I like what you did here instead of templating registry and passing it to the customer somewhere else

rosieyohannan (approved):

PR #7202, opened 2022-08-02
Updated about-circleci.md - Moved GKE to a dedicated bullet point.

4TT1L4:
# Description
Moved GKE to a dedicated bullet point. 

# Reasons
There is no reason to group GKE with AWS S3 in one bullet point.


devinmarieb (approved):
@4TT1L4 Thank you!

PR #7266, opened 2022-08-17
Fix openid-connect-tokens doc typo

tetsuya28:
# Description
Fix command parametes in openid-connect-tokens doc.

# Reasons
With the current document, `gcloud` command fails with these parameters.

### DIFF
--- a/jekyll/_cci2/openid-connect-tokens.adoc
+++ b/jekyll/_cci2/openid-connect-tokens.adoc
@@ -337,9 +337,9 @@ commands:
             echo $CIRCLE_OIDC_TOKEN > << parameters.oidc_token_file_path >>
             # Create a credential configuration for the generated OIDC ID Token
             gcloud iam workload-identity-pools create-cred-config \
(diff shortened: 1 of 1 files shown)


kelvintaywl (approved):

kelvintaywl:
thanks @tetsuya28 , this LGTM :)

kelvintaywl:
I had originally approved this PR without diving deep.

However, on a second look, I believe our original doc was correct.
This is because the following command parameters are of `env_var_name` type:

- project_id
- workload_identity_pool_id
- workload_identity_pool_provider_id
- service_account_email

Ref: https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci2/openid-connect-tokens.adoc?plain=1#L314-L332

So the way to evaluate the env var's value would be to use "${<< parameters.project_id>>}" as how our docs was showing.

tetsuya28:
@kelvintaywl Sorry, you're correct.
I tried the oidc tutorial with string parameter, so I couldn't pass a ci. In the document, as you say, using parameters with env_var_name.
I apologize you to be confused.
You can close this pr.

kelvintaywl:
@tetsuya28 no worries, I understand your points too.
I note you didnt need `${}` since you have changed your command parameters to `string` type instead.

i will close this PR as such then 👍 
but thank you so much for trying to check in with us!

PR #7328, opened 2022-09-02
make more obvious

sebastian-lerner:
# Description

# Reasons

# Content Checklist

### DIFF
--- a/jekyll/_cci2/runner-installation-docker.adoc
+++ b/jekyll/_cci2/runner-installation-docker.adoc
@@ -13,16 +13,15 @@ version:
 
 This page describes how to install CircleCI self-hosted runner with Docker using a Machine execution environment.
 
-[#strongly-recommended-method]
-== Strongly recommended method
+[#new-recommended-method-container-runner]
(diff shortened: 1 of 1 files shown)


sebastian-lerner:
same here @devinmarieb

devinmarieb:
@sebastian-lerner I think we need one more sentence in the container runner paragraph saying _why_ the container runner solution is recommended. I added a bit about it being closer to parity with the cloud offering. We can change it to whatever, but I think we need to offer a reason why, other than it's just the recommended method. I also changed the first two heading titles - this is still the Docker install page, so the headings should not imply that Docker is only an alternative solution.

sebastian-lerner:
these are the main reasons:

The ability to seamlessly define, publish, and use custom Docker images during job execution

The ability to easily manage dependencies or libraries through custom Docker images instead of enumerating dependencies as part of config.yml steps

Access to a clean build environment with every job which downloads an image and executes the job in a container; taking full advantage of the lightweightedness of container-based technologies

devinmarieb (approved):

PR #7385, opened 2022-09-20
[DD-841] write a script that transforms all the current doc files frontmatter

chantelle11:
[DD-841](https://circleci.atlassian.net/browse/DD-841)

# Description
Wrote a script that:
1. Traverses adoc and md files inside of _cci2 and _cci2_ja
2. Checks if version metadata exists
3. Rewrites file in temporary folder with new format

# Reasons
In preparation of allowing content tagging to exist in all files

# How to test
1. Navigate to scripts/frontmatter_tagging in terminal
2. run `python3 ./frontmatter_tagging.py`

Test files:
`_aws-install.adoc` ---> has no existing front matter
` about-circleci` ---> has version in front matter and other keys
` storage-lifecycle.adoc` ---> only has version in front matter
`v.2.18-overview.md` ---> does not have version in front matter

Also do various 'spot' checks

# Considerations
1. This is a Draft PR until testing can be done in `docs-platform` 
2. Files are rewritten as EditFrontMatter will not use proper encoding for Japanese so files have to be rewritten. EditFrontMatter allows for the requested template format.
I left the above inefficiency because:
    a. the script runs relatively fast still 
    b. the script will only be run locally and not very often
3. I will make another PR to update all the .md and .adoc files. There will be over 500 files updated and wanted to focus the script visibility
4. sectionTags was added inside of _config.yml for preperation of adding it in later

# Note
Tags and Section tags are set to null to be picked up easily in `docs-platform` later. In addition, is a default response for adding a None key.

# Content Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,9 @@ src-shared/
 # output for docs-platform
 json/
 
+# output of scripts
+scripts/frontmatter_tagging/tmp/*
+
 # output from building API doc
 src-api/index.html
 src-api/openapi-final.json
--- a/jekyll/_config.yml
+++ b/jekyll/_config.yml
@@ -154,6 +154,8 @@ defaults:
       suggested: false # Disable helpful resources section by default
       sitemap: true # Enable sitemap by default
       readtime: true # Enable Read Time by default
+      tags: null
+      sectionTags: null
 
 t:
     translation_in_progress:
--- a/jekyll/_config.yml
+++ b/jekyll/_config.yml
@@ -154,7 +154,7 @@ defaults:
       suggested: false # Disable helpful resources section by default
       sitemap: true # Enable sitemap by default
       readtime: true # Enable Read Time by default
-      tags: null
+      contentTags: null
 
 t:
     translation_in_progress:
--- a/jekyll/_config.yml
+++ b/jekyll/_config.yml
@@ -155,7 +155,6 @@ defaults:
       sitemap: true # Enable sitemap by default
       readtime: true # Enable Read Time by default
       tags: null
-      sectionTags: null
 
 t:
     translation_in_progress:
--- a/scripts/frontmatter_tagging/frontmatter_tagging.py
+++ b/scripts/frontmatter_tagging/frontmatter_tagging.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
-# This script iterates over our docs and converts the front matter to 
-# tags and section tags format
+# This script iterates over our docs and converts version front matter to tags format
+# sectionTags are not dealt with in this script
 
 import os
 import glob
@@ -51,17 +51,13 @@ def iterate_docs():
         proc.run({'addedVariable': version})
         with open(tmpPath, 'w') as f:
           print(proc.dumpFileData(), file=f)
-      else: ### Test case file has no front matter and file does not have version frontmatter
-        post.__setitem__('tags', None)
-        post.__setitem__('sectionTags', None)
+        ### Test case file has no front matter and file does not have version frontmatter
+        ### Covered inside of _config default value null
+
+        # This is a work around of EditFrontMatter not encoding japanese characters properly
+        # We rewrite the file with frontmatter that does encode japanese correctly
+        post = frontmatter.load(tmpPath)
         with open(tmpPath, 'w') as f:
           print(frontmatter.dumps(post), file=f)
-
-      # This is a work around of EditFrontMatter not encoding japanese characters properly
-      # We rewrite the file with frontmatter that does encode japanese correctly
-      post = frontmatter.load(tmpPath)
-      with open(tmpPath, 'w') as f:
-        print(frontmatter.dumps(post), file=f)
-
 
 iterate_docs()
--- a/scripts/frontmatter_tagging/frontmatter_tagging.py
+++ b/scripts/frontmatter_tagging/frontmatter_tagging.py
@@ -42,8 +42,10 @@ def iterate_docs():
 
       if matchFrontmatter:
         if matchFrontmatter[0].find("version:") != -1:
-          temp = open(tmpPath, "w")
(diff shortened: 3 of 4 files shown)


scripts/frontmatter_tagging/frontmatter_tagging.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+
+# This script iterates over our docs and converts the front matter to
```
chantelle11, line 3: current version front matter

scripts/frontmatter_tagging/frontmatter_template_tags.j2
```diff
@@ -0,0 +1,4 @@
+tags: 
+  platform: 
+    {{ addedVariable }}
+sectionTags: null
```
rsylvian, line 4:
```suggestion
```

this is not needed at this stage of the project :)

scripts/frontmatter_tagging/frontmatter_tagging.py
```diff
@@ -0,0 +1,67 @@
...
+        template_str = ''.join(open(os.path.abspath("frontmatter_template_tags.j2"), "r").readlines())
+        proc = EditFrontMatter(file_path=path, template_str=template_str)
+        proc.keys_toDelete = ['version']
+        proc.run()
+        proc.run({'addedVariable': version})
+        with open(tmpPath, 'w') as f:
+          print(proc.dumpFileData(), file=f)
+      else: ### Test case file has no front matter and file does not have version frontmatter
+        post.__setitem__('tags', None)
+        post.__setitem__('sectionTags', None)
```
rsylvian, line 56: we don't need to put `None` when data is not present. this will keep the content file as light as possible.

If we need a default value, it can be set globally on the Jekyll level here: https://github.com/circleci/circleci-docs/blob/master/jekyll/_config.yml#L150-L156

jekyll/_config.yml
```diff
@@ -154,6 +154,8 @@ defaults:
       suggested: false # Disable helpful resources section by default
       sitemap: true # Enable sitemap by default
       readtime: true # Enable Read Time by default
+      tags: null
+      sectionTags: null
```
chantelle11, line 158: sectionTags are not being used but I added it in for use later to not be missed

rsylvian: Is `null` a valid value in yml?

jekyll/_config.yml
```diff
@@ -154,6 +154,7 @@ defaults:
       suggested: false # Disable helpful resources section by default
       sitemap: true # Enable sitemap by default
       readtime: true # Enable Read Time by default
+      tags: null
```
chantelle11, line 157: Should we set tags default null? To always have tags be defined? Or have it as an optional variable @halie-kastl 

Version use to always be defined in src/utils/types.ts, version: string[]; but had no default state.

Should we not have a default state for tags and keep it optional? 
Or have a default state and keep it defined?

halie-kastl: This is the state I mentioned that we should verify works. We will need to make sure this is compatible all the way through re-indexing. 

If possible, I'd like it to always we present. this will make future experiments more reliable, and it will help the docs team know where to place the tags they create. The same pattern should be followed through to section tags

halie-kastl: if the field is on some meta-data and not others, we could run into more issues than an occasional null state.

jekyll/_config.yml
chantelle11: Yes https://symfony.com/legacy/doc/reference/1_2/en/02-yaml

.gitignore
```diff
@@ -47,6 +47,9 @@ src-shared/
 # output for docs-platform
 json/
 
+# output of scripts
+scripts/frontmatter_tagging/tmp/*
```
chantelle11, line 51: I put all the files into a tmp folder instead of having them replace `_cci2` or `_cci2_ja2` to allow for spot checks before committing the script.

jekyll/_config.yml
```diff
@@ -154,6 +154,7 @@ defaults:
       suggested: false # Disable helpful resources section by default
       sitemap: true # Enable sitemap by default
       readtime: true # Enable Read Time by default
+      contentTags: null
```
chantelle11, line 157: This is needed such that Algolia has a default value for `contentTags`

halie-kastl (commented):
Just one item to address. Thanks!

chantelle11:
> Just one item to address. Thanks!

Hi Halie,
All of the front matter is moved to its correct place here: https://github.com/circleci/circleci-docs/pull/7398
The reasoning behind the script still placing files in `tmp` folder is to not override the source of truth when running the script.
1. It helps with debugging. You can see the output vs input
2. When debugging you do not have to revert files to get back to the original source of truth
3. If we wanted to run the script directly on the source of truth after verification it is easier to change `tmpPath` to `currentPath` after the final result has been verified
4. It allows for the continued separation in PR review to focus on the script itself than its output of 400 files

halie-kastl (approved):
Approved! Let's have Romain take a peek, and then we can get this thing in!

scripts/frontmatter_tagging/frontmatter_tagging.py
```diff
@@ -0,0 +1,72 @@
...
+import sys
+from pathlib import Path
+import re
+
+## Select desired markdown and adoc files
+docs_paths = (
+  '../../jekyll/_cci2/*.md',
+  '../../jekyll/_cci2/*.adoc',
+  '../../jekyll/_cci2_ja/*.md',
+  '../../jekyll/_cci2_ja/*.adoc',
```
rsylvian, line 16: Can this be removed since Mayuko requested to no have the script impact the content?

rsylvian (approved):
Looks good to me once we remove the `ja` as part of the script 🙌

PR #7400, opened 2022-09-26
Make section tags unique

Mayuko-Sa:
# Description
I‘m not sure if I'm following the right naming convention but it seems that Crowdin requires section tags to be unique. (The updated file was imported successfully.)

# Reasons
Crowdin failed to import the updates with the following messages.
スクリーンショット 2022-09-26 20 37 33
スクリーンショット 2022-09-26 20 40 00


rosieyohannan (approved):
Ah thanks @Mayuko-Sa I'll make sure not to do this again! 🎉

PR #7411, opened 2022-09-28
Ja localization

Mayuko-Sa:
Updated EN -> JA documentation translations

English texts remain in this PR, which will be translated next time.

Code comments (starts with #) are no longer translated. When changes are made to the code, code comments that were previously translated to Japanese may be reverted to English, but this is allowed.


jekyll/_cci2_ja/hello-world.adoc
```diff
@@ -0,0 +1,513 @@
+---
+
+redirect_from: /ja/examples-intro/
+tags:
```
chantelle11, line 4: Use `contentTags`

saori-circleci (approved):

rosieyohannan (approved):

PR #7412, opened 2022-09-28
Fix typo in canceled state of workflow

filias:
# Description
This PR fixes a typo in the `canceled` state of a workflow.
The correct value is `canceled` and not `cancelled`.

# Reasons
We spotted this error while checking the status of a workflow. We had followed the docs with `cancelled` (with 2 `l`s) but it turned out that the real value was `canceled` (with 1 `l` only).


rosieyohannan (approved):
Thank you!!

PR #7447, opened 2022-10-11
The latest tag no longer exists its now current

owenjoliver:
# Description
This fixes the example as I ran into the issue when trying this just now

# Reasons
I ran into this issue just now and we have updated the latest tag to now be current so the build fails when using the example.

### DIFF
--- a/jekyll/_cci2/pipeline-variables.md
+++ b/jekyll/_cci2/pipeline-variables.md
@@ -1,6 +1,6 @@
 ---
 layout: classic-docs
-title: "Pipeline values and parameters"
+title: Pipeline values and parameters
 description: "Detailed information about pipeline parameters and values"
 categories: [getting-started]
 order: 1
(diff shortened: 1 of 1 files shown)


devinmarieb:
Hey, @owenjoliver, something in the pipeline failed. I don't have the ability to try and rerun it. Could you check out the error and see what's wrong?
https://app.circleci.com/pipelines/github/circleci/circleci-docs/46306

owenjoliver:
The error it is throwing is a re-auth error which is odd.

Could not find a usable config.yml, you need to reauthorize and approve the SAML-protected org.
Please sign out of CircleCI and log back in with your VCS before triggering a new pipeline.

devinmarieb:
@owenjoliver It seems your token has become invalid, can you re-authorize? There's a support article about this issue here: https://support.circleci.com/hc/en-us/articles/360043002793-Troubleshooting-CircleCI-Access-After-Enabling-Github-SSO

owenjoliver:
Thank you I have just done the re-auth so it should be ok again now

devinmarieb (approved):

PR #7448, opened 2022-10-11
Update reusing-config.md

BenjaminLaschkar:
# Description
Change in documentation according to this discussion. https://discuss.circleci.com/t/using-pipeline-parameters/34980/9

# Reasons
The default is always required for parameters.
image

### DIFF
--- a/jekyll/_cci2/reusing-config.md
+++ b/jekyll/_cci2/reusing-config.md
@@ -72,7 +72,7 @@ Key Name | Description | Default value
 ---|---|---
 description | Optional. Used to generate documentation for your orb. | N/A
 type | Required. See **Parameter Types** in the section below for details. | N/A
-default | The default value for the parameter. If not present, the parameter is implied to be required. | N/A
+default | Required. The default value for the parameter. If the parameter is required, use his own name for the default value and check if it exist with an if. | N/A
 {: class="table table-striped"}
 
 ### Parameter types


jekyll/_cci2/reusing-config.md
```diff
@@ -72,7 +72,7 @@ Key Name | Description | Default value
 ---|---|---
 description | Optional. Used to generate documentation for your orb. | N/A
 type | Required. See **Parameter Types** in the section below for details. | N/A
-default | The default value for the parameter. If not present, the parameter is implied to be required. | N/A
+default | Required. The default value for the parameter. If the parameter is required, use his own name for the default value and check if it exist with an if. | N/A
```
rosieyohannan, line 75:
```suggestion
default | Default value for the parameter. Can be overridden. If a default is not supplied the parameter becomes mandatory. | N/A
```

rosieyohannan (commented):
Thank you for helping us improve this section! How about the suggestion? Does that make it clear? We don't want to say this is required as that's not quite accurate. 🙏

BenjaminLaschkar:
> Thank you for helping us improve this section! How about the suggestion? Does that make it clear? We don't want to say this is required as that's not quite accurate. 🙏

It's a pleasure to help :) Sorry I am seeing this only now.
I find your suggestion not clear enough. The default parameter is mandatory everytime right ? So I find the "Required" crystal clear here. I don't understand why you want to remove it

rosieyohannan:
@BenjaminLaschkar Thanks for your reply. We don't want to say "Required", because that would mean the `default` is always needed, and that's not right. If the parameter is available/set then the default does not strictly need to be defined.

BenjaminLaschkar:
First sorry for my late reply.
I updated the documentation **because it is required**. If it is not set then I have the error message above in the description.
I didn't find a way to use a parameter without `default`, that's why I created this PR ^^'
If I am wrong please **tell in the documentation how to do it without the default** as it is really not explicit.

BenjaminLaschkar:
First sorry for my late reply.

I am not agree, I opened this PR because this keyword is mandatory. If it is not may be it's a bug ?
If it's not a bug, then the documentation is not clear on how to use a parameter without the keyword `default`.
If this keyword is not set for the parameter then there is an error as you can see the picture in my description.

Please read the comment from stec00 on your forum [here](https://discuss.circleci.com/t/using-pipeline-parameters/34980/11). 

He had the same issue.

PR #7478, opened 2022-10-17
Replaced privacy policy link from the global footer

yuki-sakamoto-san:
# Description
I replaced the link for the privacy policy from the old to the new (https://circleci.com/legal/privacy/)

# Reasons
We replaced these links across the Outer and Marketo because the previous URL was redirecting to our legal portal which is privacy.circleci.com. 

We made this decision a while ago that, we'd only want folks who want to take an action against their personal data to be landed on privacy.circleci.com - everyone else should land https://circleci.com/legal/privacy/

# Content Checklist

### DIFF
--- a/jekyll/_includes/global-footer.html
+++ b/jekyll/_includes/global-footer.html
@@ -15,7 +15,7 @@
 
           <li><a class="nav-link" href="https://circleci.com/terms-of-service/"
               data-analytics-child='{ "menu": "terms-of-service" }'>Terms of Service</a></li>
-          <li><a class="nav-link" href="https://circleci.com/privacy/"
+          <li><a class="nav-link" href="https://circleci.com/legal/privacy/"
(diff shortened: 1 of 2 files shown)


devinmarieb (approved):

devinmarieb:
@circleci/docs-disco Can you also approve/verify this?

devinmarieb:
@circleci/docs-disco Just an FYI in case anything else needs changing regarding this

Mayuko-Sa (approved):

j2em6h:
Looks like this might actually be coming from https://github.com/circleci/docs-devhub-ui-lib/blob/main/src/internal/translations/footer.json

j2em6h:
> Ah looks like I don't have write access to that.

I don't either.

chantelle11:
It seems like this needs to be addressed in another PR. The current code highlighted is deprecated.

PR #7489, opened 2022-10-20
update tool used for api v2 gen

rosieyohannan:
update text in config.yml to indicate tool used to generate api v2 docs


devinmarieb (approved):

chantelle11 (approved):
Good catch!

PR #7549, opened 2022-11-09
Describe how to use BATS

Peter-Darton-i2:
# Description
Adds a section that describes how to use BATS (the unit-test framework recommended by CircleCI for testing scripts in orbs).

# Reasons
This is non-trivial because CircleCI can't ingest the format output by BATS natively and the CircleCI "bats" orb doesn't do it either (see bats orb [issue 4](https://github.com/CircleCI-Public/BATS-orb/issues/4)), leaving users without a good example/easy path to follow without this documentation.
...and this is important because BATS is highlighted as "the" way of testing scripts in orbs, and orb development is core to CircleCI.

# Content Checklist

### DIFF
--- a/jekyll/_cci2/collect-test-data.md
+++ b/jekyll/_cci2/collect-test-data.md
@@ -606,6 +606,33 @@ A working `.circleci/config.yml` section for testing might look like the followi
           path: build/out.xml
 ```
 
+### BATS for Bash
+{: #bats-for-bash }
+
+While CircleCI can't ingest the TAP data created by [BATS](https://bats-core.readthedocs.io/) directly, you can post-process the data into a format that can be ingested.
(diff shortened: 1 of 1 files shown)


devinmarieb:
@Peter-Darton-i2 Thank you for this! I will double check the examples with the Orbs team 🙌

Peter-Darton-i2:
This is one of those situations where the documentation is "a bit ugly" because the UX is (currently) "a bit ugly" - we can (and should) document that to make the ugliness as painless as possible...

... but this should also be a wake-up call to get the functionality "de-uglified" so that the resulting documentation can be updated to say "Yes, this is (now) easy".
(and, if the functionality is changed to make the UX nice & easy then I'd be happy to make another PR simplifying the docs 😁)

devinmarieb:
Hey @Peter-Darton-i2, sorry for the wait. The team is actually working on updating the orb to support this. I was not given a timeframe, but it is being worked on.

Peter-Darton-i2:
No worries; updating the orb to make it so easy to use that it barely needs documenting at all is, IMO, a _much_ better option than having documentation explain a gnarly workaround 😁

I'll keep an eye on orb progress (via support) and, if/when that comes to fruition, I'll update this PR (or close it, if the orb team update the docs as part of that work).

Peter-Darton-i2:
@devinmarieb FYI [circleci/bats@1.1.0](https://circleci.com/developer/orbs/orb/circleci/bats?version=1.1.0) has been released.
I've updated my PR to take advantage of it (it's now _really_ simple 👍).

WDYT?

devinmarieb:
@Peter-Darton-i2 Nice thank you for staying on top of this! I just made some minor changes to keep with the style we use in docs wording 👍

devinmarieb (approved):

PR #7724, opened 2023-02-01
remove redirect to discuss

sebastian-lerner:
we got feedback from a couple of large customers that the redirect to Discuss for pricing makes it feel not official and ephemeral.  can we just put the pricing in the actual docs page?


devinmarieb:
@sebastian-lerner No, we don't put pricing in docs. Official pricing information needs to live in Outer somewhere. If there is somewhere in Outer it lives we can link to that.

sebastian-lerner:
it's already in the docs: https://circleci.com/docs/faq/#how-do-I-calculate-my-monthly-IP-ranges-costs

can we re-evaluate our policy for not putting pricing in docs?  as a customer, it's super confusing why we wouldn't just have the pricing right there.  

i thought the logic for not having it in the docs was because pricing changes, this hasn't changed in 1.5 years

devinmarieb:
@sebastian-lerner Yeah this one is fine then. But in general, no we won't be adding pricing and credit usage to feature pages in docs. There is a reason we took them out to begin with and those issues have not been solved (those issues lie outside of the docs team). We do have somewhat of a plan for Billing stuff when we do our content re-org, but it's not final yet.

devinmarieb (approved):

PR #7743, opened 2023-02-06
Move glossary/config ref

devinmarieb:
# Description
- Move glossary to About
- Move Config ref to Dev Toolkit, remove all duplicate entries

# Reasons
The glossary was overlooked in the re-org, and we are starting to work on removing the Reference section completely.

# Content Checklist

### DIFF
--- a/jekyll/_data/sidenav.yml
+++ b/jekyll/_data/sidenav.yml
@@ -452,12 +452,12 @@ en:
           link: how-to-use-the-circleci-local-cli
     - name: Reference
       children:
+        - name: Configuration reference
+          link: configuration-reference
         - name: API v2 reference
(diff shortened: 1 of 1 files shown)


jekyll/_data/sidenav.yml
```diff
@@ -458,6 +456,8 @@ en:
           link: https://circleci.com/docs/api/v2
         - name: API v1 reference
           link: https://circleci.com/docs/api/v1
+        - name: Configuration reference
```
kbatuigas, line 459: Should we consider moving this up before API v2 reference? I guess mostly because there's already an earlier API section and the config reference could be a bit more visible within this particular section..?

PR #7758, opened 2023-02-10
Start to sync JA sidenav with EN

kbatuigas:
# Description
This PR is to sync the JA side nav with the EN version as closely as possible.

The sidenav reorg is an ongoing project and while there are still some sections that are being worked on, some new sections have been created and pages rearranged in this new architecture.

Where a localized section or subsection title, or page title, is available, we have included these. 

Before:
image

After:
image

# Reasons
We would like to more closely sync the JA side nav at this time instead of waiting for the entire sidenav reorg project to be complete (which will likely take a few more weeks). That way we can also minimize the amount of work that the localization team has to do.

# Content Checklist


jekyll/_data/sidenav.yml
```diff
@@ -1301,25 +1250,80 @@ ja:
...
           link: ja/docker-compose
           i18n_name: docker-composeの使用
+        - name: Rename organizations and repositories
+          link: ja/rename-organizations-and-repositories
+          i18n_name: 組織名とリポジトリ名の変更
         - name: Migrate from deploy to run
           link: ja/migrate-from-deploy-to-run
           i18n_name: deployからrunへの移行
 
+- name: Optimizations
```
rosieyohannan, line 1269: @kbatuigas is there a reason we're not using the same section titles as the English here? (Also Security rather than Security and Permissions and maybe some others). I'm having trouble reviewing properly because I still can't build!

kbatuigas: @rosieyohannan ah, yes - in cases like these I just left it at the existing/"old" EN equivalent because I wasn't sure that the JA translation would be the exact same (Optimizations vs Optimize). I haven't requested actual translation work for this PR, but just a quick check that the section titles still look OK and mostly match up. Figured it would make sense to put in the localization request only when the reorg for EN is complete?

rosieyohannan (approved):

PR #7764, opened 2023-02-13
Prepare sidenav template and data file for translation via Crowdin

tfe:
This restructures the `sidenav.yml` data file to be compatible with Crowdin for translations, and updates consumers of that data file (sidebar.html and mobile-sidebar.html) to reflect the new structure.

Prior to this, the sidenav data file had a couple issues. First was that both languages were stored in the same file, in the form of not only a top level `en`/`ja` key, but also parallel keys in the data structure for e.g. `i18n_name`.

The other issue that's important for Crowdin compatibility is avoiding use of arrays/lists in favor of hashes with named keys. This is because Crowdin needs a definitive address to identify a particular string in the file, and given a list it will just use the index position. If someone adds or moves an item in the list, the address for all subsequent translations will be off by one. 

Example:

```yaml
children:
  - name: What is continuous integration?
    link: about-circleci
  - name: Benefits of CircleCI
    link: benefits-of-circleci
  - name: Concepts
    link: concepts
  - name: Glossary
    link: glossary
  - name: Web app introduction
    link: introduction-to-the-circleci-web-app
  - name: FAQ
    link: faq
```

Given a structure like this, Crowdin will address the name string “Concepts” as `children -> 2 -> name`. You can see how this will break down if "Concepts" was removed, and "Glossary" became the 2nd item... we'd have a mis-translation with Crowdin using the Japanese translation for "Concepts" for what is now "Glossary".

Instead we need to always use structures like this for Crowdin-translated YAML:

```yaml
children:
  What is continuous integration:
    name: What is continuous integration?
    link: about-circleci
  Benefits of CircleCI:
    name: Benefits of CircleCI
    link: benefits-of-circleci
  Concepts:
    name: Concepts
    link: concepts
  Glossary:
    name: Glossary
    link: glossary
  Web app introduction:
    name: Web app introduction
    link: introduction-to-the-circleci-web-app
  FAQ:
    name: FAQ
    link: faq
```

Now the address for “Concepts” will be `children -> Concepts -> name`. The actual key we use for `Concepts` doesn’t matter, except that it’s more descriptive than the index `2` we had before, and not subject to getting mixed up as the list is modified in the future. For the purposes of this commit, I just took the value for `name` and reused that as the key, but it really could be any descriptive unique (within that hash) string. Fortunately YAML is very flexible and forgiving in terms of valid keys, so almost any names work for these.

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -16,8 +16,6 @@ files:
     translation: /jekyll/_cci2_%two_letters_code%/server/overview/%original_file_name%
   - source: /jekyll/_data/sidenav.yml
     translation: /jekyll/_data/%two_letters_code%/sidenav.yml
-  - source: /jekyll/_data/sidenavstyle.yml
-    translation: /jekyll/_data/%two_letters_code%/sidenavstyle.yml
   - source: /jekyll/_includes/edit-on-github.html
     translation: /jekyll/_includes/%two_letters_code%/edit-on-github.html
   - source: /jekyll/_includes/license-notice.html
--- a/jekyll/_data/sidenav.yml
+++ b/jekyll/_data/sidenav.yml
@@ -1,2083 +1,1342 @@
-en:
-- name: About CircleCI
+About CircleCI:
+  name: About CircleCI
   icon: icons/sidebar/circle-logo.svg
   children:
-    - name: Overview
+    Overview:
+      name: Overview
       children:
-        - name: What is continuous integration?
+        What is continuous integration?:
+          name: What is continuous integration?
           link: about-circleci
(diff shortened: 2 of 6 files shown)


felicianotech:
@dre thoughts on this? This is older and there's conflicts. Should this be moved to draft? Something else?

rosieyohannan:
> @circleci/dre thoughts on this? This is older and there's conflicts. Should this be moved to draft? Something else?

This was something we were going to do once th JA and EN sections had been split into separate files. But the formatting change was to help with the Crowdin integration for these files so I think this can just be closed now.

PR #7802, opened 2023-02-28
docs: xarg typo in jest fail fast example

alejandrovelez7:
# Description
The jest example using `--fail-fast` uses `xarg` instead of `xargs`. `xarg` is not a valid bash command afaik. 

# Reasons
Simple typo I ran across after copy pasting this code 😅


rosieyohannan:
Thank you @alejandrovelez7! This actually just got fixed in another PR. Thank you so much for pointing this out 🎉

PR #7858, opened 2023-03-14
Fixing the layout of built-in-env-vars.adoc

georgekronberg:
# Description
The value for `VCS` column is missing for this row, which breaks the layout for all rows below. Not sure if `GitHub, Bitbucket, GitLab` is the correct list of supported VCSs, so please adjust if it's incorrect.


kbatuigas:
Thanks for catching this, @georgekronberg

kbatuigas (approved):

Issue #7729, opened 2023-02-02
Using docker executor, can non-primary containers communicate with each other?


thomasmatecki:
# Missing Information
If using multiple containers with the docker executor, can non-primary containers communicate with each other?

Docs currently state:
> It is possible to specify multiple images for your job. Specify multiple images if, for example, you need to use a database for your tests or for some other required service. All containers run in a common network and every exposed port will be available on localhost from a [primary container](https://circleci.com/docs/glossary/#primary-container).

If I am running cypress tests on an application that has a static frontend, an  API server, a database and also some other stuff (let's say Redis).

```
Cypress (chrome) | -> API container (e.g. django) | -> Database Container
                 |                                | -> Redis Container
                 | -> Static UI (e.g React)

```

... I need a least two containers that can access another container, so network connectivity from only the primary container doesn't really work.

If using docker compose (for example), this is accomplished by using services names as if they are hostnames. The docs don't say that I can do that though. Would that work?

# Already Looked
[Using multiple Docker images](https://circleci.com/docs/using-docker/#using-multiple-docker-images)

# Potential Locations
Same as above: [Using multiple Docker images](https://circleci.com/docs/using-docker/#using-multiple-docker-images)

rosieyohannan:
Thanks for your question! I have checked with engineering and they explained that yes, all containers can talk to each other. When using multiple Docker containers for a job, the containers share a cgroups network namespace. So, in terms of networking, it's like every container shares the same localhost. They also share the same process namespace, so each can see the other's processes, too. Hopefully this is helpful, please let me know either way and I can try and expand on this.

rosieyohannan:
Yes, running two services on the same port would cause the pipeline to fail.

Let me know how you get on, and yes I'll make an update to the page to clarify.

PR #7883, opened 2023-03-23
Docker multi container setup clarifications

### LINKED ISSUE #7800: Docs unclear if we can name docker containers
# Location
A link to the document that is unclear.

https://circleci.com/docs/configuration-reference/#docker

# Explanation
Describe what confused you about the document.
This could be syntax, word choice, sentence length, or something entirely different.

The docs for `name` say:

> name defines the name for reaching the secondary service containers. By default, all services are exposed directly on localhost. The field is appropriate if you would rather have a different host name instead of localhost, for example, if you are starting multiple versions of the same service.

I would really expect `name` to just be the name to use for the docker container, i.e. the `--name` argument value passed to `docker run`.  If that's not what this configuration parameter does, then it should really be called something else like `hostname`.

rosieyohannan:
# Description
Fixes #7800 and #7729 

# Reasons
* Not clear that all containers can communicate with eachother when using multiple containers for a job
* Use of name rather than hostname for defining the `name` key was not clear

# Content Checklist

### DIFF
--- a/jekyll/_cci2/configuration-reference.md
+++ b/jekyll/_cci2/configuration-reference.md
@@ -695,7 +695,7 @@ Configured by `docker` key which takes a list of maps:
 Key | Required | Type | Description
 ----|-----------|------|------------
 image | Y | String | The name of a custom docker image to use. The first `image` listed under a job defines the job's own primary container image where all steps will run.
-name | N | String | `name` defines the the hostname for the container (default: localhost), used for reaching secondary, service containers. By default, all services are exposed directly on `localhost`. This field is appropriate if you would rather have a different hostname instead of `localhost`, for example, if you are starting multiple versions of the same service.
+name | N | String | `name` defines the the hostname for the container (the default is `localhost`), used for reaching secondary, service containers. By default, all services are exposed directly on `localhost`. This field is appropriate if you would rather have a different hostname instead of `localhost`, for example, if you are starting multiple versions of the same service.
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/configuration-reference.md
```diff
@@ -695,7 +695,7 @@ Configured by `docker` key which takes a list of maps:
 Key | Required | Type | Description
 ----|-----------|------|------------
 image | Y | String | The name of a custom docker image to use. The first `image` listed under a job defines the job's own primary container image where all steps will run.
-name | N | String | `name` defines the name for reaching the secondary service containers.  By default, all services are exposed directly on `localhost`.  The field is appropriate if you would rather have a different host name instead of localhost, for example, if you are starting multiple versions of the same service.
+name | N | String | `name` defines the the hostname for the container (default: localhost), used for reaching secondary, service containers. By default, all services are exposed directly on `localhost`. This field is appropriate if you would rather have a different hostname instead of `localhost`, for example, if you are starting multiple versions of the same service.
```
rosieyohannan, line 698:
```suggestion
name | N | String | `name` defines the the hostname for the container (the default is `localhost`), used for reaching secondary, service containers. By default, all services are exposed directly on `localhost`. This field is appropriate if you would rather have a different hostname instead of `localhost`, for example, if you are starting multiple versions of the same service.
```

PR #7944, opened 2023-04-14
save_cache example can cache sensitive information

nanophate:
# Description

Changed the `save_cache` to only save the necessary cache. Support got a report that it can include sensitive information that should not be leaked.

# Reasons

Customer shared that we cache ` ~/.gradle` in our documentation which can include `~/.gradle/gradle.properties` from being leaked as it can contain sensitive information. Also for `~/.m2` it can include `~/.m2/settings.xml` and also can be leaked.

# Content Checklist


rosieyohannan (approved):
Thank you @nanophate!

PR #7975, opened 2023-04-26
Updates to reflect new way DLC works

rosieyohannan:
# Description
* general refresh of mentions of DLC
* Update references to how DLC works to be accurate

# Reasons
https://discuss.circleci.com/t/setup-remote-docker-architecture-change/45303

# Content Checklist


jekyll/_cci2/docker-layer-caching.md
```diff
@@ -66,52 +93,20 @@ DLC has **no** effect on Docker images used as build containers. That is, contai
...
 
-The DLC volumes are deleted after 3 days of not being used in a job.
+When none of the layers in the image change between job runs, DLC pulls the layers from the cache and reuses those instead of rebuilding the entire image.
 
-CircleCI will create a maximum of 30 DLC volumes per project, so a maximum of 30 concurrent `machine` or Remote Docker jobs per project can have access to DLC. This takes into account the parallelism of the jobs, so a maximum of 1 job with 30x parallelism will have access to DLC per project, or 2 jobs with 15x parallelism, and so on.
+If part of the Dockerfile changes (which changes part of the image), a subsequent run of the exact same job with the modified Dockerfile may still finish faster than rebuilding the entire image. This is because the cache can still be used for the first few steps that did not change in the Dockerfile. The steps that follow the change must be rerun because the Dockerfile change invalidates the cache for those layers.
 
-![Docker Layer Caching](/docs/assets/img/docs/dlc_cloud.png)
```
rosieyohannan, line 79: I removed the image but atually it might be worth just updating "Volume" to "Cache" as it's still largely correct I think: https://circleci.com/docs/docker-layer-caching/#how-dlc-works

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,58 +1,85 @@
...
 contentTags:
   platform:
   - Cloud
   - Server v4.x
   - Server v3.x
 ---
 
-Docker layer caching (DLC) can reduce Docker image build times on CircleCI. DLC is available on
-the [Free and above](https://circleci.com/pricing/) usage plans (credits are charged per run job) and on installations of [CircleCI server](https://circleci.com/enterprise/).
+Use Docker layer caching (DLC) to reduce Docker image build times on CircleCI. DLC is available on all CircleCI plans, and credits are charged per job run.
```
sebastian-lerner, line 12: do you wanna link to where they can see the credits per job run?

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,58 +1,85 @@
...
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+```
+
+Every layer built in a previous job will be accessible in the Remote Docker Environment. However, in some cases your job may run in a clean environment, even if the configuration specifies `docker_layer_caching: true`.
 
-Please note that high usage of [parallelism](/docs/configuration-reference/#parallelism) (that is, a parallelism of 30 or above) in your configuration may cause issues with DLC, notably pulling a stale cache or no cache.  For example:
+If you run many concurrent jobs for the same project that depend on the same environment, all of them will be provided with a Remote Docker environment. Docker layer caching guarantees that jobs will have exclusive Remote Docker Environments that other jobs cannot access. However, some of the jobs may have cached layers, some may not have cached layers, and not all of the jobs will have identical caches.
```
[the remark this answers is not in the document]
sebastian-lerner: i think we can get rid of this.... @brentmmarks / @joshhsoj1902 is this relevant still?

jekyll/_cci2/docker-layer-caching.md
sebastian-lerner: @brentmmarks / @joshhsoj1902 is the image still accurate if we change "volume" to "cache"

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -66,52 +93,20 @@ DLC has **no** effect on Docker images used as build containers. That is, contai
...
-Docker layer caching can also reduce job runtimes when building Docker images using the [`machine` executor](/docs/configuration-reference/#machine). Use DLC with the `machine` executor by adding `docker_layer_caching: true` below your `machine` key (as seen above in our [example](#configyml)):
-
-```yml
-machine:
-  image: ubuntu-2004:202104-01  # any available image
-  docker_layer_caching: true    # default - false
-```
+With DLC enabled, the entirety of `/var/lib/docker` is cached, which also includes any custom networks created in previous jobs.
 
 ## Examples
```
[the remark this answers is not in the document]
sebastian-lerner: we can get rid of it

sebastian-lerner:
adding @brentmmarks and @joshhsoj1902 to this review as well.  

brent, any chance when youre back next week you can take a look and see if anything is off?

bryan, i still added you for visibility even though youre out next week

jekyll/_cci2/docker-layer-caching.md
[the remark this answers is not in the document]
brentmmarks: Parallelism is no longer restricted to amount of caches

jekyll/_cci2/docker-layer-caching.md
brentmmarks: Yes and no. It will all be the same cache if all the jobs are started at the same time. Job run 3 in the docs is no longer accurate since we no longer store separate caches.

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,58 +1,85 @@
...
+    docker:
+      - image: cimg/base:2023.04
+
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+```
+
+Every layer built in a previous job will be accessible in the Remote Docker Environment. However, in some cases your job may run in a clean environment, even if the configuration specifies `docker_layer_caching: true`.
```
brentmmarks, line 45: This to me makes it feel like we are making a promise to customer that we will keep all layers for the customer which is not true. We do pruning on layers to reduce the size of cache and try our best to store as much useful pieces as possible. We did this in old world too. cc @sebastian-lerner

jekyll/_cci2/docker-layer-caching.md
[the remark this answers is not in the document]
brentmmarks: And if layer is in cache all builds will get access to same cache so this can be removed.

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -66,52 +93,20 @@ DLC has **no** effect on Docker images used as build containers. That is, contai
 ## How DLC works
 {: #how-dlc-works }
 
-DLC caches your Docker image layers by creating an external volume and attaching it to the instances that execute the `machine` and Remote Docker jobs. The volume is attached in a way that makes Docker save the image layers on the attached volume. When the job finishes, the volume is disconnected and re-used in a future job. The layers downloaded in a previous job with DLC will be available in the next job that uses the same DLC volume.
-
-One DLC volume can only be attached to one `machine` or Remote Docker job at a time. If one DLC volume exists but two jobs that request DLC are launched, CircleCI will create a new DLC volume and attach it to the second job. From that point on, the project will have two DLC volumes associated with it. This applies to parallel jobs as well: if two `machine` jobs are run in parallel, they will get different DLC volumes.
+DLC caches your Docker image layers using a sparse file within the container/virtual machine used to run your job. The same applies to jobs that use parallelism. If two `machine` jobs are run in parallel, each VM will have a separate DLC cache.
```
brentmmarks, line 96: Just want to make sure it is clear they will have separate cache but whichever is uploaded last will be stored for next build.

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -66,52 +93,20 @@ DLC has **no** effect on Docker images used as build containers. That is, contai
...
-{: #machine-executor }
-
-Docker layer caching can also reduce job runtimes when building Docker images using the [`machine` executor](/docs/configuration-reference/#machine). Use DLC with the `machine` executor by adding `docker_layer_caching: true` below your `machine` key (as seen above in our [example](#configyml)):
-
-```yml
-machine:
-  image: ubuntu-2004:202104-01  # any available image
-  docker_layer_caching: true    # default - false
-```
+With DLC enabled, the entirety of `/var/lib/docker` is cached, which also includes any custom networks created in previous jobs.
```
brentmmarks, line 109: We prune containers and networks so dont think we need the part about networks.

jekyll/_cci2/docker-layer-caching.md
sebastian-lerner: i think let's just get rid of the image...

jekyll/_cci2/docker-layer-caching.md
sebastian-lerner: can we get rid of this sentence too @rosieyohannan ?

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,58 +1,85 @@
...
-{: class="alert alert-info"}
+## Quickstart
+{: #quickstart }
 
-### Limitations
-{: #limitations }
+### Remote Docker environment
+{: #remote-docker-environment }
+
+To use DLC with the Docker execution environment, you will need to configure your job to run in the Remote Docker Environment. To do this, add `docker_layer_caching: true` under the `setup_remote_docker` key in your [`.circleci/config.yml`](/docs/configuration-reference/) file:
```
kbatuigas, line 29: "a remote Docker environment"?

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -66,52 +93,20 @@ DLC has **no** effect on Docker images used as build containers. That is, contai
...
-
-One DLC volume can only be attached to one `machine` or Remote Docker job at a time. If one DLC volume exists but two jobs that request DLC are launched, CircleCI will create a new DLC volume and attach it to the second job. From that point on, the project will have two DLC volumes associated with it. This applies to parallel jobs as well: if two `machine` jobs are run in parallel, they will get different DLC volumes.
+DLC caches your Docker image layers using a sparse file within the container/virtual machine used to run your job. The same applies to jobs that use parallelism. If two `machine` jobs are run in parallel, each VM will have a separate DLC cache.
 
-Depending on which jobs the volumes are used in, they might end up with different layers saved on them. For instance, the volumes that are used less frequently might have older layers saved on them.
+If, for example, the first run of your job takes over two minutes to build a Docker image, and nothing changes in the Dockerfile before the second run, the Docker file build steps happen instantly, in zero seconds.
```
kbatuigas, line 98: "Docker file" -> "Dockerfile"?

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -254,39 +249,12 @@ However, because our `#install jq` step is new, it as well as subsequent steps n
...
-
-This means that if you change something in the Dockerfile, all of those later steps are invalidated and the layers have to be rebuilt. When some of the steps remain the same (the steps before the one you removed), those steps can be reused. So, it is still faster than rebuilding the entire image.
+DLC was previously enabled via the `reusable: true` key. The `reusable` key has been deprecated in favor of the `docker_layer_caching` key.
 
-<div class="video-wrapper">
-  <iframe width="560" height="315" src="https://www.youtube.com/embed/AL7aBN7Olng" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen></iframe>
-</div>
+In addition, the `exclusive: true` option is deprecated and all Remote Docker VMs are now treated as exclusive. This means that when using DLC, jobs are guaranteed to have an exclusive Remote Docker environment that other jobs cannot access.
```
kbatuigas, line 257:
```suggestion
In addition, the `exclusive: true` option is deprecated and all remote Docker VMs are now treated as exclusive. This means that when using DLC, jobs are guaranteed to have an exclusive Remote Docker environment that other jobs cannot access.
```

jekyll/_cci2/glossary.adoc
```diff
@@ -30,7 +30,7 @@ A container is a running instance of an image.
 [#docker-layer-caching]
 == Docker layer caching
 
-The CircleCI Docker Layer Caching feature allows builds to reuse Docker image layers from previous builds. Image layers are stored in separate volumes in the cloud, and are not shared between projects. Layers may only be used by builds from the same project. For additional information about how the secure environment is created, see the <<docker-layer-caching#,Docker Layer Caching>> and <<building-docker-images#,Running Docker Commands>> documents.
+The CircleCI Docker Layer Caching (DLC) feature allows builds to reuse Docker image layers from previous builds. Use DLC to speed up workflows. Jobs that are configured to build docker images run in a shorter time on subsequent runs. Layers may only be used by builds from the same project. For more information and usage examples, see the xref:docker-layer-caching#[Docker Layer Caching] and xref:building-docker-images#[Running Docker Commands] pages.
```
kbatuigas, line 33:
```suggestion
The CircleCI Docker Layer Caching (DLC) feature allows builds to reuse Docker image layers from previous builds. Use DLC to speed up workflows. Jobs that are configured to build Docker images run in a shorter time on subsequent runs. Layers may only be used by builds from the same project. For more information and usage examples, see the xref:docker-layer-caching#[Docker Layer Caching] and xref:building-docker-images#[Running Docker Commands] pages.
```

sebastian-lerner:
@rosieyohannan actually one more thing...where we refer to a "sparse file", i think we can probably eliminate that implementation detail from the docs? it's not critical to how the feature works from the customer's POV

jekyll/_cci2/docker-layer-caching.md
rosieyohannan: I found this FAQ that we could link to, but this content might also need some updates: the 200 credits for DLC, and the second paragraph. https://circleci.com/docs/credits/#charge-for-docker-layer-caching Any thoughts @sebastian-lerner ?

jekyll/_cci2/docker-layer-caching.md
rosieyohannan: OK yes I'll make this clear with a subsection for DLC with parallelism

jekyll/_cci2/docker-layer-caching.md
sebastian-lerner: sorry i missed this, i was thinking https://circleci.com/pricing/

sebastian-lerner (approved):
LGTM

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,293 +1,124 @@
...
-      - checkout
-      - setup_remote_docker:
-          docker_layer_caching: true
-      - run: docker build .
-```
+### Scope of cache
+{: #scope-of-cache }
 
-When none of the layers in the image change between job runs, DLC pulls the layers from cache from the image that was built previously and reuses those instead of rebuilding the entire image.
+With DLC enabled, the entirety of `/var/lib/docker` is cached.
```
joshhsoj1902, line 112: I'm wondering if this is slightly misleading.

It is `/var/lib/docker` that we backup, but we do clean it up first (for example we wont backup volumes created during the job run). We've seen cases where customers have tried to store their own files in `/var/lib/docker`, this is not supported and depending on how they do it can corrupt the cache resulting in us throwing away the whole cache.

sebastian-lerner: do we need to include anything about /var/lib/docker? those sound like implementation details that may change in the future?

joshhsoj1902: That's sorta my feeling too. If we don't need to say it then we shouldn't.

sebastian-lerner: @rosieyohannan i think let's yank it out

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,293 +1,124 @@
...
+  browser-tools: circleci/browser-tools@1.2.3
+
 jobs:
-  build_elixir:
-    machine:
-      image: ubuntu-2004:202104-01
-      docker_layer_caching: true
+ build:
+    docker:
+      - image: cimg/node:17.2-browsers # DLC does nothing here, its caching depends on commonality of the image layers.
```
joshhsoj1902, line 77: I want to test this, I think technically this image would be cached, but I would like to confirm that before I say more. I'm leaning toward just removing this comment and not go out of our way to say that we do cache it.

I'll test this in the next hour and report back

joshhsoj1902: My tests here were somewhat inconclusive. These images were being pruned. but I'm not sure why... I might loop back to play with it more, but I feel like removing this comment may be best.

sebastian-lerner: kk, thanks josh. @rosieyohannan let's remove the comment

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,293 +1,124 @@
...
+      docker_layer_caching: true    # default - false
 
-### Config.yml
-{: #configyml }
+    steps:
+      - checkout
+```
 
-In the `.circleci/config.yml` snippet below, let's assume the `build_elixir` job is regularly building an image using the above Dockerfile.
+## Limitations
```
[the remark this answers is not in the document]
rosieyohannan: Thanks for this @joshhsoj1902! Just one question - where you say "Any customer who wants to use "buildx builder instances" with DLC should name their builder", why is that? Just wondering if I can add a reason too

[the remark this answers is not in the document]
joshhsoj1902: Hey @rosieyohannan. If a name isn't set, in the background docker will generate a random name for it (for example `trusting_elion`), and that generated name is also used in the docker volume name that that the builder instance uses.

As is if a customer doesn't set a name they would end up with a new unique docker volume created everytime they run the job and it wouldn't be very useful for them. So we will automatically clean those up.

But if a customer does properly name their builder instance, we can detect that and preserve the docker volume, then in their next job run when then create a builder instance with the same name it'll attach to the existing volume.

If you need any clarification on any of that let me know!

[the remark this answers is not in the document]
rosieyohannan: Thanks, interesting!

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,293 +1,128 @@
...
+          docker_layer_caching: true # DLC will explicitly cache layers here and try to avoid rebuilding.
+      - run: docker build .
 ```
 
-On subsequent commits, if our example Dockerfile has not changed, then DLC will pull each Docker image layer from cache during the `build Elixir image` step, and our image will theoretically build almost instantaneously.
+DLC has **no** effect on Docker images used as build containers. That is, containers that are used to _run_ your jobs are specified with the `image` key when using the [`docker` executor](/docs/using-docker/) and appear in the **Spin up Environment** step on your jobs pages.
+{: class="alert alert-info"}
 
-Now, let's say we add the following step to our Dockerfile, in between the `# use unicode` and `# install docker` steps:
+### Buildx builder instances
```
rosieyohannan, line 92: @joshhsoj1902 I've added a section here if you could take a look, also @sebastian-lerner

joshhsoj1902: That looks good to me. If support ends up getting questions regarding the pruning we can expand a little and explain how to configure the buildkit purning/garbage collection, but that feels outside the scope of DLC and I don't expect anyone to really need to use that.

joshhsoj1902 (approved):

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,293 +1,129 @@
...
+{: #buildx-builder-instances }
 
-```dockerfile
-# Install jq
-RUN JQ_URL="https://circle-downloads.s3.amazonaws.com/circleci-images/cache/linux-amd64/jq-latest" \
-  && curl --silent --show-error --location --fail --retry 3 --output /usr/bin/jq $JQ_URL \
-  && chmod +x /usr/bin/jq \
-  && jq --version
-```
+When using buildx builder instances with DLC it is important to [name your builders](https://docs.docker.com/engine/reference/commandline/buildx_create/#name). Doing so ensures CircleCI can detect and preserve the Docker volumes for subsequent job runs. If you no not name your builders, each time the job is run they will have different, randomly generated names, and the resulting volumes will be automatically cleaned up.
```
kbatuigas, line 95:
```suggestion
When using buildx builder instances with DLC, it is important to [name your builders](https://docs.docker.com/engine/reference/commandline/buildx_create/#name). Doing so ensures CircleCI can detect and preserve the Docker volumes for subsequent job runs. If you do not name your builders, each time the job is run they will have different, randomly generated names, and the resulting volumes will be automatically cleaned up.
```

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,293 +1,129 @@
...
 
-On the next commit, DLC will ensure that we still get cached image layers for the first few steps in our Dockerfile—pulling from `elixir:1.11.4` as our base image, the `# make apt non-interactive` step, the step starting with `RUN apt-get update`, the `# set timezone to UTC` step, and the `# use unicode` step.
+If you store Docker build artifacts in a Docker volume, managed by the buildkit inside buildx builder instances, the DLC feature cannot _maintain_ these artifacts, but they can still be supported. DLC is not able to prune these images/buildcache, but buildx builders do have some in-built pruning. For more information, see the [Docker docs](https://docs.docker.com/build/cache/garbage-collection/#default-policies).
```
kbatuigas, line 97: "buildcache" -> "build cache"?

kbatuigas (changes requested):

jekyll/_cci2/docker-layer-caching.md
```diff
@@ -1,293 +1,129 @@
...
-  && jq --version
-```
+When using buildx builder instances with DLC, it is important to [name your builders](https://docs.docker.com/engine/reference/commandline/buildx_create/#name). Doing so ensures CircleCI can detect and preserve the Docker volumes for subsequent job runs. If you do not name your builders, each time the job is run they will have different, randomly generated names, and the resulting volumes will be automatically cleaned up.
 
-On the next commit, DLC will ensure that we still get cached image layers for the first few steps in our Dockerfile—pulling from `elixir:1.11.4` as our base image, the `# make apt non-interactive` step, the step starting with `RUN apt-get update`, the `# set timezone to UTC` step, and the `# use unicode` step.
+If you store Docker build artifacts in a Docker volume, managed by the buildkit inside buildx builder instances, the DLC feature cannot _maintain_ these artifacts, but they can still be supported. DLC is not able to prune these images/buildcache, but buildx builders do have some in-built pruning. For more information, see the [Docker docs](https://docs.docker.com/build/cache/garbage-collection/#default-policies).
```
rosieyohannan, line 97:
```suggestion
If you store Docker build artifacts in a Docker volume, managed by the buildkit inside buildx builder instances, the DLC feature cannot _maintain_ these artifacts, but they can still be supported. DLC is not able to prune these images/build cache, but buildx builders do have some in-built pruning. For more information, see the [Docker docs](https://docs.docker.com/build/cache/garbage-collection/#default-policies).
```

kbatuigas (approved):
🎉

PR #7992, opened 2023-05-03
Update gitlab-integration.adoc

sebastian-lerner:
probably could use a link to test insights but i always get so confused with how to add links in that i just left it out for now...https://circleci.com/docs/insights-tests/


rosieyohannan (approved):
Thanks for this!

PR #8159, opened 2023-06-29
Update rerun-failed-tests-only.adoc

sebastian-lerner:
move to known limitations to make more obvious


jekyll/_cci2/rerun-failed-tests-only.adoc
```diff
@@ -247,7 +247,7 @@ The snippet above will write the list of test file names to `files.txt`.  On a n
 [#known-limitations]
 == Known limitations
 
-* When re-running only the failed tests, test splitting by timing may not be as efficient as expected the next time that job runs, as the test results being stored are only from the subset of failed tests that were run.
+* If your testing job is using parallelism and test splitting, the job will spin up the number of containers/virtual machines (VMs) that are specified with the `parallelism` key. However, the step that runs tests for each of those parallel containers/VMs will only run a subset of tests, or no tests, after the tests are split across the total number of parallel containers/VMs. For example, if parallelism is set to eight, there may only be enough tests after the test splitting occurs to "fill" the first parallel container/VM. The remaining seven containers/VMs will still start up, but they will not run any tests when they get to the test execution step.  **The vast majority of customers still observe substantial time & credit savings** despite spinning up containers/VMs that do not run tests.
```
rosieyohannan, line 250:
```suggestion
* If your testing job is using parallelism and test splitting, the job will spin up the number of containers/virtual machines (VMs) that are specified with the `parallelism` key. However, the step that runs tests for each of those parallel containers/VMs will only run a subset of tests, or no tests, after the tests are split across the total number of parallel containers/VMs. For example, if parallelism is set to eight, there may only be enough tests after the test splitting occurs to "fill" the first parallel container/VM. The remaining seven containers/VMs will still start up, but they will not run any tests when they get to the test execution step.  **The vast majority of customers still observe substantial time and credit savings** despite spinning up containers/VMs that do not run tests.  
```

rosieyohannan (approved):

PR #8162, opened 2023-06-29
Update rerun-failed-tests-only.adoc

sebastian-lerner:
feature name is dropping "only"


sebastian-lerner:
also get rid of the warning that it won't go to GA.  we are planning on GA'ing on July 11

rosieyohannan (approved):

PR #8242, opened 2023-07-22
batched updates to rerun failed tests page

sebastian-lerner:
1. added in something that some customers are missing when it comes to using the halt command
2. added to output file names section that the halt command is probably needed to avoid a failure if using parallelism
3. added to troubleshooting section an error for an edge case that 3 ppl reported last week
4. add store_test_results to all examples since i've talked to a few customers who forgot to add that because they just copied from the docs and didn't read the prereqs

### DIFF
--- a/jekyll/_cci2/rerun-failed-tests.adoc
+++ b/jekyll/_cci2/rerun-failed-tests.adoc
@@ -268,7 +268,7 @@ NOTE: Ensure that you are using version 1.34.2 or later of Playwright. Earlier v
     ./gradlew test $GRADLE_ARGS
 
 - store_test_results:
-    ...
+    path: build/test-results/test
 ```
 
 . Update the `glob` command to match your use case. **If your current job is using xref:test-splitting-tutorial#[CircleCI's intelligent test splitting], you must change the `circleci tests split` command to `circleci tests run` with the `--split-by=timings` parameter.**. If you are not using test splitting, `--split-by=timings` can be omitted.
@@ -332,7 +332,7 @@ Django takes as input test filenames with a format that uses dots ("."), however
       pipenv run coverage run manage.py test --parallel=8 --verbosity=2 $TESTFILES
 
 - store_test_results:
-    ...
+    path: test-results
 ```
 
 [#known-limitations]
--- a/jekyll/_cci2/rerun-failed-tests.adoc
+++ b/jekyll/_cci2/rerun-failed-tests.adoc
@@ -311,7 +311,7 @@ Example:
     path: test-results
 ```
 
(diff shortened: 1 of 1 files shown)


felicianotech:
We should get to this on Monday. Thanks.

jekyll/_cci2/rerun-failed-tests.adoc
```diff
@@ -287,26 +299,41 @@ Example:
...
+    path: test-results
 ```
 
-The snippet above will write the list of test file names to `files.txt`.  On a non-rerun, this list will be all of the test file names.  On a "rerun", the list will be a subset of file names (the test file names that had at least 1 test failure in the previous run).  You can pass the list of test file names from `files.txt` into, for example, your custom `makefile`.
+The snippet above will write the list of test file names to `files.txt`.  On a non-rerun, this list will be all of the test file names.  On a "rerun", the list will be a subset of file names (the test file names that had at least 1 test failure in the previous run).  You can pass the list of test file names from `files.txt` into, for example, your custom `makefile`.  If using parallelism, CircleCI spins up the same number of containers/VMs as the parallelism level that is set in config.yml, however, not all parallel containers/VMs will execute tests.  For the parallel containers/VMs that will not run tests, files.txt may not be created.  The `halt` command ensures that in the case where a parallel run is not executing tests, the parallel run is stopped immediately.
```
rosieyohannan, line 314:
```suggestion
The snippet above will write the list of test file names to `files.txt`.  On a non-rerun, this list will be all of the test file names.  On a "rerun", the list will be a subset of file names (the test file names that had at least 1 test failure in the previous run).  You can pass the list of test file names from `files.txt` into, for example, your custom `makefile`.  If using parallelism, CircleCI spins up the same number of containers/VMs as the parallelism level that is set in `.circleci/config.yml`. However, not all parallel containers/VMs will execute tests.  For the parallel containers/VMs that will not run tests, `files.txt` may not be created.  The `halt` command ensures that in the case where a parallel run is not executing tests, the parallel run is stopped immediately.
```

jekyll/_cci2/rerun-failed-tests.adoc
```diff
@@ -257,6 +266,9 @@ NOTE: Ensure that you are using version 1.34.2 or later of Playwright. Earlier v
     echo "Prepared arguments for Gradle: $GRADLE_ARGS"
 
     ./gradlew test $GRADLE_ARGS
+
+- store_test_results:
+    ...
```
rosieyohannan, line 271: As the path is required can we include for all these examples?

sebastian-lerner: added.  for a couple of them it's gonna depend what the person specifies in another configuration file, but i think it's fine to put these values

sebastian-lerner: (that's why i left them out initially)

sebastian-lerner:
@rosieyohannan any chance you can give this a stamp when you get a chance?

PR #8253, opened 2023-07-26
batched updates to rerun failed tests docs

sebastian-lerner:
1. make clear that this is a new workflow, not just a new job that gets kicked off (this may need some work). basically it's not clear (and has come up with a couple of customers) that this is a brand new workflow just like "rerun workflow from failed", not just a job in isolation (like ssh rerun)
2. typo with whitespace
3. add elixir tests example
4. another typo
5. add that kotlin and gradle can use the same mechanism to run their tests with this feature
6. add phpunit example
7. I've had ~5 ppl this week reach out that they were confused why it wasn't working for them. They all forgot the `xargs` in their --command.  Several of them said it would be good to have an explicit reminder in the language-specific examples to use `xargs`.  I tend to agree so I added that in.

### DIFF
--- a/jekyll/_cci2/rerun-failed-tests.adoc
+++ b/jekyll/_cci2/rerun-failed-tests.adoc
@@ -137,6 +137,7 @@ gem 'rspec_junit_formatter'
     path: ~/rspec
 ```
 
+. Ensure you are using `xargs` in your `circleci tests run` command to pass the list of test files/classnames via stdin to `--command`.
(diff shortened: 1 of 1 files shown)


sebastian-lerner:
im actually gonna add one more example for phpunit that im just wrapping up right now.

sebastian-lerner:
okay php unit has been added, ready for review

rosieyohannan (approved):

PR #8298, opened 2023-08-18
Improve set-environment-variable sample code to reduce chance of errors during use 

legopin:
# Description
Improve the example code used for dynamically using shell script to set ENV.
Use double quotes when command outputs are intended to be loaded into ENV, so that the command is only executed once instead of every time a new shell is spawned.

# Reasons
This reduces the chances for errors to appear which complain about not finding resources in a different directory.

We suffered from this in our CI setup and was greatly puzzled why a non related script was erroring out when we did not call the script. 

This is the original sample code
```
echo 'export VERY_IMPORTANT=$(cat important_value)' >> "$BASH_ENV"
```

After executing the .bashrc file has this line, and if the important_value is actually a filename, then when starting a shell in a directory that doesn't have the file will result in blanks or errors
```
export VERY_IMPORTANT=$(cat important_value)
```

The better example should be using double quotes to evaluate the command first, then save the value in .bashrc
```
export VERY_IMPORTANT=VALUE_CONTENT
```

# Content Checklist

- [x ] Break up walls of text by adding paragraph breaks.


legopin:
Is there anything to change?

jekyll/_cci2/set-environment-variable.adoc
```diff
@@ -35,7 +35,7 @@ jobs:
           # available in current step.
           command: |
             echo 'export PATH=/path/to/foo/bin:"$PATH"' >> "$BASH_ENV"
-            echo 'export VERY_IMPORTANT=$(cat important_value)' >> "$BASH_ENV"
+            echo "export VERY_IMPORTANT=$(cat important_value)" >> "$BASH_ENV"
```
kbatuigas, line 38: Hi @legopin , so I've gotten feedback from our engineers and their advice is to actually leave out the bash command substitution (`$(...)`) out from examples entirely. So perhaps the better edit to do here would be something like what you indicated in the PR description e.g.
```
"export VERY_IMPORTANT=VALUE_CONTENT"
```
Does that make sense?

kbatuigas (commented):
Thank you @legopin , please see clarifying comment.

kbatuigas (approved):
Thank you, @legopin ! We'll go ahead and merge.

PR #8322, opened 2023-09-06
Document setup workflows and pipeline parameters

jkzilla:
# Description

# Reasons

# Content Checklist

### DIFF
--- a/jekyll/_cci2/pipeline-variables.adoc
+++ b/jekyll/_cci2/pipeline-variables.adoc
@@ -64,6 +64,8 @@ Pipeline parameters support the following types:
 
 See xref:reusing-config#parameter-syntax[Parameter syntax] for usage details.
 
+NOTE: When using setup workflows, all pipeline parameters must be present in the subsequent configuration, including in API calls. More information on setup workflows can be found in the xref:dynamic-config#how-dynamic-config-works[How dynamic config works] section of the Dynamic Configuration overview.
+
 The example below shows a configuration with two pipeline parameters (`image-tag` and `workingdir`) defined at the top of the configuration, and then subsequently referenced in the `build` job:
 
 ```yml
--- a/jekyll/_cci2/pipeline-variables.adoc
+++ b/jekyll/_cci2/pipeline-variables.adoc
@@ -27,7 +27,6 @@ Pipeline values and parameters can be used to create reusable pipeline configura
 
 Pipeline values are available to all pipeline configurations and can be used without previous declaration.
 
-Please note: if using setup workflows, all pipeline parameters must be present in the subsequent configuration, even if calling with the API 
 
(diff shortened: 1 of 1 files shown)


jekyll/_cci2/pipeline-variables.adoc
```diff
@@ -27,6 +27,8 @@ Pipeline values and parameters can be used to create reusable pipeline configura
 
 Pipeline values are available to all pipeline configurations and can be used without previous declaration.
 
+Please note: if using setup workflows, all pipeline parameters must be present in the subsequent configuration, even if calling with the API
```
kbatuigas, line 30: I think I'd actually go for moving this note in a section further down the page - I'll add it as a suggestion :)

kbatuigas, line 30:
```suggestion

```

kbatuigas (changes requested):

rosieyohannan (approved):

kbatuigas (approved):

jkzilla (commented):
Today 05:24
That text is unclear/misleading.
e.g. "all pipeline parameters must be present in the subsequent configuration, including in API calls" makes little sense; if anything, the opposite is true - you don't need to pass any parameters in the API calls because CircleCI will include them anyway, even if you didn't want them (which is how I found this out).

It should say something like:

When using setup workflows, where control is passed to a second configuration, all pipeline parameters that are declared in the setup workflow must also be declared (ideally with the same defaults) in the second configuration.
This is because, when the setup workflow passes control to a second configuration, the parameters passed *to* the setup workflow will be merged in with any parameters specified *by* the setup workflow's code before being passed to the second configuration.
i.e. in addition to any parameters that the setup workflow's code might deliberately pass, the second configuration can also receive any parameter(s) that the setup workflow can, itself, legally accept, so the second configuration must tolerate receiving them (even if it doesn't need them).

... and you also need to explain how conflicts are resolved.
e.g. if a user triggers the setup build with MyParameter=foo but the setup workflow passes MyParameter=bar in as a parameter to the subsequent configuration, which one takes precedence?
Or does that cause the build to fail?

In my opinion, this behaviour is gnarly enough to warrant its own "Pipeline parameters in dynamic configuration" section on the https://circleci.com/docs/pipeline-variables/ page.
Describe it "nicely" there (i.e. tell folks what they need to do in order to have an easy time of things), and then go into full details (including how conflicts are resolved) in https://circleci.com/docs/dynamic-config/#how-dynamic-config-works (and add a "for more details" link to that from the new section in https://circleci.com/docs/pipeline-variables/).

... and then link to the new "Pipeline parameters in dynamic configuration" section from everywhere else that needs it (everywhere I linked to in my previous message) so that user's can't reasonably claim "I didn't see that".

@rosieyohannan @kbatuigas the customer followed up with the above, thoughts?

PR #8407, opened 2023-10-12
Update github-integration.adoc

e-grapp:
# Description

Add a mention of Env Vars for write access since we mention SSH keys and no mentions of permissions needed elsewhere

# Reasons

User wrote in about this in a ticket 
https://circleci.zendesk.com/agent/tickets/140038

# Content Checklist

### DIFF
--- a/jekyll/_cci2/github-integration.adoc
+++ b/jekyll/_cci2/github-integration.adoc
@@ -60,6 +60,7 @@ CircleCI requests the following permissions from GitHub, defined in the link:htt
 
 - Get a list of a user's repos
 - Add an SSH key to a user's account
+- Add Environment Variables to Project Settings
 
 **Admin Permissions**, needed for setting up a project
 


jekyll/_cci2/github-integration.adoc
```diff
@@ -60,6 +60,7 @@ CircleCI requests the following permissions from GitHub, defined in the link:htt
 
 - Get a list of a user's repos
 - Add an SSH key to a user's account
+- Add Environment Variables to Project Settings
```
rosieyohannan, line 63: So my understanding is this is a list of thing CircleCI can do on a user's behalf once they have these permissions. Rather than listing things the user is able to do. Maybe there's a better way to surface this 🤔 Does that make sense?

e-grapp: @rosieyohannan - apologies for delay on this!

Totally makes sense and I think you're correct that this isn't the best way to surface this as this is more of what _CircleCI can do on behalf of user_

We have an equivalent for Apps/GL but not exactly for OAuth integrations and I'm thinking an equivalent would be the best approach for this:

> https://circleci.com/docs/roles-and-permissions-overview/

rosieyohannan: Oh interesting! Thanks, I'll turn this into a Jira and then close the PR

PR #8533, opened 2023-12-18
Windows Known Issues update

e-grapp:
# Description
- Removed "Connecting to a Windows job via SSH and using the `bash` shell results in an empty terminal prompt."
- Replaced with note that `add_ssh_keys` step is required

# Reasons
This is part of a discussion in a private channel with @makotom to update known issues and limitations with Windows

- it no longer mentions "rerun with SSH with Bash" as a limitation (see screenshot below)
- it adds instead that explicit add_ssh_keys will be needed for any SSH key used in the job in any case (the key is used only in the checkout step) 
image (2)

# Content Checklist

### DIFF
--- a/jekyll/_cci2/using-windows.md
+++ b/jekyll/_cci2/using-windows.md
@@ -342,7 +342,7 @@ You can read more about using SSH in your builds [here]({{site.baseurl}}/ssh-acc
 
 These are the issues with the Windows executor that we are aware of and will address as soon as we can:
 
-* Connecting to a Windows job via SSH and using the `bash` shell results in an empty terminal prompt.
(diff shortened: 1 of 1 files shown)


makotom:
LGTM from this Windows freaks!

Overcommunicate: WRT `add_ssh_keys` , `checkout` might work without explicit `add_ssh_keys` for GitHub OAuth orgs, but it doesn't if the setup is based on GitHub Apps. Hence, for the sake of avoidance of confusions, advising customers to always include explicit `add_ssh_keys` is a solid approach in my opinion.

Great work, thank you very much @e-grapp!  @circleci/dre please feel free to bug me if we should more of the context!

rcircleci (approved):

PR #8570, opened 2024-01-16
Update pipeline-values.adoc

sebastian-lerner:
still validating if this is correct or not


sebastian-lerner:
okay confirmed that these are accurate https://circleci.slack.com/archives/C055HTSAN8K/p1706029571581379?thread_ts=1705419530.009829&cid=C055HTSAN8K

rcircleci (approved):

PR #8613, opened 2024-02-14
Updated JFrog CLI Documentation to Use v2 Standards

adrian-85:
# Description
Updated the JFrog CLI examples to use v2, as v1 is no longer in active development, and there is a known defect with Maven 3.9.0 and JFrog CLI v1.

# Reasons
[A link to a GitHub and/or JIRA issue (if applicable).](https://github.com/jfrog/jfrog-cli/issues/1863)
Current documentation does not take advantage of JFrog CLI integration with Maven, uses old v1 syntax.

# Changes
- Replaced dead link for Maven plugin
- Pointed JFrog CLI to v2 docs.
- Updated examples to use JFrog CLI v2 standards.
- Replaced native Maven commands with integrated JFrog Maven commands. 
- Cleaned up script to take advantage of built-in JFrog ability to package and upload artifacts. 
- Removed command that risks exposing sensitive creds in job output.


rosieyohannan:
@adrian-85 Thank you so much for these updates. I just need to verify them , and then we'll get these updates published 😃

adrian-85:
@rosieyohannan This has been open for quite some time.  My concern if review isn't done soon, this change itself may become out of date if newer versions of JFrog are released.

jekyll/_cci2/deploy-to-artifactory.md
```diff
@@ -39,7 +39,7 @@ Add the following to your `.circleci/config.yml`:
 ```yml
 - run:
     name: Install jFrog CLI
-    command: curl -fL https://getcli.jfrog.io | sh
+    command: curl -fL https://getcli.jfrog.io/v2-jf | sh
```
rosieyohannan, line 42: Looks like this should be `curl -fL https://install-cli.jfrog.io | sh`? https://docs.jfrog-applications.jfrog.io/jfrog-applications/jfrog-cli/install#install-with-curl

adrian-85: I believe `getcli` is the command we want in this case.  It mimics what was originally in the docs, and seems to be the approach prescribed in this blog:
https://github.com/jfrog/jfrog-cli/blob/v2/guides/getting-started-with-jfrog-using-the-cli.md

I can also share CircleCI config I successfully used this command in if that helps(private repo).

adrian-85: @rosieyohannan I know with the holiday things might have dragged a bit, wondering if you've had a chance to review my last comment?

adrian-85: @rosieyohannan Any update?

adrian-85: @rosieyohannan Can you take a minute to re-review when you have a chance?

adrian-85: Apologies, I missed hitting the re-review request button.  Something we started enforcing in our own team recently because otherwise it's just too difficult to sort through the open PR's.

adrian-85: @rosieyohannan ^^.

pete-woods: Should we instead link directly to their install docs? https://docs.jfrog-applications.jfrog.io/jfrog-applications/jfrog-cli/install#installation

Rather than recommend everyone use the (not very secure) `curl -sSL https://random-stuff | sh` method? Their docs e.g. recommend using OS package managers first, and the curl install as a lower down option.

adrian-85: There are a lot of different options on that page, OS, node, etc.  I agree, it may be better to simply present the options available through a link.  We could keep the current line as an example.  My thought when I opened this was to stick as close to the original outline as possible, which uses curl.

I'm fine proceeding either way, what do you want to see @pete-woods @rosieyohannan ?

pete-woods: https://circleci.atlassian.net/wiki/spaces/SD/pages/6340575858/Avoiding+curl+bash+in+the+product

pete-woods: Would far rather see us encourage secure practises for our customers :)

adrian-85: Updated, please re-review.

pete-woods: Looks great!

pete-woods (approved):

rosieyohannan (approved):

rosieyohannan:
Thank you for updating this for us @adrian-85, and for your patience!

PR #8680, opened 2024-03-27
Adding docs for disabling presigned URLs in server 4.3+

soulchips:
# Description
As of server 4.3, customers are able to disable presigned URLs to allow for larger artifact uploads but this has not been documented. This PR adds these instrutions

# Reasons
https://circleci.atlassian.net/browse/ONPREM-908

# Content Checklist

### DIFF
--- a/jekyll/_cci2/server/v4.3/installation/phase-1-prerequisites.adoc
+++ b/jekyll/_cci2/server/v4.3/installation/phase-1-prerequisites.adoc
@@ -556,7 +556,7 @@ aws iam create-role --role-name circleci-s3 --assume-role-policy-document file:/
 }
 ----
 +
-NOTE: If you wish to store artifacts which are larger than 5GB, you will need to enable non-presigned URLs. To do this you will need your IRSA role to assume itself. Replace your trust policy above with the contents below.
+NOTE: If you wish to store artifacts which are larger than 5GB, you will need to enable non-presigned URLs. To do this you will need your IRSA role to be able to assume itself. Replace your trust policy above with the contents below.
 +
 [source, json]
 ----
--- a/jekyll/_cci2/server/v4.3/installation/phase-1-prerequisites.adoc
+++ b/jekyll/_cci2/server/v4.3/installation/phase-1-prerequisites.adoc
@@ -568,15 +568,19 @@ NOTE: If you wish to store artifacts which are larger than 5GB, you will need to
       "Principal": {
         "Federated": "<OIDC_PROVIDER_ARN>"
       },
-      "Action": [
-        "sts:AssumeRoleWithWebIdentity",
-        "sts:AssumeRole"
-      ],
+      "Action": "sts:AssumeRoleWithWebIdentity",
       "Condition": {
         "StringEquals": {
           "<OIDC_PROVIDER_URL>:sub": "system:serviceaccount:<K8S_NAMESPACE>:object-storage"
         }
       }
+    },
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "<ROLE_ARN>"
+      },
+      "Action": "sts:AssumeRole"
     }
   ]
 }
--- a/jekyll/_cci2/server/v4.3/installation/phase-1-prerequisites.adoc
+++ b/jekyll/_cci2/server/v4.3/installation/phase-1-prerequisites.adoc
@@ -703,7 +703,7 @@ aws iam attach-role-policy --role-name <OBJECT_STORAGE_ROLE_NAME> --policy-arn=<
 --
 **Option 2:** IAM access keys
 
-NOTE: If you wish to store artifacts which are larger than 5GB, you will need to disable presigned mode which requires an AWS role. We recommend you follow the instructions for creating an IRSA role in this case.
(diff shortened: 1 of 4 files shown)


jekyll/_cci2/server/v4.3/installation/phase-1-prerequisites.adoc
```diff
@@ -555,6 +555,32 @@ aws iam create-role --role-name circleci-s3 --assume-role-policy-document file:/
...
+        "sts:AssumeRole"
+      ],
+      "Condition": {
+        "StringEquals": {
+          "<OIDC_PROVIDER_URL>:sub": "system:serviceaccount:<K8S_NAMESPACE>:object-storage"
+        }
+      }
+    }
+  ]
+}
```
DomParfitt, line 582: I don't think this is the correct Trust Policy for non-presigned Role. It looks like it just adds the `sts:AssumeRole` `Action` to the existing statement but it should be in an additional statement:
```json
{
  "Effect": "Allow",
  "Principal": {
    "AWS": "<ROLE_ARN>"
  },
  "Action": "sts:AssumeRole"
}
```
So the whole thing would be:
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "<OIDC_PROVIDER_ARN>"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "<OIDC_PROVIDER_URL>:sub": "system:serviceaccount:<K8S_NAMESPACE>:object-storage"
        }
      }
    },
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "<ROLE_ARN>"
      },
      "Action": "sts:AssumeRole"
    }
  ]
}
```

jekyll/_cci2/server/v4.3/installation/phase-2-core-services.adoc
```diff
@@ -679,6 +677,18 @@ object_storage:
     region: "<role-region>"
     irsaRole: "<irsa-arn>"
 ----
+
+**Non-presigned URLs (Optional)**
+If you wish to store artifacts larger than 5GB, you will need to disable presigned URLS. Add the following to the `object_storage.s3` section:
```
DomParfitt, line 682: Should this section include an explicit link back to the IAM setup for non-presigned? I can see a customer skipping over that first section and then attempting to enable non-presigned mode just using this section and things not working.

jekyll/_cci2/server/v4.3/installation/phase-1-prerequisites.adoc
```diff
@@ -555,6 +555,32 @@ aws iam create-role --role-name circleci-s3 --assume-role-policy-document file:/
   ]
 }
 ----
++
+NOTE: If you wish to store artifacts which are larger than 5GB, you will need to enable non-presigned URLs. To do this you will need your IRSA role to assume itself. Replace your trust policy above with the contents below.
```
DomParfitt, line 559:
```suggestion
NOTE: If you wish to store artifacts which are larger than 5GB, you will need to disable presigned URLs. To do this you will need your IRSA role to assume itself. Replace your trust policy above with the contents below.
```
`non-presigned URLs` doesn't really correctly describe how the non-presigned mode works.

DomParfitt, line 559:
```suggestion
NOTE: If you wish to store artifacts which are larger than 5GB, you will need to enable non-presigned URLs. To do this you will need your IRSA role to be able to assume itself. Replace your trust policy above with the contents below.
```

jekyll/_cci2/server/v4.3/installation/phase-2-core-services.adoc
```diff
@@ -679,6 +677,18 @@ object_storage:
     region: "<role-region>"
     irsaRole: "<irsa-arn>"
 ----
+
+**Non-presigned URLs (Optional)**
```
DomParfitt, line 681:
```suggestion
**Non-presigned Mode (Optional)**
```
As above, `Non-presigned URLs` doesn't really fit correctly. I think `Non-presigned Mode` is probably the best description, we've also used `Direct Access` internally but I think that might be slightly less clear.

jekyll/_cci2/server/v4.3/installation/phase-2-core-services.adoc
DomParfitt, line 682: It might be worth including something about what is actually happening when you run with presigned URLs disabled (i.e. that we perform a multi-part upload to S3), as I can see someone misreading this as just disabling the presigning of the URLs (i.e. disabling the auth), which it isn't (the security between the 2 is essentially equivalent).

May also be good to point out that disabling presigned URLs may also offer better resilience and throughput (i.e. faster uploads) compared with using presigned URLS, in addition to enabling uploads greater than 5GB.

DomParfitt (changes requested):
I think all my comments also apply to `4.4`. This should all be true for `4.2` as well, so I think we could add these pieces to those docs.

kelvintaywl (approved):

DomParfitt:
> > I think all my comments also apply to `4.4`. This should all be true for `4.2` as well, so I think we could add these pieces to those docs.
> 
> the 4.3 and 4.4 docs have been updated. this isn't an option in 4.2 however: https://github.com/circleci/server/blob/server-4.2/values.yaml#L111-L130

Yep, got my versions confused, this was introduced in `4.3`

DomParfitt (approved):

PR #8744, opened 2024-05-02
more specific

sebastian-lerner:
anyone with push access to a repo can trigger a build so this is wrong. gitlab/github app triggers cant be run via the CCI UI or API today. so im making this more specific to say "reruns"

https://circleci.slack.com/archives/C055HTSAN8K/p1714667491334609

### DIFF
--- a/jekyll/_cci2/roles-and-permissions-overview.adoc
+++ b/jekyll/_cci2/roles-and-permissions-overview.adoc
@@ -261,7 +261,7 @@ The table below shows the permissions associated with each CircleCI organization
 ^| icon:check-circle[]
 ^| icon:check-circle[]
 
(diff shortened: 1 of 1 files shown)


sebastian-lerner:
uhh it looks like there is a lint error. @rosieyohannan any idea if i should fix that?

nathanfishcircleci (commented):
it includes api or UI interaction correct? 

Maybe the better wording is "CircleCI Application (via UI or API)"

My main concern is people not understanding its possible via two methods.

sebastian-lerner:
@nathanfishcircleci you can't trigger a pipeline via the UI or the API for gitlab/github app?

only reruns via the UI

PR #8745, opened 2024-05-02
Update ip-ranges.adoc show deprecation of macOS IPs

labsquared:
# Description
Add note to indicate deprecation of known macOS IPs.

# Reasons
macOS well-known IPs are no longer supported after April 24, 2024.

# Content Checklist

### DIFF
--- a/jekyll/_cci2/ip-ranges.adoc
+++ b/jekyll/_cci2/ip-ranges.adoc
@@ -164,6 +164,11 @@ CircleCI _does not recommend_ configuring an IP-based firewall based on the AWS 
 
 In addition to AWS and GCP (see above), CircleCI's macOS cloud hosts jobs executed by machines. The following IP address ranges are used by CircleCI macOS Cloud:
 
+[NOTE]
+====
+The macOS cloud IPs listed below are deprecated as of April 24, 2024. 
+====
+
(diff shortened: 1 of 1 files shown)


kelvintaywl (changes requested):
i'm sorry to request changes here, but thinking on behalf of readers/customers, I wonder if we can announce this change better?
I hope you can hear me out 🙇 

1. Leaving the listed IP ranges around can cause confusion; I'd argue removing the listed IP ranges would be less confusing.
2. this change would mean that the following text is **still** there, which can also confuse

> macOS builds are automatically restricted within the IP ranges listed here. In other words, you do not have to explicitly set circleci_ip_ranges: true for macOS builds.
([ref](https://circleci.com/docs/ip-ranges/#:~:text=macOS%20builds%20are%20automatically%20restricted%20within%20the%20IP%20ranges%20listed%20here.%20In%20other%20words%2C%20you%20do%20not%20have%20to%20explicitly%20set%20circleci_ip_ranges%3A%20true%20for%20macOS%20builds.))

If we prefer setting the <Note> on its deprecation, can we be verbose and say something like "the earlier list of IP ranges are no longer supported/guaranteed for MacOS builds".
ideally, a documented next-step / workaround would also be great.

PR #8859, opened 2024-07-16
Chore: 7/15 CVEs

rcircleci:
This takes care of the following CVEs:

- minimatch
- inflight
- rack

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,6 @@ gem "kramdown-parser-gfm"
 gem 'liquid-c', '~> 4.0.1'
 gem 'htmlcompressor'
 gem 'htmlentities', '~> 4.3', '>= 4.3.4'
-gem 'rack', '~> 3.0.0'
 gem 'nokogiri', '~> 1.16.5'
 
 group :jekyll_plugins do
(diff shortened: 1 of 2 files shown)


tfe (commented):
I wouldn't recommend adding httparty to the Gemfile as it's not actually a direct dependency of the project. It'd likely be confusing to someone working on the project later. And I already updated it in an earlier PR, so adding this line actually downgrades it from 0.22 to 0.21.

tfe (approved):

tfe:
Actually as far as I can tell rack isn't even used in this project. It was added in [this commit](https://github.com/circleci/circleci-docs/commit/72a3cc3453e21a0cd75cd0b09ea21b071171288c), but did not exist in the Gemfile.lock prior to that. A false positive from that time? I think we should try just removing it and seeing if the project builds.

szymczdm (approved):

tfe:
@rcircleci that works: https://app.circleci.com/pipelines/github/circleci/circleci-docs/54319/workflows/c41d1995-f5d5-48c8-bbc5-23439c9a2c60

And also I looked at the project itself and do not see anything that loads rack.

PR #8900, opened 2024-08-08
fix: Air-gap should use :start to download assets

ryanwohara:
The :start command was failing when there is no REPL to connect to. Using :start allows the shell to download assets to be committed via Docker and then utilized within an air-gap environment.


christian-stephen (approved):

kelvintaywl (approved):
thank you for the fix! 🙇

PR #8935, opened 2024-08-30
fix: Add server versions to page descriptions

ryanwohara:
Added server versions to the titles that were missing them. I did a simple regex so it's _very_ likely I missed some.

These were the settings I had when I did the replace:
image

I manually removed the duplicate version in the title from the post-installation docs and fixed the Mongo versions that were seemingly replaced by accident in 4.5/4.6 (which makes sense - I'm sure we just did a `sed 's|v4\.4|v4.5|g` and then 4.5->4.6 on them).

### DIFF
--- a/jekyll/_cci2/server/v4.5/operator/managing-user-accounts.adoc
+++ b/jekyll/_cci2/server/v4.5/operator/managing-user-accounts.adoc
@@ -8,7 +8,7 @@ noindex: true
 = Managing user accounts
 :page-layout: classic-docs
 :page-liquid:
-:page-description: This section provides information to help CircleCI server perators manage user accounts.
+:page-description: This section provides information to help CircleCI server v4.5 perators manage user accounts.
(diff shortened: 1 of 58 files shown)


jekyll/_cci2/server/v4.5/operator/managing-user-accounts.adoc
```diff
@@ -8,7 +8,7 @@ noindex: true
 = Managing user accounts
 :page-layout: classic-docs
 :page-liquid:
-:page-description: This section provides information to help CircleCI server perators manage user accounts.
+:page-description: This section provides information to help CircleCI server v4.5 perators manage user accounts.
```
rosieyohannan, line 11:
```suggestion
:page-description: This section provides information to help CircleCI server v4.5 operators manage user accounts.
```

rosieyohannan (approved):

PR #9073, opened 2024-11-13
Docker compose updates

rosieyohannan:
# Description
* Update instruction for installing Docker Compose in a non cimg image
* Refer to the latest version of compose by using `docker compose`

# Reasons
Customer confusion around `docker-compose` / `docker compose`

# Content Checklist


jekyll/_cci2/concepts.adoc
```diff
@@ -281,7 +281,7 @@ jobs:
   build2:
     machine: # Specifies a machine image that uses
       # an Ubuntu version 20.04 image with Docker 20.10.12
-      # and docker-compose 1.29.2, follow CircleCI Discuss Announcements
+      # and docker compose 1.29.2, follow CircleCI Discuss Announcements
       # for new image releases.
       image: ubuntu-2004:202201-02
```
DominicLavery, line 286: Not directly related to your change, but whilst we are in the area:
We probably shouldn't recommend pinned version and instead should give examples with the `current` tag instead

jekyll/_cci2/docker-compose.md
```diff
@@ -6,44 +6,56 @@ contentTags:
...
+
+If you want to use Docker Compose to manage a multi-container setup with a Docker Compose file, use the `machine` key in your `.circleci/config.yml` file and use `docker compose` as you would normally (see Linux VM execution environment documentation [here]({{site.baseurl}}/using-linuxvm) for more details). That is, if you have a Docker Compose file that shares local directories with a container, this will work as expected.
+
+
+## Using Docker Compose with docker executor
+{: #using-docker-compose-with-docker-executor }
+
+Using the `docker` execution environment combined with `setup_remote_docker` provides a remote engine similar to the one created with running Docker commands and the machine execution environment, but volume mounting and port forwarding do **not** work the same way in this setup. The remote docker daemon runs on a different system than the docker CLI and docker compose, so you must move data around to make this work. Mounting can usually be solved by making content available in a docker volume. It is possible to load data into a Docker volume by using `docker cp` to get the data from the CLI host into a container running on the docker remote host.
```
DominicLavery, line 24: ```The remote docker daemon runs on a different system than the docker CLI and docker compose```
This isn't strictly true now. Remote docker jobs now run on a machine executor, and that is where we spin up the jobs containers. The daemon is running on the same executor. I don't know if this can be reworded somehow to leave out such details 😖 Maybe it doesn't matter for your purposes here

rosieyohannan: Yes I wondered about this - so is the whole explanation of moving data aropund to make it accessible outdated now?

This section is there to point out differences (that users need to be aware of) around using docker compose and the docker executor as an alternative to using docker compose and the machine executor. If this info is now irrelevant I can just remove it

DominicLavery: I suppose it does do that. It is perhaps a bit academic. I see the section has only moved as well. So, I'll give this a stamp

DominicLavery (approved):

sebastian-lerner (approved):
🚀

jekyll/_cci2/docker-compose.md
sebastian-lerner: 👋 i am talking to a prospective customer and they pointed me to this paragraph as a reason for why they couldnt achieve the thing they wanted with remote docker.  

i had the same reaction as you @DominicLavery which is that this wording is outdated because remote docker is basically machine now.  if i open up a PR to edit the language, any guidance for what to remove Dom? definitley the pieces about "remote engine" and "remote docker daemon runs on a different system".  any idea if the person still needs to do the `docker cp` stuff? i would assume no but the user i'm talking to said that it didnt work for them

"The issue I was referring to about docker is that as part of the deploy we run a command in a docker container (which has encapsulated some generic infrastructure setup) that requires as input a configuration file accessed via a bind mounted volume. Something like:

`docker run -v ./config.yaml:/app/config.yaml ... image`

This did not work"

also let me know if you want me to post and tag ask-execution for this question instead.  the user did manage to get things working using `docker cp` so maybe that is still needed for some reason.

DominicLavery: RE: docker cp
It doesn't really change anything around where the data lives. They still run in a primary docker container and so still might need to copy data around. A volume mount would occur on the hosts file system, not the primary containers so the config file or other data they would be after likely wouldn't be there

sebastian-lerner: https://github.com/circleci/circleci-docs/pull/9170 lmk what you think

PR #9079, opened 2024-11-17
Update configuration-reference.adoc

HennaAbbas:
Adding depreciation on when :always

# Description
Added that when:  always is no longer supported and a use must use when: true

# Reasons
Clarity for customers 

# Content Checklist

### DIFF
--- a/jekyll/_cci2/configuration-reference.adoc
+++ b/jekyll/_cci2/configuration-reference.adoc
@@ -2550,10 +2550,9 @@ workflows:
 [#using-when-in-workflows]
 ==== *Using `when` in workflows*
 
-NOTE: Using `when` or `unless` under `workflows` is supported in `version: 2.1` configuration. Workflows are always run unless there is a when or unless filter that prevents the workflow from being run. If you want a workflow to run in every pipeline, do not add a when or unless filter.. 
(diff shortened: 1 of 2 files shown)


jekyll/_cci2/configuration-reference.adoc
```diff
@@ -2550,10 +2550,11 @@ workflows:
 [#using-when-in-workflows]
 ==== *Using `when` in workflows*
 
-NOTE: Using `when` or `unless` under `workflows` is supported in `version: 2.1` configuration.
+NOTE: Using `when` or `unless` under `workflows` is supported in `version: 2.1` configuration. Using `when: always` is no longer supported as of December 3,2024. You must use `when: true`.
```
gordonsyme, line 2553: `when: always` has never been supported, we aren't deprecating it. `always` does not mean anything special in the context of a workflow `when`, it's just a non-empty string.

The thing to do is to delete the `when` key, the default behaviour is to always run a workflow _unless_ there is a `when` (or `unless`) key that says when not to run the workflow.

gordonsyme: How about

> Workflows are always run unless there is a `when` or `unless` filter that prevents the workflow from being run. If you want a workflow to run in every pipeline, do not add a `when` or `unless` filter.

rosieyohannan: @HennaAbbas do you want to make this change in this PR?

PR #9093, opened 2024-11-21
Update when always to when true

nick-ohhh:
Based on the incoming change to when condition, this updates `when: always` references in `configuration-reference` to `when: true`.

# Description
Based on the incoming change to when condition, this updates `when: always` references in `configuration-reference` to `when: true`.

# Reasons
Internal customer feedback.

# Content Checklist

### DIFF
--- a/jekyll/_cci2/configuration-reference.adoc
+++ b/jekyll/_cci2/configuration-reference.adoc
@@ -1530,7 +1530,7 @@ NOTE: the `run` step replaces the deprecated `deploy` step. If your job has a pa
 | `when`
 | N
 | String
-| <<the-when-attribute,Specify when to enable or disable the step>>. Takes the following values: `always`, `on_success`, `on_fail` (default: `on_success`)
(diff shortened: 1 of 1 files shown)


gordonsyme:
@f440 that's correct, the `when` attribute of steps is not affected.

---

For clarity, what's happening to workflow `when` is a tightening up of validation to avoid unexpected outcomes.

We're seeing some folks using redundant `when` values, and others using values that do not do what the config author intends.

There have never been any special keywords for workflow `when`, non-empty strings are currently considered truthy which is equivalent to leaving the `when` out, the default behaviour is to always run workflows.

`when: always`, `when: never`, `when: tuesdays` are all non-empty string constants and all result in the workflow always running, despite the clear intent in some cases not to do so.

This leads to unintentional, and unexpected, credit usage.

The additional validation will forbid string constants and ensure that the value of a workflow `when` is either a parameter (e.g. `when: << pipeline.parameters.run-special-workflow >>`), or an expression (e.g. `when: {equal: [<< pipeline.git.branch >>, "main"]}`), or a boolean (`when: false`).

gordonsyme:
@nick-ohhh I'm going to close this for now.

PR #9170, opened 2025-01-09
remove implementation details, make accurate with new remote docker arch

sebastian-lerner:
tries to remove some (i think unecessary) implementation details about how our system works. also tries to remove references to the word "remote" since that's a bit of a relic from a previous architecture. 

that being said, it keeps the bulk of the content in place because the workarounds described are still needed when using remote docker even with the new architecture


DominicLavery (approved):
This is totally an improvement and happy for it to be merged as is. 

The "different location" statement could maybe be improved. I'm thinking the location of the daemon is the implementation detail. The fact that a job is executed in a container that has access to a daemon is the feature. 
So something along the lines of
> A jobs commands are executed in a container that has access to an external docker daemon, so whilst using docker CLI or compose you must move data around...

DominicLavery:
LGTM :shipit:

sebastian-lerner:
@rosieyohannan can you take a peek when you get a chance please 🙏 

thanks!

jekyll/_cci2/docker-compose.md
```diff
@@ -21,7 +21,7 @@ If you want to use Docker Compose to manage a multi-container setup with a Docke
...
+Using the `docker` execution environment combined with `setup_remote_docker` enables you to run Docker commands similar to how you run Docker commands in a machine execution environment, however, volume mounting and port forwarding do **not** work the same way when using the `docker` execution environment. When using the `docker` execution environment with `setup_remote_docker`, a job's commands are executed in a container that has access to an external Docker daemon. Therefore, to use the Docker CLI or Docker Compose, you must move data around. Mounting can typically be solved by making content available in a Docker volume. It is possible to load data into a Docker volume by using `docker cp` to get the data from the CLI host into a location running on the Docker host.
```
rosieyohannan, line 24:
```suggestion
Using the `docker` execution environment combined with `setup_remote_docker` enables you to run Docker commands similarly to how you run Docker commands in a machine execution environment. However, volume mounting and port forwarding do **not** work the same way when using the `docker` execution environment. When using the `docker` execution environment with `setup_remote_docker`, a job's commands are executed in a container that has access to an external Docker daemon. Therefore, to use the Docker CLI or Docker Compose, you must move data around. Mounting can typically be solved by making content available in a Docker volume. It is possible to load data into a Docker volume by using `docker cp` to get the data from the CLI host into a location running on the Docker host.
```

rosieyohannan (approved):

PR #9184, opened 2025-01-18
Remove redundant folder tmp for files.txt

vinzid:
Remove redundant folder `tmp` for `files.txt`, since no `/tmp/` prepend when writing to `files.txt` in the previous command.


rosieyohannan:
Thank you for spotting this and fixing for us! 👍

PR #9212, opened 2025-02-08
Update Doc for rerun SSH code snippets

b23prodtm:
Add circleci as namespace in command snippets

# Description
Reading self-host runner doc, update snippets, that way I can install self-host runner with SSH


ryanwohara:
Users can do this themselves. The names don't need to be what is implied in this PR.

PR #9259, opened 2025-03-05
Add serial-groups section to config reference

rosieyohannan:
# Description
Add details of the serial-group key to the configuration reference

Screenshot 2025-03-07 at 15 44 42

# Reasons

# Content Checklist

### DIFF
--- a/jekyll/_cci2/configuration-reference.adoc
+++ b/jekyll/_cci2/configuration-reference.adoc
@@ -2654,6 +2654,48 @@ A job can have the keys `requires`, `name`, `context`, `type`, and `filters`.
 A job name that exists in your `config.yml`.
 
 '''
+====== `serial-group`
+
+The `serial-group` key is used to add a property to a job to allow a group of jobs to run in series, rather than concurrently, across an organization. Serial groups control serialization of jobs across an organization, not just within projects and pipelines.
+
+The value of the `serial-group` key is a string that is used to group jobs together to run one after another. The key must meet the following requirements:
+
+* Must be less than or equal to (≤) 512 characters, once compiled.
+* Must not be blank.
+
+You can use pipeline values and parameters in the `serial-group` key.
+
+[.table.table-striped]
+[cols=4*, options="header", stripes=even]
+|===
+| Key | Required | Type | Description
+
+| `serial-group`
+| N
+| String
+| A string that is used across an org to group jobs together to run one after another. Can include pipeline values and parameters. Use this same serial group across multiple pipelines to control serialization.
+|===
+
+*Example:*
+
+[,yml]
+----
+# Creating multiple pipelines at the same time with the below config will results in
+# all pipelines running test and build but only a single pipeline will run deploy at a time.
+
+workflows:
+  main-workflow:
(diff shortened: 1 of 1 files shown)


nathanfishcircleci (commented):
I believe right now we only support the individual job and not a group of jobs. We should be clear that this is specific for one job and that we will add "groups of jobs" in the future.

rosieyohannan:
> I believe right now we only support the individual job and not a group of jobs. We should be clear that this is specific for one job and that we will add "groups of jobs" in the future.

Does this work: "The `serial-goup` key is configurable per job. It is not possible to configure the key for a group of jobs at this time."

jekyll/_cci2/configuration-reference.adoc
```diff
@@ -2653,6 +2653,51 @@ A job can have the keys `requires`, `name`, `context`, `type`, and `filters`.
 
 A job name that exists in your `config.yml`.
 
+'''
+====== `serial-group`
+
+The `serial-group` key is used to add a property to a job to allow a group of jobs to run in series, rather than concurrently, across an organization. Serial groups control the orchestration of jobs across an organization, not just within projects and pipelines.
```
Fernando-Abreu, line 2659: We should note that if jobs with the same serial-group key within the same project have lower pipeline numbers, they will be skipped if a job from a higher pipeline number has already been executed.

cc: @nathanfishcircleci @liamclarkedev

liamclarkedev: I agree, that might even deserve its own subsection or somehow highlighted as an important detail.

Something along the lines of:

Serial groups are started in order of pipeline number, if a group is currently waiting or running and another group in the same project attempts to join the queue with a lower pipeline number, it will be `not_run` to maintain order integrity.

If there are no serial groups waiting/running, a pipeline with a lower number can start, such as restoring back to an previous pipeline via a rerun workflow.

jekyll/_cci2/configuration-reference.adoc
```diff
@@ -2653,6 +2653,51 @@ A job can have the keys `requires`, `name`, `context`, `type`, and `filters`.
 
 A job name that exists in your `config.yml`.
 
+'''
+====== `serial-group`
```
liamclarkedev, line 2657: I think it's worth mentioning somewhere that serial groups will wait for a maximum of 5 hours before being cancelled.

This only applies to groups waiting in the queue, the standard limits apply to the jobs inside the queue (e.g. max run time of build jobs).

jekyll/_cci2/configuration-reference.adoc
rosieyohannan: We can add a caution section

jekyll/_cci2/configuration-reference.adoc
rosieyohannan: What does "cancelled" mean in this scenario? Like the serial group cancels so all the waiting jobs will just run concurrently after 5 hours?

nathanfishcircleci (approved):

yaningo (approved):

PR #9523, opened 2025-08-20
Simplify wording in cache retention to add clarity

2shortplanks:
The original text is misworded - it says "Job retention reduces the organization-level retention" but job retention has no effect on the overall organization-level retention.


2shortplanks:
None of the lint errors were caused by my changes.

rosieyohannan (approved):
Thank you!

PR #9529, opened 2025-08-20
[EXEC-5670] Update Xcode release & deprecation policy

OompahLoompah:
# Description
Update the Xcode release & deprecation policy with the new deprecation policy and ensure the rest of the document is consistent with current practices.

# Reasons
Current practice with regards to Xcode image release and deprecation diverged from our policy, especially regarding how we announce new images.

# Content Checklist

Please take a moment to check through the following items when submitting your PR (this is just a guide so will not be relevant for all PRs):

### DIFF
--- a/docs/guides/modules/execution-managed/pages/xcode-policy.adoc
+++ b/docs/guides/modules/execution-managed/pages/xcode-policy.adoc
@@ -96,7 +96,7 @@ Each Xcode image is built on top of a clean macOS install. We aim to keep the ma
 
 We aim to keep the macOS version aligned across our different macOS executors, however this may not always be the case. Check the xref:test:testing-ios.adoc#supported-xcode-versions[Software Manifest] file for the image for the most accurate information.
 
-When a new major version of macOS (ex: `12.0`) is released, we will start to use this version after a minimum of two minor Xcode releases have passed.
+When a new major version of macOS (ex: `26.0`) is released, we will start to use this version after a minimum of two minor Xcode releases have passed.
 
 [#exceptions]
 == Exceptions
--- a/docs/guides/modules/execution-managed/pages/xcode-policy.adoc
+++ b/docs/guides/modules/execution-managed/pages/xcode-policy.adoc
@@ -13,7 +13,7 @@ You will find the most up to date information on these images on the link:https:
 [#xcode-image-retention-and-deprecation]
 == Xcode image retention and deprecation
 
-We aim to retain four major versions of Xcode, with more recent versions having a larger number of minor versions to choose from.
+We will retain the three latest minor versions of the latest major version of Xcode. In addition we will also retain the last minor version of the three previous major versions of Xcode.
 
 For example, with Xcode 16 being the latest major version being released:
 
@@ -22,16 +22,16 @@ For example, with Xcode 16 being the latest major version being released:
 | Xcode Version | Action
 
 | Xcode 16
-| We will retain all `major.minor` versions at the latest patch version
+| We will retain the three latest major.minor versions at the latest patch version
 
 | Xcode 15
-| We will retain the 4 latest `major.minor` versions at the latest patch version
+| We will retain the last major.minor version
 
 | Xcode 14
-| We retain a single image which will be the last stable release of Xcode 14
+| We will retain the last major.minor version
 
 | Xcode 13
-| We retain a single image which will be the last stable release of Xcode 13
+| We will retain the last major.minor version
 |===
 
 Future example, when Xcode 26 enters Beta:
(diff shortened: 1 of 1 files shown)


cadecairos (approved):

docs/guides/modules/execution-managed/pages/xcode-policy.adoc
```diff
@@ -22,16 +22,16 @@ For example, with Xcode 16 being the latest major version being released:
 | Xcode Version | Action
 
 | Xcode 16
-| We will retain all `major.minor` versions at the latest patch version
+| We will retain the three latest major.minor versions at the latest patch version
```
brentmmarks, line 25: @labsquared thought we were going to change this model to use edge/current etc instead of continuing with the specific minor versions and setting up xcode current/edge

OompahLoompah: My understanding was that `current`/`edge` was going to supplement specifying the version. What's the reasoning for removing the ability of users to specify the Xcode version?

brentmmarks: specifically looking to remove specifying minor versions - as this would reduce customers pinning to minor versions and having to keep all the minor versions etc. This would also help with deprecations.

TLDR: helps with not needing to support as many skus in turn run smaller fleet etc -> better time to first step

OompahLoompah: I'd be hesitant to remove all older images just because I know in the past there have been customers who depended on them. I'd just want to know ahead of time that this isn't going to be a problem before we commit to it especially given how large of a change it would be going from 12 images to 2.

OompahLoompah: FWIW I'm not saying we won't do that, I think reducing the image count to 6 as in this new policy would be a good intermediate step to help gauge customer response. If we see pushback from customers then further reducing the image count probably doesn't make sense. If we don't get any negative reaction then further reducing the image count might be fine

brentmmarks: We are doing this with other image families -> I would like to keep MacOS similar if not the same. Doesnt need to be as aggressive as only 2 images. can do previous as well to support 3 versions but idea is to reduce pinning for customers.

brentmmarks: Github does something similar: https://github.com/actions/runner-images (they do allow you to access specific versions within the image but still moving away from specific images per version believe)

brentmmarks: I hear what you are saying about doing this incrementally but this is something we have decided to do for other images families and we know some customers might have pushback.

brentmmarks: Based on our last meeting it seemed like customers are less concerned about minor versions and more concerned about major versions?

brentmmarks: not saying we only necessarily keep only 2 major versions, could keep more but thing would ideally do is remove minor versions

OompahLoompah: If we were to consolidate Xcode versions into one or two images I think it would be an easier sell but that'll take time and we need to be deprecating some images very soon to make room for Xcode 26. @labsquared do you know if there is any indication whether customers on older images require them?

OompahLoompah: Made an attempt to clarify the retention policy following what was discussed earlier today. Could definitely use a new review as I confused myself a few times trying to write something legible. It's been a long time since I had to word-smith anything

docs/guides/modules/execution-managed/pages/xcode-policy.adoc
```diff
@@ -87,16 +87,16 @@ To read about our customer support policy regarding beta images, check out this
...
 
-We aim to keep the macOS version aligned across our different macOS executors (for example, Intel and M1 platforms), however this may not always be the case. Check the xref:test:testing-ios.adoc#supported-xcode-versions[Software Manifest] file for the image for the most accurate information.
+We aim to keep the macOS version aligned across our different macOS executors, however this may not always be the case. Check the xref:test:testing-ios.adoc#supported-xcode-versions[Software Manifest] file for the image for the most accurate information.
 
-When a new major version of macOS (ex: `12.0`) is released, we will start to use this version after a minimum of two minor Xcode releases have passed. This will allow for any major bugs and issues to be resolved. Release timing for this is entirely dependent on Apple's own release cycle, but will always be announced on our link:https://discuss.circleci.com/c/announcements/39[Discuss forum].
+When a new major version of macOS (ex: `12.0`) is released, we will start to use this version after a minimum of two minor Xcode releases have passed.
```
cadecairos: We might as well modernize this to "26.0", no?

OompahLoompah: Done

PR #9868, opened 2026-01-08
Remove read time but allow opt in for some how-to pages

rosieyohannan:
Right now all pages show read time.

The read time calculation is unhelpful, especially when there are many code blocks and when there is content in tabs (reader would likely pick one tab rather than read through all) because all words are counted equally in the calculation.

**This change removes the read time from all pages but allows it to be added back for specific pages - I've added it back to the quickstart guide.**

Other change is just to fix a path that was broken for loacl builds which was preventing the Last Updated section from being shown in a local build.

### DIFF
--- a/extensions/page-metadata-extension.js
+++ b/extensions/page-metadata-extension.js
@@ -30,8 +30,8 @@ module.exports.register = function () {
         pages.forEach((page) => {
           let { abspath, origin, path } = page.src
           const { branch, startPath, webUrl } = origin
-          const index = abspath?.indexOf('docs/')
-          const pagePath = abspath ? (index > -1 ? abspath.substring(index) : abspath ) :  startPath + '/' + path
+          // Use startPath + path for the git path (relative to repo root)
+          const pagePath = startPath + '/' + path
 
           const key = `${component}:${version}:${page.path}`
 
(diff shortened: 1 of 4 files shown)


extensions/page-metadata-extension.js
```diff
@@ -30,8 +30,8 @@ module.exports.register = function () {
         pages.forEach((page) => {
           let { abspath, origin, path } = page.src
           const { branch, startPath, webUrl } = origin
-          const index = abspath?.indexOf('docs/')
-          const pagePath = abspath ? (index > -1 ? abspath.substring(index) : abspath ) :  startPath + '/' + path
+          // Use startPath + path for the git path (relative to repo root)
```
rosieyohannan, line 33: This change is to fix the ability to see last update time in a local docs build

ui/src/partials/article-info-bar.hbs
```diff
@@ -1,24 +1,26 @@
...
           {{format-time-ago lastUpdate}}
         </a>
-        <span class="text-xl px-2">&middot;</span>
       {{/if}}
       <!-- Read Time -->
-      {{#if readingTime}}
-        <span class="flex min-w-fit items-center">
-          {{readingTime}}
-        </span>
+      {{#if @root.page.attributes.show-readtime}}
```
rosieyohannan, line 13: Only show readtime with `:page-show-readtime: true` set for page

lavin-ds (approved):

PR #9952, opened 2026-02-03
Cross repo triggers how-to

rosieyohannan:
_Written with Claude code for help with structure, use case and examples_

Adds a how-to guide for setting up cross-repo triggers with two pathways, one in which the config is stored with the code and one in which a centralised config repo is used.

Diagrams included render like this:
Screenshot 2026-03-12 at 11 39 28
Screenshot 2026-03-12 at 11 39 36
Screenshot 2026-03-12 at 11 39 55


docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,373 @@
...
+:page-platform: Cloud
+:page-description: Learn how to automatically trigger consumer repository pipelines when you tag a new library release.
+:experimental:
+:page-show-readtime: true
+
+In this guide you will learn how to set up cross-repo triggers. A cross-repo trigger watches for events in a repository other than the one that contains the pipeline configuration. *Cross-repo triggers allow you to run a pipeline to build code from one repository when an event occurs in another repository*.
+
+*For this how-to guide we have chosen the following scenario:*
+
+When you tag a new release of an internal library, CircleCI automatically runs tests in the repositories that consume that library. This validates that your library changes work correctly with dependent applications before you publish the release.
```
BeFunes: This may be just me, but when I first read it I thought it sounded like we're saying that this already happens by default. 
As opposed to - "we want CircleCI to automatically run tests..."

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,373 @@
+= Set up cross-repo triggers for library releases
```
BeFunes: Title says "for library **releases**", the navbar says "for library **consumers**"

Maybe that's fine, just flagging

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,373 @@
...
+
+This section covers some elements you might want to include in your configuration file. If you just want to follow along you can use the example config file provided at the end of this section.
+
+When a change to your library triggers your consumer pipeline, you need to know which _version_ of the library initiated the trigger so you can test against that specific version. CircleCI provides built-in pipeline values that capture details about the trigger event, including the tag name, branch name, commit SHA, and repository name.
+
+CircleCI automatically populates the following pipeline values when a cross-repo trigger fires:
+
+* `pipeline.git.tag`: The tag name that was created (for tag triggers). Empty string if not triggered by a tag.
+* `pipeline.git.branch`: The branch name where the event occurred in the library repository.
+* `pipeline.git.revision`: The specific commit SHA from the library repository.
```
BeFunes: Have you tested these? I don't think they are supposed to be right. If they are, there is something wrong :( 
Afaik it should be:
- `pipeline.event.github.ref` (this is valid both TAG and for BRANCH, it just depends on which one you are pushing. But I believe it's pushed with the prefix `tag/{tagname}` and `refs/heads/{branchname}`. 
- `pipeline.event.context.github.sha`  (`pipeline.event.context.github.after`)


Everything that is prefixed with `pipeline.git.*` refers to the **checkout** source. Here you are trying to get data from the **trigger** source

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,373 @@
...
+* `pipeline.trigger_parameters.github_app.repo_name`: The name of the library repository that triggered the pipeline.
```
BeFunes: This one is technically correct but we want to deprecate trigger parameters, we should use :
`"pipeline.event.github.repository.name"`

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,373 @@
...
+[source,shell]
+----
+my-library/
+my-consumer-app/
+  └── src/
+circleci-configs/
+  ├── .circleci/
+  │   ├── consumer-config.yml
+  │   └── other-config.yml
+----
```
BeFunes, line 138: I wonder if it'd be helpful to put a screenshot of what the trigger looks like at the bottom of the table. It would render very small, but I'm assuming users would be able to click and zoom in?

BeFunes: Actually, now I've read more of the page - maybe it's better suited to the tabs where you have "Option 1: Config in consumer repository" and walk the user through setting it up

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,373 @@
...
++
+If you are using a centralized config repository, you will also need to enter a branch name for the checkout source, typically `main`.
+
+. Select btn:[Save].
+
+=== 4.b. Configure tag filters in your workflow
+
+The trigger is now configured in CircleCI, but you also need to update your configuration file for your integration tests to run workflows for tags.
+
+By default, *CircleCI does not run workflows for tags unless you explicitly specify tag filters*.
```
BeFunes: is it worth a link to the config guide that talks about this?

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
rosieyohannan: Oh good catch, defo a mistake

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
rosieyohannan: 👍 👍 👍 I'm going to hold off on this PR until we have documented these [here](https://circleci.com/docs/reference/variables/#pipeline-values)

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
rosieyohannan: I've updated these now

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
rosieyohannan: Updated

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
rosieyohannan: Done

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
rosieyohannan: Done

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+
+|**Repository structure**
+a|
+[source,shell]
+----
+my-library/
+my-consumer-app/
+  ├── .circleci/
+  │   └── config.yml
+  └── src/
```
BeFunes, line 126: I still wonder if something like this would make it clearer that this doesn't need to be the CORE build-test-deploy.yml of the consumer, but can be a totally separate YAML file.
```suggestion
my-consumer-app/
  ├── .circleci/
  │   └── config.yml
  │   └── on-library-update.yml  
  └── src/
```

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+
+NOTE: This guide demonstrates both approaches. The steps below use tabs to show the differences between Option 1 and Option 2.
+
+== 1. Create configuration file for your consumer pipeline
+
+To follow this how-to guide you will need a CircleCI configuration file for your consumer pipeline. This file will define the steps you want to run when the library changes.
+
+This section covers some elements you might want to include in your configuration file. If you just want to follow along you can use the example config file provided at the end of this section.
+
+When a change to your library triggers your consumer pipeline, you need to know which _version_ of the library initiated the trigger so you can test against that specific version. CircleCI provides built-in pipeline values that capture details about the trigger event, including the tag name, branch name, commit SHA, and repository name.
```
BeFunes: Now that our pipeline values docs is a bit more tidy, would it make sense to link to it from the [pipeline values] mention?

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+--
+
+NOTE: This guide demonstrates both approaches. The steps below use tabs to show the differences between Option 1 and Option 2.
+
+== 1. Create configuration file for your consumer pipeline
+
+To follow this how-to guide you will need a CircleCI configuration file for your consumer pipeline. This file will define the steps you want to run when the library changes.
+
+This section covers some elements you might want to include in your configuration file. If you just want to follow along you can use the example config file provided at the end of this section.
+
```
BeFunes, line 158: Sorry if you say this below, but shouldn't we specify that they need to confiure a pipeline definition + A trigger with option "Tag pushes"?

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+          filters:
+            tags:
+              only: /.*/
+----
+
+IMPORTANT: The `filters` section is required because CircleCI does not run workflows for tags unless you explicitly specify tag filters. The pattern `/.*/` matches all tags. Both lightweight and annotated tags are supported.
+
+In this example:
+
+* When a tag like `v1.2.3` is created in the library repository, `<< pipeline.event.github.ref >>` will contain `v1.2.3`.
```
BeFunes: So actually it's a bit more complicated than this. 
The format of `pipeline.event.github.ref` is
- refs/heads/<branch-name> for branch pushes
- refs/tags/<tag-name> for tag pushes
We need to do some wizardry in config to remove the "refs/tags". As long as the trigger is configured to fire only on tags, it should be relatively straight forward, I can _try_ to help you come up with that config.

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+    docker:
+      - image: cimg/node:18.0
+    steps:
+      - checkout # checks out the consumer repository code
+
+      - run:
+          name: Display trigger information
+          command: |
+            echo "=== Cross-repo trigger information ==="
+            echo "Library version/tag: << pipeline.event.github.ref >>"
```
BeFunes: We need something extra here otherwise this will show as
`Library version/tag: refs/tags/v1.2.3`

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
BeFunes: ahhhh sorry I just read it all, you don't actually want to filter. You want to use the commit SHA if the tag isn't present.

BeFunes: maybe instruct to set up an "all pushes" trigger  then?

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+
+      - run:
+          name: Install library version from trigger
+          command: |
+            # In a real scenario, this would install from npm registry:
+            # npm install @myorg/my-library@<< pipeline.event.github.ref >>
+
+            echo "Installing library from trigger..."
+            LIBRARY_VERSION="<< pipeline.event.github.ref >>"
+            if [ -z "$LIBRARY_VERSION" ]; then
```
BeFunes: this too, basically this will always be populated because whether it's a tag or a branch :( So we need to do some magic. Hopefully Claude can help.

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+----
+
+IMPORTANT: The `filters` section is required because CircleCI does not run workflows for tags unless you explicitly specify tag filters. The pattern `/.*/` matches all tags. Both lightweight and annotated tags are supported.
+
+In this example:
+
+* When a tag like `v1.2.3` is created in the library repository, `<< pipeline.event.github.ref >>` will contain `v1.2.3`.
+* The workflow filter ensures the workflow runs for tag events.
+* The configuration installs that specific version using `npm install @myorg/my-library@v1.2.3`.
+* If triggered by a push (not a tag), it falls back to using the commit SHA.
```
BeFunes:
```suggestion
* If triggered by a push to a branch (not a tag), it falls back to using the commit SHA.
```

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+.. In the **Config path** field, enter the path to the specific configuration file for this consumer, for example `.circleci/consumer-app-config.yml`.
++
+TIP: If your central config repository contains configurations for multiple projects, use descriptive file names like `.circleci/consumer-app-integration-tests.yml` or organize them in subdirectories like `.circleci/consumers/my-app/config.yml`.
+
+.. For **Checkout source**, select your consumer repository from the dropdown menu (this is the repository whose code will be tested).
+.. Select **Save**.
++
+NOTE: When using centralized configuration, the `checkout` step will check out code from the consumer repository (the checkout source), not from the central config repository. This allows you to maintain configurations separately from application code.
+
+.Example of a pipeline configured for a centralized config repository
```
BeFunes: Would it be better for this image to be shown in point #4 (add a trigger for library events) given that it shows a pipeline with a trigger? OTherwise the user will see stuff that you have't explained yet

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
BeFunes: Finally got to reading the instruction to set up a trigger, and see you did actually suggested "Tag pushes". In this case, this is no longer true because the pipeline will ONLY trigger on tag pushes. Maybe that's an easier way to approach things? 
> If triggered by a push (not a tag), it falls back to using the commit SHA.

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
BeFunes:
```suggestion
            RAW_REF="<< pipeline.event.github.ref >>"
            LIBRARY_VERSION="${RAW_REF#refs/tags/}"
            echo "Library version/tag: ${LIBRARY_VERSION}"
```

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,343 @@
...
+      - run:
+          name: Install library version from trigger
+          command: |
+            # In a real scenario, this would install from npm registry:
+            # npm install @myorg/my-library@<< pipeline.event.github.ref >>
+
+            echo "Installing library from trigger..."
+            LIBRARY_VERSION="<< pipeline.event.github.ref >>"
+            if [ -z "$LIBRARY_VERSION" ]; then
+              LIBRARY_VERSION="<< pipeline.event.github.after >>"
```
BeFunes:
```suggestion
            RAW_REF="<< pipeline.event.github.ref >>"

            # Strip refs/tags/ or refs/heads/ prefix
            LIBRARY_VERSION="${RAW_REF#refs/tags/}"

            if [ -z "$LIBRARY_VERSION" ] || [ "$LIBRARY_VERSION" = "$RAW_REF" ]; then
              # Fallback if ref wasn't a tag
              LIBRARY_VERSION="<< pipeline.event.github.after >>"
```

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
rosieyohannan: @BeFunes I think you are saying we should just suggest using a tag pushes trigger and remove the bits in the tutorial that talk about falling back to the commit SHA on a branch push? Or do you think we should recommend all pushes and accommodate pushes to a branch in this guide? I think maybe we keep it simple and use a tag push trigger. Like we are only interested in running these integration tests when a tag is created for the library?

BeFunes: Yes exactly, that's what I think!!

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
BeFunes: This still needs to be adjusted. 
```suggestion
* When a tag like `v1.2.3` is created in the library repository, `<< pipeline.event.github.ref >>` will contain `refs/tags/v1.2.3`.
```

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,415 @@
...
+To follow this how-to guide you need a CircleCI configuration file for your consumer pipeline. This file defines the steps you want to run when the library changes.
+
+This section covers some elements you might want to include in your configuration file. If you just want to follow along you can use the example config file provided at the end of this section.
+
+When a change to your library triggers your consumer pipeline, you need to know which _version_ of the library initiated the trigger so you can test against that specific version. CircleCI provides built-in pipeline values that capture details about the trigger event, including the tag name, branch name, commit SHA, and repository name. For a full list of pipeline values, see the xref:reference:ROOT:variables.adoc[Pipeline Values] guide.
+
+CircleCI automatically populates the following pipeline values when a cross-repo trigger fires:
+
+* `pipeline.event.github.ref`: The full Git ref that was pushed, this will be a tag name for tag triggers.
+* `pipeline.event.github.after`: The SHA of the most recent commit on the branch after the push.
```
BeFunes: This is no longer needed here

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,415 @@
...
+workflows:
+  validate-library-integration:
+    jobs:
+      - test-with-library:
+          filters:
+            tags:
+              only: /.*/
+----
+
+IMPORTANT: The `filters` section is required because CircleCI does not run workflows for tags unless you explicitly specify tag filters. The pattern `/.*/` matches all tags. Both lightweight and annotated tags are supported.
```
BeFunes: I've tested, and this is actually not needed because the TAG is **triggering** the pipeline, but the checkout reference is a branch, not a tag. 
It's super confusing, but basically no tag filter is needed

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
rosieyohannan:
```suggestion
```

docs/guides/modules/orchestrate/pages/set-up-cross-repo-triggers-for-library-consumers.adoc
```diff
@@ -0,0 +1,415 @@
...
+    jobs:
+      - your-job-name:
+          filters:
+            tags:
+              only: /.*/ # Matches all tags
+----
+
+The pattern `/.*/` matches all tags. Both lightweight and annotated tags are supported.
+
+Your cross-repo trigger is now fully configured. When a tag is created in your library repository, CircleCI will trigger the consumer pipeline and run your workflows.
```
BeFunes: So all of this can be removed too. I'm almost tempted to say we should leave it to avoid confusing customers, but truly - it's not needed 😢

rosieyohannan: How about if I remove but then add a note that if you don't use the tag trigger but instead choose all pushes you need it? and just point people to where we explain in the workflows doc: https://circleci.com/docs/guides/orchestrate/workflows/#executing-workflows-for-a-git-tag but actually that section needs updating too to reference the ability to use a tag trigger?

BeFunes: Using the "Tag pushes" triggers is not the reason why this is not needed, it's more nuanced than that. 
It's still true that when you use "Tag pushes", in a single repo situation you will need that filter. 

The reason is that the `filter:  \n tags:` filter behind the scenes looks at the pipeline value `pipeline.git.tag`, which is only populated if the **checkout ref** is a tag. 
In this situation, `pipeline.git.tag` is empty, because we're checking out a branch (main). The fact that you're triggering through tag does not affect it. 

I don't think this is how things should work, but at the moment it's the reality, unfortunately.

PR #9965, opened 2026-02-06
Update openid-connect-tokens.adoc

cmolenda:
# Description
Fix misuse of term "AWS" in GCP section.

# Reasons
Avoid unnecessary possible confusion.


rosieyohannan:
Thank you for noticing this and fixing it for us 🫶

PR #10118, opened 2026-04-01
use expressions as first choice in all docs exmples

rosieyohannan:
# Description

Updated multiple docs pages and example files to make expression-based syntax the primary, recommended approach for conditionals in CircleCI configuration, with logic statements repositioned as a secondary option.

> **Note:** Deprecating job filters was split out into #10591 so it can be reviewed as its own decision, separate from this syntax-promotion cleanup.

Example YAML files were reorganised: expression-based examples moved into a dedicated `expression-examples/` directory, and logic statement examples moved into `logic-statements/` (removing the word "legacy" from the directory name).

# Reasons

* Expressions are the recommended way to write conditionals in CircleCI config. 
* Previously, docs introduced logic statements first and treated expressions as an alternative. This change ensures readers encounter the recommended approach first, while logic statements remain documented for users with existing config that uses them.


stig (approved):
LGTM.

docs/guides/modules/ROOT/examples/expression-examples/step-when/when-step-nested.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  test:
+    docker:
+      - image: cimg/node:19.0.1
+    steps:
+      - checkout
+      - when:
+# This nested expression demonstrates complex conditional logic with multiple parameters and pipeline values
+# It evaluates to true when: ((branch is "main" AND param1 is false) OR param3 is false) AND (param3 is false OR (param1 OR param2))
+          condition: << ((pipeline.git.branch == "main" and not pipeline.parameters.param1) or not pipeline.parameters.param3) and (not pipeline.parameters.param3 or (pipeline.parameters.param1 or pipeline.parameters.param2)) >>
```
stig: The comments starting at column 0 is a bit odd.
```suggestion
          # This nested expression demonstrates complex conditional logic with multiple parameters and pipeline values
          # It evaluates to true when: ((branch is "main" AND param1 is false) OR param3 is false) AND (param3 is false OR (param1 OR param2))
          condition: << ((pipeline.git.branch == "main" and not pipeline.parameters.param1) or not pipeline.parameters.param3) and (not pipeline.parameters.param3 or (pipeline.parameters.param1 or pipeline.parameters.param2)) >>
```

docs/reference/modules/ROOT/pages/configuration-reference.adoc
```diff
@@ -3594,93 +3599,6 @@ workflows:
 
 '''
 
-[#logic-statements]
-== Logic statements
```
rosieyohannan: Just moved below the section on expressions

docs/reference/modules/ROOT/pages/configuration-reference.adoc
```diff
@@ -3111,50 +3111,6 @@ An approval job can have any name. In the example above the approval job is name
 
 '''
 
-[#jobfilters]
-==== *`filters`*
```
rosieyohannan: Just moved below section on using expressions

docs/guides/modules/deploy/pages/deploy-to-cloud-foundry.adoc
```diff
@@ -81,22 +81,16 @@ workflows:
       - dark-deploy:
           requires:
             - test
-          filters:
-            branches:
-              only: main
+          filters: pipeline.git.branch == main # run the job on the main branch only
```
stig: I think we need to quote the branch, e.g. 
```suggestion
          filters: pipeline.git.branch == "main" # run the job on the main branch only
```

(This occurs several other times too.)

docs/guides/modules/deploy/pages/deploy-to-google-cloud-platform.adoc
```diff
@@ -69,9 +69,7 @@ workflows:
       - deploy-job:
           requires:
             - build-job # Only deploy once the build job has completed
-          filters:
-            branches:
-              only: main # Only deploy on the main branch
+          filters: pipeline.git.branch == main # run the job on the main branch only
```
stig: quote again:
```suggestion
          filters: pipeline.git.branch == "main" # run the job on the main branch only
```

docs/guides/modules/deploy/pages/deploy-to-pypi-registry.adoc
```diff
@@ -230,10 +227,7 @@ workflows:
       - publish:
           context:
             - trusted-publishing-guard
-          filters:
-            branches:
-              only:
-                - main
+          filters: pipeline.git.branch == main # run the job on the main branch only
```
stig: Again this needs quoting. This is the last of these I'm calling out, but there are others. 
```suggestion
          filters: pipeline.git.branch == "main" # run the job on the main branch only
```

docs/guides/modules/ROOT/examples/expression-examples/workflow-when/boolean-dynamic-syntax.yml
```diff
@@ -1,3 +1,10 @@
+version: 2.1
+
+parameters:
+  run_integration_tests:
+    type: boolean
+    default: false
+
 workflows:
   integration_tests:
 # The "integration_tests" workflow only runs if the pipeline parameter "run_integration_tests" is true.
```
gordonsyme, line 10: ~\~ the expression below could be simplified to `when: pipeline.parameters.run_integration_tests`

docs/guides/modules/ROOT/examples/expression-examples/workflow-when/exclude-branch-dynamic.yml
```diff
@@ -0,0 +1,6 @@
+workflows:
+  my-workflow:
+# The "my-workflow" workflow will NOT run when the branch name matches the pattern "feature-.*"
+    when: not (pipeline.git.branch matches "^feature-.*$")
```
gordonsyme: Patterns need to be delimited with `/`s so the interpreter knows they aren't strings, so
```
when: not (pipeline.git.branch matches /^feature-.*$/)
```

docs/guides/modules/ROOT/partials/using-expressions/operators.adoc
```diff
@@ -16,6 +16,10 @@
 | `starts-with`
 | String prefix equality, `"hello world" starts-with "hello"` evaluates as `true`. It is an error to use a non-string type as an operand.
 
+| Equality
+| `matches`
+| String pattern matching, `"hello world" matches "hello.*"` evaluates as `true`. It is an error to use a non-string type as an operand.
```
gordonsyme: Patterns have `/` delimiters: `"hello world" matches /hello.*/`

docs/guides/modules/orchestrate/pages/workflows.adoc
```diff
@@ -467,11 +468,7 @@ workflows:
     triggers:
       - schedule:
           cron: "0 0 * * *"
-          filters:
-            branches:
-              only:
-                - main
-                - /^release\/.*/
+          filters: pipeline.git.branch == "main" or pipeline.git.branch matches /^release\/.*/ # only run the nightly workflow on the main branch and any branch that starts with `release/`
```
gordonsyme: I am not certain that we changed config processing for the scheduled workflows case (since we want scheduled workflows to go away). Will need to check up on this one.

rosieyohannan: I'm struggling to verify this one - even using old style filters I'm having issues actually 👎

stig (changes requested):
Good direction overall — promoting expressions as the primary syntax makes sense. But there are several correctness issues that need fixing before merge, ranging from invalid YAML to wrong semantics. Details below, roughly by severity.

---

**Summary of blockers**

There are two broken YAML constructs, one wrong field name (branch vs tag), one unquoted identifier, and a `matches` example with a stray `"` in the regex — all of which will mislead readers or produce invalid config. There are also several places where the semantic meaning has quietly changed from the original (tags-only → all-branches-plus-tags) without acknowledgment. The note about `pipeline.event.*` availability has been narrowed in a way that may be inaccurate. These need careful fixes before the PR lands.

docs/guides/modules/test/pages/testing-ios.adoc
```diff
@@ -154,8 +154,7 @@ workflows:
       - build-and-test
       - adhoc:
           filters:
-            branches:
-              only: development
+            pipeline.git.branch == "development" # only deploy on the development branch
```
stig: **CRITICAL — structurally broken YAML**

```yaml
      - adhoc:
          filters:
            pipeline.git.branch == "development" # only deploy on the development branch
          requires:
```

`filters:` is followed by a block-mapping child key `pipeline.git.branch == "development"`, which is not valid. The expression string needs to be on the same line as `filters:`:

```yaml
      - adhoc:
          filters: pipeline.git.branch == "development" # only deploy on the development branch
          requires:
```

This is the pattern used consistently everywhere else in the PR.

docs/guides/modules/ROOT/partials/using-expressions/env-vars-in-conditional-caveat.adoc
```diff
@@ -1,5 +1,3 @@
-* Environment variables cannot be used in logic statements. You can only use pipeline values or pipeline parameters.
+* Environment variables cannot be used in expressions and logic statements. You can only use pipeline values or pipeline parameters.
 +
-Logic statements are evaluated at compilation time whereas the actual values of environment variables only become available at run time.
-+
-See the link:https://support.circleci.com/hc/en-us/articles/21200178128155-Can-I-use-environment-variables-in-conditional-statements["Can I use environment variables in conditional statements?"] article for more information.
+Expressions and logic statements are evaluated at compilation time whereas the actual values of environment variables only become available at run time.
```
stig: **MINOR — support article link removed**

The PR drops the link to the ["Can I use environment variables in conditional statements?"](https://support.circleci.com/hc/en-us/articles/21200178128155-Can-I-use-environment-variables-in-conditional-statements) support article. The explanation in the partial is accurate without it, but the article likely contains more detail and troubleshooting help for readers who hit this limitation. Was this intentional? If the article is still live and relevant, worth keeping.

docs/orbs/modules/author/pages/testing-orbs.adoc
```diff
@@ -280,19 +283,15 @@ workflows:
           requires:
             - integration-tests-default-profile
           context: orb-publisher
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              only: /^v[0-9]+\.[0-9]+\.[0-9]+$/
+          filters: pipeline.git.branch matches /^v[0-9]+\.[0-9]+\.[0-9]+$/
```
stig: **CRITICAL — filters branch instead of tag**

```yaml
          filters: pipeline.git.branch matches /^v[0-9]+\.[0-9]+\.[0-9]+$/
```

The original filter was `tags: { only: /^v[0-9]+\.[0-9]+\.[0-9]+$/ }` — a **tag** filter. This new version checks the **branch name** against a semver pattern, which will never match. Should be:

```yaml
          filters: pipeline.git.tag matches /^v[0-9]+\.[0-9]+\.[0-9]+$/
```

docs/guides/modules/ROOT/examples/expression-examples/workflow-when/boolean-complex.yml
```diff
@@ -0,0 +1,6 @@
+workflows:
+  my-workflow:
+# This workflow runs when: branch is "main" AND param1 is false AND (param1 OR param2 is true)
```
stig: **MINOR — misleading comment**

```
# This workflow runs when: branch is "main" AND param1 is false AND (param1 OR param2 is true)
```

The comment says "(param1 OR param2 is true)" but the condition is `(pipeline.parameters.param1 or pipeline.parameters.param2)`. This means "param1 is truthy OR param2 is truthy" — both are being evaluated as booleans, not just param2. The phrase "param2 is true" implies only param2 is the boolean check.

Also worth noting: `param1 is false AND (param1 OR param2)` is logically equivalent to `param1 is false AND param2 is true` (since param1 is already known false). The comment is technically accurate but the example as written could be simplified to `... and pipeline.parameters.param2`. If the intent is to show complex boolean composition, consider using two distinct independent parameters.

docs/guides/modules/execution-managed/pages/private-images.adoc
```diff
@@ -192,17 +192,11 @@ workflows:
   main:
     jobs:
       - build:
-          filters:
-            tags:
-              only: /^\d{4}\.\d+$/
+          filters: pipeline.git.branch or pipeline.git.tag matches /^\d{4}\.\d+$/ # run the job on all branches and tags that match the pattern
       - deploy:
           requires:
             - build
```
stig, line 198: **MAJOR — semantic change, and operator precedence ambiguity**

Original `build` job filter: `tags: { only: /^\d{4}\.\d+$/ }` — runs on **tags only** matching the pattern.

New: `pipeline.git.branch or pipeline.git.tag matches /^\d{4}\.\d+$/` — due to operator precedence, `matches` binds tighter than `or`, so this parses as:

```
pipeline.git.branch or (pipeline.git.tag matches /^\d{4}\.\d+$/)
```

Meaning: run on **any branch**, OR on tags matching the pattern. This is a significant expansion of the original intent (tags-only) and will run the `build` job on every branch push.

The same issue applies to the `deploy` job just below — the original had `branches: ignore: /.*/` (no branches at all) plus `tags: only: /^\d{4}\.\d+$/`. The new expression also runs on all branches.

If the original intent was "tags matching the pattern only", use:
```yaml
filters: pipeline.git.tag matches /^\d{4}\.\d+$/
```

If running on all branches + matching tags is genuinely the desired new behaviour, the comment should be updated to say so explicitly, and it's worth a note explaining the change.

docs/reference/modules/ROOT/pages/configuration-reference.adoc
```diff
@@ -2618,8 +2618,6 @@ workflows:
                only:
                  - main
                  - beta
-     jobs:
```
stig, line 2621: **MINOR — incomplete scheduled workflow examples**

Four places in this file remove `jobs: - test` from the scheduled workflow examples, leaving the `schedule:` block without a `jobs:` key. The examples now show a `triggers:` block with no `jobs:` section, which is an incomplete (and technically invalid) workflow definition. Readers won't know what jobs the schedule would run.

Either keep the `jobs:` key in each example, or add a comment like `# ... jobs: - your-job-name` to make clear it's required.

stig (changes requested):
Good direction overall — promoting expressions as the primary syntax is the right call. There are several correctness issues that need fixing before merge though, ranging from invalid YAML to a wrong field name. Details in the inline comments.

**Blockers (must fix):**

1. `workflows.adoc` — YAML anchor `&filters-staging` / `&filters-production` appended to scalar expression strings is invalid YAML. Anchors only work on mapping nodes; the anchor will be treated as literal string text, and the downstream `<<: *filters-staging` merge keys will break. Remove the anchor pattern and repeat the expression strings, or restructure the example.
2. `workflows.adoc` — `pipeline.git.branch != main` — `main` must be quoted: `"main"`.
3. `testing-ios.adoc` — `filters:` with the expression on the next line as an indented child key is invalid. The expression must be inline: `filters: pipeline.git.branch == "development"`.
4. `testing-orbs.adoc` — `orb-tools/publish` job filter uses `pipeline.git.branch matches /^v.../` but the original was a **tag** filter. Should be `pipeline.git.tag matches /^v.../`.

**Should fix:**

5. `operators.adoc` — `matches` example has a stray `"` inside the regex: `/hello.*"/` should be `/hello.*/`.
6. `workflows.adoc` / `private-images.adoc` — multiple places use `pipeline.git.branch or pipeline.git.tag matches /pattern/`. Due to operator precedence, `matches` binds tighter than `or`, so this means "any branch, OR tags matching the pattern" — not "branches or tags that match the pattern". Where the original was tags-only, the semantics have changed silently. Either fix the logic or document the intent change explicitly.
7. `orchestration-cookbook.adoc` — The `pipeline.event.*` availability note was changed from "only available for GitHub App" to "not available for GitLab". This may mislead GitHub OAuth users. Please verify the actual platform support.

**Nice to fix:**

8. `configuration-reference.adoc` — Four scheduled workflow examples had `jobs: - test` removed, leaving the `triggers:` block without a `jobs:` key. Add a placeholder or restore the line so readers know `jobs:` is still required.
9. `boolean-complex.yml` — The comment "param1 OR param2 is true" is ambiguous; "param1 is truthy OR param2 is truthy" would be clearer. The example is also logically redundant (can be simplified to `... and param2`) — consider two independent parameters instead.
10. `env-vars-in-conditional-caveat.adoc` — Support article link was removed. Intentional?

docs/guides/modules/orchestrate/pages/workflows.adoc
```diff
@@ -780,15 +759,11 @@ workflows:
   build: # This workflow will run on all branches except 'main' and will not run on tags
     jobs:
       - test:
-          filters:
-            branches:
-              ignore: main
+          filters: pipeline.git.branch != main # only run the workflow on any branch that is not the main branch
```
stig: YAML anchor syntax is invalid here. `value &anchor` puts `&filters-staging` inside the scalar string — it does not create a YAML anchor. Anchors must precede the node they name.

```suggestion
          filters: &filters-staging pipeline.git.branch == "main" # this yaml anchor is setting these values to "filters-staging"
```

Same issue on the `&filters-production` line below.

docs/guides/modules/orchestrate/pages/workflows.adoc
```diff
@@ -797,11 +772,7 @@ workflows:
   production: # This workflow will only run on tags (specifically starting with 'v.') and will not run on branches
     jobs:
       - test:
-          filters: &filters-production # this yaml anchor is setting these values to "filters-production"
-            branches:
-              ignore: /.*/
-            tags:
-              only: /^v.*/
+          filters: pipeline.git.tag matches /^v.*/ &filters-production # this yaml anchor is setting these values to "filters-production"
```
stig: Same YAML anchor problem — `&filters-production` must precede the scalar value:

```suggestion
          filters: &filters-production pipeline.git.tag matches /^v.*/ # this yaml anchor is setting these values to "filters-production"
```

docs/orbs/modules/author/pages/testing-orbs.adoc
stig: Wrong pipeline value — this checks the branch name against a semver pattern, but the original filter was tag-only (`branches: ignore: /.*/ tags: only: /^v.../`). Orb publishing should only trigger on version tags, not on branches named `v1.2.3`.

```suggestion
          filters: pipeline.git.tag matches /^v[0-9]+\.[0-9]+\.[0-9]+$/
```

docs/guides/modules/orchestrate/pages/workflows.adoc
stig: Unquoted string. `main` is not a keyword — every other comparison in this PR uses quoted strings. Should be `"main"`.

```suggestion
          filters: pipeline.git.branch != "main" # only run the workflow on any branch that is not the main branch
```

docs/guides/modules/ROOT/examples/expression-examples/workflow-when/boolean-complex.yml
stig: The comment is misleading and the example has dead code. Because `not pipeline.parameters.param1` is required by the third clause, `param1` is always `false` at that point — so `(pipeline.parameters.param1 or pipeline.parameters.param2)` simplifies to `pipeline.parameters.param2`. The `param1` inside the `or` is unreachable.

Consider simplifying the condition and updating the comment, or replacing this with an example that genuinely exercises all three parameters independently.

stig (changes requested):
The strategic direction is solid — expressions are the right first choice, and the bulk of the simple `branches: only: main` → `pipeline.git.branch == "main"` conversions are clean and correct. A handful of bugs are concentrated in the more complex filter conversions.

**Critical**

- **YAML anchor syntax** (×2, `workflows.adoc`): `value &anchor` is not valid YAML — the anchor becomes part of the scalar string. Fix: `&anchor value` (anchor must precede the node). See inline comments.
- **`testing-ios.adoc` indentation**: expression is on the line below `filters:` as a block child, not its scalar value — invalid YAML. Should be `filters: pipeline.git.branch == "development"` inline.
- **`testing-orbs.adoc` `orb-tools/publish`**: `pipeline.git.branch matches /^v.../` checks the branch name, not the tag. Should be `pipeline.git.tag`. This would prevent orb publish from running on version tags.
- **Operator precedence** (×6 instances): `matches` binds tighter than `or`, so `pipeline.git.branch or pipeline.git.tag matches /^v.*/` parses as `pipeline.git.branch or (pipeline.git.tag matches /^v.*/)` — runs unconditionally on every branch. All six instances have wrong inline comments too. Affected: `workflows.adoc` build-deploy (×2), build-test-deploy (×2), `private-images.adoc` (×2).

**Major**

- **Unquoted `main`** (`workflows.adoc`): `pipeline.git.branch != main` — `main` is not a keyword, should be `"main"`.
- **`pipeline.event.*` availability** (`orchestration-cookbook.adoc`): Changed from "only available for GitHub App" to "not available for GitLab". If these variables are also unavailable for GitHub OAuth, the new wording silently drops a restriction. Needs product verification.
- **Stray `"` in regex** (`operators.adoc`): `/hello.*"/` → `/hello.*/`.
- **Dead code in `boolean-complex.yml`**: `not param1` is required, so `param1` inside the subsequent `or` is always false — the subexpression reduces to `param2`. Comment is also wrong.

**Minor**

- **`^` stripped from legacy regex** (`workflows.adoc` line ~475, also `configuration-reference.adoc`): `/release\/.*/` instead of `/^release\/.*/` — changes "starts with" to "contains".

stig:
Apologies for the spam, I used Claude to help me review your Pr and it ended up submitting it before I had reviewed and edited its findings :-(

I'll review and close unnecessary comments.

docs/guides/modules/execution-managed/pages/private-images.adoc
```diff
@@ -192,17 +192,11 @@ workflows:
   main:
     jobs:
       - build:
-          filters:
-            tags:
-              only: /^\d{4}\.\d+$/
+          filters: pipeline.git.branch or (pipeline.git.tag matches /^\d{4}\.\d+$/) # run the job on all branches and tags that match the pattern
```
stig: I wonder if this can be made more clear by use of comma: 

```suggestion
          filters: pipeline.git.branch or (pipeline.git.tag matches /^\d{4}\.\d+$/) # run the job on all branches, and tags that match the pattern
```

rosieyohannan: Yeah I realised after your review there are some comments that could definitely be improved like this. I will go through them again, thank you!

stig:
PS: overall I'm really happy with where this is going :-) Nice diff stat too.

rosieyohannan:
> PS: overall I'm really happy with where this is going :-) Nice diff stat too.

🧹 🗑️

stig (approved):
I skimmed some of the changes since my preview reviews, and the full diff. I did not go through everything with a fine tooth comb again, but this looks like a nice improvement.

meeech (approved):

docs/guides/modules/execution-managed/pages/private-images.adoc
bosniltd-a11y: {
  "state": "created",
  "created_at": "2019-08-24T14:15:22Z",
  "number": 25,
  "id": "497f6eca-6276-4993-bfeb-53cbbbba6f08"
}

bosniltd-a11y: {
  "state": "created",
  "created_at": "2019-08-24T14:15:22Z",
  "number": 25,
  "id": "497f6eca-6276-4993-bfeb-53cbbbba6f08"
}

bosniltd-a11y: > {
>   "state": "created",
>   "created_at": "2019-08-24T14:15:22Z",
>   "number": 25,
>   "id": "497f6eca-6276-4993-bfeb-53cbbbba6f08"
> }

PR #10397, opened 2026-05-26
Document default machine images and resource classes (Server 4.10)

soulchips:
## Summary

- Add developer-facing subsections (nested under AWS and GCP) on `docs/server-admin-4.10/.../manage-virtual-machines-with-machine-provisioner.adoc` that list the default machine executor image names (incl. `default`/`docker-default` aliases and the GCP-only `android-default`), a sample `.circleci/config.yml`, and resource-class tables per executor type.
- Refresh `Default AWS AMI lists` from a flat 2-tab block into a 3-tab block keyed by Ubuntu version (22.04 / 24.04 / 26.04). Each tab holds two stacked, captioned tables (`.x86 AMI list`, `.Arm AMI list`).
- Update all 22.04 AMI IDs to current values, add full 24.04 and 26.04 matrices, and add `eu-west-3` rows to every Arm table (it was missing from the prior 22.04 Arm list).

## Test plan

- [ ] Build the docs locally and confirm the page renders without AsciiDoc errors.
- [ ] Verify the Default AWS AMI lists tab group shows three tabs (Ubuntu 22.04 / 24.04 / 26.04), each with two captioned tables of 17 rows.
- [ ] Verify the new `Default machine executor images`, `Reference default images in .circleci/config.yml`, and `Available resource classes` subsections appear nested under both AWS and GCP.
- [ ] Spot-check a few AMI IDs against the server-4.10 source maps.
- [ ] CI Vale check passes.


atulsingh0 (approved):

rosieyohannan:
The resource class tables actually already exist in the configuration reference and are stored as reusable partials so we can just pull them in: https://circleci.com/docs/reference/configuration-reference/#resourceclass.

Can you check they are correct for server and I can update them in this branch?

soulchips:
> The resource class tables actually already exist in the configuration reference and are stored as reusable partials so we can just pull them in: https://circleci.com/docs/reference/configuration-reference/#resourceclass.
> 
> Can you check they are correct for server and I can update them in this branch?

so they aren't consistent between cloud and server. I do know that resource classes got added and dropped between server versions. so we don't match.

eg: on cloud there is small and medium+ but we [don't have those resource classes on server](https://github.com/circleci/server/blob/main/templates/machine-provisioner/configmap.yaml#L71-L90)

rosieyohannan (approved):
Thank you for doing this, looks great!

PR #10408, opened 2026-05-27
Replace byte-chunking with section-based Algolia indexing

rosieyohannan:
## Summary

- **Root cause of poor search**: The previous indexing split pages at arbitrary byte boundaries (~8,500 bytes), producing ~520 large records with no awareness of document structure. Chunk boundaries fell mid-phrase (e.g. splitting "Pipeline states" across two records), and huge records diluted Algolia relevance scores so the right section ranked below unrelated pages.
- **New approach**: One Algolia record per `h2`/`h3`/`h4` section. The section heading is stored in a dedicated `heading` field with higher search priority than body content. The section anchor (`#pipeline-states`) is included in the record URL so clicking a result links directly to the right section on the page.
- **Expected record count**: ~5,000–15,000 (from ~520), average size ~1KB (from ~5.5KB) — matching what the old Algolia crawler produced.
- **Search results display**: Shows `Page title / Section heading` so users can see both which page and which section matched.

## Changes

**`extensions/export-content-extension.js`**
- Replace `collectPages` + `chunkText` with section-based extraction using `extractSections`
- `extractSections` splits page HTML at `h2`/`h3`/`h4` boundaries, extracts heading text and `id` attribute (anchor) per section
- `indexToAlgolia` simplified: one record per section, no chunking loop
- `searchableAttributes` updated to `[title, heading, content, path]` — section heading matches rank above body text

**`ui/src/js/07-search.js`**
- Add `heading` to `attributesToRetrieve`
- Result title displays as `Page title / Section heading` when a section heading exists

## Test plan

- [ ] Run a local build with `SKIP_INDEX_SEARCH=true` and inspect `.temp/site-content.json` — records should have `heading` field and `relUrl` values with `#` anchors
- [ ] After merge and CI index rebuild: search "pipeline states" — should return the Pipeline states section of the pipelines overview page as a top result
- [ ] Verify search result displays as `Pipelines overview and setup / Pipeline states`
- [ ] Verify clicking the result lands on the `#pipeline-states` anchor
- [ ] Check Algolia dashboard record count is substantially higher than 520

### DIFF
--- a/extensions/export-content-extension.js
+++ b/extensions/export-content-extension.js
@@ -5,7 +5,9 @@ const path = require('path');
 const { algoliasearch } = require('algoliasearch');
 
 /**
- * An Antora extension that exports page content to JSON and indexes it in Algolia v5 with chunking support.
+ * An Antora extension that exports page content to JSON and indexes it in Algolia.
+ * One record per section (h2/h3/h4 heading), with the section heading stored as a
+ * dedicated field and the section anchor included in the record URL.
  */
 module.exports.register = function () {
   this.once('navigationBuilt', async ({ playbook, contentCatalog }) => {
@@ -18,8 +20,8 @@ module.exports.register = function () {
     console.log(`Wrote JSON to ${outPath}`);
 
     if (shouldSkipIndexing()) {
-        console.log('Skipping Algolia indexing (requested by configuration)');
-        return;
+      console.log('Skipping Algolia indexing (requested by configuration)');
+      return;
     }
 
     if (!hasIndexed(tempDir) && hasAlgoliaCredentials()) {
@@ -37,12 +39,11 @@ module.exports.register = function () {
 };
 
 /**
- * Collect pages from contentCatalog and extract text, including component, version, and relative URL
+ * Collect all sections across all pages, one entry per h2/h3/h4 section.
  */
 function collectPages(contentCatalog, siteUrl) {
   const all = [];
 
-  // Define which server-admin versions to exclude from indexing
   const excludedServerAdminVersions = [
     'server-4.2',
     'server-4.3',
@@ -56,10 +57,9 @@ function collectPages(contentCatalog, siteUrl) {
 
   contentCatalog.getComponents().forEach(({ name: comp, versions }) => {
     versions.forEach(({ version }) => {
-      // Skip indexing for excluded server-admin versions
       if (comp === 'server-admin' && excludedServerAdminVersions.includes(version)) {
         console.log(`Skipping Algolia indexing for server-admin version: ${version}`);
-        return; // Skip this entire component version
+        return;
       }
 
       const compVer = contentCatalog.getComponentVersion(comp, version);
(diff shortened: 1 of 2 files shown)


extensions/export-content-extension.js
```diff
@@ -68,36 +68,72 @@ function collectPages(contentCatalog, siteUrl) {
...
+    const parsed = parseHTML(`<div>${part}</div>`);
+    const headingEl = parsed.querySelector('h2, h3, h4');
+
+    const heading = headingEl
+      ? headingEl.textContent.trim()
+      : (index === 0 ? pageTitle : '');
+
+    const anchor = headingEl?.getAttribute('id') || '';
+    const relUrl = anchor ? `${pageRelUrl}#${anchor}` : pageRelUrl;
+    const content = parsed.textContent.trim().replace(/\s+/g, ' ');
```
meeech: `textContent` here runs over the heading **and** body together. Asciidoctor emits `</h2><div class="sectionbody">` with no whitespace between them, so stripping the tags fuses the heading's last word with the body's first word — e.g. `Pipeline states` + `A pipeline...` → `"Pipeline statesA pipeline..."`. The `\s+ → ' '` collapse can't fix it because there was never a space there to collapse.

Simple fix: drop the heading node before reading the body, then join with a space. Keeps the heading words searchable in `content` without the glue.

```suggestion
    const anchor = headingEl?.getAttribute('id') || '';
    const relUrl = anchor ? `${pageRelUrl}#${anchor}` : pageRelUrl;
    // Remove the heading node first — textContent inserts no separator where a
    // tag is stripped, so heading text would otherwise fuse to the first body word.
    headingEl?.remove();
    const body = parsed.textContent.trim();
    const content = `${heading} ${body}`.trim().replace(/\s+/g, ' ');
```

extensions/export-content-extension.js
```diff
@@ -68,36 +68,72 @@ function collectPages(contentCatalog, siteUrl) {
...
+
+  parts.forEach((part, index) => {
+    if (!part.trim()) return;
+
+    const parsed = parseHTML(`<div>${part}</div>`);
+    const headingEl = parsed.querySelector('h2, h3, h4');
+
+    const heading = headingEl
+      ? headingEl.textContent.trim()
+      : (index === 0 ? pageTitle : '');
```
meeech, line 115: The badge extension renders a `<span class="subsection-badge">Preview</span>` inside the `<h2>`, so `textContent` pulls the badge word into the heading: `Slack notifications` → `"Slack notifications Preview"`. That badge is page chrome, not part of the section name — but it gets indexed as high-priority search text and shown in the `Page title / Section heading` result.

The span is class-tagged, so strip it before reading the heading:

```suggestion
    const headingEl = parsed.querySelector('h2, h3, h4');
    headingEl?.querySelector('.subsection-badge')?.remove();

    const heading = headingEl
      ? headingEl.textContent.trim()
      : (index === 0 ? pageTitle : '');
```

meeech (changes requested):

meeech (approved):

PR #10566, opened 2026-07-08
Add PyPI OIDC trusted publishing guide

meeech:
# Description
Adds a new how-to guide, **Publish to PyPI using OIDC trusted publishing**, and lists it in the deploy nav next to the npm trusted publishing guides.

The recommended flow uses twine's native trusted publishing (twine 6.1.0+), where twine detects the CircleCI OIDC credentials and performs the exchange automatically. A manual mint-token exchange is included as a fallback tab. The guide also covers protected-context lockdown, self-hosted runners, and deploy markers, mirroring the npm guide.

# Reasons
CircleCI is being added as a trusted publisher on PyPI. This documents the tokenless publishing path.

> [!NOTE]
> Screenshots are placeholders pending the feature landing on pypi.org.

# Content checks

**Preview your changes:**
- [ ] View the Vale linter results, select the `ci/circleci: lint` job at the bottom of your PR.
- [ ] Preview your changes, select the `ci/circleci: build` job and open `index.html` in the Artifacts tab.

**Formatting:**
- [x] Check all headings h1-h6 are in sentence case.


docs/guides/modules/deploy/pages/deploy-to-pypi-registry.adoc
```diff
@@ -0,0 +1,269 @@
+= Publish to PyPI using OIDC trusted publishing
+:page-badge: Preview
+:page-platform: Cloud
+:page-description: Publish packages to PyPI from CircleCI using OIDC trusted publishing, the recommended tokenless approach.
+:experimental:
+
+In this how-to guide, you will learn how to configure CircleCI to publish to the Python Package Index (PyPI) using trusted publishing. Trusted publishing uses OpenID Connect (OIDC) to exchange a short-lived CircleCI token for a PyPI API token at publish time. No long-lived `PYPI_API_TOKEN` exists to store, rotate, or leak.
+
+For a functional example you can clone and adapt, see the link:https://github.com/CircleCI-Public/sign-and-publish-examples/tree/main/pypi[CircleCI PyPI trusted publishing example].
```
rosieyohannan, line 11: fork rather than clone?

Also are we pointing to the right place? Should it be https://github.com/CircleCI-Public/sign-and-publish-examples/tree/pypi ?

docs/guides/modules/deploy/pages/deploy-to-pypi-registry.adoc
```diff
@@ -0,0 +1,269 @@
...
+. Save the publisher.
+
+To publish a project name that does not exist yet, create a pending publisher instead from menu:Your projects[Publishing], then publish your first release through the workflow.
+
+[#lock-it-down]
+== 3. Lock it down
+
+We recommend adding some additional information to further secure your publishing pipeline. Contexts will allow you to control who can run pipelines that publish. You can also add information about your source repo and branch.
+
+=== 3.1 VCS origin and VCS ref
```
rosieyohannan: I think this para would be better up in the step where we show people to gather this info. Then this lock it down step can just be for getting people to set up some context restrictions (if they choose to do that). I can make this change

rosieyohannan: done

docs/guides/modules/deploy/pages/deploy-to-pypi-registry.adoc
meeech: no, clone is fine. they dont need to fork it. 

Also, this will be my final destination of it, not on a branch. i just have to merge it. I want to do a new test of it first

rosieyohannan (approved):

docs/guides/modules/deploy/pages/deploy-to-pypi-registry.adoc
meeech: i guess - we can discuss. its there because thats part of locking it down. anyways, we can adjust after.

rosieyohannan: @meeech Yeah so the reason I moved it was to get the info in the step where you actually get and then use the VCS origin and ref. 

As a step 3 the contexts restriction makes sense to me because you could get to actually doing that, setting up the restruction after the setup step before, so it is actually a next step? But the VCS optional info is added in step 2 already so telling you to do it in step 3 seemed off. Happy to rework this though. Or put a note or something in the lock it down section about what the two VCS options are for.

PR #10623, opened 2026-07-29
fix: correct broken server-terraform release tag links in upgrade guides (4.8, 4.9, 4.10)

hidetaka-cci:
## Summary

The server-terraform GitHub release tag naming convention changed starting from version 4.8.
Tags no longer use bare version numbers (e.g. `4.9.4`) but now include a `server-` prefix
(e.g. `server-4.9.4`). As a result, the Terraform module upgrade links in the upgrade
guides for Server 4.8, 4.9, and 4.10 were pointing to 404 URLs.

## Changes

Updated the `server-terraform` release URLs in the following pages to include the `server-`
prefix in the tag path:

- `docs/server-admin-4.8/modules/installation/pages/upgrade-server.adoc`
- `docs/server-admin-4.9/modules/installation/pages/upgrade-server.adoc`
- `docs/server-admin-4.10/modules/installation/pages/upgrade-server.adoc`

**Before:**
```
https://github.com/CircleCI-Public/server-terraform/releases/tag/4.9.4  (404)
```

**After:**
```
https://github.com/CircleCI-Public/server-terraform/releases/tag/server-4.9.4  (✓)
```

Note: The `{serverversion*}` attributes in `antora-playbook.yml` intentionally retain bare
version numbers (e.g. `4.9.4`) as they are also used in Helm commands where the prefix is
not required. The fix was applied at the URL template level instead.

## Testing

Verified the following URLs resolve correctly:
- https://github.com/CircleCI-Public/server-terraform/releases/tag/server-4.8.11
- https://github.com/CircleCI-Public/server-terraform/releases/tag/server-4.9.4
- https://github.com/CircleCI-Public/server-terraform/releases/tag/server-4.10.0

Made with [Cursor](https://cursor.com)

PR #10624, opened 2026-07-29
docs: add note on pipeline number priority vs. lock arrival order in serial groups

hidetaka-cci:
## Summary

- Adds a NOTE block to the "Pipeline number priority" section of the serial groups page
- Clarifies that priority is determined by pipeline number, not job arrival order at the lock
- Highlights a dynamic configuration-specific risk: setup workflow duration varies between pipelines, so continuation jobs may reach the lock out of merge order, causing older pipeline jobs to be skipped rather than queued

## Motivation

The existing documentation stated the priority rule but did not address a common source of confusion in dynamic config setups: when setup workflow duration differs between pipelines, lock arrival order and pipeline number order can diverge. This note makes the distinction explicit and warns users not to rely on merge order as a guarantee.

## Changes

- `docs/guides/modules/orchestrate/pages/controlling-serial-execution-across-your-organization.adoc`: added NOTE block under "Pipeline number priority" section

Made with [Cursor](https://cursor.com)


rosieyohannan:
@hidetaka-cci I think the commits from you server docs branch are in this branch too. Could you separate them? Then we can get a review from the pipelines team for the serial groups note

hidetaka-cci:
@rosieyohannan 
Thanks for catching that — commits are now separated (server-terraform fix has its own PR: #10623). Also fixed a Vale lint error the new note introduced.

Two things still show in CI but look unrelated to this change: a pre-existing Vale error elsewhere in the file, and a flaky build job (Antora choking on an unrelated branch, v3-api-docs-link) — main's recent builds are all green. Want me to fix the pre-existing Vale error here too, or leave it for later? Otherwise this should be ready to review.

docs/guides/modules/orchestrate/pages/controlling-serial-execution-across-your-organization.adoc
```diff
@@ -49,6 +49,13 @@ CircleCI uses a distributed locking system to control execution order. When a jo
 
 CircleCI uses pipeline numbers to determine execution priority. Jobs from newer pipelines (higher pipeline numbers) take precedence over jobs from older pipelines within the same project. If a job from an older pipeline tries to run while a newer pipeline's job is waiting or running, the older job is automatically skipped.
 
+[NOTE]
+====
+Serial group priority is based on pipeline number, not on when a job reaches the lock. When a job attempts to acquire a serial group lock, CircleCI checks for jobs from higher-numbered pipelines in the same serial group. If one is already queued or running, CircleCI skips the job.
+
+With dynamic configuration, setup workflow duration varies between pipelines, so the order in which continuation jobs reach the lock is not guaranteed to match pipeline number (that is, merge) order. A job from an older pipeline may therefore be skipped rather than queued. Do not rely on merge order to guarantee that every pipeline's serially grouped job runs.
```
stig, line 56: I don't think it's guaranteed that hooks don't reach us in order either.

hidetaka-cci: @stig Thanks for flagging this — I want to make sure I understand the concern correctly before revising.
Is the point that pipeline number order isn't a perfectly reliable proxy for merge order to begin with (e.g. webhook delivery to CircleCI isn't guaranteed to be strictly ordered), independent of the dynamic config timing issue I described? If so, I agree the note as written overstates it by attributing the mismatch solely to setup workflow duration.

If that's the concern, would it work to soften the note to say something like "pipeline number order may not always align with merge order, for reasons including but not limited to variable setup workflow duration under dynamic config"? Let me know if I'm still missing something.

Thanks

stig: web hooks arriving (or being processed) out of order either from the provider, or because of retries on our end, before they arrive at builds-service (which allocates the pipeline number) is my concern. Dynamic config is not relevant: continuations use the same pipeline number as the setup workflow.

PR #10659, opened 2026-08-18
Fix structure, audience, and cross-doc consistency for OAuth DCR page

rosieyohannan:
## Summary

**Intro and audience**
- Added explicit audience statement so developers can self-identify and regular users know to look elsewhere
- Reframed value proposition from "avoids X" to a concrete benefits list (no long-lived secrets, user-controlled scopes, tokens are revocable)

**Prerequisites**
- Replaced four constraint bullets (which were rules, not prerequisites) with actual prerequisites: CircleCI account (standard xref pattern), `curl`, `python3`, `jq` (optional), each with a `brew install` command
- Moved constraints into the steps where they are relevant:
  - Redirect URI loopback rule → Step 1 IMPORTANT admonition
  - S256 required → Step 2 PKCE sub-section intro
  - `state` required → Step 3 intro sentence
- Removed the OAuth endpoints sub-section (every endpoint already appears in the step that uses it)

**Step structure**
- Merged the standalone "Understand PKCE values" step into Step 2 as an optional "Understanding PKCE values" sub-section — action comes first, reference material follows
- Moved access levels table inline to the consent screen in Step 3, where the user actually makes the choice; removed the standalone Access levels section
- Added `CLIENT_ID` variable assignment to Step 1 immediately after the registration response
- Removed static URL template from Step 3; replaced with the `printf` command readers will actually run, followed by a "URL parameter reference" sub-section for readers who want to understand the parameters
- Clarified substitutable vs fixed values in Step 1 (`<your-tool-name>`, port, path) vs fields that must remain as shown
- Removed stale "Before opening the authorization URL" sentence from Step 2 intro
- Renamed "Token lifecycle" → "Manage your token's expiry and rotation"; added section intro paragraph
- Condensed duplicate token expiry prose in Step 4 to a single cross-reference

### managing-api-tokens.adoc
- Updated personal token description to clarify that manually-created tokens carry full permissions, while OAuth-issued tokens are scoped to the access level chosen during authorization

### api-developers-guide.adoc
- Updated the personal tokens note to mention OAuth Dynamic Client Registration as an alternative token creation path, with a link to the new page

## Test plan

- [ ] Preview locally with `npm run start:dev` and confirm page structure and inline access levels table render correctly
- [ ] Verify all xrefs resolve across the three changed files
- [ ] Confirm `[#token-lifecycle]` anchor still resolves after heading rename


nathanfishcircleci (commented):
perfect

PR #10753, opened 2026-09-17
Add how-to guide for using a GitHub Actions workflow file as pipeline config

rosieyohannan:
Adds a how-to guide for the preview feature that lets a pipeline definition use an existing GitHub Actions workflow file instead of a CircleCI config file.

Linear: [DOC-264](https://linear.app/circleci/issue/DOC-264)

## Scope

**GitHub App pipelines on CircleCI Cloud, and nothing else.** The page says support for more pipeline types is planned, rather than listing what is unsupported. Cursor Origin lands in the coming days once the UI and CLI/API changes are in, and the preview note is worded so that adding it needs no rewrite.

## What is here

**New page** at `guides:orchestrate:use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc`:

- Setup for a new project and for an existing project, each with Web app and CircleCI CLI tabs
- A short "How it works" mental model: CircleCI reads the workflow file, translates it into a CircleCI workflow with `needs:` becoming `requires:`, and runs the jobs with GitHub Actions semantics
- How the CircleCI trigger and the workflow file's `on:` key interact, with the event mapping table. Two separate gates, which is the most likely thing to confuse a reader
- A friendly version of the compatibility contract from `circleci/actions-runner`, using four support levels: Supported, Partial, No effect, Not supported
- Contexts table, limits, and the runtime values that differ from GitHub Actions, including the CircleCI-issued OIDC tokens

**Two nav entries**, under Orchestrate and under Migrate to CircleCI. The second is deliberate, so readers on the GitHub Actions migration path find it. It does break the alphabetical run in the Migrate list.

**Cross-links** from four pages: the config source list and the GitHub App pipeline steps in `pipelines.adoc`, a TIP in `migrating-from-github.adoc`, a new row in the pipeline type feature matrix, and a note in the GitHub tab of `create-project.adoc`.

**A new `circle-amber` status colour** in `ui/src/css/`, for the Partial support level. `#f7c007` is 1.68:1 on white, so light mode uses a darkened sibling at 4.90:1 and dark mode uses `#f7c007` at 10.95:1. This follows the pattern the existing `circle-green` and `circle-red` pairs already use. Needs `npm run build:ui` to see locally.

**`AGENTS.md`**: the worktree rule now says local agents work in the main checkout and ask first, while cloud agents still always create a worktree.

## Why placement went to Orchestrate

The feature reads as a GitHub thing, which pulls toward `integration` or `migrate`, but it is really a new kind of config source for a pipeline definition. `pipelines.adoc` already defines a pipeline as config source plus checkout source and lists what a config source can be, so this extends that list. Notes from working that out are on [DOC-254](https://linear.app/circleci/issue/DOC-254/working-towards-an-ia-skill) as input to the IA skill.

## Two things reviewers should look hardest at

**`--config-file-type github-actions`.** This is the flag that makes the CLI path work, and omitting it makes CircleCI parse the workflow file as CircleCI syntax. The failure is confusing rather than obvious, so it carries an IMPORTANT admonition with the actual error text.

**The "No effect" support level.** Several keys are accepted and then do nothing: `concurrency` does not serialize runs, `fail-fast` does not cancel siblings, job-level `container:` is ignored. A workflow relying on any of these for correctness or safety behaves differently here while still going green. The page calls this out in the support-level key and gives `actions/setup-python` its own subsection, since the disabled `add-path` command means later `run:` steps silently use the wrong tool version.

Every CLI command on the page was run against the installed CLI to confirm it exists and behaves as described. The steps run from the repository directory, because `--project` defaults to the git remote on all of them, which removes the project slug entirely.

## Open questions from the source material

- `compatibility.md` says "All four runtimes supported" for JavaScript actions but lists three. Written up as Node 16, 20, and 24.
- `compatibility.md` notes that the build emits a warning claiming `timeout-minutes` has no effect, and that the warning is wrong. The page documents the real behaviour and tells readers to ignore the warning. Fixing the warning may be preferable to documenting it.


docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,782 @@
...
+CAUTION: **There is no CLI teardown.** The CLI has no command to delete a project or a pipeline definition, and no command to delete or disable a trigger. Check your flags before you create anything, and remove unwanted pipelines and triggers in the web app.
+--
+====
+
+== How it works
+
+You point a pipeline at a GitHub Actions workflow file, and CircleCI does the rest each time the pipeline is triggered:
+
+. **CircleCI reads your workflow file** from the config source repository, at the path you gave the pipeline.
+. **CircleCI translates the workflow into a CircleCI workflow.** Each GitHub Actions job becomes a CircleCI job, and `needs:` relationships become `requires:` relationships, so your job graph is preserved.
```
briceicle: "...so your **workflow** graph is preserved."

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,782 @@
...
+. **Each job runs on the CircleCI actions runner.** The runner executes your steps using GitHub Actions semantics. This covers `run:` and `uses:` steps, `${{ }}` expression evaluation, step and job outputs, and secret masking.
```
briceicle, line 218: re: `CircleCI actions runner` - pinging @michael-webster / @nathanfishcircleci  to confirm this is the name we want to use publicly.

Buildkite calls their product `GitHub Actions Buildkite plugin`

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,782 @@
...
+--
+[cols="1,3", options="header"]
+|===
+| Level
+| Meaning
+
+| [.circle-green]#**Supported**#
+| Works as it does on GitHub Actions.
+
+| **Partial**
```
briceicle: this could be yellow/orange color

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,782 @@
...
+--
+[cols="2,1,3", options="header"]
+|===
+| Feature
+| Support
+| What to know
+
+| `secrets.*`
+| **Partial**
+| Resolved from the job environment, matching names without regard to case. Set these values as CircleCI environment variables or in a xref:security:contexts.adoc[Context]. Secrets are not available in `if:` conditions, or inside composite action children.
```
briceicle, line 580: @michael-webster can correct me if I'm wrong, but I think the guidance should be to add secrets to project environment variables only for now. 

During translation of the GHA yaml, we wouldn't know which contexts customers would have migrated their secrets to.

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,782 @@
...
+
+[.table-scroll]
+--
+[cols="2,1,1", options="header"]
+|===
+| Context
+| Job `if`
+| Step `if`
+
+| `github.actor`, `github.event_name`, `github.ref`, `github.ref_name`, `github.ref_type`, `github.repository`, `github.sha`, `github.base_ref`, and `github.head_ref`
```
briceicle, line 657: There are a few more `github.*` contexts available. The full list is [here](https://github.com/circleci/actions-runner/blob/main/docs/github-context-mapping.md#mapping-table)

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,782 @@
...
+| [.circle-green]#**Yes**#
+| [.circle-green]#**Yes**#
+
+| `github.event.*`
```
briceicle, line 661: The entire github event payload is not available. Only some fields will be available in the preview: see [here](https://github.com/circleci/actions-runner/blob/main/docs/github-context-mapping.md#githubevent-mapping)

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,782 @@
...
+| [.circle-red]#**No**#
+| [.circle-red]#**No**#
+
+| `secrets.*`
+| [.circle-red]#**No**#
+| [.circle-red]#**No**#
+|===
+--
+
+The `github.event.*` context is built from the CircleCI pipeline values, so it is a partial payload rather than a complete GitHub webhook payload.
```
briceicle, line 711: It is also worth nothing that `github.run_id` is a UUID. GitHub's `run_id` is a numeric integer, but CircleCI's pipeline ID is a UUID and there is no numeric equivalent available

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,782 @@
...
+--
+
+== What to expect when your workflow runs on CircleCI
+
+Your workflow runs on CircleCI compute, so some values your workflow reads at runtime differ from the values GitHub would report.
+
+* Jobs run on Linux x86-64. The `runner.os` context reports `Linux`, `runner.arch` reports `X64`, and `runner.environment` reports `self-hosted`.
+* All standard `GITHUB_*` and `RUNNER_*` environment variables are set, with the exception of `ACTIONS_RUNTIME_TOKEN`.
+* OIDC tokens are issued by CircleCI. See <<oidc-tokens,OIDC Tokens>>.
+* Set the secrets your workflow reads as CircleCI environment variables or in a xref:security:contexts.adoc[Context], rather than as GitHub repository secrets.
```
briceicle, line 755: Ditto about using contexts

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
rosieyohannan: 🙇‍♀️

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
rosieyohannan: Yeah that makes sense - I'll wait on word from webster then update

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,762 @@
...
+| [.circle-amber]#**Partial**#
+| Workflow, job, and step values use normal precedence, so the most specific value wins. `NODE_OPTIONS` cannot be set through `GITHUB_ENV`.
+
+| `concurrency`
+| **No effect**
+| Parsed, and a warning is emitted. Concurrent runs are not serialized, so a workflow that relies on a concurrency group to prevent overlapping deploys is not protected.
+
+| `environment`
+| **No effect**
+| Parsed, but no approval gate is added and no environment-scoped secrets or variables are resolved. A job that relies on a manual approval gate runs without one.
```
briceicle, line 343: I just implemented partial support for environments, see [here](https://github.com/circleci/actions-runner/blob/main/docs/compatibility.md#deployment-environments). When a job declares `environment:`, we inserts a CircleCI approval job before it.

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
nathanfishcircleci: Github Actions run_on CircleCI or something like that. Great question. 

@vineetantilcirclci ^^^

rosieyohannan: I'll just remove the reference to the runner and say CircleCI runs your Actions

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
rosieyohannan: I'll just recommend env vars for now

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
rosieyohannan: follow up quesiton:

Three things I couldn't verify from engineering's note, which affect what the page should say:

1. Environment-scoped secrets and variables. I kept the existing limitation, on the basis that "partial support" implies the approval job is what landed. But if they're now resolved too, that sentence is wrong and the row understates the support.
2. Who can approve. GitHub environments carry required reviewers. A CircleCI approval job is gated by CircleCI permissions instead, so "someone approves it" is deliberately vague. If the required-reviewers list is honoured, that's worth stating; if it isn't, it's arguably a security-relevant caveat a reader deploying to production would want flagged.
3. Wait timers and deployment branch rules. The other two things a GitHub environment can enforce. Silently ignoring a deployment branch rule would be the sharper of the two.

Worth a follow-up question to engineering, because the behaviour change cuts both ways: a workflow that declared environment: and previously ran straight through will now pause waiting for approval. That's the right behaviour, but it will surprise anyone who tried this last week — and it's the kind of thing worth a line in the preview release note as well as the docs.

briceicle: 1. Environment-scoped secrets and variables are not resolved. The approval job is a workaround this to gate jobs that require an environment
2. Anyone who has access to the project can approve the job. 
3. Those are ignored

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,844 @@
...
+
+CircleCI implements GitHub Actions semantics rather than calling GitHub, so only a subset of features is available. See <<github-actions-feature-support,GitHub Actions Feature Support>>.
+
+[#how-circleci-triggers-your-workflow]
+=== How CircleCI triggers your workflow
+
+Two things decide whether your workflow runs:
+
+. **The CircleCI trigger** decides whether the pipeline runs at all. You configure this in CircleCI, in the same way as any other pipeline.
+. **The `on:` key in your workflow file** decides whether your workflow accepts the event that CircleCI passes to it. A workflow that declares only unsupported events never runs.
```
briceicle: The two-gate framing here overclaims what `on:` does today.

We currently do not apply `on.<event>.branches` / `paths` / `types` filters — `On` is only read for `workflow_dispatch.inputs` and `workflow_call`. So e.g. a CircleCI `all-pushes` trigger + `on.push.branches: [main]` still runs on feature branches.

Schedule note below already says cron is not evaluated so it's the same here for filters.

Suggested rewrite:
- CircleCI trigger decides whether the pipeline runs
- `on:` filters (branches, paths, types) are not applied
- Keep the event-mapping table as “how CircleCI maps events into GHA context,” not as a second acceptance gate

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,844 @@
...
+
+| `runs-on`
+| [.circle-amber]#**Partial**#
+| Use `ubuntu-latest`, `ubuntu-24.04`, `ubuntu-22.04`, `ubuntu-20.04`, or `self-hosted`. All other labels are rejected, including all macOS and Windows labels.
+
+| `name`
+| [.circle-green]#**Supported**#
+| Retained at workflow and job level. Job names can use matrix values.
+
+| `run-name`
```
michael-webster: let's delete this.

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,844 @@
...
+| [.circle-amber]#**Partial**#
+| Literal combinations work. An `include` entry whose value is an expression that cannot be resolved when the workflow is parsed is skipped without an error.
+
+| `fail-fast`
+| **No effect**
+| A failing matrix instance does not cancel its siblings.
+
+| `max-parallel`
+| **No effect**
+| Parsed, but not enforced. Use CircleCI concurrency controls instead.
```
michael-webster: let's get rid of the Parsed, but not enforced. can be a follow up

docs/guides/modules/orchestrate/pages/use-a-github-actions-workflow-file-as-your-pipeline-configuration.adoc
```diff
@@ -0,0 +1,844 @@
...
+| GitHub Packages, Releases, Checks, and Deployments
+| [.circle-red]#**Not supported**#
+| These GitHub services are not emulated.
+|===
+--
+
+[#oidc-tokens]
+=== OIDC tokens
+
+A job that declares `permissions: id-token: write`, or `permissions: write-all`, receives an OIDC token endpoint. CircleCI injects `ACTIONS_ID_TOKEN_REQUEST_URL` and `ACTIONS_ID_TOKEN_REQUEST_TOKEN` into `run` steps and JavaScript action steps, so `core.getIDToken()` and a direct `curl` both work.
```
michael-webster: Let's change this to

```
A job that declares `permissions: id-token: write`, or `permissions: write-all`, can request an OIDC token. See the CircleCI docs on OIDC token formatt to migrate OIDC trust relationship
```

michael-webster (approved):

